PR #16, opened 2022-09-07
fix: event for ownership transfer #44

H3xept:


ezynda3:
@H3xept compilation is failing due to invalid variable names.

PR #30, opened 2022-09-09
fix: internal modifier for certain executor funcs

H3xept:
Not sure if this plays well with the changes to the Axelar Facet, please check.


ezynda3:
Deprecated https://github.com/lifinance/contracts/pull/73

PR #31, opened 2022-09-09
fix: added fees to bridge struct #24

H3xept:


H3xept:
@ezynda3 Are we not going to support callbacks? If we plan that, maybe we should also take the callback data from the struct.

PR #53, opened 2022-09-13
fix: arbitrum check ether balance #97

H3xept:


ezynda3:
https://github.com/spearbit-audits/LiFi/issues/97#issuecomment-1246177363

PR #60, opened 2022-09-13
refactor: moved generic errors to generic errors file  #19

H3xept:
Some were kept in the original file as they arguably they are not generic. @ezynda3 please let me know if you have other preferences here.


ezynda3:
Looks like one test is broken from this

PR #65, opened 2022-09-14
chore: bumped solidity version #25

H3xept:


ezynda3:
https://github.com/spearbit-audits/LiFi/issues/25#issuecomment-1249413323

H3xept:
Would you like to use another compiler release because of the bug the auditor mentioned?

PR #72, opened 2022-09-15
fix: separated token pulling from swapping

H3xept:


H3xept:
Closed in favour of https://github.com/lifinance/contracts/pull/75

PR #73, opened 2022-09-16
Add Receiver contract to fix issue with when post call fails on the destination chain #86

satoshi-lifi:


ezynda3:
@satoshi-lifi please check https://github.com/spearbit-audits/LiFi/issues/86

PR #74, opened 2022-09-16
refactor: new eth deposit method #115

H3xept:


ezynda3:
@H3xept  https://github.com/spearbit-audits/LiFi/issues/115#issuecomment-1249439656

PR #76, opened 2022-09-19
feat(wormhole): now supporting native token transfers #105

H3xept:


ezynda3:
@H3xept https://github.com/spearbit-audits/LiFi/issues/105#issuecomment-1236108448

PR #79, opened 2022-09-20
Hardcode bridge addresses via immutable #98

satoshi-lifi:


ezynda3:
@satoshi-lifi https://github.com/spearbit-audits/LiFi/issues/79#issuecomment-1254949953

PR #84, opened 2022-09-23
chore: array length into local variable #117

H3xept:


ezynda3:
https://github.com/spearbit-audits/LiFi/issues/117#issuecomment-1256926311

PR #85, opened 2022-09-23
fix: added check for native fee subtraction #118

H3xept:


ezynda3:
https://github.com/spearbit-audits/LiFi/issues/118#issuecomment-1256404690

PR #62, opened 2022-09-14
Issue 83 axelar claim #83

H3xept:

PR #58, opened 2022-09-13
refactor: using unchecked increments for all loops #26

H3xept:

PR #94, opened 2022-09-28
Processing of initial, final balances, remaining tokens #58 #62 #82

satoshi-lifi:


ezynda3:
https://github.com/spearbit-audits/LiFi/issues/82#issuecomment-1261660648

PR #97, opened 2022-09-29
fix(across): variable quote time

H3xept:


src/Facets/AcrossFacet.sol
```diff
@@ -38,9 +45,14 @@ contract AcrossFacet is ILiFi, ReentrancyGuard, SwapperV2, Validatable {
...
     /// @param _spokePool The contract address of the spoke pool on the source chain.
-    constructor(IAcrossSpokePool _spokePool, address _weth) {
+    constructor(
+        IAcrossSpokePool _spokePool,
+        address _weth,
+        uint256 _maxQuoteTime
+    ) {
         spokePool = _spokePool;
         weth = _weth;
+        getStorage().maxQuoteTime = _maxQuoteTime;
```
Jonah246, line 55: storage should not be set in the facet's constructor.
It only changes the storage in the implementation address.

H3xept: Yep, updated.

PR #107, opened 2022-09-30
feat: added receiver address to generic swap facet

H3xept:
I used the receiver also for excess tokens, please make sure it's ok


maxklenk (changes requested):
Please also adjust the tests.

H3xept:
Updated.

PR #110, opened 2022-09-30
feat: access control for withdraw

H3xept:
@maxklenk Im not sure if we want to move the contractOwner checks in the LibAccess impl -- this would entail that now LibAccess requires LibDiamond and imo libraries should be as standalone as possible.


maxklenk (approved):
Great.

Moving the whole in one function is not that important.

PR #111, opened 2022-09-30
fix: removed useless check and changed deprecated method signature

H3xept:


H3xept:
Not needed anymore after satoshis update

PR #112, opened 2022-09-30
fix: remove duplicate receiver check

maxklenk:
This check is already done by `validateBridgeData`.


H3xept:
please pull to fix the forge test

PR #124, opened 2022-10-19
Update documents of facets

satoshi-lifi:


docs/CBridgeFacet.md
```diff
@@ -53,7 +41,7 @@ The swap library can be found [here](../src/Libraries/LibSwap.sol).
 
 ## LiFi Data
```
H3xept, line 42: This is present in all facets, should this be ## BridgeData? 
Also the text below says "some methods" where in fact we always pass the bridge data as param?

ezynda3:
Can we move the `memory` to `calldata` changes to a different PR? We've already deployed all of these contracts to production and we need to figure out if it's even worth it to redeploy those at the moment.

ezynda3 (commented):
Can we move the memory to calldata changes to a different PR? We've already deployed all of these contracts to production and we need to figure out if it's even worth it to redeploy those at the moment.

satoshi-lifi:
> Can we move the memory to calldata changes to a different PR? We've already deployed all of these contracts to production and we need to figure out if it's even worth it to redeploy those at the moment.

I think even we can do it in this PR since it's only change with the `memory` to `calldata` and no logic change.
or do you want me to do it later in other PR?

PR #134, opened 2022-11-15
added receive() function to executor + 1 test case

danielb1986:


danielb1986:
> Are the console.log(1/2/3/4...) needed?

No I thought I had removed all of them. Sorry for that.
I sent another commit which removes them.

H3xept (approved):

PR #135, opened 2022-11-16
Add extra guards for gelato and reserve gas for recovery

ezynda3:


src/Periphery/Receiver.sol
```diff
@@ -132,7 +164,7 @@ contract Receiver is ILiFi, ReentrancyGuard, TransferrableOwnership {
         bool success;
 
         if (LibAsset.isNativeAsset(assetId)) {
-            try executor.swapAndCompleteBridgeTokens{ value: amount }(_transactionId, _swapData, assetId, receiver) {
+            try executor.swapAndCompleteBridgeTokens{ value: amount, gas: gasleft() - recoverGas }(_transactionId, _swapData, assetId, receiver) {
```
H3xept, line 167: This could lead to a negative calcualtion revert, check gasLeft() > recoverGas & if false execute the fallback early

H3xept (changes requested):

H3xept (approved):

PR #141, opened 2022-11-23
DeBridge integration

satoshi-lifi:


config/debridge.json
```diff
@@ -0,0 +1,44 @@
...
+    "UNWRAP_ETH": 0,
+    "REVERT_IF_EXTERNAL_FAIL": 1,
+    "PROXY_WITH_SENDER": 2,
+    "SEND_HASHED_DATA": 3,
+    "SEND_EXTERNAL_CALL_GAS_LIMIT": 4,
+    "MULTI_SEND": 5
+  },
+  "config": {
+    "hardhat": {
+      "chainId": 1,
```
danielb1986, line 12: Is this correct that hardhat has same chainID as mainnet?
Usually hardhat has like 31337 or so

deploy/035_deploy_debridge_facet.ts
```diff
@@ -0,0 +1,46 @@
+import { ethers } from 'hardhat'
```
danielb1986, line 1: If I understand correctly, hardhat is deprecated and we do not need a hardhat deploy script anymore.

script/UpdateDeBridgeFacet.s.sol
```diff
@@ -0,0 +1,38 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.17;
+
+import { UpdateScriptBase } from "./utils/UpdateScriptBase.sol";
+import { stdJson } from "forge-std/StdJson.sol";
+import "forge-std/console.sol";
```
danielb1986, line 6: not so important but could be removed

scripts/demoDeBridge.ts
```diff
@@ -0,0 +1,179 @@
+import { providers, Wallet, utils, constants, Contract } from 'ethers'
```
danielb1986, line 1: same as with deploy script. Not sure if we still need .ts files.

src/Facets/DeBridgeFacet.sol
```diff
@@ -0,0 +1,139 @@
...
+        bytes data;
+    }
+
+    /// @param permit deadline + signature for approving the spender by signature.
+    /// @param useAssetFee Use assets fee for pay protocol fix (work only for specials token)
+    /// @param nativeFee Native fee for the bridging when useAssetFee is false.
+    /// @param referralCode Referral code.
+    /// @param autoParams Structure that enables passing arbitrary messages and call data.
+    struct DeBridgeData {
+        bytes permit;
```
danielb1986, line 46: I think you could save some gas by rearranging the parameters (cause in the struct they will be tightly packed).
I think the uint256 parameter should be first and then the other...i dont know in which order exactly, would recommend to try...

src/Facets/DeBridgeFacet.sol
```diff
@@ -0,0 +1,139 @@
...
+    /// @notice Bridges tokens via DeBridge
+    /// @param _bridgeData the core information needed for bridging
+    /// @param _deBridgeData data specific to DeBridge
+    function startBridgeTokensViaDeBridge(ILiFi.BridgeData calldata _bridgeData, DeBridgeData calldata _deBridgeData)
+        external
+        payable
+        refundExcessNative(payable(msg.sender))
+        validateBridgeData(_bridgeData)
+        doesNotContainSourceSwaps(_bridgeData)
+        doesNotContainDestinationCalls(_bridgeData)
```
danielb1986, line 72: I dont understand why the doeNotContainDestinationCalls modifier applies here.....I thought deBridge supports message transfers? If they do, then this flag will be an issue as soon as we set bridgeData.hasDestinationCalls to true

src/Facets/DeBridgeFacet.sol
```diff
@@ -0,0 +1,139 @@
...
+        ILiFi.BridgeData memory _bridgeData,
+        LibSwap.SwapData[] calldata _swapData,
+        DeBridgeData memory _deBridgeData
+    )
+        external
+        payable
+        nonReentrant
+        refundExcessNative(payable(msg.sender))
+        containsSourceSwaps(_bridgeData)
+        doesNotContainDestinationCalls(_bridgeData)
```
danielb1986, line 93: same as in function above

test/solidity/Facets/DeBridgeFacet.t.sol
```diff
@@ -0,0 +1,242 @@
...
+    function addDex(address _dex) external {
+        LibAllowList.addAllowedContract(_dex);
+    }
+
+    function setFunctionApprovalBySignature(bytes4 _signature) external {
+        LibAllowList.addAllowedSelector(_signature);
+    }
+}
+
+contract DeBridgeFacetTest is DSTest, DiamondTest {
```
danielb1986, line 29: Could you maybe update your test file so that it inherits from TestbaseFacet.sol?
Please have a look at the test files of the other facets in order to understand how to do that. I am happy to support you in case you need any help

deploy/035_deploy_debridge_facet.ts
satoshi-lifi: yeah, but we still have the `deploy` folder, so I added it.
when we remove the folder, it will be removed together ;)

script/UpdateDeBridgeFacet.s.sol
satoshi-lifi: I think it should be removed in other PR since other scripts have it in them as well.
so we can create a new PR to remove this line from all scripts

scripts/demoDeBridge.ts
satoshi-lifi: it's a demo test script

test/solidity/Facets/DeBridgeFacet.t.sol
satoshi-lifi: updated it

script/UpdateDeBridgeFacet.s.sol
danielb1986: as you prefer

test/solidity/Facets/DeBridgeFacet.t.sol
```diff
@@ -76,167 +71,61 @@ contract DeBridgeFacetTest is DSTest, DiamondTest {
...
-
-        ILiFi.BridgeData memory bridgeData = validBridgeData;
-        bridgeData.sendingAssetId = address(0);
-        bridgeData.minAmount = 3e18;
-
-        vm.expectRevert(InvalidAmount.selector);
-        deBridgeFacet.startBridgeTokensViaDeBridge{ value: 2e18 }(bridgeData, validDeBridgeData);
-
-        vm.stopPrank();
+    function testBase_Revert_BridgeWithInvalidDestinationCallFlag() public override {
```
danielb1986, line 103: Even if the facet supports destination calls, this test should still pass. 
In Stargate and NXTP facet we have checks that make sure that if _bridgeData.hasDestinationCall is true then there must also be callData

We can discuss to not do these checks but in order to have all facets behave similar, I would recommend to add such a check and make sure the test passes


Please review the test in TestBaseFacet.sol

test/solidity/Facets/DeBridgeFacet.t.sol
```diff
@@ -76,167 +71,61 @@ contract DeBridgeFacetTest is DSTest, DiamondTest {
...
+        // facet supports destination call
     }
 
-    function testRevertToBridgeTokensWhenInformationMismatch() public {
-        vm.startPrank(USDC_HOLDER);
-
-        usdc.approve(address(deBridgeFacet), 10_000 * 10**usdc.decimals());
+    function testBase_CanBridgeTokens_fuzzed(uint256 amount) public override {
```
danielb1986, line 107: You could simplify this override similar to the one in stargate facet:

    function testBase_CanBridgeTokens_fuzzed(uint256 amount) public override {
        // fails otherwise with "slippage too high" from Stargate router contract
        vm.assume(amount > 100);
        super.testBase_CanBridgeTokens_fuzzed(amount);
    }

test/solidity/Facets/DeBridgeFacet.t.sol
satoshi-lifi: updated the DeBridge facet with destination call flag validation and updated the foundry test

test/solidity/Facets/DeBridgeFacet.t.sol
satoshi-lifi: updated the test

danielb1986 (approved):

PR #148, opened 2022-11-30
Update wormhole facet with init and config

satoshi-lifi:


ezynda3:
@satoshi-lifi 
Can you remove the following from the PR?
- `deploy/012_deploy_wormhole_facet.ts`
- `script/DeployArbitrumBridgeFacet.ts`
- `src/Facets/HopFacet.sol`
- `src/Facets/WithdrawFacet.sol`

Also please update this script to handle the setting of chainIds
- `script/UpdateWormholeFacet.sol`

PR #181, opened 2023-01-27
feat(insurance): add insurance fee collector

0xksure:
Insurance fees are collected when a user wants to bridge tokens.


0xksure:
Will be incorporated in the gasCollector.

PR #182, opened 2023-01-28
Begin implementing service fee collection

ezynda3:


src/Periphery/ServiceFeeCollector.sol
```diff
@@ -0,0 +1,144 @@
...
+            receiver,
+            feeAmount
+        );
+    }
+
+    /// @notice Withdraws fees
+    /// @param tokenAddress The address of the token to withdraw fees for
+    function withdrawFees(address tokenAddress) external onlyOwner {
+        uint256 balance = LibAsset.getOwnBalance(tokenAddress);
+        if (balance == 0) {
```
danielb1986, line 116: I dont think such a check is required in an admin-only function. If balance is 0, it will just fail but we can trust (ourselves) to know what we are doing and save deployment gas here

src/Periphery/ServiceFeeCollector.sol
```diff
@@ -0,0 +1,144 @@
...
+        // Prevent extra native token from being locked in the contract
+        if (remaining > 0) {
+            (bool success, ) = payable(msg.sender).call{ value: remaining }(
+                ""
+            );
+            if (!success) {
+                revert TransferFailure();
+            }
+        }
+        emit GasFeesCollected(LibAsset.NULL_ADDRESS, receiver, feeAmount);
```
danielb1986, line 71: I imagine it would be cheaper to use address(0) instead of using Library code for this, but I might be wrong

src/Periphery/ServiceFeeCollector.sol
```diff
@@ -0,0 +1,144 @@
...
+    /// @param tokenAddress The address of the token to collect
+    /// @param feeAmount The amount of fees to collect
+    /// @param receiver The address to send gas to on the destination chain
+    function collectTokenGasFees(
+        address tokenAddress,
+        uint256 feeAmount,
+        address receiver
+    ) external {
+        LibAsset.depositAsset(tokenAddress, feeAmount);
+        emit GasFeesCollected(tokenAddress, receiver, feeAmount);
```
danielb1986, line 50: Do we really have to emit an event every time we collect fees? 
Just challenging this because of the latest gas-saving discussions

test/solidity/Periphery/ServiceFeeCollector.t.sol
```diff
@@ -0,0 +1,179 @@
...
+        feeCollector.transferOwnership(newOwner);
+    }
+
+    function testFailOwnerCanConfirmPendingOwnershipTransfer() public {
+        address newOwner = address(0x1234567890123456789012345678901234567890);
+        feeCollector.transferOwnership(newOwner);
+        feeCollector.confirmOwnershipTransfer();
+    }
+
+    function testFailOwnershipTransferToSelf() public {
```
danielb1986, line 175: I would suggest that in cases where we know which error to expect, we should always catch the exact revert reason and make sure that the tests fail for the reason we expect (example in TestBase). I had situations before where an issue slipped through my tests because of that.

src/Periphery/ServiceFeeCollector.sol
ezynda3: This event is needed in order for our trusted Gas worker to know when to credit gas to someone.

test/solidity/Periphery/ServiceFeeCollector.t.sol
ezynda3: Good call.

src/Periphery/ServiceFeeCollector.sol
```diff
@@ -0,0 +1,144 @@
...
+/// @author LI.FI (https://li.fi)
+/// @notice Provides functionality for collecting service fees (gas/insurance)
+contract ServiceFeeCollector is TransferrableOwnership {
+    /// Errors ///
+    error TransferFailure();
+    error NotEnoughNativeForFees();
+
+    /// Events ///
+    event GasFeesCollected(
+        address indexed _token,
```
satoshi-lifi, line 17: nit: I think we can remove underscore of the parameters in events

src/Periphery/ServiceFeeCollector.sol
```diff
@@ -0,0 +1,144 @@
...
+
+    /// @notice Collects gas fees in native token
+    /// @param feeAmount The amount of native token to collect
+    /// @param receiver The address to send gas to on destination chain
+    function collectNativeGasFees(uint256 feeAmount, address receiver)
+        external
+        payable
+    {
+        if (msg.value < feeAmount) revert NotEnoughNativeForFees();
+        uint256 remaining = msg.value - (feeAmount);
```
satoshi-lifi, line 61: is there any reason that parentheses are used here?

PR #187, opened 2023-02-03
fix(amarok): use new parameters

maxklenk:


deployments/AmarokFacet.staging.json
```diff
@@ -0,0 +1,3 @@
+{
+  "": ""
+}
```
0xDEnYO, line 3: this file should not exist imho (dont even know how it got into existence)

0xDEnYO (approved):

PR #191, opened 2023-02-09
fixed script

0xDEnYO:


ezynda3 (commented):
Need to change `address(0)` to the facet address or it will fail.

PR #195, opened 2023-02-13
withdraw function added

0xDEnYO:


src/Periphery/RelayerCBridge.sol
```diff
@@ -418,6 +418,22 @@ contract RelayerCBridge is ILiFi, ReentrancyGuard, TransferrableOwnership {
         }
     }
 
+    /// @notice Sends remaining token to given receiver address (for refund cases)
+    /// @param assetId address of the token to be withdrawn
```
satoshi-lifi, line 422: nit: Can we make the first character uppercase to be more readability?
`/// @param assetId Address of the token to be withdrawn`

satoshi-lifi (approved):
LGTM

src/Periphery/RelayerCBridge.sol
0xDEnYO: done

PR #198, opened 2023-02-16
domainId will get passed as parameter to facet now

0xDEnYO:


src/Facets/AmarokFacet.sol
```diff
@@ -25,25 +25,23 @@ contract AmarokFacet is ILiFi, ReentrancyGuard, SwapperV2, Validatable {
 
     /// Errors ///
```
ezynda3, line 26: Maybe remove the headers if there is nothing underneath?

0xDEnYO: ok

PR #203, opened 2023-02-18
chore: deploy ServiceFeeCollector to all supported chains

maxklenk:
I have deployed the `ServiceFeeCollector` to all our networks and registered it in the Diamond.

- verification failed: Gnosis, OKX, aurora, boba
- verification key missing: celo, fuse, mumbai, nova, evmos, velas, harmony


script/DeployServiceFeeCollector.s.sol
```diff
@@ -13,7 +13,10 @@ contract DeployScript is DeployScriptBase {
     {
         vm.startBroadcast(deployerPrivateKey);
 
-        constructorArgs = abi.encode(deployerAddress);
+        address owner = address(0x8932eb23BAD9bDdB5cF81426F78279A53c6c3b71);
+        // or for testing:
+        // address owner = deployerAddress
+        constructorArgs = abi.encode(owner);
```
maxklenk, line 19: This is not so clean, perhaps you know a better way to specify that address in env? I wanted to avoid having to move access for every contract manually after deployment.

networks
```diff
@@ -17,7 +17,7 @@ harmony
 evmos
 aurora
 boba
+nova
+
 goerli
-sepolia
```
maxklenk, line 21: I removed sepolia from the list for now, as we don't have our contract deployed there.

script/DeployServiceFeeCollector.s.sol
0xDEnYO: How about you use the foundry cheatcode to derive the address from your private key? That way the script should work for both TesterWallet and DeployerWallet

https://book.getfoundry.sh/cheatcodes/addr

0xDEnYO (approved):

script/DeployServiceFeeCollector.s.sol
0xDEnYO: Also I noticed that you want to merge this branch into master directly. 
Does our management of the various branches also include that anything that gets merged into master directly will also be merged into (staging and) dev?
It should imho....otherwise we will have diverging codebases...

maxklenk: Accessing the deployer wallet is no problem, that is what was done before using the `deployerAddress` address. I want to use a third address which should later collect the fees.

maxklenk: I created this PR agains master because I added the contracts to the production diamond. We have to manually merge master into the other branches afterwards. I chose this direction as everything from master should be good for the other environments, but we don't want to merge everything from dev/staging to master right away.

PR #207, opened 2023-02-21
feat: create Allbridge facet

0xksure:


docs/AllBridgeFacet.md
```diff
@@ -0,0 +1,82 @@
...
+## Bridge Specific Parameters
+
+Some of the methods listed above take a variable labeled `_allBridgeData`.
+
+This data is specific to allBridgefacet and is represented as the following struct type:
+
+```solidity
+/// @notice The struct for the AllBridge data.
+/// @param fees The amount of token to pay the messenger and the bridge
+/// @param recipient The address of the token receiver after bridging.
```
0xDEnYO, line 43: recipient and destinationChainId are part of the bridgeData...what is the reasoning for having duplicates in the AllBridgeData?

docs/AllBridgeFacet.md
```diff
@@ -0,0 +1,82 @@
...
+/// @param destinationChainId The destination chain id.
+/// @param receiveToken The token to receive on the destination chain.
+/// @param nonce A random nonce to associate with the tx.
```
0xDEnYO, line 46: Maybe "unique" instead of random?

docs/AllBridgeFacet.md
```diff
@@ -0,0 +1,82 @@
...
+
+Swapping is performed by a swap specific library that expects an array of calldata to can be run on variaous DEXs (i.e. Uniswap) to make one or multiple swaps before performing another action.
+
+The swap library can be found [here](../src/Libraries/LibSwap.sol).
+
+## LiFi Data
+
+Most of the methods accept a `ILiFiBridgeData _bridgeData` parameter.
+
+In the AllBridge contract call the fields `minAmount` and `sendingAssetId` are used for the transfer amount and the asset to be sent. Since the AllBridge bridge does not support native token briding (it's mainly a stablecoin bridge) the methods will failed if native assets is tried bridged.
```
0xDEnYO, line 76: typos: 
"briding" => "bridging"
"will failed if native assets is tried bridged" => "will fail if native assets are tried to be bridged"

script/UpgradeAllBridge.s.sol
```diff
@@ -0,0 +1,34 @@
...
+
+contract DeployScript is UpdateScriptBase {
+    using stdJson for string;
+
+    function run() public returns (address[] memory facets) {
+        address facet = json.readAddress(".AllBridgeFacet");
+
+        vm.startBroadcast(deployerPrivateKey);
+
+        // Across
```
0xDEnYO, line 17: Across => AllBridge (or remove entirely since the comment doesnt really say anything

src/Facets/AllBridgeFacet.sol
```diff
@@ -0,0 +1,148 @@
...
+    /// @param fees The amount of token to pay the messenger and the bridge
+    /// @param recipient The address of the token receiver after bridging.
+    /// @param destinationChainId The destination chain id.
+    /// @param receiveToken The token to receive on the destination chain.
+    /// @param nonce A random nonce to associate with the tx.
+    /// @param messenger The messenger protocol enum
+    struct AllBridgeData {
+        uint256 fees;
+        bytes32 recipient;
+        uint8 destinationChainId;
```
0xDEnYO, line 52: as mentioned before, not sure why recipient and destinationChainId are part of the AllBridgeData

docs/AllBridgeFacet.md
```diff
@@ -0,0 +1,82 @@
...
+
+```mermaid
+graph LR;
+    D{LiFiDiamond}-- DELEGATECALL -->C(AllBridgeFacet);
+    C(AllBridgeFacet) -- CALL --> C(AllBridge Core)
+```
+
+## Public Methods
+
+- `function startBridgeTokensViaBridgeFacet(BridgeData calldata _bridgeData, AllBridgeData calldata _allBridgeData)`
```
0xDEnYO, line 29: The method name does not match with the facet

here: startBridgeTokensViaBridgeFacet()
facet: startBridgeTokensViaAllBridge() << correct

src/Facets/AllBridgeFacet.sol
```diff
@@ -0,0 +1,148 @@
...
+    /// @notice Initializes the AllBridge contract
+    /// @param _allBridge The address of the AllBridge contract
+    constructor(IAllBridge _allBridge) {
+        allBridge = _allBridge;
+    }
+
+    /// @notice Bridge tokens to another chain via AllBridge
+    /// @param _bridgeData The bridge data struct
+    function startBridgeTokensViaAllBridge(
+        ILiFi.BridgeData memory _bridgeData,
```
0xDEnYO, line 67: _bridgeData can be "calldata" instead of memory to save gas

src/Facets/AllBridgeFacet.sol
```diff
@@ -0,0 +1,148 @@
...
+    /// @param a The address to convert
+    function _toBytes(address a) internal pure returns (bytes memory) {
+        return abi.encodePacked(a);
+    }
+
+    /// @notice Bridge tokens to another chain via AllBridge
+    /// @param _bridgeData The bridge data struct
+    /// @param _allBridgeData The allBridge data struct for AllBridge specicific data
+    function _startBridge(
+        ILiFi.BridgeData memory _bridgeData,
```
0xDEnYO, line 121: same as above, can be calldata imho

src/Facets/AllBridgeFacet.sol
```diff
@@ -0,0 +1,148 @@
...
+        AllBridgeData calldata _allBridgeData
+    ) internal {
+        bool isNative = _bridgeData.sendingAssetId == LibAsset.NATIVE_ASSETID;
```
0xDEnYO, line 124: doesnt really make a difference but just for convention...LibAsset has a function called "isNativeAsset(_bridgeData.sendingAssetId)" that we would normally use in such case

And since the variable "isNative" is only used once, I would suggest to simplify the code and put the "isNativeAsset" function directly in the if condition

src/Facets/AllBridgeFacet.sol
```diff
@@ -0,0 +1,148 @@
...
+            _bridgeData.minAmount,
+            _swapData,
+            payable(msg.sender)
+        );
+        _startBridge(_bridgeData, _allBridgeData);
+    }
+
+    /// @notice converts an address to bytes
+    /// @param a The address to convert
+    function _toBytes(address a) internal pure returns (bytes memory) {
```
0xDEnYO, line 113: Is this function still used anywhere? Seems outdated and could be removed

test/solidity/Facets/AllBridgeFacet.t.sol
```diff
@@ -0,0 +1,116 @@
...
+        LibAllowList.addAllowedContract(_dex);
+    }
+
+    function setFunctionApprovalBySignature(bytes4 _signature) external {
+        LibAllowList.addAllowedSelector(_signature);
+    }
+}
+
+contract AllBridgeFacetTest is TestBaseFacet {
+    // These values are for Optimism_Kovan
```
0xDEnYO, line 22: is this comment accurate? Seems like they are mainnet addresses

test/solidity/Facets/AllBridgeFacet.t.sol
```diff
@@ -0,0 +1,116 @@
...
+        bridgeData.destinationChainId = 137;
+
+        uint256 testPolygonMessengerFees = 69857330070000;
+        uint256 testPolygonBridgeFees = 28764782970000;
+        uint256 fees = testPolygonBridgeFees + testPolygonMessengerFees;
+        // produce valid AllBridgeData
+        validAllBridgeData = AllBridgeFacet.AllBridgeData({
+            fees: fees,
+            recipient: 0x00000000000000000000000012561cc3ea2a60aa158b0421010859a983bf3c96,
+            destinationChainId: 5,
```
0xDEnYO, line 74: Why does this differ from the destinationChainId in the bridgeData?

test/solidity/Facets/AllBridgeFacet.t.sol
```diff
@@ -0,0 +1,116 @@
...
+            messenger: MessengerProtocol.Allbridge
+        });
+
+        ERC20(ADDRESS_USDC).approve(ALLBRIDGE_POOL, type(uint256).max);
+    }
+
+    function initiateBridgeTxWithFacet(bool isNative) internal override {
+        if (isNative) {
+            allBridgeFacet.startBridgeTokensViaAllBridge{
+                value: bridgeData.minAmount + validAllBridgeData.fees
```
0xDEnYO, line 86: just FYI:

the testbase contract has a variable "addToMessageValue" that you can initialize with the fees amount in the init function.
Your bridge does not support native bridging so it's not a problem here but one testcase in the TestBaseFacet.sol contract uses that variable

docs/AllBridgeFacet.md
0xksure: Because `BridgeData.destinationChainId` is uin256 and `AllBridgeData.destinationChainId` is uint8. Does it makes sense to just cast the `BridgeData.destinationChainId` to Uint8? Also, the chainIds are completely different between AllBridge and LiFi

docs/AllBridgeFacet.md
0xksure: Yes!

src/Facets/AllBridgeFacet.sol
0xksure: See ☝️ for `destinationChainId`. Recipient is bytes32 in AllBridgeData because it supports Solana which uses 32bytes addresses. While `LiFi.BridgeData` uses addresses which is 20bytes. So I can't cast upwards

src/Facets/AllBridgeFacet.sol
0xksure: I remember vaguely that the reason why `BridgeData` is memory is because it mutated in `depositAsset`. This is similar to AcrossFacet which also requires `BridgeData` to be in memory.

src/Facets/AllBridgeFacet.sol
0xksure: 👍

src/Facets/AllBridgeFacet.sol
0xksure: Removed

docs/AllBridgeFacet.md
ezynda3: I would keep them separate because they are indeed different.

test/solidity/Facets/AllBridgeFacet.t.sol
0xksure: no they are not, just sloppy copy paste

test/solidity/Facets/AllBridgeFacet.t.sol
0xksure: It's of type uint8. Is it cheaper to cast the `LiFi.BridgeData.DestinationChainId` to a uint8 than passing an extra field in the struct?

test/solidity/Facets/AllBridgeFacet.t.sol
ezynda3: As far as I'm aware, Allbridge does support native tokens so those tests should be used.

0xksure: Thank you, started using `addToMessageValue` instead

docs/AllBridgeFacet.md
0xksure: updated

script/UpgradeAllBridge.s.sol
0xksure: rm

docs/AllBridgeFacet.md
0xksure: 👍

docs/AllBridgeFacet.md
0xDEnYO: OK, makes sense if they differ. I would suggest to call the variable "allBridgeDestinationChain" or so to make it clear that this is a custom chainId and not the same as standard chainIds

test/solidity/Facets/AllBridgeFacet.t.sol
```diff
@@ -0,0 +1,115 @@
...
+            }(bridgeData, validAllBridgeData);
+        } else {
+            allBridgeFacet.startBridgeTokensViaAllBridge{
+                value: addToMessageValue
+            }(bridgeData, validAllBridgeData);
+        }
+    }
+
+    function testBase_CanBridgeNativeTokens() public override {
+        // facet does not support bridging of native assets
```
0xDEnYO, line 95: According to Ed's comment, the bridge supports native bridging so the tests should be activated here

0xDEnYO (changes requested):

test/solidity/Facets/AllBridgeFacet.t.sol
0xksure: I don't see that they support native tokens here https://docs-core.allbridge.io/product/how-does-allbridge-core-work/allbridge-core-contracts

docs/AllBridgeFacet.md
0xksure: Isn't it clear from the parent struct?

test/solidity/Facets/AllBridgeFacet.t.sol
0xksure: It says here https://docs-core.allbridge.io/ `Allbridge Core is a cross-chain swap built specifically for dollar-pegged tokens (stablecoins)`

test/solidity/Facets/AllBridgeFacet.t.sol
0xksure: Check out https://docs-core.allbridge.io/ it says `Allbridge Core is a cross-chain swap built specifically for dollar-pegged tokens (stablecoins). Operates without wrapping tokens and works by keeping liquidity pools for the supported tokens on each blockchain. Incoming tokens go into the pool, and the value of those tokens is sent via the messaging protocol to another chain, where it is converted back to the tokens to be paid to the user.`

docs/AllBridgeFacet.md
0xDEnYO: It's just a suggestion...

if it was super clear, I would not have stumbled upon this in the first place ;)

0xksure: haha true, sure, can update it

src/Facets/AllBridgeFacet.sol
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    /// @notice Bridge tokens to another chain via AllBridge
+    /// @param _bridgeData The bridge data struct
+    /// @param _allBridgeData The allBridge data struct for AllBridge specicific data
+    function _startBridge(
+        ILiFi.BridgeData memory _bridgeData,
+        AllBridgeData calldata _allBridgeData
+    ) internal {
+        if (!LibAsset.isNativeAsset(_bridgeData.sendingAssetId)) {
```
ezynda3, line 118: This can be removed since we don't handle native assets.

PR #210, opened 2023-02-23
Optimized Hop config script

ezynda3:
fix typos and redeploy to staging on Polygon

### DIFF
--- a/config/hop.json
+++ b/config/hop.json
@@ -219,15 +219,21 @@
     "chainId": 5,
     "tokens": [
       {
+        "name": "ETH",
+        "bridge": "0xC8A4FB931e8D77df8497790381CA7d228E68a41b",
+        "token": "0x0000000000000000000000000000000000000000",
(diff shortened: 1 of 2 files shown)


satoshi-lifi (approved):
LGTM

config/hop.json
```diff
@@ -193,35 +246,5 @@
         "ammWrapper": "0x6329d42Cc5517ba1d913303E019a556d84B5a05d"
       }
     }
```
maxklenk, line 248: Can you please also update the testnet configs based on this file:
https://github.com/hop-protocol/hop/blob/develop/packages/core/src/addresses/goerli.ts

maxklenk (changes requested):

config/hop.json
ezynda3: @maxklenk updated

maxklenk (approved):

PR #211, opened 2023-02-23
make swap functions of optimizedHop payable

maxklenk:
The swap functions have to be able to accept any token, including native token that are then swapped into an ERC20.

- [x] Add L1 Test: swap ETH in USDC and bridge with hopOptimized
- [x] Add L2 Test: swap MATIC in USDC and bridge with hopOptimized


maxklenk (commented):
Perfect, looks good to me.
We should deploy it and replace the current optimized hop facets with it.

PR #214, opened 2023-02-27
Clear up facet contracts

satoshi-lifi:


src/Facets/AmarokFacet.sol
```diff
@@ -42,10 +37,8 @@ contract AmarokFacet is ILiFi, ReentrancyGuard, SwapperV2, Validatable {
 
     /// @notice Initialize the contract.
     /// @param _connextHandler The contract address of the connext handler on the source chain.
-    /// @param _srcChainDomain The domain of source chain.
-    constructor(IConnextHandler _connextHandler, uint32 _srcChainDomain) {
+    constructor(IConnextHandler _connextHandler) {
         connextHandler = _connextHandler;
-        srcChainDomain = _srcChainDomain;
```
0xDEnYO, line 48: if you remove this (which is correct) then you also need to adjust the deploy script

0xDEnYO (changes requested):

0xDEnYO (approved):

src/Facets/AmarokFacet.sol
satoshi-lifi: you are right.
I updated the deployment script of Amarok facet.
https://github.com/lifinance/contracts/pull/214/commits/31a195a828391171d0fb240aa0a39f700d6256df

PR #216, opened 2023-02-27
Implement Squid Bridge

ezynda3:


src/Facets/SquidFacet.sol
```diff
@@ -0,0 +1,133 @@
...
+    enum SquidCallType {
+        BridgeCall,
+        CallBridge,
+        CallBridgeCall
+    }
+
+    struct SquidData {
+        SquidCallType callType;
+        bytes callData;
+    }
```
maxklenk, line 27: Instead of passing callData I would use a type similar to `callBridgeCall`, but pass token, amount, destinationAddress from the bridge data. We want to be able to use as much of the bridgeData as possible, to be able to rely on that that, e.g. to validate what will happen or to adjust the amount without knowing the full callData.

src/Facets/SquidFacet.sol
```diff
@@ -0,0 +1,133 @@
...
+                    sendingAssetId,
+                    address(squidRouter),
+                    _bridgeData.minAmount
+                );
+            }
+        }
+
+        (bool success, bytes memory res) = address(squidRouter).call{
+            value: msgValue
+        }(_squidData.callData);
```
maxklenk, line 125: If the _bridgeData.minAmount for an ERC20 was adjusted we do not pass it here yet.

src/Facets/SquidFacet.sol
```diff
@@ -0,0 +1,133 @@
...
+        ILiFi.BridgeData memory _bridgeData,
+        SquidData memory _squidData
+    )
+        external
+        payable
+        nonReentrant
+        refundExcessNative(payable(msg.sender))
+        validateBridgeData(_bridgeData)
+        doesNotContainSourceSwaps(_bridgeData)
+        doesNotContainDestinationCalls(_bridgeData)
```
maxklenk, line 54: This is not really true, as we may do both in squid, so the check should be based on that too, comparing the `_bridgeData` flag with the parameters passed in `_squidData`.

src/Facets/SquidFacet.sol
maxklenk:
```
    function bridgeCall(
        string calldata destinationChain,
        string calldata bridgedTokenSymbol,
        uint256 amount, <= _bridgeData.minAmount
        ISquidMulticall.Call[] calldata calls,
        address refundRecipient, <= _bridgeData.recipient
        bool forecallEnabled
    ) external payable;

    function callBridge(
        address token, <= _bridgeData.sendingAssetId
        uint256 amount, <= _bridgeData.minAmount
        string calldata destinationChain,
        string calldata destinationAddress, <= _bridgeData.recipient
        string calldata bridgedTokenSymbol,
        ISquidMulticall.Call[] calldata calls
    ) external payable;

    function callBridgeCall(
        address token, <= _bridgeData.sendingAssetId
        uint256 amount, <= _bridgeData.minAmount
        string calldata destinationChain,
        string calldata bridgedTokenSymbol,
        ISquidMulticall.Call[] calldata sourceCalls,
        ISquidMulticall.Call[] calldata destinationCalls,
        address refundRecipient, <= _bridgeData.recipient
        bool forecallEnabled
    ) external payable;
```

maxklenk (changes requested):

src/Facets/SquidFacet.sol
ezynda3: The problem is it is very difficult to rebuild the Call[] array from the information passed by the API alone.

docs/SquidFacet.md
```diff
@@ -0,0 +1,79 @@
...
+
+## Bridge Specific Parameters
+
+Some of the methods listed above take a variable labeled `_squidData`.
+
+This data is specific to Squid and is represented as the following struct type:
+
+```solidity
+/// @notice Contains the data needed for bridging via Squid squidRouter
+/// @param RouteType the type of route to use
```
satoshi-lifi, line 35: nit: It might be better if the description starts Uppercase for readability

src/Facets/SquidFacet.sol
```diff
@@ -0,0 +1,178 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.17;
+
+import { ILiFi } from "../Interfaces/ILiFi.sol";
+import { ISquidRouter } from "../Interfaces/ISquidRouter.sol";
+import { ISquidMulticall } from "../Interfaces/ISquidMulticall.sol";
+import { LibAsset, IERC20 } from "../Libraries/LibAsset.sol";
+import { LibSwap } from "../Libraries/LibSwap.sol";
+import { LibUtil } from "../Libraries/LibUtil.sol";
```
satoshi-lifi, line 9: nit: it's not used

src/Facets/SquidFacet.sol
```diff
@@ -0,0 +1,178 @@
...
+    /// Types ///
+
+    enum RouteType {
+        BridgeCall,
+        CallBridge,
+        CallBridgeCall
+    }
+
+    /// @dev Contains the data needed for bridging via Squid squidRouter
+    /// @param RouteType the type of route to use
```
satoshi-lifi, line 29: nit: It might be better if the description starts Uppercase for readability

src/Facets/SquidFacet.sol
```diff
@@ -0,0 +1,178 @@
...
+
+    /// Constructor ///
+    constructor(ISquidRouter _squidRouter) {
+        squidRouter = _squidRouter;
+    }
+
+    /// External Methods ///
+
+    /// @notice Bridges tokens via Squid Router
+    /// @param _bridgeData the core information needed for bridging
```
satoshi-lifi, line 57: nit: same as above

src/Facets/SquidFacet.sol
```diff
@@ -0,0 +1,178 @@
...
+        LibAsset.depositAsset(
+            _bridgeData.sendingAssetId,
+            _bridgeData.minAmount
+        );
+
+        _startBridge(_bridgeData, _squidData);
+    }
+
+    /// @notice Swaps and bridges tokens via Squid Router
+    /// @param _bridgeData the core information needed for bridging
```
satoshi-lifi, line 91: nit: same as above

src/Facets/SquidFacet.sol
```diff
@@ -0,0 +1,178 @@
...
+            _swapData,
+            payable(msg.sender)
+        );
+        _startBridge(_bridgeData, _squidData);
+    }
+
+    /// Internal Methods ///
+
+    /// @dev Contains the business logic for the bridge via Squid Router
+    /// @param _bridgeData the core information needed for bridging
```
satoshi-lifi, line 125: nit: same as above

src/Facets/SquidFacet.sol
```diff
@@ -0,0 +1,178 @@
...
+    /// @param _squidData data specific to Squid Router
+    function _startBridge(
+        ILiFi.BridgeData memory _bridgeData,
+        SquidData memory _squidData
+    ) internal {
+        IERC20 sendingAssetId = IERC20(_bridgeData.sendingAssetId);
+        bool isNative = LibAsset.isNativeAsset(address(sendingAssetId));
```
satoshi-lifi, line 132: nit: This variable is used once. So I think we can use the comparison directly.

src/Interfaces/ISquidMulticall.sol
```diff
@@ -0,0 +1,24 @@
...
+        address target;
+        uint256 value;
+        bytes callData;
+        bytes payload;
+    }
+
+    error AlreadyRunning();
+    error CallFailed(uint256 callPosition, bytes reason);
+
+    function run(Call[] calldata calls) external payable;
```
satoshi-lifi, line 23: nit: I think these are not used anywhere

src/Interfaces/ISquidRouter.sol
```diff
@@ -0,0 +1,42 @@
...
+
+import { ISquidMulticall } from "./ISquidMulticall.sol";
+
+interface ISquidRouter {
+    event CrossMulticallExecuted(bytes32 indexed payloadHash);
+    event CrossMulticallFailed(
+        bytes32 indexed payloadHash,
+        bytes reason,
+        address indexed refundRecipient
+    );
```
satoshi-lifi, line 12: nit: It seems these are not used

test/solidity/Facets/SquidFacet.t.sol
```diff
@@ -0,0 +1,171 @@
...
+        LibAllowList.addAllowedContract(_dex);
+    }
+
+    function setFunctionApprovalBySignature(bytes4 _signature) external {
+        LibAllowList.addAllowedSelector(_signature);
+    }
+}
+
+contract SquidFacetTest is TestBaseFacet {
+    // These values are for Optimism_Kovan
```
satoshi-lifi, line 26: Is this test for optimism kovan?

PR #220, opened 2023-03-01
added onERC721Received to Executor

0xDEnYO:


src/Periphery/Executor.sol
```diff
@@ -8,6 +8,7 @@ import { LibAsset } from "../Libraries/LibAsset.sol";
 import { ILiFi } from "../Interfaces/ILiFi.sol";
 import { IERC20Proxy } from "../Interfaces/IERC20Proxy.sol";
 import { TransferrableOwnership } from "../Helpers/TransferrableOwnership.sol";
+import { IERC721Receiver } from "@openzeppelin/contracts/token/ERC721/IERC721Receiver.sol";
```
ezynda3, line 11: Is there any reason to actually import this?

src/Periphery/Executor.sol
```diff
@@ -261,6 +262,20 @@ contract Executor is ILiFi, ReentrancyGuard, TransferrableOwnership {
         return balances;
     }
 
+    /// @dev Will be called as part of an ERC721 token transfer to this contract (see IERC721Receiver for more details)
+    /// @param operator The address of the tx initiator
+    /// @param from The address from which the ERC721 token was transferred to this contract
+    /// @param tokenId The ID of the token that was transferred
+    /// @return uint256[] Array of token balances.
```
satoshi-lifi, line 269: return value is not balances(uint256[]) but selector(bytes4)

src/Periphery/Executor.sol
0xDEnYO: for explicity purpose and the comments in the file I would tend to include it. But I forgot to let the Executor inherit from it, which I changed now.

src/Periphery/Executor.sol
0xDEnYO: updated

PR #221, opened 2023-03-01
Create Stargate update script

ezynda3:


satoshi-lifi (approved):
LGTM

scripts/update-sg-config.sh
```diff
@@ -0,0 +1,63 @@
...
+
+  if [ $attempts -eq 11 ]; then
+    echo "Failed to execute $SCRIPT"
+    exit 1
+  fi
+
+  echo "$SCRIPT successfully executed on network $NETWORK"
+}
+
+checkFailure() {
```
0xDEnYO, line 55: could be removed since it's unused

0xDEnYO (approved):

scripts/update-sg-config.sh
ezynda3: Done

PR #234, opened 2023-03-13
Synapse Bridge integration

satoshi-lifi:


src/Facets/SynapseBridgeFacet.sol
```diff
@@ -0,0 +1,126 @@
...
+/// @author LI.FI (https://li.fi)
+/// @notice Provides functionality for bridging through SynapseBridge
+contract SynapseBridgeFacet is ILiFi, ReentrancyGuard, SwapperV2, Validatable {
+    /// Storage ///
+
+    address internal constant NETH_ADDRESS =
+        0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE;
+
+    /// @notice The contract address of the SynapseRouter on the source chain.
+    ISynapseRouter private immutable synapseRouter;
```
ezynda3, line 21: We can make this public just so it's easier to verify after deployment

ezynda3 (commented):
Please add a simple doc page and update `docs/README.md`

src/Facets/SynapseBridgeFacet.sol
satoshi-lifi: updated it.
https://github.com/lifinance/contracts/pull/234/commits/d4f9c11982b08173f86a9d9a4b04ebb67df9b7ef

src/Facets/SynapseBridgeFacet.sol
```diff
@@ -0,0 +1,126 @@
...
+    /// Internal Methods ///
+
+    /// @dev Contains the business logic for the bridge via Synapse Bridge
+    /// @param _bridgeData the core information needed for bridging
+    /// @param _synapseData data specific to Synapse Bridge
+    function _startBridge(
+        ILiFi.BridgeData memory _bridgeData,
+        SynapseData calldata _synapseData
+    ) internal {
+        uint256 nativeAssetAmount;
```
0xDEnYO, line 101: I dont really see the added benefit by this variable. Could also just use _bridgeData.minAmount directly if the value is not manipulated.

satoshi-lifi: you mean we can check if the sending asset is native asset again when we call `bridge` function?

0xDEnYO: ahh... .sorry, I misread the code. Forget my comment, please :)

PR #237, opened 2023-03-14
fix: stargate pool info passed in calldata

H3xept:
as opposed to reading from storage


0xDEnYO (approved):

ezynda3:
Looks good but please remake this PR into dev

PR #255, opened 2023-03-22
#73: match swap result with allowed token

maxklenk:


satoshi-lifi:
Also it resolve below issue
LF-2637 [#82 Redundant checks in CircleBridgeFacet](https://github.com/spearbit-audits/review-lifi-retainer1/issues/82)

PR #283, opened 2023-03-25
feat: Add HopFacetPacked

maxklenk:


src/Facets/HopFacetPacked.sol
```diff
@@ -0,0 +1,435 @@
...
+        string calldata integrator,
+        address receiver,
+        uint256 destinationChainId,
+        uint256 bonderFee,
+        uint256 amountOutMin,
+        uint256 destinationAmountOutMin,
+        address hopBridge
+    ) external pure returns (bytes memory) {
+        return bytes.concat(
+            abi.encodeWithSignature("startBridgeTokensViaHopL2NativePacked()"),
```
gpersoon, line 85: A more robust version as it doesn't rely on strings:
`abi.encodeCall(HopFacetPacked.startBridgeTokensViaHopL2NativePacked,());`

src/Facets/HopFacetPacked.sol
```diff
@@ -0,0 +1,435 @@
...
+            integrator: string(abi.encodePacked(getCalldataValue(12, 16))), // bytes16 > string
+            receiver: address(uint160(getCalldataValue(28, 20))), // bytes20 > address
+            destinationChainId: getCalldataValue(48, 4), // bytes4(uint32) > uint256
+            bonderFee: getCalldataValue(52, 16), // bytes16(uint128) > uint256
+            amountOutMin: getCalldataValue(68, 16), // bytes16(uint128) > uint256
+            destinationAmountOutMin: getCalldataValue(84, 16), // bytes16(uint128) > uint256
+            hopBridge: address(uint160(getCalldataValue(100, 20))) // bytes20 > address
+            // => total calldata length required: 120
+        });
+    }
```
gpersoon, line 32: An alternative way to decode using only solidity:
```solidity
(bytes8 transactionId,
bytes16 integrator,
address receiver,
bytes4 destinationChainId,
bytes16 bonderFee,
bytes16 amountOutMin,
bytes16 destinationAmountOutMin,
address hopBridge) 
= abi.decode(msg.data[4:], 
(bytes8,
bytes16,
address,
bytes4,
bytes16,
bytes16,
bytes16,
address));
```

src/Facets/HopFacetPacked.sol
```diff
@@ -0,0 +1,435 @@
...
+
+    /// Internal Methods ///
+
+    /// @notice Validate raw callData length
+    /// @param length Total required callData length
+    function checkCalldataLength(uint length) private pure {
+            uint _calldatasize;
+            assembly {
+                _calldatasize := calldatasize()
+            }
```
gpersoon, line 263: Could also use `msg.data.length`

docs/HopFacetPacked.md
```diff
@@ -0,0 +1,91 @@
...
+- `sendingAssetId`: Address of the source asset to bridge (only needed when passing ERC20 tokens)
+- `amount`: Amount of the source asset to bridge (only needed when passing ERC20 tokens)
+
+Additional parameters are passed to track transfers on chain:
+- `transactionId`: Custom transaction ID for tracking
+- `integrator`: LI.FI partner name
+
+## Packed Parameters
+
+In order to pass as little data as possible when calling the `*Packed` methods, the parameters are restricted in length and concatinated manually to avoid any additional padding. The contract expects them to have those fixed lengths to be able to extract data from the passed callData. When generating the parameter the backend has to check that the values fit into the restricted length to ensure a lossless compression.
```
gpersoon, line 50: typo: `concatinated -> concatenated`

src/Facets/HopFacetPacked.sol
```diff
@@ -0,0 +1,435 @@
...
+            abi.encodeWithSignature("startBridgeTokensViaHopL2NativePacked()"),
+            bytes8(transactionId),
+            bytes16(bytes(integrator)),
+            bytes20(receiver),
+            bytes4(uint32(destinationChainId)),
+            bytes16(uint128(bonderFee)),
+            bytes16(uint128(amountOutMin)),
+            bytes16(uint128(destinationAmountOutMin)),
+            bytes20(hopBridge)
+        );
```
gpersoon, line 94: Perhaps check here the values fit within the destination bytes,
When this call is done off chain gas cost wouldn't be an issue

src/Facets/HopFacetPacked.sol
Jonah246: @maxklenk We can avoid low-level assembly with this succinct method.

I'm okay with both approaches. Solidity would revert if there are extra call data to decode.
Use `abi.decode(msg.data[4:120]` if we want to attach extra information in the calldata. https://github.com/spearbit-audits/review-lifi-retainer1/issues/45#issuecomment-1485414924

gpersoon: Doesn't seem to work after all

src/Facets/HopFacetPacked.sol
maxklenk: Good hint, it is even easier, we can just access the `selector` directly:
`HopFacetPacked.startBridgeTokensViaHopL2NativePacked.selector`

src/Facets/HopFacetPacked.sol
maxklenk: @gpersoon it would be great to be able to decode in solidity, but I was not able to get the suggested `abi.decode` function to work. As `abi.decode` is the counterpart to `abi.encode` which encodes all values in `bytes32` length, I assume the the decoding also expects that the data is formatted that way.

src/Facets/CBridgeFacetPacked.sol
```diff
@@ -29,15 +29,16 @@ contract CBridgeFacetPacked is ILiFi {
     /// No params, all data will be extracted from manually encoded callData
     function startBridgeTokensViaCBridgeNativePacked(
     ) external payable {
-        checkCalldataLength(60);
+        require(msg.data.length >= 60, "CallData length smaler than required.");
```
gpersoon, line 32: typo `smaler` -> `smaller`

docs/HopFacetPacked.md
```diff
@@ -0,0 +1,91 @@
...
+    bytes16(bytes(INTEGRATOR_STRING)), // integrator
+    bytes20(RECEIVER_ADDRESS), // receiver
+    bytes16(uint128(amountBonderFeeNative)), // bonderFee
+    bytes16(uint128(amountOutMinNative)), // amountOutMin
+    bytes2(uint16(137)), // destinationChainId
+    bytes16(uint128(amountOutMinNative)), // destinationAmountOutMin
+    bytes20(HOP_NATIVE_BRIDGE) // hopBridge
+);
+bytes memory packedNative = bytes.concat(
+    abi.encodeWithSignature("startBridgeTokensViaHopL2NativePacked()"),
```
gpersoon, line 79: could use the `selector` version here too

docs/HopFacetPacked.md
```diff
@@ -0,0 +1,91 @@
...
+    packedNativeParams
+);
+
+// execute call
+(bool success, ) = address(diamond).call{value: amountNative}(packedNative);
+```
+
+## Getting Sample Calls to interact with the Facet
+
+In order to interact with this optimized facet is is best to only use requests directly returned by the LI.FI API to ensure the packed parameters are formatted right.
```
gpersoon, line 89: typo `is is` -> `it is`

PR #305, opened 2023-04-07
fix: amarok should reserve native for fees when swapping from native

H3xept:
We should reserve the fee whenever we swap from native, otherwise funds are sent back to the user.


ezynda3 (changes requested):
Please update the version on line 15. Should be `1.0.1` in this case.

PR #313, opened 2023-04-14
Handle facet upgrades

ezynda3:


script/UpgradeAllBridge.s.sol
0xDEnYO, line 1: why was this script deleted (and the deploy file was kept)? Erroneous delete?

0xDEnYO (approved):

script/UpgradeAllBridge.s.sol
ezynda3: The file was already nenamed but the old file was kept around somehow. `UpdateAllBridge` still exists.

PR #314, opened 2023-04-16
fix(hop): do not pass destinationDeadline to L1 (HopFacetPacked v1.0.0 > v1.0.1)

maxklenk:


src/Facets/HopFacetPacked.sol
```diff
@@ -367,7 +369,7 @@ contract HopFacetPacked is ILiFi {
             amountOutMin,
             deadline,
             destinationAmountOutMin,
-            deadline
+            toL1 ? 0 : deadline
```
0xDEnYO, line 372: I think you could probably save some gas by removing the bool variable. It is only used once, therefore I do not see a need to store the result of "destinationChainId == 1 || destinationChainId == 5" in a variable.

I would propose to change line 322 and 372 to:


`destinationChainId == 1 || destinationChainId == 5 ? 0 : deadline`

0xDEnYO (approved):

src/Facets/HopFacetPacked.sol
maxklenk: Thank you. I tried that first but but an `CompilerError: Stack too deep.` error prevented that implementation. I now moved also the deadline calculation inline to solve this problem.

The changes saved stunning 20 gas...
Accessing `block.timestamp` twice only added an overhead of 2-5 gas. 
But by not calculating the deadline buffer for transfers to L1 saved 100 gas in those calls.

PR #328, opened 2023-04-21
chore: set approval for packed facets

maxklenk:


script/UpdateCBridgeFacetPackedToken.s.sol
```diff
@@ -0,0 +1,31 @@
+// SPDX-License-Identifier: UNLICENSED
```
0xDEnYO, line 1: I just noticed this. We usually have MIT License on all files. Here we have UNLICENSED. Why? And should it be coherent?

script/UpdateCBridgeFacetPackedToken.s.sol
```diff
@@ -0,0 +1,31 @@
...
+pragma solidity ^0.8.17;
```
0xDEnYO, line 2: Also I see a problem with the naming of the deploy & update script.
The scripts should always be named Deploy<FacetName> and Update<FacetName>, otherwise the deploy scripts will probably run into errors.

If you have different versions of these scripts, I would still combine them into one and then pass a bool (or whatever) parameter that defines which mode/features to be used.

0xDEnYO (changes requested):

ezynda3:
@maxklenk any updates on @0xDEnYO's comments?

script/UpdateCBridgeFacetPackedToken.s.sol
maxklenk: Many of the Update scripts contain this License Identifier, if we want to change that we should do it in one PR on all files.

script/UpdateCBridgeFacetPackedToken.s.sol
maxklenk: The scripts are planned to be executed manually if needed, so I used different names. My idea would be that the master deploy script ignores them if the name does not match the facet name exactly.

maxklenk:
@ezynda3 in my opinion out of scope of this PR.

script/UpdateCBridgeFacetPackedToken.s.sol
0xDEnYO: I do not fully understand why we are creating such "exceptional" cases if we could with minimum effort align them with our established procedures, such as suggested above.

I understand your opinion but it doesnt improve the maintainability of our code base and I dont see any strong reason for such an exception.

And lastly, I feel that we are not dealing correctly with review comments here. They are not resolved but the PR is merged anyway.
Same as with the comment above. I think I had a fair point there (about the license identifier) but the comment does not get resolved (e.g. by creating a new JIRA task for it), instead it's ignored or deemed out of scope and will just get forgotten that way. I feel we could collaborate better here.

PR #331, opened 2023-04-24
feat(celerCircle): Copy circleFacet for celer proxy for circle bridge (CelerCircleBridgeFacet > v1.0.0)

maxklenk:


src/Facets/CelerCircleBridgeFacet.sol
```diff
@@ -0,0 +1,104 @@
...
+    ReentrancyGuard,
+    SwapperV2,
+    Validatable
+{
+    /// Storage ///
+
+    /// @notice The address of the TokenMessenger on the source chain.
+    ICircleBridgeProxy private immutable circleBridgeProxy;
+
+    /// @notice The USDC address on the source chain.
```
0xDEnYO, line 26: maybe replace "source chain" with "current chain" as it might be misleading

src/Facets/CelerCircleBridgeFacet.sol
```diff
@@ -0,0 +1,104 @@
...
+        LibAsset.maxApproveERC20(
+            IERC20(usdc),
+            address(circleBridgeProxy),
+            _bridgeData.minAmount
+        );
+
+        // initiate bridge transaction
+        circleBridgeProxy.depositForBurn(
+            _bridgeData.minAmount,
+            uint64(_bridgeData.destinationChainId),
```
0xDEnYO, line 97: Maybe we should check before if this value is within uint64 range and throw a meaningful error since gas consumption does not seem to be a worry for this facet

src/Interfaces/ICircleBridgeProxy.sol
```diff
@@ -0,0 +1,24 @@
...
+    ///      - given destinationDomain has no TokenMessenger registered.
+    ///      - transferFrom() reverts. For example, if sender's burnToken balance
+    ///        or approved allowance to this contract is less than `amount`.
+    ///      - burn() reverts. For example, if `amount` is 0.
+    ///      - MessageTransmitter returns false or reverts.
+    /// @param _amount Amount of tokens to burn.
+    /// @param _dstChid Destination domain.
+    /// @param _mintRecipient Address of mint recipient on destination domain.
+    /// @param _burnToken Address of contract to burn deposited tokens, on local domain.
+    /// @return nonce Unique nonce reserved by message.
```
0xDEnYO, line 17: "..reserved by message" is not clear to me. What message?

test/solidity/Facets/CelerCircleBridgeFacet.t.sol
```diff
@@ -0,0 +1,116 @@
+// SPDX-License-Identifier: Unlicense
+pragma solidity 0.8.17;
+
+import { LibSwap, LibAllowList, TestBaseFacet, console } from "../utils/TestBaseFacet.sol";
+import { InsufficientBalance } from "src/Errors/GenericErrors.sol";
+import { CelerCircleBridgeFacet } from "lifi/Facets/CelerCircleBridgeFacet.sol";
+import { ICircleBridgeProxy } from "lifi/Interfaces/ICircleBridgeProxy.sol";
+
+// import { DiamondTest, LiFiDiamond } from "../utils/DiamondTest.sol";
```
0xDEnYO, line 9: can be removed?

src/Facets/CelerCircleBridgeFacet.sol
```diff
@@ -0,0 +1,104 @@
+// SPDX-License-Identifier: MIT
```
0xDEnYO, line 1: the .md file for this new bridge is missing in the PR

0xDEnYO (approved):

src/Interfaces/ICircleBridgeProxy.sol
maxklenk: This is how the official CCTP contract describes that parameter: https://goerli.etherscan.io/address/0xd0c3da58f55358142b8d3e06c1c30c5c6114efe8#code
I would assume every transfer is packed as a message and this will be some identifier for it.

PR #334, opened 2023-04-25
feat(hop): Handle swap in hop facet (HopFacetPacked 1.02 > 1.0.3)

ezynda3:


src/Facets/HopFacetPacked.sol
```diff
@@ -6,12 +6,40 @@ import { ERC20 } from "solmate/utils/SafeTransferLib.sol";
...
+        uint256 dx,
+        uint256 minDy,
+        uint256 deadline
+    ) external returns (uint256);
+}
 
 /// @title Hop Facet (Optimized for Rollups)
 /// @author LI.FI (https://li.fi)
 /// @notice Provides functionality for bridging through Hop
 /// @custom:version 1.0.2
```
0xDEnYO, line 34: The version needs to be bumped here (and in PR title)

src/Facets/HopFacetPacked.sol
```diff
@@ -20,7 +48,23 @@ contract HopFacetPacked is ILiFi, TransferrableOwnership {
 
     /// @notice Initialize the contract.
     /// @param _owner The contract owner to approve tokens.
-    constructor(address _owner) TransferrableOwnership(_owner) {}
+    constructor(
+        address _owner,
+        address _wrapper
+    ) TransferrableOwnership(_owner) {
+        l2CanonicalToken = _wrapper != address(0)
```
0xDEnYO, line 55: If I understand correctly, we allow this facet to be initiated with  `_wrapper == address(0)` since on L1 (=ETH) there is no wrapper, right?

Maybe we could add a check that allows _wrapper to be only address(0) on non-ETH networks (to make sure that the facet is functional.

0xDEnYO (changes requested):

src/Facets/HopFacetPacked.sol
```diff
@@ -6,12 +6,40 @@ import { ERC20 } from "solmate/utils/SafeTransferLib.sol";
...
 /// @author LI.FI (https://li.fi)
 /// @notice Provides functionality for bridging through Hop
 /// @custom:version 1.0.2
 contract HopFacetPacked is ILiFi, TransferrableOwnership {
+    /// Storage ///
+
+    address public immutable bridge;
+    address public immutable l2CanonicalToken;
+    address public immutable hToken;
+    address public immutable exchangeAddress;
```
maxklenk, line 41: I would rename the variables to highlight that they are for native token only. ERC20 tokens use different addresses.

src/Facets/HopFacetPacked.sol
```diff
@@ -56,16 +100,26 @@ contract HopFacetPacked is ILiFi, TransferrableOwnership {
         // hopBridge: address(bytes20(msg.data[88:108]))
         // => total calldata length required: 108
 
+        uint256 amountOutMin = uint256(uint128(bytes16(msg.data[52:68])));
```
maxklenk, line 103: as the variable is not reused it can be inlined in the function call.

maxklenk (changes requested):

maxklenk (approved):
LGTM

PR #340, opened 2023-04-26
add new multichain config

maxklenk:


src/Facets/MultichainFacet.sol
```diff
@@ -65,16 +65,8 @@ contract MultichainFacet is ILiFi, SwapperV2, ReentrancyGuard, Validatable {
 
         Storage storage s = getStorage();
 
-        if (anyNative == address(0)) {
-            revert InvalidConfig();
-        }
-
         s.anyNative = anyNative;
 
-        if (s.initialized) {
```
ezynda3, line 74: I would remove line 89 as well. Also would add a comment next to line 31 saying this var is no longer used but any new vars should be added below it to maintain the same storage layout and not corrupt anything.

PR #350, opened 2023-05-02
feat(hop): add additional hop parameters (HopFacet 1.0.0 > 2.0.0)

maxklenk:


deployments/_deployments_log_file.json
```diff
@@ -1013,62 +1017,6 @@
       }
     }
   },
-  "WormholeFacet": {
```
0xDEnYO, line 1016: Why are WormholeFacet and CelerCircleBridgeFacet log entries removed here? Doesnt seem right to me

foundry.toml
```diff
@@ -39,7 +39,7 @@ nova = "${ETH_NODE_URI_NOVA}"
 # testnets
 goerli = "${ETH_NODE_URI_GOERLI}"
 mumbai = "${ETH_NODE_URI_MUMBAI}"
-consensys-zkevm-testnet = "${ETH_NODE_URI_CONSENSYS_ZKEVM_TESTNET}"
```
0xDEnYO, line 42: Would be good to also adjust the env.example with the new network

networks
```diff
@@ -18,8 +17,10 @@ harmony
 evmos
 aurora
 boba
+nova
+
```
0xDEnYO, line 21: I am not sure if a newline could cause trouble cause the scripts are reading the list of networks from this file. I know you wanted to separate mainnets and testnets but I would suggest to remove the newline to avoid error sources.

test/solidity/Facets/HopFacet.t.sol
```diff
@@ -79,15 +77,18 @@ contract HopFacetTest is TestBaseFacet {
...
         validHopData = HopFacet.HopData({
             bonderFee: 0,
             amountOutMin: 0,
             deadline: block.timestamp + 60 * 20,
             destinationAmountOutMin: 0,
-            destinationDeadline: block.timestamp + 60 * 20
+            destinationDeadline: block.timestamp + 60 * 20,
+            relayer: address(0),
+            relayerFee: 0,
+            nativeFee: 0
```
0xDEnYO, line 91: Is this really valid data (I thought a nativeFee has to be sent with every tx...so I would have expected a value here)?

test/solidity/Facets/HopFacetL1Testnet.t.sol
```diff
@@ -0,0 +1,157 @@
...
+    function initiateBridgeTxWithFacet(bool isNative) internal override {
+        if (isNative) {
+            // fee parameter Native
+            validHopData.relayerFee = 10000000000000000;
+            validHopData.relayer = 0x81682250D4566B2986A2B33e23e7c52D401B7aB7;
+        } else {
+            // fee parameter ERC20
+            validHopData.relayer = 0xB47dE784aB8702eC35c5eAb225D6f6cE476DdD28;
+        }
+
```
0xDEnYO, line 122: These two if-statements could be merged imho

test/solidity/Facets/HopFacetL1Testnet.t.sol
```diff
@@ -0,0 +1,157 @@
...
+                bridgeData,
+                validHopData
+            );
+        }
+    }
+
+    function initiateSwapAndBridgeTxWithFacet(
+        bool isNative
+    ) internal override {
+        if (bridgeData.sendingAssetId == address(0)) {
```
0xDEnYO, line 138: Is this supposed to be different here (not checking if "isNative")? If so, would be nice to add a comment to explain the reasoning

deployments/_bytecode_storage.json
```diff
@@ -39,7 +39,8 @@
...
...151610100818187015280870151915050610120612bd78187018315159052565b90950151151593019290925250919050565b600060208284031215612bfb57600080fd5b5051919050565b600060208284031215612c1457600080fd5b813561209b816126ec565b60008083357fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffe1843603018112612c5457600080fd5b83018035915067ffffffffffffffff821115612c6f57600080fd5b602001915036819003821315612c8457600080fd5b9250929050565b60008085851115612c9b57600080fd5b83861115612ca857600080fd5b5050820193919092039150565b7fffffffff000000000000000000000000000000000000000000000000000000008135818116916004851015612cf55780818660040360031b1b83161692505b505092915050565b8183823760009101908152919050565b60208152600061209b6020830184612a8c565b600060208284031215612d3257600080fd5b815167ffffffffffffffff811115612d4957600080fd5b8201601f81018413612d5a57600080fd5b8051612d686126b582612650565b818152856020838501011115612d7d57600080fd5b612d8e826020830160208601612a68565b95945050505050565b600060208284031215612da957600080fd5b815161209b816126ec565b80820180821115610f2057610f20612941565b60008251612a5e818460208701612a6856fea26469706673582212207ccb78ab9459cec277fd30a891a7ead0dc8de5c978c8cfe1d0b8835d98cc5ded64736f6c63430008110033"
```
0xDEnYO, line 42: Why is the code of 1.0.0 replaced?
Maybe this is an error (= tried to deploy before the version was bumped)?

I would expect this PR to only add a new version but not change the code of an existing one

0xDEnYO (changes requested):

test/solidity/Facets/HopFacet.t.sol
maxklenk: Yes, those values are currently only needed for Goerli/Linea transfers.

test/solidity/Facets/HopFacetL1Testnet.t.sol
maxklenk: Yes, I moved them out to have a similar handling as in the function below.

deployments/_bytecode_storage.json
maxklenk: The version did not get replaced. Just a comma was added at the end to be able to add the next version.

.github/workflows/types.yaml
```diff
@@ -5,6 +5,7 @@ on:
     branches:
       - "master"
       - "staging"
+      - "develop"
```
0xDEnYO, line 8: is this correct? Cause the name of the branch is "dev", not "develop" (as in the backend)

foundry.toml
maxklenk: Added

.github/workflows/types.yaml
maxklenk: fixed

networks
maxklenk: fixed

PR #372, opened 2023-05-10
Implement calldata verification facet v1.0.0

ezynda3:


src/Facets/CalldataVerificationFacet.sol
```diff
@@ -0,0 +1,240 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.17;
+
+import { ILiFi } from "../Interfaces/ILiFi.sol";
+import { LibSwap } from "../Libraries/LibSwap.sol";
+import { AmarokFacet } from "./AmarokFacet.sol";
+import { StargateFacet } from "./StargateFacet.sol";
+import { CelerIMFacet } from "./CelerIMFacet.sol";
+
+/// @title A title that should describe the contract/interface
```
satoshi-lifi, line 10: Let's add actual title ;)

src/Facets/CalldataVerificationFacet.sol
```diff
@@ -0,0 +1,240 @@
...
+        ILiFi.BridgeData memory bridgeData;
+        (bridgeData) = abi.decode(data[4:], (ILiFi.BridgeData));
+
+        if (bridgeData.hasSourceSwaps) {
+            LibSwap.SwapData[] memory swapData;
+            (, swapData) = abi.decode(
+                data[4:],
+                (ILiFi.BridgeData, LibSwap.SwapData[])
+            );
+            sendingAssetId = swapData[0].sendingAssetId;
```
satoshi-lifi, line 96: I think we always return the `bridgeData.sendingAssetId` for the `sendingAssetId`.
When we do swaps before the bridging, the final asset of swaps is `sendingAssetId`.

src/Facets/CalldataVerificationFacet.sol
```diff
@@ -0,0 +1,240 @@
...
+pragma solidity 0.8.17;
+
+import { ILiFi } from "../Interfaces/ILiFi.sol";
+import { LibSwap } from "../Libraries/LibSwap.sol";
+import { AmarokFacet } from "./AmarokFacet.sol";
+import { StargateFacet } from "./StargateFacet.sol";
+import { CelerIMFacet } from "./CelerIMFacet.sol";
+
+/// @title A title that should describe the contract/interface
+/// @author Li.Finance (https://li.finance)
```
maxklenk, line 11: /// @author LI.FI (https://li.fi)

src/Facets/CalldataVerificationFacet.sol
maxklenk: Yes, but we want to return which asset the user starts with in this case.

src/Facets/CalldataVerificationFacet.sol
```diff
@@ -0,0 +1,240 @@
...
+    function extractMainParameters(
+        bytes calldata data
+    )
+        public
+        pure
+        returns (
+            string memory bridge,
+            address sendingAssetId,
+            address receiver,
+            uint256 minAmount,
```
maxklenk, line 81: I think we can name this `amount` as exactly this amount has to be passed.

src/Facets/CalldataVerificationFacet.sol
```diff
@@ -0,0 +1,240 @@
...
+                keccak256(abi.encodePacked(bridge))) &&
+            (sendingAssetId == 0xFFfFfFffFFfffFFfFFfFFFFFffFFFffffFfFFFfF ||
+                bridgeData.sendingAssetId == sendingAssetId) &&
+            (receiver == 0xFFfFfFffFFfffFFfFFfFFFFFffFFFffffFfFFFfF ||
+                bridgeData.receiver == receiver) &&
+            (amount == type(uint256).max || bridgeData.minAmount == amount) &&
+            (destinationChainId == type(uint256).max ||
+                bridgeData.destinationChainId == destinationChainId) &&
+            bridgeData.hasSourceSwaps == hasSourceSwaps &&
+            bridgeData.hasDestinationCall == hasDestinationCall;
```
maxklenk, line 160: Splitting this in more lines with comments would help to understand what we check here.

src/Facets/CalldataVerificationFacet.sol
```diff
@@ -0,0 +1,240 @@
...
+                    LibSwap.SwapData[],
+                    CelerIMFacet.CelerIMData
+                )
+            );
+            return keccak256(dstCalldata) == keccak256(celerIMData.callData);
+        }
+
+        // All other cases
+        return false;
+    }
```
maxklenk, line 239: We should add a function to parse calls to the `GenericSwapFacet` facet.

maxklenk (changes requested):

src/Facets/CalldataVerificationFacet.sol
ezynda3: @maxklenk What would we like to return from this?
```
reciever
amount
sendingAssetId
```

maxklenk: @ezynda3 I think input and output are important
```
sendingAssetId <= first swap asset
sendingAmount <= first swap amount
receiver 
receivingAssetId <= last swap asset
receivingMinAmount <= passed minAmount
```

PR #403, opened 2023-05-25
Parse target state sheet

0xDEnYO:


scripts/deploy/resources/deployHelperFunctions.sh
```diff
@@ -1035,6 +1035,133 @@ function getOptimizerRuns() {
       echo "$VERSION"
 
     }
+function parseTargetStateGoogleSpreadsheet() {
+  # read function arguments into variables
+  SPREADSHEET_ID="1FWAcFBNHK1ZNd5QF5Cq8_6lpmLiIMUFgxaym8_Yq-1s"
```
satoshi-lifi, line 1040: We should store it in env, shouldn't we?

ezynda3: We could also store this in the config.sh we created since that is git ignored.

satoshi-lifi (approved):

PR #418, opened 2023-06-07
deprecated contracts moved to archive, tests deleted & unused networks excluded and removed from networks list

0xDEnYO:


scripts/config.example.sh
```diff
@@ -59,7 +59,7 @@ DEBUG=true
 VERIFY_CONTRACTS=false
 
 # contract verification will be deactivated for any network listed here
-DO_NOT_VERIFY_IN_THESE_NETWORKS="gnosis,testNetwork,aurora"
+DO_NOT_VERIFY_IN_THESE_NETWORKS="gnosis,testNetwork,aurora,nova,harmony"
```
ezynda3, line 62: Same as above

PR #453, opened 2023-06-29
Update AllBridgeFacet to support new SM interface << Version 2.0.0

ezynda3:


maxklenk (approved):

src/Facets/AllBridgeFacet.sol
```diff
@@ -31,6 +31,7 @@ contract AllBridgeFacet is ILiFi, ReentrancyGuard, SwapperV2, Validatable {
         bytes32 receiveToken;
         uint256 nonce;
         IAllBridge.MessengerProtocol messenger;
+        bool payFeeWithSendingAsset;
     }
```
maxklenk, line 36: Why does the `destinationChainId` need to be define here instead of using what's in bridgeData? If we need another number here it should be uint256 as well.

maxklenk (changes requested):

maxklenk (approved):

PR #456, opened 2023-06-30
Add documentation and codegen for implementing new contracts

ezynda3:


templates/facet.template.hbs
```diff
@@ -0,0 +1,98 @@
...
+
+import { ILiFi } from "../Interfaces/ILiFi.sol";
+import { LibAsset } from "../Libraries/LibAsset.sol";
+import { LibSwap } from "../Libraries/LibSwap.sol";
+import { ReentrancyGuard } from "../Helpers/ReentrancyGuard.sol";
+import { SwapperV2 } from "../Helpers/SwapperV2.sol";
+import { Validatable } from "../Helpers/Validatable.sol";
+
+/// @title {{titleCase name}} Facet
+/// @author LI.FI (https://li.fi)
```
0xDEnYO, line 12: should the author be the actual author?

templates/facet.template.hbs
```diff
@@ -0,0 +1,98 @@
...
+/// @custom:version 1.0.0
+contract {{titleCase name}}Facet is ILiFi, ReentrancyGuard, SwapperV2, Validatable {
+    /// Storage ///
+    
+    address public immutable example;
+
+
+    /// Types ///
+
+    /// @param exampleParam Example paramter
```
0xDEnYO, line 23: Maybe we could explain somewhere that this is optional

templates/facet.template.hbs
```diff
@@ -0,0 +1,98 @@
...
+    struct {{titleCase name}}Data {
+      string exampleParam;
+    }
+
+    /// Constructor ///
+
+    /// @notice Initialize the contract.
```
0xDEnYO, line 30: imho not entirely accurate. 
The constructor does not and should not initialize the contract. I think this could be misleading. 
The constructor should only assign immutable params. Any other initialization (such as token lists, domainIDs, etc.) should be done in a separate init Method (that is missing here).

I think this pattern should also be carefully explained to third parties

templates/facetConfig.template.hbs
0xDEnYO, line 1: I think it could be beneficial for understanding by third parties to also have some sort of token list or so here, just a little more complex example and how this is being used by the deploy and/or update script

templates/facetTest.template.hbs
```diff
@@ -0,0 +1,94 @@
...
+
+contract {{titleCase name}}FacetTest is TestBaseFacet {
+    {{titleCase name}}Facet.{{titleCase name}}Data internal valid{{titleCase name}}Data;
+    Test{{titleCase name}}Facet internal {{camelCase name}}Facet;
+
+    function setUp() public {
+        customBlockNumberForForking = 17130542;
+        initTestBase();
+
+        {{camelCase name}}Facet = new Test{{titleCase name}}Facet(address(0xb33f));
```
0xDEnYO, line 30: why are we passing a hardcode address here?

templates/facetTest.template.hbs
0xDEnYO, line 1: I think the general setup with our inherited test cases should also be explained somewhere so that they know they will need to either ensure these tests pass or overwrite them with a reasoning why these tests do not apply for this facet

0xDEnYO (approved):

PR #465, opened 2023-07-05
added LIFuelFacet and GenericSwapFacet as core facet in scripts etc.

0xDEnYO:


script/deploy/facets/UpdateCoreFacets.s.sol
```diff
@@ -107,19 +118,25 @@ contract DeployScript is UpdateScriptBase {
...
+            if (cut.length > 0) {
+                cutData = abi.encodeWithSelector(
+                    DiamondCutFacet.diamondCut.selector,
+                    cut,
+                    address(0),
+                    ""
+                );
+            }
+            return (facets, cutData);
+        }
```
satoshi-lifi, line 132: It seems this part is duplicated.

script/tasks/solidity/transferOwnershipOfPeripheryContractsForImmutable.s.sol
```diff
@@ -42,17 +42,14 @@ contract DeployScript is Script {
             "json"
         );
         networkLogJSON = vm.readFile(path);
-        console.log("A");
```
satoshi-lifi, line 45: nit: `console` should be removed from import as well.

0xDEnYO: removed

satoshi-lifi (approved):

PR #466, opened 2023-07-06
SGEthVaultWrapper v1.0.0

ezynda3:


src/Facets/StargateFacet.sol
```diff
@@ -184,28 +187,46 @@ contract StargateFacet is ILiFi, ReentrancyGuard, SwapperV2, Validatable {
...
         StargateData calldata _stargateData
-    ) private noNativeAsset(_bridgeData) {
-        LibAsset.maxApproveERC20(
-            IERC20(_bridgeData.sendingAssetId),
-            address(router),
-            _bridgeData.minAmount
-        );
+    ) private {
+        if (LibAsset.isNativeAsset(_bridgeData.sendingAssetId)) {
+            nativeRouter.swapETHAndCall{ value: _bridgeData.minAmount }(
```
satoshi-lifi, line 192: Not sure but when bridge native asset, Stargate doesn't ask `lzFee`?

test/solidity/Facets/StargateFacet.t.sol
```diff
@@ -7,12 +7,16 @@ import { StargateFacet } from "lifi/Facets/StargateFacet.sol";
 import { IStargateRouter } from "lifi/Interfaces/IStargateRouter.sol";
 import { FeeCollector } from "lifi/Periphery/FeeCollector.sol";
 import { ILiFi } from "lifi/Interfaces/ILiFi.sol";
+import { LibSwap } from "lifi/Libraries/LibSwap.sol";
 
 // Stub CBridgeFacet Contract
```
satoshi-lifi, line 12:
```suggestion
// Stub StargateFacet Contract
```

test/solidity/Facets/StargateFacet.t.sol
```diff
@@ -7,12 +7,16 @@ import { StargateFacet } from "lifi/Facets/StargateFacet.sol";
...
 contract TestStargateFacet is StargateFacet {
     /// @notice Initialize the contract.
     /// @param _router The contract address of the stargatefacet router on the source chain.
```
satoshi-lifi, line 15:
```suggestion
    /// @param _router The contract address of the stargate router on the source chain.
    /// @param _nativeRouter The contract address of the native stargate router
    ///                      on the source chain.
```

satoshi-lifi:
Here are additional suggestions
- Remove `InvalidConfig` from imports
- Remove `PoolIdConfig` struct
- Remove `InvalidStargateRouter` custom error
- Update `_bridgeData` from `memory` to `calldata` in `startBridgeTokensViaStargate`.
   I recommend to update `_bridgeData` to `calldata` in `startBridgeTokensVia...` functions when we update facets.

src/Facets/StargateFacet.sol
ezynda3: From what I understand in order to behave like other bridges we would like to deduct the fee from the minAmount. Correct me if I'm wrong @maxklenk?

satoshi-lifi (approved):

src/Facets/StargateFacet.sol
satoshi-lifi: maybe we can add `payFeeWithSendingAsset` to the `StargateData`.
and when the flag is set, then we include the fee in `minAmount`, otherwise, we separate the fee from `minAmount`

maxklenk: The `payFeeWithSendingAsset` flag would make it clearer. But as we can't really use it for other tokens, because stargate does not allow to pay the fee with them, it could be confusing.

I see the beauty of using the including the fee into the transferred amount, as e.g. it would be possible to start with some ERC20 only, swap all of it into native, pay the fees with it and bridge the rest.

I have to check in the backend how the fees are actually calculated, I assume it works the same as with ERC20 tokens, that we specify the amount we want to transfer and the fees are added. This would make it difficult to start with an exact amount that includes both fee and transfer amount.

PR #484, opened 2023-07-18
Add ServiceFeeCollector to approved DEXs

ezynda3:


satoshi-lifi:
@ezynda3  Is there a reason that `artifacts-zk` and `cache-zk` are pushed?

ezynda3:
> @ezynda3 Is there a reason that `artifacts-zk` and `cache-zk` are pushed?

Fixed

satoshi-lifi (approved):
LGTM

PR #486, opened 2023-07-20
Update OFTWrapperFacet with proxy OFT >> OFTWrapperFacet v2.0.0

satoshi-lifi:


src/Facets/OFTWrapperFacet.sol
```diff
@@ -13,7 +14,7 @@ import { Validatable } from "../Helpers/Validatable.sol";
 /// @title OFTWrapper Facet
 /// @author LI.FI (https://li.fi)
 /// @notice Provides functionality for bridging through OFTWrapper
-/// @custom:version 1.0.0
+/// @custom:version 1.0.1
```
0xDEnYO, line 17: This is a breaking change, should be 2.0.0

src/Interfaces/IOFTV2.sol
0xDEnYO, line 1: Shouldnt the interface be named IProxyOFT.sol ?

src/Facets/OFTWrapperFacet.sol
```diff
@@ -148,6 +158,12 @@ contract OFTWrapperFacet is ILiFi, ReentrancyGuard, SwapperV2, Validatable {
         doesNotContainDestinationCalls(_bridgeData)
         noNativeAsset(_bridgeData)
     {
+        _checkProxyOFTAddress(
```
0xDEnYO, line 161: Could move this into _startBridge to not have code duplicate imho

0xDEnYO (changes requested):

src/Interfaces/IOFTV2.sol
satoshi-lifi: It's from layer zero's repo. They named the interface `IOFTV2`

src/Facets/OFTWrapperFacet.sol
satoshi-lifi: I think if we put it to the `_startBridge`, then it will waste gas for `swapAndBridge...`.
but I agree with you, since the above case is just for reverts.

src/Interfaces/IOFTV2.sol
0xDEnYO: not sure what you are referring to but the IOFTV2 interface looks different:
https://github.com/LayerZero-Labs/solidity-examples/blob/main/contracts/token/oft/v2/IOFTV2.sol

also, this is a contract in our repo. It describes the interface for a contract that is a ProxyOFT....so I would name it accordingly, otherwise it's just confusing/misleading

A V2 OFT is not necessarily a proxy (as far as I understand)

satoshi-lifi: I was talking about this part of their code.
image
I will update the interface

0xDEnYO (approved):

PR #492, opened 2023-07-26
Add script to check whether Executor match on Diamond and Receiver

satoshi-lifi:


ezynda3:
Getting a jq error when running on production for the Immutable diamond.
```
jq: error: Invalid numeric literal at EOF at line 1, column 3 (while parsing '.0.') at <top-level>, line 1:
.0.value
jq: error: syntax error, unexpected IDENT, expecting $end (Unix shell quoting issues?) at <top-level>, line 1:
.0.value   
jq: 2 compile errors
```

ezynda3 (approved):

PR #504, opened 2023-08-18
feat: Add StandardizedCallFacet 1.0.0

ezynda3:


script/deploy/facets/DeployStandardizedCallFacet.s.sol
```diff
@@ -0,0 +1,51 @@
...
+                        type(StandardizedCallFacet).creationCode,
+                        constructorArgs
+                    )
+                )
+            )
+        );
+
+        vm.stopBroadcast();
+    }
+}
```
0xb10ckdev, line 51:
```suggestion
        constructorArgs = getConstructorArgs();

        deployed = StandardizedCallFacet(
            deploy(type(StandardizedCallFacet).creationCode)
        );
    }

    function getConstructorArgs() internal override returns (bytes memory) {
        string memory path = string.concat(
            root,
            "/config/standardizedCallFacet.json"
        );
        string memory json = vm.readFile(path);

        address example = json.readAddress(
            string.concat(".", network, ".example")
        );

        return abi.encode(example);
    }
}
```

script/deploy/facets/UpdateStandardizedCallFacet.s.sol
```diff
@@ -0,0 +1,41 @@
...
+
+        vm.startBroadcast(deployerPrivateKey);
+        if (cut.length > 0) {
+            cutter.diamondCut(cut, address(0), "");
+        }
+        facets = loupe.facetAddresses();
+
+        vm.stopBroadcast();
+    }
+}
```
0xb10ckdev, line 41:
```suggestion
import { UpdateScriptBase } from "./utils/UpdateScriptBase.sol";

contract DeployScript is UpdateScriptBase {
    function run()
        public
        returns (address[] memory facets, bytes memory cutData)
    {
        return update("StandardizedCallFacet");
    }
}
```

src/Facets/StandardizedCallFacet.sol
```diff
@@ -0,0 +1,49 @@
...
+
+/// @title StandardizedCallFacet Facet
+/// @author LIFI https://li.finance ed@li.finance
+/// @notice Allows calling different facet methods through a single standardized entrypoint
+/// @custom:version 1.0.0
+contract StandardizedCallFacet {
+    /// External Methods ///
+
+    // @notice Make a standardized call to a facet
+    // @param callData The calldata to forward to the facet
```
0xb10ckdev, line 14:
```suggestion
    /// @notice Make a standardized call to a facet
    /// @param callData The calldata to forward to the facet
```

src/Facets/StandardizedCallFacet.sol
```diff
@@ -0,0 +1,49 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.17;
+
+import { LibDiamond } from "../Libraries/LibDiamond.sol";
+
+/// @title StandardizedCallFacet Facet
```
0xDEnYO, line 6: could remove the last "Facet" here to match with the titles of our other contracts

script/deploy/facets/DeployStandardizedCallFacet.s.sol
```diff
@@ -0,0 +1,51 @@
...
+    function run()
+        public
+        returns (StandardizedCallFacet deployed, bytes memory constructorArgs)
+    {
+        string memory path = string.concat(
+            vm.projectRoot(),
+            "/config/standardizedCallFacet.json"
+        );
+        string memory json = vm.readFile(path);
+        address example = json.readAddress(
```
0xDEnYO, line 22: This seems to be erroneously used from the template to me. 
The facet does not have a constructor and the whole example variable should be removed imho (same for line 26)

src/Facets/StandardizedCallFacet.sol
```diff
@@ -0,0 +1,49 @@
...
+/// @author LIFI https://li.finance ed@li.finance
+/// @notice Allows calling different facet methods through a single standardized entrypoint
+/// @custom:version 1.0.0
+contract StandardizedCallFacet {
+    /// External Methods ///
+
+    // @notice Make a standardized call to a facet
+    // @param callData The calldata to forward to the facet
+    function standardizedCall(bytes memory callData) external payable {
+        LibDiamond.DiamondStorage storage ds = LibDiamond.diamondStorage();
```
0xDEnYO, line 16: I would propose to call the facetAddress function of the DiamondLoupeFacet here since this is duplicated code from this facet and the code would be much easier to understand if it's just that one function call

test/solidity/Facets/StandardizedCallFacet.t.sol
```diff
@@ -0,0 +1,64 @@
...
+        addFacet(
+            tmpDiamond,
+            address(standardizedCallFacet),
+            functionSelectors
+        );
+
+        diamond = Diamond(address(tmpDiamond));
+    }
+
+    function testCanCallOtherFacet() public {
```
0xDEnYO, line 45: I do not really understand this test at all. 

In my understanding, the StandardizedFacet is supposed to forward calls to FACETS, not to periphery. 
This test seems to me like a test of the peripheryRegistryFacet. 

Shouldnt the test setup more be like:

- create diamond
- add FooFacet
- make call to FooFacet.fooFunction() through standardizedFacet with selector of fooFunction
?

test/solidity/Facets/StandardizedCallFacet.t.sol
```diff
@@ -0,0 +1,64 @@
...
+        address result = diamond.getPeripheryContract("Foobar");
+        assertEq(result, address(0xf00));
+    }
+
+    function testFailWhenCallingANonExistentFunction() public {
+        bytes memory data = abi.encodeWithSelector(
+            NotAContract.notAFunction.selector
+        );
+
+        diamond.standardizedCall(data);
```
0xDEnYO, line 62: We should make our tests more explicit and add the exact type of error we are expecting. This test could fail for a number of reasons and we would not see that here if it failed for the wrong reason.

0xDEnYO (changes requested):

src/Facets/StandardizedCallFacet.sol
ezynda3: Calling the DiamondLoupeFacet would incur more gas as it is an external function call. Diamonds should try to use libraries for any internal logic.

test/solidity/Facets/StandardizedCallFacet.t.sol
ezynda3: PeripheryRegistry facet is a facet. All the test cares about is whether we can make a call that uses the same context as any other call to the diamond would.

src/Facets/StandardizedCallFacet.sol
0xDEnYO: Good point. 
A comment would help to make this clearer.

test/solidity/Facets/StandardizedCallFacet.t.sol
0xDEnYO: Right, I understand your thinking.

I find it confusing though that we test this facet in a way that it's not intended to being used. 
The intended use is to initiate a bridge tx through this standardized entry point. 
A test case that would mimic the actual usage would make more sense imho, even though the effects may be the same. 
It would also be nice to showcase this for a couple of different facets, not just for one to make the test more robust.

test/solidity/Facets/StandardizedCallFacet.t.sol
0xDEnYO: Hmm....dont see how this was resolved

0xDEnYO (approved):

PR #507, opened 2023-08-24
chore: Deploy contracts to Linea Mainnet

ezynda3:


config/amarok.json
```diff
@@ -55,6 +55,10 @@
     "connextHandler": "0x0000000000000000000000000000000000000000",
     "domain": "0"
   },
+  "linea": {
```
0xDEnYO, line 58: why do we add this if Amarok is not even deployed in this network?

config/stargate.json
```diff
@@ -26,6 +26,7 @@
     "goerli": "0x7612aE2a34E5A363E137De748801FB4c86499152",
     "bsc-testnet": "0x0000000000000000000000000000000000000000",
     "lineatest": "0x0000000000000000000000000000000000000000",
+    "linea": "0x0000000000000000000000000000000000000000",
```
0xDEnYO, line 29: same here...why do we add address 0 if we dont deploy the facet on this network?

deployments/linea.diamond.immutable.json
```diff
@@ -0,0 +1,58 @@
...
+        "Version": "1.0.0"
+      },
+      "0xcB62A1B9Ab262df717926B4b24D468bb5f18F24b": {
+        "Name": "PeripheryRegistryFacet",
+        "Version": "1.0.0"
+      },
+      "0x45A208ab68A7A18533bbbb2649a6Acd5f9B12d24": {
+        "Name": "LIFuelFacet",
+        "Version": "1.0.0"
+      },
```
0xDEnYO, line 35: the standardizedCallFacet is missing

0xDEnYO (approved):

config/amarok.json
ezynda3: To get rid of the warning in the deploy scripts.

config/stargate.json
ezynda3: Same as above

config/amarok.json
0xDEnYO: Well....it doesnt make sense that the deploy script warns you for not having information about a facet that it's not even planning to deploy. I wasnt aware that it does that.. 

If that is indeed the case then I would suggest we adjust the script/warnings instead of adding meaningless and misleading config information into our repo....

ezynda3: The warning is for the Receiver deploy script. It needs these values in order to even deploy.

ezynda3: We can discuss further at the SC sync this week.

PR #510, opened 2023-08-25
feat: CallDataVerificationFacet 1.0.1 > 1.1.0 - Add handling for calldata wrapped in a StandardizedCallFacet.standardizedCall()

ezynda3:


src/Facets/CalldataVerificationFacet.sol
0xDEnYO, line 1: Please rename the PR to reflect the version change in the facet

src/Facets/CalldataVerificationFacet.sol
```diff
@@ -6,12 +6,17 @@ import { LibSwap } from "../Libraries/LibSwap.sol";
 import { AmarokFacet } from "./AmarokFacet.sol";
 import { StargateFacet } from "./StargateFacet.sol";
 import { CelerIMFacetBase, CelerIM } from "lifi/Helpers/CelerIMFacetBase.sol";
+import { StandardizedCallFacet } from "lifi/Facets/StandardizedCallFacet.sol";
+import { console } from "../../lib/forge-std/src/console.sol";
```
0xDEnYO, line 10: may wanna remove this

src/Facets/CalldataVerificationFacet.sol
```diff
@@ -117,10 +122,23 @@ contract CalldataVerificationFacet {
...
+            bytes memory unwrappedData = abi.decode(data[4:], (bytes));
+            (, , , receiver, receivingAmount, swapData) = abi.decode(
+                unwrappedData.slice(4, unwrappedData.length - 4),
+                (bytes32, string, string, address, uint256, LibSwap.SwapData[])
+            );
+        } else {
+            // normal call
+            (, , , receiver, receivingAmount, swapData) = abi.decode(
+                data[4:],
+                (bytes32, string, string, address, uint256, LibSwap.SwapData[])
```
0xDEnYO, line 139: This is duplicated code and could be optimized. 
the abi.decode() is exactly the same, just the bytes parameter differs. So the if statement could just prepare the bytes parameter and then the decode follows after the if statement

test/solidity/Facets/CalldataVerificationFacet.t.sol
```diff
@@ -51,11 +52,22 @@ contract CallVerificationFacetTest is TestBase {
...
         calldataVerificationFacet.extractBridgeData(fullCalldata);
         calldataVerificationFacet.extractSwapData(fullCalldata);
         calldataVerificationFacet.extractData(fullCalldata);
         calldataVerificationFacet.extractMainParameters(fullCalldata);
+
+        fullCalldata = bytes.concat(standardizedCallData, "extra stuff");
+        calldataVerificationFacet.extractBridgeData(fullCalldata);
+        calldataVerificationFacet.extractSwapData(fullCalldata);
+        calldataVerificationFacet.extractData(fullCalldata);
+        calldataVerificationFacet.extractMainParameters(fullCalldata);
```
0xDEnYO, line 70: checkBridgeData and checkSwapData calls are missing here

test/solidity/Facets/CalldataVerificationFacet.t.sol
```diff
@@ -271,6 +422,38 @@ contract CallVerificationFacetTest is TestBase {
...
+        validCallWithSwap = calldataVerificationFacet
+            .validateDestinationCalldata(
+                standardizedCallDataWithSwap,
+                abi.encode(address(0xdeadbeef)),
+                bytes("foobarbytes")
+            );
+
+        badCall = calldataVerificationFacet.validateDestinationCalldata(
+            standardizedCallData,
+            abi.encode(address(0xdeadbeef)),
```
0xDEnYO, line 450: not sure why you are using this address for valid and invalid calls.....maybe use USER_RECEIVER instead for calls that you expect to pass (i.e. have valid data)?

0xDEnYO (changes requested):

test/solidity/Facets/CalldataVerificationFacet.t.sol
ezynda3: Here we only care that it doesn't blow-up with extra bytes added

src/Facets/CalldataVerificationFacet.sol
ezynda3: The versions match.

0xDEnYO: I thought we had established a pattern along those lines:

(CallDataVerificationFacet 1.0.1 > 1.1.0) <some text that describes what changed>

0xDEnYO (approved):

src/Facets/CalldataVerificationFacet.sol
ezynda3: Ah nm. Didn't understand what you meant at first

PR #518, opened 2023-09-01
fix: Update gnosis safe scripts

ezynda3:


script/deploy/gnosisSAFE/proposeTx.ts
0xDEnYO, line 1: some comments in this file could be helpful ;)

ezynda3: Added :)

PR #525, opened 2023-09-07
fix: add and allow fee collectors on base

maxklenk:


ezynda3:
@maxklenk Everything seems to be in order

PR #526, opened 2023-09-07
feat(opBNB): OpBNBBridgeFacet 1.0.0

ezynda3:


foundry.toml
```diff
@@ -41,7 +41,7 @@ velas = "${ETH_NODE_URI_VELAS}"
 
 # testnets
 goerli = "${ETH_NODE_URI_GOERLI}"
-bsc-testnet = "${ETH_NODE_URI_BSCTEST}"
+bsc-testnet = "${ETH_NODE_URI_BSC_TESTNET}"
```
maxklenk, line 44: Please also update `.env.example`

ezynda3: Done

deployments/bsc-testnet.diamond.json
```diff
@@ -1,16 +1,54 @@
...
+        "Version": "1.0.0"
+      },
+      "0xc40572B9d7B61CCB70Ab05267CcF42fB57e1ad29": {
+        "Name": "OpBNBBridgeFacet",
+        "Version": "1.0.0"
+      }
+    },
     "Periphery": {
-      "AxelarExecutor": "",
       "ERC20Proxy": "",
```
0xDEnYO, line 46: how come core facets are not deployed/registered?

script/helperFunctions.sh
```diff
@@ -872,17 +872,10 @@ function checkRequiredVariablesInDotEnv() {
   fi
 
   local PRIVATE_KEY="$PRIVATE_KEY"
-  local RPC="ETH_NODE_URI_$(tr '[:lower:]' '[:upper:]' <<<"$NETWORK")"
+  local RPC="ETH_NODE_URI_$(tr '[:lower:]' '[:upper:]' <<<"$NETWORK" | sed s/-/_/g)"
```
0xDEnYO, line 875: just FYI: there is a helper function to get the RPC_URL for a given network

src/Facets/OpBNBBridgeFacet.sol
```diff
@@ -0,0 +1,196 @@
...
+    bytes32 internal constant NAMESPACE = keccak256("com.lifi.facets.opbnb");
+
+    /// Types ///
+
+    struct Storage {
+        mapping(address => IL1StandardBridge) bridges;
+        IL1StandardBridge standardBridge;
+    }
+
+    struct Config {
```
0xDEnYO, line 30: Could benefit from a more descriptive naming or comments. Does every asset has its own bridge address?

src/Facets/OpBNBBridgeFacet.sol
```diff
@@ -0,0 +1,196 @@
...
+    /// @notice Register token and bridge
+    /// @param assetId Address of token
+    /// @param bridge Address of bridge for asset
+    function registerOpBNBBridge(address assetId, address bridge) external {
+        LibDiamond.enforceIsContractOwner();
+
+        Storage storage s = getStorage();
+
+        if (bridge == address(0)) {
+            revert InvalidConfig();
```
0xDEnYO, line 83: InvalidParameter would be more appropriate imho (but it's not that important)

src/Facets/OpBNBBridgeFacet.sol
```diff
@@ -0,0 +1,196 @@
...
+        }
+
+        s.bridges[assetId] = IL1StandardBridge(bridge);
+
+        emit OptimismBridgeRegistered(assetId, bridge);
+    }
+
+    /// @notice Bridges tokens via OpBNB Bridge
+    /// @param _bridgeData Data contaning core information for bridging
+    /// @param _bridgeData Data specific to OpBNB Bridge
```
0xDEnYO, line 93: copy/paste error

src/Facets/OpBNBBridgeFacet.sol
```diff
@@ -0,0 +1,196 @@
...
+    function startBridgeTokensViaOpBNBBridge(
+        ILiFi.BridgeData memory _bridgeData,
```
0xDEnYO, line 95: this can be calldata (even if you call _startBridge which requires memory), it works (as used in OftWrapperFacet

src/Facets/OpBNBBridgeFacet.sol
```diff
@@ -0,0 +1,196 @@
...
+
+    /// @dev Contains the business logic for the bridge via OpBNB Bridge
+    /// @param _bridgeData Data contaning core information for bridging
+    /// @param _bridgeData Data specific to OpBNB Bridge
+    function _startBridge(
+        ILiFi.BridgeData memory _bridgeData,
+        OpBNBData calldata _opBNBData
+    ) private {
+        Storage storage s = getStorage();
+        IL1StandardBridge nonStandardBridge = s.bridges[
```
0xDEnYO, line 149: storing this in a variable seems unnecessary. 
Could integrate that into the next (conditional) statement

0xDEnYO (approved):

src/Facets/OpBNBBridgeFacet.sol
ezynda3: Here storing in a var saves gas as we would need to read from storage twice otherwise.

src/Facets/OpBNBBridgeFacet.sol
ezynda3: Will keep to stay consistent with older facets for now. Would need to add a new Error to `GenericErrors.sol` otherwise

script/helperFunctions.sh
ezynda3: I just left this as is because there are quite a lot of lines like this. Better to refactor as part of another PR.

deployments/bsc-testnet.diamond.json
ezynda3: I didn't deploy any periphery as I didn't need them at the moment.

0xDEnYO: mhm....that could cause errors in the future though. 
Maybe it would be good to make it a habit to say "if we deploy a diamond somewhere, we make sure it has at the very least all the core functionality and is correctly configured"

deployments/bsc.diamond.staging.json
```diff
@@ -41,9 +41,9 @@
         "Name": "LIFuelFacet",
         "Version": "1.0.0"
       },
-      "0x3004db169fa7956609A872736452E4951D4BDA8b": {
-        "Name": "OFTWrapperFacet",
-        "Version": "2.0.0"
```
maxklenk, line 46: Why does the script replace this other deployed facet, both should be listed in the file.

src/Facets/OpBNBBridgeFacet.sol
```diff
@@ -0,0 +1,200 @@
...
+        mapping(address => IL1StandardBridge) bridges;
+        IL1StandardBridge standardBridge;
+    }
+
+    // @notice OpBNB Config
+    // @notice some assets but not all require a non-standard bridge
+    // @param assetId Address of token
+    // @param bridge Address of non-standard bridge for asset
+    struct Config {
+        address assetId;
```
0xDEnYO, line 35: what is the difference between this assetId and the _bridgeData.sendingAssetId?

PR #543, opened 2023-09-26
feat(stargate): Use composer for all calls v2.2.0

ezynda3:


src/Periphery/Receiver.sol
```diff
@@ -13,7 +13,7 @@ import { ExternalCallFailed, UnAuthorized } from "../Errors/GenericErrors.sol";
 /// @title Receiver
 /// @author LI.FI (https://li.fi)
 /// @notice Arbitrary execution contract used for cross-chain swaps and message passing
-/// @custom:version 2.0.2
+/// @custom:version 2.0.3
```
0xDEnYO, line 16: why did the version change if nothing in the code changed?

0xDEnYO (approved):

PR #550, opened 2023-10-26
Redeploy ServiceFeeCollector

ezynda3:


deploy/013_deploy_lifuel_fee_collector.ts
```diff
@@ -64,4 +64,4 @@ export default func
 
 func.id = 'deploy_lifuel_fee_collector'
 func.tags = ['DeployLiFuelFeeCollector']
-// func.dependencies = ['DeployPeripheryRegistryFacet']
```
0xDEnYO, line 67: interesting...in the last PR I reviewed you removed that dependency and here you added it. 
What is the rationale behind that?

0xDEnYO (approved):

PR #566, opened 2023-12-08
Refactor to use encrypted keystores instead of plaintext private keys

ezynda3:


script/helperFunctions.sh
```diff
@@ -722,7 +722,7 @@ function saveDiamondPeriphery_MULTICALL_NOT_IN_USE() {
   while [ $attempts -lt 11 ]; do
     echo "Trying to execute multicall now - attempt ${attempts}"
     # try to execute call
-    MULTICALL_RESULTS=$(cast send "$MULTICALL_ADDRESS" "aggregate((address,bytes)[]) returns (uint256,bytes[])" "$MULTICALL_DATA" --private-key $(getPrivateKey "$NETWORK" "$ENVIRONMENT") --rpc-url "https://polygon-rpc.com" --legacy)
+    MULTICALL_RESULTS=$(cast send "$MULTICALL_ADDRESS" "aggregate((address,bytes)[]) returns (uint256,bytes[])" "$MULTICALL_DATA" --account $(getPrivateKey "$NETWORK" "$ENVIRONMENT") --rpc-url "https://polygon-rpc.com" --legacy)
```
0xDEnYO, line 725: I dont really understand these changes. 
I would have expected that instead of `--private-key 4136723423....` we would now use `--account <KeyName>`  but the helper function `getPrivateKey` is not adapted by this PR so it would still return a private key, wouldnt it?

ezynda3: I just did a bunch of find/replace so overlooked this. Will fix. Thanks!

script/deploy/deploySingleContract.sh
```diff
@@ -190,7 +189,7 @@ deploySingleContract() {
     doNotContinueUnlessGasIsBelowThreshold "$NETWORK"
 
     # try to execute call
-    RAW_RETURN_DATA=$(DEPLOYSALT=$DEPLOYSALT NETWORK=$NETWORK FILE_SUFFIX=$FILE_SUFFIX DEFAULT_DIAMOND_ADDRESS_DEPLOYSALT=$DEFAULT_DIAMOND_ADDRESS_DEPLOYSALT DEPLOY_TO_DEFAULT_DIAMOND_ADDRESS=$DEPLOY_TO_DEFAULT_DIAMOND_ADDRESS PRIVATE_KEY=$(getPrivateKey "$NETWORK" "$ENVIRONMENT") DIAMOND_TYPE=$DIAMOND_TYPE forge script "$FULL_SCRIPT_PATH" -f $NETWORK -vvvv --json --silent --broadcast --skip-simulation --legacy)
+    RAW_RETURN_DATA=$(DEPLOYSALT=$DEPLOYSALT NETWORK=$NETWORK FILE_SUFFIX=$FILE_SUFFIX DEFAULT_DIAMOND_ADDRESS_DEPLOYSALT=$DEFAULT_DIAMOND_ADDRESS_DEPLOYSALT DEPLOY_TO_DEFAULT_DIAMOND_ADDRESS=$DEPLOY_TO_DEFAULT_DIAMOND_ADDRESS ETH_KEYSTORE_ACCOUNT=$(getAccount "$NETWORK" "$ENVIRONMENT") PASSWORD=$PASSWORD DIAMOND_TYPE=$DIAMOND_TYPE forge script "$FULL_SCRIPT_PATH" -f $NETWORK -vvvv --json --silent --broadcast --skip-simulation --legacy)
```
0xDEnYO, line 192: I am missing an update in the actual deploy scripts, e.g. here:
https://github.com/lifinance/contracts/blob/main/script/deploy/facets/utils/ScriptBase.sol

This file still expects an env variable PRIVATE_KEY but this will now only contain the name under which the privKey is stored, not the privKey itself. The script will try to generate a deployerAddress from it. It doesnt look right to me.

script/helperFunctions.sh
```diff
@@ -722,7 +722,7 @@ function saveDiamondPeriphery_MULTICALL_NOT_IN_USE() {
   while [ $attempts -lt 11 ]; do
     echo "Trying to execute multicall now - attempt ${attempts}"
     # try to execute call
-    MULTICALL_RESULTS=$(cast send "$MULTICALL_ADDRESS" "aggregate((address,bytes)[]) returns (uint256,bytes[])" "$MULTICALL_DATA" --private-key $(getPrivateKey "$NETWORK" "$ENVIRONMENT") --rpc-url "https://polygon-rpc.com" --legacy)
+    MULTICALL_RESULTS=$(cast send "$MULTICALL_ADDRESS" "aggregate((address,bytes)[]) returns (uint256,bytes[])" "$MULTICALL_DATA" --account $(getAccount "$NETWORK" "$ENVIRONMENT") --rpc-url "https://polygon-rpc.com" --legacy)
```
0xDEnYO, line 725: no need for --password here?

script/helperFunctions.sh
```diff
@@ -738,7 +738,7 @@ function saveDiamondPeriphery_MULTICALL_NOT_IN_USE() {
     exit 1
   fi
 
-  #MULTICALL_RESULTS=$(cast send "$MULTICALL_ADDRESS" "aggregate((address,bytes)[]) returns (uint256,bytes[])" "$MULTICALL_DATA" --private-key "$PRIV_KEY" --rpc-url  "https://opt-mainnet.g.alchemy.com/v2/4y-BIUvj_mTGWHrsHZncoJyNolNjJrsT" --legacy)
+  #MULTICALL_RESULTS=$(cast send "$MULTICALL_ADDRESS" "aggregate((address,bytes)[]) returns (uint256,bytes[])" "$MULTICALL_DATA" --account "$PRIV_KEY" --rpc-url  "https://opt-mainnet.g.alchemy.com/v2/4y-BIUvj_mTGWHrsHZncoJyNolNjJrsT" --legacy)
```
0xDEnYO, line 741: could remove this I guess

script/helperFunctions.sh
```diff
@@ -1541,10 +1541,10 @@ function removeFacetFromDiamond() {
     # call diamond
     if [[ "$DEBUG" == *"true"* ]]; then
       # print output to console
-      cast send "$DIAMOND_ADDRESS" "$ENCODED_ARGS" --private-key "$(getPrivateKey "$NETWORK" "$ENVIRONMENT")" --rpc-url "${!RPC}" --legacy
+      cast send "$DIAMOND_ADDRESS" "$ENCODED_ARGS" --account "$(getAccount "$NETWORK" "$ENVIRONMENT")" --rpc-url "${!RPC}" --legacy
```
0xDEnYO, line 1544: no need for --password?

script/helperFunctions.sh
```diff
@@ -1541,10 +1541,10 @@ function removeFacetFromDiamond() {
...
     else
       # do not print output to console
-      cast send "$DIAMOND_ADDRESS" "$ENCODED_ARGS" --private-key "$(getPrivateKey "$NETWORK" "$ENVIRONMENT")" --rpc-url "${!RPC}" --legacy >/dev/null 2>&1
+      cast send "$DIAMOND_ADDRESS" "$ENCODED_ARGS" --account "$(getAccount "$NETWORK" "$ENVIRONMENT")" --rpc-url "${!RPC}" --legacy >/dev/null 2>&1
```
0xDEnYO, line 1547: no need for --password?

script/helperFunctions.sh
```diff
@@ -1583,7 +1583,7 @@ function confirmOwnershipTransfer() {
   while [ $attempts -lt "$MAX_ATTEMPTS_PER_SCRIPT_EXECUTION" ]; do
     echo "Trying to confirm ownership transfer on contract with address ($address) - attempt ${attempts}"
     # try to execute call
-    cast send "$address" "confirmOwnershipTransfer()" --rpc-url "$rpc_url" --private-key "$private_key" 2>/dev/null
+    cast send "$address" "confirmOwnershipTransfer()" --rpc-url "$rpc_url" --account "$private_key" 2>/dev/null
```
0xDEnYO, line 1586: no need for --password?

script/tasks/diamondMakeImmutable.sh
```diff
@@ -17,22 +17,21 @@ diamondMakeImmutable() {
...
   # get diamond address from path (finds any key that contains "LiFiDiamondImmutable"
-	DIAMOND_ADDRESS=$(jq -r '.LiFiDiamondImmutable' "$ADDRS")
+  DIAMOND_ADDRESS=$(jq -r '.LiFiDiamondImmutable' "$ADDRS")
 
-	# check if all required env variables have values
-	if [[ -z "$PRIVATE_KEY_REFUND_WALLET" || -z "$PRIVATE_KEY_WITHDRAW_WALLET" ]]; then
-	  error "your .env file is missing either PRIVATE_KEY_REFUND_WALLET or PRIVATE_KEY_WITHDRAW_WALLET values. Script cannot continue."
-	  exit 1
+  # check if all required env variables have values
+  if [[ -z "$PRIVATE_KEY_REFUND_WALLET" || -z "$PRIVATE_KEY_WITHDRAW_WALLET" ]]; then
```
0xDEnYO, line 26: I think this needs to be updated, too, in order to work with the new solution, since those values also will just contain the name under which the privateKey is stored in the key store. 

I think a helper function getPrivateKey() could still helpful for such cases.

script/tasks/diamondMakeImmutable.sh
```diff
@@ -42,17 +41,14 @@ diamondMakeImmutable() {
...
-	attempts=1
+  attempts=1
 
-  while [ $attempts -lt 11 ]
-  do
+  while [ $attempts -lt 11 ]; do
     echo "[info] >>>>>>>>>>>>>>>>>>>>>>>>>>>> Trying to transfer ownership of periphery contracts now - attempt ${attempts}"
     # try to execute call
-    RAW_RETURN_DATA=$(NETWORK=$NETWORK FILE_SUFFIX=$FILE_SUFFIX PRIVATE_KEY=$(getPrivateKey "$NETWORK" "$ENVIRONMENT") forge script script/tasks/solidity/TransferOwnershipOfPeripheryContractsForImmutable.s.sol -f $NETWORK -vvvv --json --silent --broadcast --verify --skip-simulation --legacy --tc DeployScript)
+    RAW_RETURN_DATA=$(NETWORK=$NETWORK FILE_SUFFIX=$FILE_SUFFIX PRIVATE_KEY=$(getAccount "$NETWORK" "$ENVIRONMENT") forge script script/tasks/solidity/TransferOwnershipOfPeripheryContractsForImmutable.s.sol -f $NETWORK -vvvv --json --silent --broadcast --verify --skip-simulation --legacy --tc DeployScript)
```
0xDEnYO, line 51: Seems wrong...

PRIVATE_KEY will only contain the name of the stored key, not the private key itself

I think we should differentiate the calls where we use --account (>> use getAccount helper function and add --password flag) and the calls where we need the actual privateKey (>> use getPrivateKey helper function and no --password flag)

script/tasks/diamondMakeImmutable.sh
```diff
@@ -91,11 +86,10 @@ diamondMakeImmutable() {
   echo "PART 2 - SETTER FUNCTION REMOVAL IN FACETS"
   # remove all function selectors that will be unusable in the immutable diamond
   attempts=1
-  while [ $attempts -lt 11 ]
-  do
+  while [ $attempts -lt 11 ]; do
     echo "[info] >>>>>>>>>>>>>>>>>>>>>>>>>>>> Trying to remove unusable function selectors from (pre-)immutable diamond ($DIAMOND_ADDRESS) now - attempt ${attempts}"
     # try to execute call
-    RAW_RETURN_DATA=$(NETWORK=$NETWORK SALT="" FILE_SUFFIX=$FILE_SUFFIX USE_DEF_DIAMOND=false PRIVATE_KEY=$(getPrivateKey "$NETWORK" "$ENVIRONMENT") NO_BROADCAST=false forge script script/tasks/solidity/RemoveUnusableSelectorsFromImmutableDiamond.s.sol -f $NETWORK -vvvv --json --silent --broadcast --verify --skip-simulation --legacy --tc DeployScript)
+    RAW_RETURN_DATA=$(NETWORK=$NETWORK SALT="" FILE_SUFFIX=$FILE_SUFFIX USE_DEF_DIAMOND=false PRIVATE_KEY=$(getAccount "$NETWORK" "$ENVIRONMENT") NO_BROADCAST=false forge script script/tasks/solidity/RemoveUnusableSelectorsFromImmutableDiamond.s.sol -f $NETWORK -vvvv --json --silent --broadcast --verify --skip-simulation --legacy --tc DeployScript)
```
0xDEnYO, line 92: Seems like we dont even need to send an ENV-variable PRIVATE_KEY here since you updated the script to obtain the privKey from the storage, right?

(that might also apply to comments above)

script/tasks/diamondMakeImmutable.sh
```diff
@@ -300,11 +290,10 @@ diamondMakeImmutable() {
   # execute selected script
   attempts=1
 
-  while [ $attempts -lt 11 ]
-  do
+  while [ $attempts -lt 11 ]; do
     echo "[info] >>>>>>>>>>>>>>>>>>>>>>>>>>>> Trying to remove DiamondCutFacet from diamond $DIAMOND_ADDRESS and transfer ownership to address(0) now - attempt ${attempts}"
     # try to execute call
-    RAW_RETURN_DATA=$(NETWORK=$NETWORK SALT="" FILE_SUFFIX=$FILE_SUFFIX USE_DEF_DIAMOND=false PRIVATE_KEY=$(getPrivateKey "$NETWORK" "$ENVIRONMENT") NO_BROADCAST=false forge script script/tasks/solidity/MakeLiFiDiamondImmutable.s.sol -f $NETWORK -vvvv --json --silent --broadcast --verify --skip-simulation --legacy --tc DeployScript)
+    RAW_RETURN_DATA=$(NETWORK=$NETWORK SALT="" FILE_SUFFIX=$FILE_SUFFIX USE_DEF_DIAMOND=false ETH_KEYSTORE_ACCOUNT=$(getAccount "$NETWORK" "$ENVIRONMENT") NO_BROADCAST=false forge script script/tasks/solidity/MakeLiFiDiamondImmutable.s.sol -f $NETWORK -vvvv --json --silent --broadcast --verify --skip-simulation --legacy --tc DeployScript)
```
0xDEnYO, line 296: no --password flag needed?

script/tasks/diamondSyncDEXs.sh
```diff
@@ -120,10 +120,10 @@ function diamondSyncDEXs {
         # call diamond
         if [[ "$DEBUG" == *"true"* ]]; then
           # print output to console
-          cast send "$DIAMOND_ADDRESS" "batchAddDex(address[])" "${PARAMS[@]}" --rpc-url "$RPC_URL" --private-key $(getPrivateKey "$NETWORK" "$ENVIRONMENT") --legacy
+          cast send "$DIAMOND_ADDRESS" "batchAddDex(address[])" "${PARAMS[@]}" --rpc-url "$RPC_URL" --account $(getAccount "$NETWORK" "$ENVIRONMENT") --legacy
```
0xDEnYO, line 123: no --password flag needed?

same applies for all other cast send calls in this file

script/tasks/diamondSyncSigs.sh
```diff
@@ -94,10 +94,10 @@ function diamondSyncSigs {
       # call diamond
       if [[ "$DEBUG" == *"true"* ]]; then
         # print output to console
-        cast send "$DIAMOND_ADDRESS" "batchSetFunctionApprovalBySignature(bytes4[],bool)" "[${PARAMS::${#PARAMS}-1}]" true --rpc-url $RPC_URL --private-key $(getPrivateKey "$NETWORK" "$ENVIRONMENT") --legacy
+        cast send "$DIAMOND_ADDRESS" "batchSetFunctionApprovalBySignature(bytes4[],bool)" "[${PARAMS::${#PARAMS}-1}]" true --rpc-url $RPC_URL --account $(getAccount "$NETWORK" "$ENVIRONMENT") --legacy
```
0xDEnYO, line 97: no --password flag needed?

same applies for all other cast send calls in this file

script/tasks/diamondUpdateFacet.sh
```diff
@@ -108,10 +108,10 @@ diamondUpdateFacet() {
...
-      RAW_RETURN_DATA=$(NETWORK=$NETWORK FILE_SUFFIX=$FILE_SUFFIX USE_DEF_DIAMOND=$USE_MUTABLE_DIAMOND NO_BROADCAST=false PRIVATE_KEY=$(getPrivateKey "$NETWORK" "$ENVIRONMENT") forge script "$SCRIPT_PATH" -f $NETWORK -vvvv --json --silent --broadcast --skip-simulation --legacy)
+      RAW_RETURN_DATA=$(NETWORK=$NETWORK FILE_SUFFIX=$FILE_SUFFIX USE_DEF_DIAMOND=$USE_MUTABLE_DIAMOND NO_BROADCAST=false PRIVATE_KEY=$(getAccount "$NETWORK" "$ENVIRONMENT") forge script "$SCRIPT_PATH" -f $NETWORK -vvvv --json --silent --broadcast --skip-simulation --legacy)
     else
       # do not print output to console
-      RAW_RETURN_DATA=$(NETWORK=$NETWORK FILE_SUFFIX=$FILE_SUFFIX USE_DEF_DIAMOND=$USE_MUTABLE_DIAMOND NO_BROADCAST=false PRIVATE_KEY=$(getPrivateKey "$NETWORK" "$ENVIRONMENT") forge script "$SCRIPT_PATH" -f $NETWORK -vvvv --json --silent --broadcast --skip-simulation --legacy) 2>/dev/null
+      RAW_RETURN_DATA=$(NETWORK=$NETWORK FILE_SUFFIX=$FILE_SUFFIX USE_DEF_DIAMOND=$USE_MUTABLE_DIAMOND NO_BROADCAST=false PRIVATE_KEY=$(getAccount "$NETWORK" "$ENVIRONMENT") forge script "$SCRIPT_PATH" -f $NETWORK -vvvv --json --silent --broadcast --skip-simulation --legacy) 2>/dev/null
```
0xDEnYO, line 114: This seems wrong.
PRIVATE_KEY will contain the name of the stored key, not the key itself. Without an update of the ScriptBase.sol, this should imho not work.

script/tasks/diamondUpdatePeriphery.sh
```diff
@@ -185,11 +185,11 @@ register() {
       doNotContinueUnlessGasIsBelowThreshold "$NETWORK"
 
       # print output to console
-      cast send "$DIAMOND" 'registerPeripheryContract(string,address)' "$CONTRACT_NAME" "$ADDR" --private-key $(getPrivateKey "$NETWORK" "$ENVIRONMENT") --rpc-url "${!RPC}" --legacy
-      #      cast send 0xd37c412F1a782332a91d183052427a5336438cD3 'registerPeripheryContract(string,address)' "Executor" "0x68895782994F1d7eE13AD210b63B66c81ec7F772" --private-key "0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80" --rpc-url "${!RPC}" --legacy
+      cast send "$DIAMOND" 'registerPeripheryContract(string,address)' "$CONTRACT_NAME" "$ADDR" --account $(getAccount "$NETWORK" "$ENVIRONMENT") --rpc-url "${!RPC}" --legacy
```
0xDEnYO, line 188: no --password flag needed?

same applies for all other cast send calls in this file

0xDEnYO (changes requested):

ezynda3:
Going to close this as I think because I think it would be more efficient to include this as part of the V2 of our deployment process.

PR #569, opened 2023-12-17
Validate data offset positions in calldata to prevent malicious calldata

ezynda3:
We were made aware of a weakness in our verification facet by a partner that uses it in their own contracts. A finding was submitted as part of an audit contest which explains the vulnerability in detail here: https://superform.notion.site/A-malicious-user-can-craft-valid-calldata-to-call-packed-version-of-some-of-LiFi-endpoints-177-e35ac227c89143ee8d6f0dcb3580ff47


ezynda3:
The contest is being run on Cantina which is run by SpearBit so will wait to hear what they think before merging this.

PR #579, opened 2024-01-19
Implement Non Standard Selectors Registry Facet

ezynda3:


0xDEnYO (approved):

test/solidity/Facets/NonStandardSelectorsRegistryFacet.t.sol
```diff
@@ -0,0 +1,56 @@
+// SPDX-License-Identifier: MIT
```
0xDEnYO, line 1: Tests for the batch function of this contract are missing. To increase test coverage, it would make sense to add them.

src/Facets/NonStandardSelectorsRegistryFacet.sol
```diff
@@ -0,0 +1,68 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.17;
+
+import { LibDiamond } from "../Libraries/LibDiamond.sol";
+
+/// @title Non Standard Selectors Registry Facet
+/// @author LIFI (https://li.finance)
+/// @notice Registry for non-standard selectors
+/// @custom:version 1.0.0
```
0xDEnYO, line 9: According to our convention the version of this facet should be in the PR title

test/solidity/Facets/NonStandardSelectorsRegistryFacet.t.sol
ezynda3: The batch test is on lines 35-48

PR #582, opened 2024-01-21
Revamped deploy scripts

ezynda3:


script/modules/deployContract.mjs
```diff
@@ -0,0 +1,106 @@
...
+    initial: 0,
+  })
+
+  // Get the contract item from the object array
+  const deployScript = deployScripts.find((c) => c.name === choice)
+
+  // Using fetch get the LIFI supprted chains from https://li.quest/v1/chains
+  // then map the result to an object by extracting chains[i].name and chains[i].metamask.rpcUrls[0]
+  let result = await spinner('Fetching LIFI supported chains', async () =>
+    (await fetch('https://li.quest/v1/chains')).json()
```
0xDEnYO, line 51: I think this  potentially would only return the currently enabled chains and not allow us to deploy to chains that are currently disabled or entirely new.

script/modules/deployContract.mjs
```diff
@@ -0,0 +1,106 @@
...
+    initial: 0,
+  })
+
+  // Get the chain item from the object array
+  const chain = chains.find((c) => c.name === chainChoice)
+
+  // Run the deploy script with forge e.g await $`forge script script/deploy/facets/DeploySomeScript.sol`
+  // TODO: Add the account parameter to the forge script command
+  try {
+    process.env.PRIVATE_KEY = DEPLOYER_PRIVATE_KEY
```
0xDEnYO, line 70: Looks like we dont have any repeat functionality here. I think this would be important for stable execution.

script/deploy/facets/DeployDummy.sol
```diff
@@ -0,0 +1,29 @@
+// SPDX-License-Identifier: UNLICENSED
```
0xDEnYO, line 1: not sure what this file is needed for...!?

ezynda3: This was just s test to make sure I could get forge working. The other scripts keep failing because we do a lot of config via ENV vars which is pretty messy. This won't stay in the final version.

script/modules/deployContract.mjs
ezynda3: Yes. Planning to add it.

script/modules/deployContract.mjs
ezynda3: Was thinking to add `Custom` as an option.

PR #585, opened 2024-01-25
Update fee collectors ServiceFeeCollector v1.0.1, LiFuelFeeCollector v1.0.1, LIFuelFacet v1.0.1

ezynda3:


src/Periphery/LiFuelFeeCollector.sol
```diff
@@ -7,7 +7,7 @@ import { TransferrableOwnership } from "../Helpers/TransferrableOwnership.sol";
 /// @title LiFuelFeeCollector
 /// @author LI.FI (https://li.fi)
 /// @notice Provides functionality for collecting fees for LiFuel
-/// @custom:version 1.0.0
+/// @custom:version 1.0.1
```
0xDEnYO, line 10: need to update PR title, too, please

src/Periphery/ServiceFeeCollector.sol
```diff
@@ -7,7 +7,7 @@ import { TransferrableOwnership } from "../Helpers/TransferrableOwnership.sol";
 /// @title Service Fee Collector
 /// @author LI.FI (https://li.fi)
 /// @notice Provides functionality for collecting service fees (gas/insurance)
-/// @custom:version 1.0.0
+/// @custom:version 1.0.1
```
0xDEnYO, line 10: same as above

test/solidity/Periphery/ServiceFeeCollector.t.sol
```diff
@@ -50,7 +50,11 @@ contract ServiceFeeCollectorTest is DSTest {
         uint256 fee = 0.015 ether;
 
         // Act
-        feeCollector.collectNativeGasFees{ value: fee }(137, address(0xb33f));
+        feeCollector.collectNativeGasFees{ value: fee }(
```
0xDEnYO, line 53: I think it could be nice to have a test case that tests the behaviour we just had issues with to ensure that things work as intended

src/Periphery/LiFuelFeeCollector.sol
```diff
@@ -53,15 +53,17 @@ contract LiFuelFeeCollector is TransferrableOwnership {
...
         address receiver
     ) external payable {
         emit GasFeesCollected(
             LibAsset.NULL_ADDRESS,
             chainId,
             receiver,
-            msg.value
+            feeAmount
         );
+        msg.sender.call{ value: msg.value - feeAmount }("");
```
maxklenk, line 66: Add check to validate if there is a amount to send back. e.g. the lifuel bridge sends the exact amount needed.

src/Periphery/ServiceFeeCollector.sol
```diff
@@ -68,6 +69,7 @@ contract ServiceFeeCollector is TransferrableOwnership {
             receiver,
             msg.value
         );
+        msg.sender.call{ value: msg.value - feeAmount }("");
```
maxklenk, line 72: Also check here

maxklenk (changes requested):

src/Periphery/ServiceFeeCollector.sol
```diff
@@ -68,6 +69,10 @@ contract ServiceFeeCollector is TransferrableOwnership {
             receiver,
             msg.value
         );
```
maxklenk, line 71: Why do we need that function still? We collect gas fees in the LiFuelFeeCollector.
Also the event emits `msg.value` instead of the `feeAmount`.

maxklenk (changes requested):
Please update the `LIFuelFacet.sol` The version should be increased and the linked Fee collector has to be updated as we have a dedicated lifuel fee collector.
Update line 20
`string internal constant FEE_COLLECTOR_NAME = "ServiceFeeCollector";`

maxklenk (approved):
Nice, that looks like what I expected.

PR #593, opened 2024-02-08
Implement DeBridge DLN Facet v1.0.0

ezynda3:


src/Facets/DeBridgeDlnFacet.sol
```diff
@@ -0,0 +1,182 @@
...
+contract DeBridgeDlnFacet is ILiFi, ReentrancyGuard, SwapperV2, Validatable {
+    /// Storage ///
+
+    address internal constant NON_EVM_ADDRESS =
+        0x11f111f111f111F111f111f111F111f111f111F1;
+    IDlnSource public immutable dlnSource;
+
+    /// Types ///
+
+    // struct BridgeData {
```
0xDEnYO, line 26: remove?

src/Facets/DeBridgeDlnFacet.sol
```diff
@@ -0,0 +1,182 @@
...
+    //     address referrer;
+    //     address sendingAssetId;
+    //     address receiver;
+    //     uint256 minAmount;
+    //     uint256 destinationChainId;
+    //     bool hasSourceSwaps;
+    //     bool hasDestinationCall;
+    // }
+
+    /// @dev Optional bridge specific struct
```
0xDEnYO, line 39: are they all optional?? If yes, a description in which case they are needed would be good

and param descriptions are missing

src/Facets/DeBridgeDlnFacet.sol
```diff
@@ -0,0 +1,182 @@
...
+    /// Internal Methods ///
+
+    /// @dev Contains the business logic for the bridge via DeBridgeDLN
+    /// @param _bridgeData The core information needed for bridging
+    /// @param _deBridgeDlnData Data specific to DeBridgeDLN
+    function _startBridge(
+        ILiFi.BridgeData memory _bridgeData,
+        DeBridgeDlnData calldata _deBridgeDlnData
+    ) internal {
+        uint256 fee = dlnSource.globalFixedNativeFee();
```
0xDEnYO, line 127: This is done twice for swapAnd.. cases. Or pass it into _startBridge() as parameter (and pull it in both startBridgeTokensViaDeBridgeDln and swapAndStartBridgeTokensViaDeBridgeDln) to avoid having to do it twice.
 
Or could even consider to move this into backend and make it a parameter of the DeBridgeDlnData

src/Facets/DeBridgeDlnFacet.sol
```diff
@@ -0,0 +1,182 @@
...
+        IDlnSource.OrderCreation memory orderCreation = IDlnSource
+            .OrderCreation({
+                giveTokenAddress: _bridgeData.sendingAssetId,
+                giveAmount: _bridgeData.minAmount,
+                takeTokenAddress: abi.encodePacked(
```
0xDEnYO, line 132: This encoding could be done in the backend and instead of address parameter we use bytes parameter. Saves the (unnecessary) encoding step on-chain

test/solidity/Facets/DeBridgeDlnFacet.t.sol
```diff
@@ -0,0 +1,182 @@
...
+            minAmountOut: (defaultUSDCAmount * 95) / 100
+        });
+
+        vm.label(address(DLN_SOURCE), "DLN_SOURCE");
+        FIXED_FEE = DLN_SOURCE.globalFixedNativeFee();
+    }
+
+    function initiateBridgeTxWithFacet(bool isNative) internal override {
+        if (isNative) {
+            validDeBridgeDlnData.minAmountOut =
```
0xDEnYO, line 80: would be helpful to have a comment here explaining the rationale behind this code (why are we adding slippage here and why only for native?)

test/solidity/Facets/DeBridgeDlnFacet.t.sol
```diff
@@ -0,0 +1,182 @@
...
+                value: defaultNativeAmount + FIXED_FEE
+            }(bridgeData, swapData, validDeBridgeDlnData);
+        } else {
+            deBridgeDlnFacet.swapAndStartBridgeTokensViaDeBridgeDln{
+                value: FIXED_FEE
+            }(bridgeData, swapData, validDeBridgeDlnData);
+        }
+    }
+
+    function testBase_CanSwapAndBridgeTokensFromNative()
```
0xDEnYO, line 107: This is not a "base" test and therefore it should be renamed

0xDEnYO (changes requested):

0xDEnYO (approved):

PR #600, opened 2024-02-27
fix: Update amarok config

0xDEnYO:


maxklenk:
AmarokFacetPacked needs to be deployed to Metis as well

ezynda3 (approved):

PR #604, opened 2024-03-05
feat: Token wrapper contract

H3xept:
Contract that will interact with external token wrappers and behave like a DEX in our generic swap flow.


config/tokenwrapper.json
```diff
@@ -0,0 +1,5 @@
+{
+    "hardhat": "0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2",
+    "mainnet": "0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2",
+    "polygon": "0x0d500B1d8E8eF31E21C99d1Db9A6444d3ADf1270"
```
H3xept, line 4: Needs to be extended

src/Periphery/TokenWrapper.sol
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    /// External Methods ///
+
+    /// @notice Wraps the native token
+    function deposit(
+    ) external payable {
+        IWrapper(wrappedToken).deposit{value: msg.value}();
+        IERC20(wrappedToken).transfer(msg.sender, msg.value);
+        emit Wrapped(msg.value);
```
maxklenk, line 43: Do we need events here? Our contracts already expose assetSwapped events for each executed swap. I don't think the event here is needed. It would save some gas (also `Unwrapped(data)` event)

maxklenk (approved):

src/Periphery/TokenWrapper.sol
```diff
@@ -0,0 +1,65 @@
...
+interface IWrapper {
+    function deposit() external payable;
+    function withdraw(uint wad) external;
+}
+
+/// @title TokenWrapper
+/// @author LI.FI (https://li.fi)
+/// @notice Provides functionality for wrapping and unwrapping tokens
+/// @custom:version 1.0.0
+contract TokenWrapper is TransferrableOwnership {
```
maxklenk, line 18: Why is this contract ownable? I don't see any call an owner could do.

config/tokenwrapper.json
```diff
@@ -0,0 +1,16 @@
+{
+    "hardhat": "0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2",
```
ezynda3, line 2: We don't actually use hardhat so it's not needed

src/Periphery/TokenWrapper.sol
H3xept: Other periphery contracts are ownable - I assume is to avoid re-deploy if we change diamond? I can remove it's not really needed here.

src/Periphery/TokenWrapper.sol
H3xept: Sure

ezynda3 (approved):

src/Periphery/TokenWrapper.sol
```diff
@@ -0,0 +1,59 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity 0.8.17;
+
+import { LibAsset } from "../Libraries/LibAsset.sol";
+import { TransferrableOwnership } from "../Helpers/TransferrableOwnership.sol";
```
maxklenk, line 5: Remove this as well please

config/tokenwrapper.json
```diff
@@ -0,0 +1,15 @@
...
+    "base": "0x4200000000000000000000000000000000000006",
+    "boba": "0xDeadDeAddeAddEAddeadDEaDDEAdDeaDDeAD0000",
+    "bsc": "0xbb4CdB9CBd36B01bD1cBaEBF2De08d9173bc095c",
+    "fantom": "0x21be370D5312f44cB42ce377BC9b8a0cEF1A4C83",
+    "gnosis": "0xe91D153E0b41518A2Ce8Dd3D7944Fa863463a97d",
+    "linea": "0xe5D7C2a44FfDDf6b295A15c148167daaAf5Cf34f",
+    "optimism": "0x4200000000000000000000000000000000000006",
+    "polygonzkevm": "0x4F9A0e7FD2Bf6067db6994CF12E4495Df938E6e9",
+    "zksync": "0x5AEa5775959fBC2557Cc8789bC1bf90A239D9a91",
+}
```
maxklenk, line 15: This list looks incomplete compared to the amount of chains we support. We should be able to extract the list from `wrappedTokens` in https://github.com/lifinance/data-types/blob/main/src/coins/coins.ts#L1730

src/Periphery/TokenWrapper.sol
H3xept: Too used to typescript linter removing imports 💀

PR #606, opened 2024-03-08
feat(mayan): MayanBridgeFacet v1.0.0

ezynda3:


script/demoScripts/demoMayan.ts
```diff
@@ -0,0 +1,132 @@
...
+    toToken: '0x2791Bca1f2de4661ED88A30C99A7a9449Aa84174',
+    fromChain: 'bsc',
+    toChain: 'polygon',
+    slippage: 3,
+  })
+
+  const swapFee = getAmountOfFractionalAmount(
+    quote.swapRelayerFee,
+    Math.min(8, quote.fromToken.decimals)
+  )
```
maxklenk, line 46: Why do we need to do that parsing are the parameters not present in the result of `getSwapFromEvmTxPayload`?

maxklenk (approved):
The implementation looks good, just added a few cleanup questions.

PR #608, opened 2024-03-14
Deprecate Wormhole

ezynda3:


0xDEnYO:
If I am not mistaken then we also need to adjust:

1) deploy_requirements.json
2) remove IWormholeRouter.sol

ezynda3:
> If I am not mistaken then we also need to adjust:
> 
> 1. deploy_requirements.json
> 2. remove IWormholeRouter.sol

done

0xDEnYO (approved):

PR #610, opened 2024-03-15
remove unused chains

ezynda3:


networks
```diff
@@ -5,27 +5,21 @@ avalanche
...
 boba
 bsc
-celo
-cronos
-evmos
 fantom
 fuse
 gnosis
 linea
 localanvil
```
maxklenk, line 12: maybe move this to the testnets

0xDEnYO (approved):

networks
```diff
@@ -5,27 +5,21 @@ avalanche
...
 moonriver
-okx
-opbnb
 optimism
 polygon
 polygonzkevm
-velas
+scroll
+zksync
 goerli
```
0xDEnYO, line 20: why are we keeping goerli? This network is not functional anymore. Would recommend to remove it

0xDEnYO (approved):

PR #618, opened 2024-03-20
refactor: Hop use SSLOAD2

H3xept:
⚠️ DO NOT MERGE THIS IS JUST A SHOWCASE ⚠️

Long story short: 
- SLOAD is one of the most expensive opcodes, 2100 gas per evm word
- A possible way to improve gas usage is start using smart-contract code as storage since `EXTCODECOPY` is cheaper for reads bigger than 1 word (https://www.evm.codes/#3c?fork=cancun - the caveat is that data will be write-once )

Reads: 
https://zefram.xyz/posts/how-i-almost-cheesed-the-evm/
https://github.com/0xsequence/sstore2

Ideas: 
(a) We could have facets read their configuration from contract code - this will only be efficient for big configs (not sure if we have such configs). The effect here could also be achieved by "flattening" the config into immutable variables.
(b) Create a "fork" of the diamond standard where the diamond storage is replaced by a contract-code-storage. This could lead to us reading the jump location for diamond dispatch + facet configurations in one go using EXTCODECOPY and avoiding the cost of multiple SLOADs. 

⚠️ The code of this PR is just meant as a showcase on how SSLOAD2(Map) is used (I couldnt use the simple SSLOAD version without modifying a lot of logic, which was beyond the scope of this). The actual gas used by this change is higher than what the original hop facet uses.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -19,3 +19,9 @@
 [submodule "lib/openzeppelin-contracts"]
 	path = lib/openzeppelin-contracts
 	url = https://github.com/openzeppelin/openzeppelin-contracts
+[submodule "lib/sstore2"]
+	path = lib/sstore2
+	url = https://github.com/0xsequence/sstore2
+[submodule "lib/create3"]
+	path = lib/create3
+	url = https://github.com/0xsequence/create3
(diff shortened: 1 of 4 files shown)


maxklenk:
A closed PR will be enough for reference.

PR #622, opened 2024-03-26
Generic swap facet v2

0xDEnYO:


src/Facets/GenericSwapFacet.sol
```diff
@@ -7,38 +10,261 @@ import { ReentrancyGuard } from "../Helpers/ReentrancyGuard.sol";
...
+        IERC20(_swapData.sendingAssetId).safeTransferFrom(
+            msg.sender,
+            address(this),
+            _swapData.fromAmount
+        );
+
+        // ensure that contract (callTo) and function selector are whitelisted
+        if (
+            // !((LibAsset.isNativeAsset(_swapData.sendingAssetId) || // TODO: confirm that we can safely remove this
+            // LibAllowList.contractIsAllowed(_swapData.approveTo)) && // TODO: confirm that we can safely remove this
```
ezynda3, line 54: I believe we can safely remove these in all cases.

src/Facets/GenericSwapFacet.sol
```diff
@@ -7,38 +10,261 @@ import { ReentrancyGuard } from "../Helpers/ReentrancyGuard.sol";
...
+    /// @param _minAmountOut the minimum amount of the final asset to receive
+    /// @param _swapData an object containing swap related data to perform swaps before bridging
+    function swapTokensSingleNativeToERC20(
+        bytes32 _transactionId,
+        string calldata _integrator,
+        string calldata _referrer,
+        address payable _receiver,
+        uint256 _minAmountOut,
+        LibSwap.SwapData calldata _swapData
+    ) external payable refundExcessNative(_receiver) {
```
ezynda3, line 193: I think it's still a good idea to keep this.

src/Facets/GenericSwapFacet.sol
```diff
@@ -7,38 +10,261 @@ import { ReentrancyGuard } from "../Helpers/ReentrancyGuard.sol";
...
+    /// @param _minAmountOut the minimum amount of the final asset to receive
+    /// @param _swapData an object containing swap related data to perform swaps before bridging
+    function swapTokensSingleERC20ToERC20(
+        bytes32 _transactionId,
+        string calldata _integrator,
+        string calldata _referrer,
+        address payable _receiver,
+        uint256 _minAmountOut,
+        LibSwap.SwapData calldata _swapData // TODO: do we need refundExcessNative here?
+    ) external payable returns (uint256 amountOut) {
```
ezynda3, line 43: I realize payable can reduce gas cost but maybe it's best to prevent people from accidentally sending native along with an ERC20 only TX as it will just get stuck.

ezynda3 (approved):

src/Facets/GenericSwapFacet.sol
0xDEnYO: removed

src/Facets/GenericSwapFacet.sol
0xDEnYO: Well, I guess we have to ask ourselves what we want....gas-optimized contracts or "foolproof" UX-friendly contracts. 
I feel our goal so far was more to be on the gas-optimized side since a) this is what we compete with in the market and b) we expect people to interact with our contracts using calldata prepared by our backend anyway
It's not like a Uniswap contract that is integrated by thousands of contracts.
Also we have ways to refund for the unusual case that someone accidentally sends a significant amount.

I put it there for a reason and would vouch to keep it. But I am happy to hear your opinion and find agreement on the way forward with you.

Do you have a strong opinion on this topic? What's your take?
@ezynda3

src/Facets/GenericSwapFacet.sol
0xDEnYO: ok, will do.
Do we need it for both cases that involve a native (ERC>Native , Native > ERC) or only for one of them?
@ezynda3

src/Facets/GenericSwapFacet.sol
0xDEnYO: We've agreed to leave it in there and prioritize gas savings

src/Facets/GenericSwapFacet.sol
```diff
@@ -3,42 +3,328 @@ pragma solidity 0.8.17;
...
+
+    event AssetSwapped(
+        bytes32 transactionId,
+        address dex,
+        address fromAssetId,
+        address toAssetId,
+        uint256 fromAmount,
+        uint256 toAmount,
+        uint256 timestamp
+    );
```
maxklenk, line 34: We should use `LibSwap.AssetSwapped` instead to make sure we emit the same event.

src/Facets/GenericSwapFacet.sol
```diff
@@ -3,42 +3,328 @@ pragma solidity 0.8.17;
...
-/// @title Generic Swap Facet
+/// @title GenericSwapFacet
 /// @author LI.FI (https://li.fi)
 /// @notice Provides functionality for swapping through ANY APPROVED DEX
-/// @dev Uses calldata to execute APPROVED arbitrary methods on DEXs
-/// @custom:version 1.0.0
+/// @dev Can only execute calldata for APPROVED function selectors
+/// @custom:version 2.0.0
 contract GenericSwapFacet is ILiFi, ReentrancyGuard, SwapperV2, Validatable {
+    using SafeERC20 for IERC20;
```
maxklenk, line 22: Why don't we use `import { ERC20, SafeTransferLib } from "solmate/utils/SafeTransferLib.sol";` as in the packed facets?

src/Facets/GenericSwapFacet.sol
```diff
@@ -3,42 +3,328 @@ pragma solidity 0.8.17;
...
+                address(this),
+                _swapData.approveTo
+            ) < _swapData.fromAmount
+        ) {
+            // allowance insufficient - register max approval
+            SafeERC20.safeApprove(
+                IERC20(_swapData.sendingAssetId),
+                _swapData.approveTo,
+                0
+            );
```
maxklenk, line 83: Ideally this is never needed as we have either no approval or max approval. Does it make sense to handle that case explicitly
```
if (approval === 0)
  set max
else if (approval < fromAmount)
  reset + set max
```

src/Facets/GenericSwapFacet.sol
```diff
@@ -3,42 +3,328 @@ pragma solidity 0.8.17;
...
 /// @author LI.FI (https://li.fi)
 /// @notice Provides functionality for swapping through ANY APPROVED DEX
-/// @dev Uses calldata to execute APPROVED arbitrary methods on DEXs
-/// @custom:version 1.0.0
+/// @dev Can only execute calldata for APPROVED function selectors
+/// @custom:version 2.0.0
 contract GenericSwapFacet is ILiFi, ReentrancyGuard, SwapperV2, Validatable {
+    using SafeERC20 for IERC20;
+
+    error InsufficientSwapOutput();
```
maxklenk, line 24: Can be removed as it is not used.

src/Facets/GenericSwapFacet.sol
```diff
@@ -3,42 +3,328 @@ pragma solidity 0.8.17;
...
+        // ensure that contract (callTo) and function selector are whitelisted
+        if (
+            !(LibAllowList.contractIsAllowed(_swapData.callTo) &&
+                LibAllowList.selectorIsAllowed(bytes4(_swapData.callData[:4])))
+        ) revert ContractCallNotAllowed();
+
+        // get initial token balance
+        uint256 initialBalance = IERC20(_swapData.receivingAssetId).balanceOf(
+            address(this)
+        );
```
maxklenk, line 69: Can be removed as discussed.

src/Facets/GenericSwapFacet.sol
```diff
@@ -3,42 +3,328 @@ pragma solidity 0.8.17;
...
+        if (!success) revert NativeAssetTransferFailed();
+
+        // emit events (both required for tracking)
+        emit AssetSwapped(
+            _transactionId,
+            _swapData.callTo,
+            _swapData.sendingAssetId,
+            _swapData.receivingAssetId,
+            _swapData.fromAmount,
+            _minAmountOut,
```
maxklenk, line 216: In `LibSwap` we emit here:
```
newBalance > initialReceivingAssetBalance
  ? newBalance - initialReceivingAssetBalance
  : newBalance
```

To have the same expected value it should be `amountReceived` here.

src/Facets/GenericSwapFacet.sol
```diff
@@ -3,42 +3,328 @@ pragma solidity 0.8.17;
...
+            _swapData.sendingAssetId,
+            _swapData.receivingAssetId,
+            _swapData.fromAmount,
+            amountReceived
+        );
+    }
+
+    //----------------------------------------------------------------------------------------------------------------
+    //----------------------------------------------------------------------------------------------------------------
+    //----------------------------------------------------------------------------------------------------------------
```
maxklenk, line 308: What are those lines for?

maxklenk (commented):
Some small improvments

src/Facets/GenericSwapFacet.sol
0xDEnYO: I dont think that really makes a difference (for the event itself) as long as the event definitions are exactly the same (which they are). But I will update it to make the relationship clearer.

src/Facets/GenericSwapFacet.sol
0xDEnYO: We use both implementations in various places. Never thought about it really. 
I ran some tests...the solmate implementation is more gas-efficient. I will change it and also write down that we will only use solmate for future implementations. 
Good feedback, thank you !

src/Facets/GenericSwapFacet.sol
0xDEnYO: yes I like that. Thank you.

src/Facets/GenericSwapFacet.sol
0xDEnYO: They were just for me, differentiating between v1 and v2. 
Removed

src/Facets/GenericSwapFacet.sol
0xDEnYO: Good spotting...updated.

src/Facets/GenericSwapFacet.sol
maxklenk: In both cases the exposed event would be the same. It just keeps the code cleaner and we avoid duplication or mismatches. 

Please remove the event here.

src/Facets/GenericSwapFacet.sol
maxklenk: Still shows up

maxklenk (approved):

src/Facets/GenericSwapFacet.sol
```diff
@@ -3,42 +3,232 @@ pragma solidity 0.8.17;
...
+            revert CumulativeSlippageTooHigh(_minAmountOut, amountReceived);
+
+        // transfer funds to receiver
+        // solhint-disable-next-line avoid-low-level-calls
+        (bool success, ) = _receiver.call{ value: amountReceived }("");
+        if (!success) revert NativeAssetTransferFailed();
+
+        // emit events (both required for tracking)
+        emit LibSwap.AssetSwapped(
+            _transactionId,
```
ezynda3, line 124: nit: I would move this to the internal `_depositAndSwapERC20` method to be more consistent with the rest of the codebase.

ezynda3 (approved):

src/Facets/GenericSwapFacet.sol
0xDEnYO: Can't do this as it needs the amountOut which varies per function.

PR #640, opened 2024-05-02
feat(GasZip): GasZip 1.0.0

H3xept:


config/gaszip.json
maxklenk, line 1: Needs to be synced with https://dev.gas.zip/gas/chain-support/inbound

src/Periphery/GasZip.sol
```diff
@@ -0,0 +1,78 @@
...
+    modifier inboundTokenIsAllowed(address token) {
+        if (!allowedInboundTokens[token]) revert InboundTokenDisallowed();
+        _;
+    }
+
+    constructor(
+        address _owner,
+        address _gasZipRouter
+    ) TransferrableOwnership(_owner) {
+        gasZipRouter = IGasZip(_gasZipRouter);
```
maxklenk, line 46: should check that address is not `ZERO`

0xDEnYO: our deploy scripts will check that the parameter is available and not a zero address. 
But for this the file ```deployRequirements.json``` need to be updated respectively

config/gaszip.json
0xDEnYO, line 1: the doc file is missing for this contract (check folder ```/docs```)

0xDEnYO, line 1: Also this contract does not have any test cases. Please add unit tests.

melianessa: Updated

src/Periphery/GasZip.sol
melianessa: @0xDEnYO updated deployRequirements. Please check 🙂
Also, please advice if we need to define it in DeployGasZip, or is it enough to have both `_owner` and `_gasZipRouter` defined in deployRequirements?
image

config/gaszip.json
```diff
@@ -0,0 +1,20 @@
+{
+  "withdrawWallet": "0x0000000000000000000000000000000000000000",
```
H3xept, line 2: I think you can reuse the global config one, can remove from here

src/Periphery/GasZip.sol
```diff
@@ -0,0 +1,68 @@
...
+        uint256 destinationChain,
+        address recipient
+    ) public {
+        LibSwap.swap(0, _swap);
+        uint256 availableNative = LibAsset.getOwnBalance(ZERO);
+        gasZipRouter.deposit{ value: availableNative }(
+            destinationChain,
+            recipient
+        );
+        (bool success, ) = msg.sender.call{ value: address(this).balance }("");
```
H3xept, line 52: this only sends native back, presumably we should send back the source sending asset from swap

H3xept (commented):
Comments

src/Periphery/GasZip.sol
maxklenk: Only the `gasZipRouters`. You can remove all the owner handling because the inboundToken handling is no longer needed.

PR #645, opened 2024-05-14
Add convenience scripts for dealing with SAFE transactions

ezynda3:


script/deploy/deployUpgradesToSAFE.sh
```diff
@@ -31,24 +31,9 @@ deployUpgradesToSAFE() {
     CLEAN_RETURN_DATA=$(echo $RAW_RETURN_DATA | sed 's/^.*{\"logs/{\"logs/')
     FACET_CUT=$(echo $CLEAN_RETURN_DATA | jq -r '.returns.cutData.value')
     if [ "$FACET_CUT" != "0x" ]; then
-      CUTS+=("$FACET_CUT")
+      echo "Proposing facet cut for $script..."
+      ts-node script/deploy/safe/propose-to-safe.ts --to "$DIAMOND_ADDRESS" --calldata "$FACET_CUT" --network "$NETWORK" --rpcUrl $(getRPCUrl $NETWORK) --privateKey "$PRIVATE_KEY"
```
maxklenk, line 35: Would it make sense to propose upgrades as a dev instead of using the deployer `PRIVATE_KEY` here. I think another env variable for the developers key would make sense.

package.json
```diff
@@ -75,25 +76,31 @@
     "execute": "node ./_scripts.js run",
     "abi:generate": "make clean && forge build --skip script --skip test --skip Base --skip Test && hardhat diamondABI",
     "typechain": "make clean && forge build --skip script --skip test --skip Base --skip Test && typechain --target ethers-v5 'out/**/*.json' --out-dir typechain",
-    "codegen": "plop"
+    "codegen": "plop",
+    "propose-safe-tx": "ts-node ./script/deploy/safe/propose-to-safe.ts",
+    "confirm-safe-tx": "ts-node ./script/deploy/safe/confirm-safe-tx.ts",
```
maxklenk, line 81: A helper to load environment variables would be nice. e.g. loading the private key and selecting the chain in a dropdown.

maxklenk (changes requested):
Please a README describing the flows, what needs to be configured and how to use the scripts.

maxklenk (approved):

PR #648, opened 2024-05-20
feat: adds gas-optimized code for multiple swaps (GenericSwapFacet v2.0.0 > v.3.0.0)

0xDEnYO:


ezynda3 (approved):

src/Facets/GenericSwapFacetV3.sol
```diff
@@ -0,0 +1,515 @@
...
+    function _returnPositiveSlippage(
+        ERC20 fromToken,
+        address receiver
+    ) private {
+        // if a balance exists in fromToken, it must be positive slippage
+        if (address(fromToken) != address(0)) {
+            uint256 fromTokenBalance = fromToken.balanceOf(address(this));
+            if (fromTokenBalance > 0) {
+                fromToken.safeTransfer(receiver, fromTokenBalance);
+            }
```
maxklenk, line 512: Why does it not handle native tokens?

src/Facets/GenericSwapFacetV3.sol
```diff
@@ -0,0 +1,515 @@
...
+            _referrer,
+            _receiver,
+            _minAmountOut,
+            _swapData
+        );
+    }
+
+    /// Private helper methods ///
+
+    function _depositMultipleERC20Tokens(
```
maxklenk, line 292: Depositing multiple tokens is a very advanced case, I don't really think the optimized version does need to support it. Instead we could deposit only tokens of the first swap (if an optimization like this saves us more gas)

src/Facets/GenericSwapFacetV3.sol
0xDEnYO: @maxklenk: 
Because we send native as msg.value in full to DEXs.....are you aware of any cases where a DEX would not use that msg.value in full but send a part back as a unused refund? If we have such a case, then it would make sense to add handling for it. I just wasnt aware of any such cases.
Please let me know.

src/Facets/GenericSwapFacetV3.sol
0xDEnYO: @maxklenk 
Yes I thought about this, too and made several tests. 
Since we currently mostly have two entries in the LibSwap array, the gas savings would be tiny (~0,15%) but we are more future-proof with this version. Especially in the light of Permit2 coming up which will also enable gasless multiswaps. And Jumper team had interest in some sort of panic button functionality that swaps all tokens to stables, for which this would also be a good use case. 

I would leave it in there but am also happy to remove it if you have strong feelings about it.

src/Facets/GenericSwapFacetV3.sol
ezynda3: If we want to use Uniswap V3 in the future we will need to be able to handle native token refunds. 0x actually facilitates UniV3 swaps for us sometimes which result in refunds of native token.

maxklenk: The only case where refunds can happen is with uniswapV3, and as Ed said I would expect it for all tokens. I think it does not hurt to have a fallback for the native case here as well.

src/Facets/GenericSwapFacetV3.sol
maxklenk: Thanks for the details, if the gas savings are that small we can keep it as it is.

src/Facets/GenericSwapFacetV3.sol
0xDEnYO: I've added handling for native sendingAsset refunds by a DEX

PR #657, opened 2024-05-24
test: adds test helpers (dex mock)

0xDEnYO:


maxklenk (approved):
Nice idea, this will help to simulate different cases.
I would be good to use the mock in a actual test before merging, so we can be sure the setup works.

0xDEnYO:
> Nice idea, this will help to simulate different cases. I would be good to use the mock in a actual test before merging, so we can be sure the setup works.

It is being used in the GenericSwapFacetV3 for which I coded it so I know it works as intended. The tests are just in a different branch.

PR #658, opened 2024-05-24
Forward Revert From External Calls

ezynda3:


maxklenk (approved):
Nice, the errors are now bubbled up even when using custom errors:

e.g. https://www.tdly.co/tx/42161/0xac6f53c72af8990dd3b264b4429ef08bb367fd2bcdc9024380d7a12729885051

PR #659, opened 2024-05-27
Deploy contracts to Rootstock

ezynda3:


deployments/_deployments_log_file.json
```diff
@@ -571,6 +571,20 @@
...
+    "rootstock": {
+      "production": {
+        "1.0.0": [
+          {
+            "ADDRESS": "0xaD50118509eB4c8e3E39a370151B0fD5D5957013",
+            "OPTIMIZER_RUNS": "1000000",
+            "TIMESTAMP": "2024-05-24 15:08:15",
+            "CONSTRUCTOR_ARGS": "0x",
+            "SALT": "",
+            "VERIFIED": "true"
```
0xDEnYO, line 584: It would be good if you could check verification status after deployments to ensure that contracts are in fact verified. This one does not seem to be verified according to: 
https://explorer.rootstock.io/address/0xad50118509eb4c8e3e39a370151b0fd5d5957013?__ctab=Code

We should keep an eye on that.

ezynda3: Their explorer doesn't work with foundry for verification therefore it is verified on sourcify. I also just check this on Louper as the names will show as long as it's verified somewhere.

PR #668, opened 2024-05-31
Post deployment checklist

ezynda3:


script/deploy/healthCheck.mts
```diff
@@ -0,0 +1,367 @@
...
+        )
+      } else {
+        consola.success(`Facet ${facet} registered in Diamond`)
+      }
+    }
+
+    // Check that core periphery facets are deployed
+    consola.box('Checking periphery contracts...')
+    for (const contract of corePeriphery) {
+      if (!deployedContracts[contract]) {
```
0xDEnYO, line 142: I would suggest to check the addresses for deployed bytecode....just relying on the logs seems risky.
Same for the check if diamond is deployed above.

script/deploy/healthCheck.mts
```diff
@@ -0,0 +1,367 @@
...
+          publicClient
+        ).read.owner()
+        if (owner !== refundWallet) {
+          logError(`Receiver owner is ${owner}, expected ${refundWallet}`)
+        } else {
+          consola.success('Receiver owner is correct')
+        }
+      }
+
+      // ServiceFeeCollector
```
0xDEnYO, line 268: could move this in a function to avoid duplicate code

0xDEnYO (approved):

script/deploy/healthCheck.mts
0xDEnYO: did you see this comment @ezynda3 ?

PR #670, opened 2024-05-31
Deploy AllBridge to Celo

ezynda3:


maxklenk (approved):
Is deployed, but please also verify it so the name is shown on Louper:
https://louper.dev/diamond/0x1231DEB6f5749EF6cE6943a275A1D3E7486F4EaE?network=celo

PR #672, opened 2024-06-03
Update health check script

ezynda3:


script/deploy/healthCheck.mts
```diff
@@ -33,7 +32,6 @@ const corePeriphery = [
   'Receiver',
```
0xDEnYO, line 32: I am wondering if it makes sense to read these from the config.sh (the config for our scripts) cause there we maintain this info already and it doesnt make so much sense to maintain it in different places. 
Alternatively we could consider to have subfolders "core" in our facets and periphery folder. Then we dont need to maintain the info anywhere but could just extract all names from the folder. What do you think?

script/deploy/healthCheck.mts
```diff
@@ -356,6 +310,26 @@ const getOwnablContract = (address: Address, client: PublicClient) => {
   })
 }
 
+const checkOwnership = async (
+  name: string,
+  expectedOwner: Address,
+  deployedContracts: Record<string, Address>,
+  publicClient: PublicClient
+) => {
+  if (deployedContracts[name]) {
```
0xDEnYO, line 319: I think I raised this before, not sure if you saw my comment. 
I feel that just reading from the deploy log can be dangerous. If a deployment goes wrong for some reason but the log is written anyway then we will not find this out.
I feel this script here should at least make sure that code is deployed at each contract address.

0xDEnYO (approved):

script/deploy/healthCheck.mts
ezynda3: I agree with having a separate folder for core.

script/deploy/healthCheck.mts
ezynda3: Yes I plan to update this in a later PR

PR #681, opened 2024-06-07
Deploy LIFI Diamond and Periphery to Sei

ezynda3:


script/helperFunctions.sh
```diff
@@ -1357,16 +1357,16 @@ function verifyContract() {
     if [ "$ARGS" = "0x" ]; then
       # only show output if DEBUG flag is activated
       if [[ "$DEBUG" == *"true"* ]]; then
-        forge verify-contract --watch --chain "$NETWORK" "$ADDRESS" "$FULL_PATH" -e "${!API_KEY}"
+        forge verify-contract --watch --chain "$CHAIN_ID" "$ADDRESS" "$FULL_PATH" -e "${!API_KEY}"
```
0xDEnYO, line 1360: Why only change it here?
Was there an issue?
Should we switch to --chain throughout all scripts?

0xDEnYO (approved):

script/helperFunctions.sh
ezynda3: It should already be $CHAIN_ID throughout. I just did a search.

PR #683, opened 2024-06-13
feat: enable odos scroll

myz1237:
https://lifi.atlassian.net/browse/LF-8779


myz1237:
why the test failed? Did I miss some setting?

ezynda3:
@myz1237 can you get rid of the yarn.lock change in this PR?

myz1237:
@ezynda3 thanks, ed. I did not notice that change

ezynda3:
I'm going to approve this as it's not the main cause of the error. This started after merging another PR and we still need to debug it. DEX has been approved on Scroll.

myz1237:
@ezynda3 It works now, thanks!

PR #685, opened 2024-06-14
Validate non evm addresses

ezynda3:


src/Facets/CalldataVerificationFacet.sol
```diff
@@ -100,6 +100,35 @@ contract CalldataVerificationFacet {
...
+        bytes calldata data
+    ) external pure returns (bytes32 nonEVMAddress) {
+        bytes memory callData = data;
+        ILiFi.BridgeData memory bridgeData = _extractBridgeData(data);
+
+        if (
+            bytes4(data[:4]) == StandardizedCallFacet.standardizedCall.selector
+        ) {
+            // standardizedCall
+            callData = abi.decode(data[4:], (bytes));
```
0xDEnYO, line 116: just wondering....would the following code be the same and if no, why?

```solidity
callData = data[4:];
```

src/Facets/CalldataVerificationFacet.sol
```diff
@@ -100,6 +100,35 @@ contract CalldataVerificationFacet {
...
+        }
+
+        if (bridgeData.hasSourceSwaps) {
+            assembly {
+                let offset := mload(add(callData, 0x64)) // Get the offset of the bridge specific data
+                nonEVMAddress := mload(add(callData, add(offset, 0x24))) // Get the non-EVM address
```
0xDEnYO, line 122: I think some explanation would be helpful to understand how this works.
Does this imply/require that the non-EVM address parameter in the facetData is always at the first(?) position? Or how do we identify it? 
Such rationale should be added here as a comment imho

0xDEnYO (changes requested):

src/Facets/CalldataVerificationFacet.sol
ezynda3: Yes this would be the same I believe.

ezynda3: So when you wrap the calldata in this function call it actually adds two extra 32 bytes words
1. The new offset for the header of the bytes
2. The number of total bytes being passed as a parameter. `abi.decode` knows this so will take the correct value and push it into `callData`

0xDEnYO (approved):

PR #687, opened 2024-06-17
feat(AcrossV3): adds support for AcrossV3 (AcrossFacetV3 v1.0.0, ReceiverAcrossV3 v1.0.0)

0xDEnYO:


deploy/9999_utils.ts
```diff
@@ -194,7 +194,7 @@ export const deployFacet = async function (
   const facet = await ethers.getContract(name)
   const diamond = await ethers.getContract(diamondContractName)
 
-  await addOrReplaceFacets([facet], diamond.address)
+  // await addOrReplaceFacets([facet], diamond.address)
```
0xDEnYO, line 197: is this intended @ezynda3  ?

archive/src/Facets/DeBridgeDlnFacet.sol
ezynda3, line 1: Looks like this might get added again. Maybe just delete it here as well?

0xDEnYO: updated (removed)

ezynda3 (approved):

PR #695, opened 2024-06-24
Sync only missing signatures

ezynda3:


script/tasks/diamondSyncSigs.ts
```diff
@@ -0,0 +1,116 @@
...
+    const chainName = chainNameMappings[network] || network
+    const chain: Chain = chainMap[chainName]
+
+    const deployedContracts = await import(
+      `../../deployments/${network.toLowerCase()}.json`
+    )
+
+    const rpcUrl = args.rpcUrl || chain.rpcUrls.default.http[0]
+
+    const publicClient = createPublicClient({
```
0xDEnYO, line 59: a little more commenting would improve the readability of the file and allow concepts like "what is a public client" here  to be easier/faster understood

0xDEnYO (approved):

PR #700, opened 2024-06-27
feat: implements GasZipFacet (v1.0.0)

0xDEnYO:


docs/GasZipFacet.md
```diff
@@ -0,0 +1,37 @@
+# GasZipFacet
+
+## Description
+
+The GasZipFacet provides function to deposit ERC20 and native tokens to the Gas.zip protocol (https://www.gas.zip/)
+
+## How To Use
```
maxklenk, line 7: Please add the new bridge functions to the docs

src/Facets/GasZipFacet.sol
```diff
@@ -0,0 +1,156 @@
...
+    constructor(address _gasZipRouter) {
+        gasZipRouter = IGasZip(_gasZipRouter);
+    }
+
+    /// @notice Bridges tokens using the gas.zip protocol
+    /// @param _bridgeData The core information needed for bridging
+    /// @param _gasZipData GasZip-specific bridge data
+    function startBridgeTokensViaGasZip(
+        ILiFi.BridgeData memory _bridgeData,
+        GasZipData calldata _gasZipData
```
maxklenk, line 49: I would not pass `GasZipData` including `gasZipSwapData` to the bridge functions as we are not using that at all.

maxklenk (changes requested):

src/Facets/GasZipFacet.sol
0xDEnYO: mhm....not sure if I can agree.

There are two cases: 
1) gaszip native(here we dont have calldata in `_gasZipData`, it's just an empty parameter, e.g. `0x`)
2) gaszip ERC20 (here we have calldata in `_gasZipData`)

Both of these cases can either be just a bridge tx, or a with a swapStep beforehand.

For me it makes sense the way it is.
Could you please reconsider and elaborate (or call me)? 

What I did notice though is that we dont need the destinationChains parameter in the `gasZipData` if we will never use it as a combined value for several chains. So I removed this and use `bridgeData.destinationChainId` instead which will save us some gas

docs/GasZipFacet.md
0xDEnYO: updated. Thanks for pointing it out

src/Facets/GasZipFacet.sol
maxklenk: Hye, "not using" was perhaps the wrong term. We don't need it. `swapAndStartBridgeTokensViaGasZip` should be used to swap tokens to native and has all the handling to pass along the swap result.

I would recommend to use `startBridgeTokensViaGasZip` only for native tokens and `swapAndStartBridgeTokensViaGasZip` for all other ERC20s. 

Please note that the chainIds gas.zip is using differ from the normal chainIds we use. We either have to maintain a mapping or pass their id:
https://dev.gas.zip/gas/chain-support/outbound

0xDEnYO: Updated as discussed yesterday. Please re-review and approve @maxklenk

maxklenk (changes requested):
We are nearly done.

src/Facets/GasZipFacet.sol
```diff
@@ -0,0 +1,156 @@
...
+contract GasZipFacet is ILiFi, ReentrancyGuard, SwapperV2, Validatable {
+    using SafeTransferLib for address;
+
+    /// @dev GasZip-specific bridge data
+    /// @param gasZipChainId The Gas.zip-specific chainId of the chain on which gas should be received on (https://dev.gas.zip/gas/chain-support/outbound)
+    /// @param gasZipSwapData (only required for ERC20 tokens): the swapData that swaps from ERC20 to native before depositing to gas.zip
+    struct GasZipData {
+        uint256 gasZipChainId;
+        LibSwap.SwapData gasZipSwapData;
+    }
```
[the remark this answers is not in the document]
0xDEnYO: Removed

src/Facets/GasZipFacet.sol
```diff
@@ -0,0 +1,156 @@
...
+                _gasZipData.gasZipChainId,
+                _bridgeData.receiver
+            );
+        else
+            depositToGasZipERC20(
+                _gasZipData.gasZipSwapData,
+                _gasZipData.gasZipChainId,
+                _bridgeData.receiver
+            );
+
```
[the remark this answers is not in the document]
0xDEnYO: yeah makes sense.  Thanks for pointing it out. 
PR and tests are updated and facet redeployed

maxklenk (approved):

PR #726, opened 2024-07-26
feat: explicitly set evm_version

kraikov:
This change will help us track the evm_version used during deployment any of the contracts.


foundry.toml
```diff
@@ -1,6 +1,7 @@
 [profile.default]
 test = 'test/solidity'
 solc_version = '0.8.17'
+evm_version='paris'
```
0xDEnYO, line 4: I would suggest that we use london here

kraikov: The default version of Foundry is `paris`, which means you've been using it up until now

kalote (approved):

PR #728, opened 2024-07-27
Fund wallet on all chains script

0xDEnYO:

- [ ] If this requires a contract version change, I have updated the version number in the source file


script/tasks/fundNewWalletOnAllChains.ts
```diff
@@ -0,0 +1,239 @@
...
+    doNotFundChains: {
+      type: 'string',
+      description: 'An array with chainIds that should not be funded',
+      required: true,
+    },
+    fundAmountUSD: {
+      type: 'string',
+      description: 'The amount of USD that should be sent to every chain',
+      required: true,
+    },
```
maxklenk, line 65: Gas costs on different networks can be quite different, would it make sense to have a default value for each chain? e.g. 1$ to Polygon but 5$ to BSC ...

0xDEnYO: I thought about this but neglected it since we have so many places where we already have to maintain code when adding a new network.
With 5 USD on every network at least one transaction can be executed for sure and it's still not too much money bound.

By the way....POL hat >12000 gas price today, transactions were costing real money. In such an environment, a wallet funded with varying amounts may would not be able to act. I would rather play it safe, that's why I decided to go for this implementation. 


Would you prefer if I hardcoded values for each network?

maxklenk: ok

maxklenk (approved):

PR #817, opened 2024-10-03
Update networks.json

ezynda3:


config/global.json
```diff
@@ -94,7 +94,7 @@
     "gravity": "https://safe.gravity.xyz/txs/api",
     "immutablezkevm": "https://transaction.safe.immutable.com/api",
     "kaia": "https://docs-safe.kaia.io/txs-cypress/api",
-    "linea": "https://transaction.safe.linea.build/api",
+    "linea": "https://safe-transaction-linea.safe.global/api",
```
0xDEnYO, line 97: if the safe API url is now part of `networks.json`, why do we also still have it in `global.json`?

config/networks.json
```diff
@@ -15,7 +22,14 @@
     "nativeCurrency": "ETH",
```
0xDEnYO, line 22: do we not need decimals here, too? What if there ever is a network that does not have 18 decimals for its native currency?

config/networks.json
```diff
@@ -258,7 +470,14 @@
     "nativeCurrency": "OKB",
     "wrappedNativeAddress": "0xe538905cf8410324e03a5a23c1c177a474d59b2b",
     "status": "active",
-    "type": "mainnet"
+    "type": "mainnet",
+    "rpcUrl": "https://rpc.xlayer.tech",
+    "explorerType": "etherscan",
```
0xDEnYO, line 475: just thinking....would it maybe make sense to rename this to verificationType instead of explorerType? 
I am not sure....

script/utils/viemScriptHelpers.ts
```diff
@@ -1,27 +1,49 @@
...
-  if (!chain)
+  if (!network)
     throw new Error(
-      `Chain ${networkName} (aka '${chainName}', if a mapping exists) not supported by viem or requires name mapping. Check if you can find your chain here: https://github.com/wevm/viem/tree/main/src/chains/definitions`
+      `Chain ${networkName} does not exist. Please check that the network exists in 'config/networks.json'`
     )
 
+  const chain = defineChain({
+    id: network.chainId,
+    name: network.name,
```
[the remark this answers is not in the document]
ezynda3: No but it's useful for setting up batch calls in the future.

config/networks.json
ezynda3: This most likely won't happen and it's easy to add later if need be. We don't use any functionality in Viem that actually depends on the `nativeCurrency` setup at the moment either

script/utils/viemScriptHelpers.ts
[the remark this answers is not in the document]
0xDEnYO: maybe we can add it or add a comment ?

0xDEnYO (approved):

PR #945, opened 2025-01-22
LF-11761 - Implement glacis [GlacisFacet v1.0.0,IGlacisAirlift v1.0.0]

mirooon:
[LF-11761]

The addLiquidity function was added to TaskBase because, in the case of Glacis, they do not provide routes for stablecoins like USDT or USDC. As a result, I had to work with more custom tokens that lack liquidity on Uniswap V2 and Sushiswap V2. This made it challenging to simulate the swapping process without having sufficient liquidity. To address this, I introduced an abstracted addLiquidity method in the parent TaskBase test suite. This abstraction allows for simulating liquidity addition upfront, enabling proper testing of swap functionality in such scenarios.

- [x] For new facets: I have checked all points from this list: https://www.notion.so/lifi/New-Facet-Contract-Checklist-157f0ff14ac78095a2b8f999d655622e

[LF-11761]: https://lifi.atlassian.net/browse/LF-11761?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[LF-11761]: https://lifi.atlassian.net/browse/LF-11761?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


script/demoScripts/demoGlacisAirlift.ts
```diff
@@ -0,0 +1,122 @@
...
+import { zeroPadValue } from 'ethers6'
+import dotenv from 'dotenv'
+dotenv.config()
+
+async function main() {
+  const RPC_URL = process.env.ETH_NODE_URI_ARBITRUM
+  const PRIVATE_KEY = process.env.PRIVATE_KEY
+  const LIFI_ADDRESS = deployments.LiFiDiamond
+  const WORMHOLE_ADDRESS = '0xB0fFa8000886e57F86dd5264b9582b2Ad87b2b91' // Wormhole token on Arbitrum
+  const destinationChainId = 10 // Optimism
```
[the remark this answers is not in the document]
mirooon: It's demo script. Values can be hardcoded

script/deploy/facets/DeployGlacisFacet.s.sol
```diff
@@ -0,0 +1,32 @@
...
+    function getConstructorArgs() internal override returns (bytes memory) {
+        string memory path = string.concat(root, "/config/glacis.json");
+        string memory json = vm.readFile(path);
+
+        address airlift = json.readAddress(
+            string.concat(".", network, ".airlift")
+        );
+
+        return abi.encode(airlift);
+    }
```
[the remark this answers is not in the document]
mirooon: Ignore it

test/solidity/Facets/GlacisFacet.t.sol
```diff
@@ -0,0 +1,286 @@
...
+        uint256 initialETHBalance = USER_SENDER.balance;
+        initiateSwapAndBridgeTxWithFacet(false);
+
+        // check balances after call
+        assertEq(
+            dai.balanceOf(USER_SENDER),
+            initialDAIBalance - swapData[0].fromAmount
+        );
+        assertEq(USER_SENDER.balance, initialETHBalance - addToMessageValue);
+    }
```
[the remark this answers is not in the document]
mirooon: 1. There is no need to test for invalid token pairs because if a token is invalid, it will only be identified by the third-party contract we interact with. Since we are working on an off-chain routing system, the contract will only receive routes with valid token pairs.
2. Liquidity checks are performed off-chain during the route discovery process. By the time the contract is involved, we assume the tokens already have sufficient liquidity, so testing this scenario is unnecessary.
3. This scenario does not require testing as it is not applicable in our setup.
4.Failed transaction scenarios are already comprehensively covered in the existing test cases. Test contract inherits from TestBaseFacet where pther failed transaction cases are covered like testBase_Revert_BridgeWithInvalidReceiverAddress, testBase_Revert_BridgeWithInvalidAmount, testBase_Revert_SwapAndBridgeWithInvalidAmount, testBase_Revert_BridgeToSameChainId, testBase_Revert_CallerHasInsufficientFunds etc

script/demoScripts/demoGlacisAirlift.ts
```diff
@@ -0,0 +1,122 @@
...
+    const gasPrice = await provider.getGasPrice()
+    const tx = await token
+      .connect(signer)
+      .approve(LIFI_ADDRESS, amount, { gasPrice })
+    await tx.wait()
+
+    console.info('Approval transaction complete. New allowance set.')
+  } else {
+    console.info('Sufficient allowance already exists. No need to approve.')
+  }
```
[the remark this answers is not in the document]
mirooon: nonce is automaticly added once it's executed, default has limit is enough.

No need for deadline variable. You proposed it but it's not used anywhere.

Good point with error handling.

script/demoScripts/demoGlacisAirlift.ts
```diff
@@ -0,0 +1,122 @@
...
+
+  // calculate native fee
+  const estimatedFees = await airlift.connect(signer).callStatic.quoteSend(
+    WORMHOLE_ADDRESS,
+    amount,
+    zeroPadValue(address, 32), // address to bytes32
+    destinationChainId,
+    address, //refund
+    utils.parseEther('1')
+  )
```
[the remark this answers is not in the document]
mirooon: Good point!

src/Interfaces/IGlacisAirlift.sol
```diff
@@ -0,0 +1,87 @@
+// SPDX-License-Identifier: MIT
+/// @custom:version 1.0.0
+pragma solidity ^0.8.17;
+
+struct QuoteSendInfo {
+    Fee gmpFee;
+    uint256 amountSent;
+    uint256 valueSent;
+    AirliftFeeInfo AirliftFeeInfo;
+}
```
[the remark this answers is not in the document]
mirooon: good point!

config/glacis.json
0xDEnYO, line 1: This will probably need some updates once they have deployed their contracts to PROD

docs/GlacisFacet.md
```diff
@@ -0,0 +1,94 @@
+# Glacis Facet
+
+## How it works
+
+The Glacis Facet works by ...
```
0xDEnYO, line 5: There is text missing...

src/Facets/GlacisFacet.sol
```diff
@@ -0,0 +1,112 @@
...
+
+    /// @notice The contract address of the glacis airlift on the source chain.
+    IGlacisAirlift public immutable airlift;
+
+    /// Types ///
+
+    /// @param refund Refund address
+    /// @param nativeFee The fee amount in native token required by the Glacis Airlift.
+    struct GlacisData {
+        address refund;
```
0xDEnYO, line 27: for consistency to other facets I would suggest to rename this variable to "refundAddress" and better parameter description

src/Facets/GlacisFacet.sol
```diff
@@ -0,0 +1,112 @@
...
+
+    /// @dev Contains the business logic for the bridge via Glacis
+    /// @param _bridgeData The core information needed for bridging
+    /// @param _glacisData Data specific to Glacis
+    function _startBridge(
+        ILiFi.BridgeData memory _bridgeData,
+        GlacisData calldata _glacisData
+    ) internal {
+        SafeERC20.safeTransfer(
+            IERC20(_bridgeData.sendingAssetId),
```
0xDEnYO, line 98: Even though this is simple code, I would suggest to comment at least on the main steps (one comment for the transfer, here you could explain that we are calling a function that requires the tokens to be sent to the contract already) and leave a blank line between unrelated code blocks.

src/Interfaces/IGlacisAirlift.sol
```diff
@@ -0,0 +1,87 @@
+// SPDX-License-Identifier: MIT
+/// @custom:version 1.0.0
+pragma solidity ^0.8.17;
+
+struct QuoteSendInfo {
```
0xDEnYO, line 5: if there are parameter descriptions available (e.g. in the original interface contract of the third party) then it would be nice to copy them over here, too.

src/Interfaces/IGlacisAirlift.sol
```diff
@@ -0,0 +1,87 @@
...
+    uint256 correctedAmount;
+    uint256 correctedValue;
+}
+
+struct Fee {
+    uint256 nativeFee;
+    uint256 tokenFee;
+}
+
+error GlacisAirlift__NotEnoughValueFee();
```
0xDEnYO, line 23: We dont need all these errors, etc.
Only put code in our codebase that is required, keep interfaces as small as (reasonably) possible cause this bloats up our codebase

test/solidity/Facets/GlacisFacet.t.sol
```diff
@@ -0,0 +1,284 @@
...
+
+        setFacetAddressInTestBase(address(glacisFacet), "GlacisFacet");
+
+        // adjust bridgeData
+        bridgeData.bridge = "glacis";
+        bridgeData.sendingAssetId = ADDRESS_WORMHOLE_TOKEN;
+        bridgeData.minAmount = defaultWORMHOLEAmount;
+        bridgeData.destinationChainId = 10;
+
+        QuoteSendInfo memory quoteSendInfo = IGlacisAirlift(address(airlift))
```
0xDEnYO, line 84: a comment would help here to understand why you make this call

Please try to comment extensively, they dont hurt but help to read files much faster

test/solidity/Facets/GlacisFacet.t.sol
```diff
@@ -0,0 +1,284 @@
...
+                REFUND_WALLET,
+                payableAmount
+            );
+        addToMessageValue =
+            quoteSendInfo.gmpFee.nativeFee +
+            quoteSendInfo.airliftFeeInfo.airliftFee.nativeFee;
+        //prepare check for events
+        vm.expectEmit(true, true, true, true, address(glacisFacet));
+        emit LiFiTransferStarted(bridgeData);
+        initiateBridgeTxWithFacet(false);
```
0xDEnYO, line 171: leave blank lines between unconnected code, e.g.

        vm.expectEmit(true, true, true, true, address(glacisFacet));
        emit LiFiTransferStarted(bridgeData);
        
        these two are closely connected...ok to not have a blank line
        
        
        emit LiFiTransferStarted(bridgeData);
        initiateBridgeTxWithFacet(false);
        
        these two are not related at all, blank line pls
        same before vm.stopPrank();

test/solidity/Facets/GlacisFacet.t.sol
```diff
@@ -0,0 +1,284 @@
...
+    }
+
+    function testBase_CanSwapAndBridgeTokens()
+        public
+        override
+        assertBalanceChange(ADDRESS_DAI, USER_RECEIVER, 0)
+        assertBalanceChange(ADDRESS_WORMHOLE_TOKEN, USER_SENDER, 0)
+        assertBalanceChange(ADDRESS_WORMHOLE_TOKEN, USER_RECEIVER, 0)
+    {
+        // add liquidity for dex pair
```
0xDEnYO, line 219: here you could also explain what you explained to us in the call...there was no liquidity for these tokens in a DEX available so to simulate a swap you need to add liquidity first.
Helps to put such coding considerations and assumptions in a comment...

test/solidity/Facets/GlacisFacet.t.sol
```diff
@@ -0,0 +1,284 @@
...
+            ADDRESS_DAI,
+            ADDRESS_WORMHOLE_TOKEN,
+            100_000 * 10 ** ERC20(ADDRESS_DAI).decimals(),
+            100_000 * 10 ** wormhole.decimals()
+        );
+
+        uint256 initialDAIBalance = dai.balanceOf(USER_SENDER);
+        vm.startPrank(USER_SENDER);
+        // prepare bridgeData
+        bridgeData.hasSourceSwaps = true;
```
0xDEnYO, line 230: blank line after this and before // prepare bridgeData comment

test/solidity/Facets/GlacisFacet.t.sol
```diff
@@ -0,0 +1,284 @@
...
+            address(uniswap),
+            ADDRESS_DAI,
+            ADDRESS_WORMHOLE_TOKEN,
+            swapData[0].fromAmount,
+            bridgeData.minAmount,
+            block.timestamp
+        );
+
+        vm.expectEmit(true, true, true, true, _facetTestContractAddress);
+        emit LiFiTransferStarted(bridgeData);
```
0xDEnYO, line 250: blank lines

script/demoScripts/demoGlacisAirlift.ts
```diff
@@ -0,0 +1,140 @@
...
+
+  const balance = await token.balanceOf(address)
+  console.info(`Token balance for connected wallet: ${balance.toString()}`)
+  if (balance.eq(0)) {
+    console.error(`Connected account has no funds.`)
+    console.error(`Exiting...`)
+    process.exit(1)
+  }
+
+  console.info('Sending WORMHOLE...')
```
0xDEnYO, line 42: this log is a bit early...you are still checking allowance, not yet sending

script/demoScripts/demoGlacisAirlift.ts
```diff
@@ -0,0 +1,140 @@
...
+      await tx.wait()
+    } catch (error) {
+      console.error('Approval failed:', error)
+      process.exit(1)
+    }
+    console.info('Approval transaction complete. New allowance set.')
+  } else {
+    console.info('Sufficient allowance already exists. No need to approve.')
+  }
+  console.info('Sent WORMHOLE')
```
0xDEnYO, line 64: log is not true...you havent sent any wormhole yet, only checked/registered allowance

script/demoScripts/demoGlacisAirlift.ts
```diff
@@ -0,0 +1,140 @@
...
+  const PRIVATE_KEY = process.env.PRIVATE_KEY
+  const LIFI_ADDRESS = deployments.LiFiDiamond
+  const WORMHOLE_ADDRESS = '0xB0fFa8000886e57F86dd5264b9582b2Ad87b2b91' // Wormhole token on Arbitrum
+  const destinationChainId = 10 // Optimism
+
+  const provider = new ethers.providers.JsonRpcProvider(RPC_URL)
+  const signer = new ethers.Wallet(PRIVATE_KEY as string, provider)
+  const glacis = GlacisFacet__factory.connect(LIFI_ADDRESS, provider) as any
+
+  const address = await signer.getAddress()
```
0xDEnYO, line 25: variable naming could be improved

script/demoScripts/demoGlacisAirlift.ts
```diff
@@ -0,0 +1,140 @@
...
+    referrer: constants.AddressZero,
+    sendingAssetId: WORMHOLE_ADDRESS,
+    receiver: address,
+    destinationChainId: destinationChainId,
+    minAmount: amount,
+    hasSourceSwaps: false,
+    hasDestinationCall: false,
+  }
+
+  const airlift = new Contract(config.arbitrum.airlift, [
```
0xDEnYO, line 79: I like to have variable names more explicit....like 'airliftContract' or 'airliftAddress'. So they are easier to distinguish.
But that's a personal thing, just for you to consider

mirooon: good point!

src/Interfaces/IGlacisAirlift.sol
mirooon: also looked at this but there are no descriptions for these structs :/

src/Interfaces/IGlacisAirlift.sol
mirooon: ok!

test/solidity/Facets/GlacisFacet.t.sol
mirooon: good point

test/solidity/Facets/GlacisFacet.t.sol
```diff
@@ -0,0 +1,284 @@
...
+
+    function setFunctionApprovalBySignature(bytes4 _signature) external {
+        LibAllowList.addAllowedSelector(_signature);
+    }
+}
+
+contract GlacisFacetTest is TestBaseFacet {
+    GlacisFacet.GlacisData internal validGlacisData;
+    TestGlacisFacet internal glacisFacet;
+    ERC20 internal wormhole;
```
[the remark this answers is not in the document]
mirooon: good point
just added tests for LINK on Base

src/Facets/GlacisFacet.sol
```diff
@@ -0,0 +1,116 @@
...
+        containsSourceSwaps(_bridgeData)
+        doesNotContainDestinationCalls(_bridgeData)
+        validateBridgeData(_bridgeData)
+    {
+        _bridgeData.minAmount = _depositAndSwap(
+            _bridgeData.transactionId,
+            _bridgeData.minAmount,
+            _swapData,
+            payable(msg.sender)
+        );
```
maxklenk, line 84: Please check that we are handling the native fee right, e.g. when the user starts with native token, we have to ensure enough native stays in the contract to pay the bridge fee. e.g. check StargateFacetV2

test/solidity/Facets/GlacisFacet.t.sol
```diff
@@ -0,0 +1,384 @@
...
+                0
+            )
+        );
+
+        initiateBridgeTxWithFacet(false);
+        vm.stopPrank();
+    }
+}
+
+contract GlacisFacetWormholeTest is GlacisFacetTestBase {
```
0xDEnYO, line 358: I feel the ordering could be improved as these two smaller contracts with their setup functions are more closely connected to the setup function above. How do you feel about that?

config/glacis.json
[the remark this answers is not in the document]
0xDEnYO: Also I thought I commented on this before....where did this comment go?

0xDEnYO (changes requested):

config/glacis.json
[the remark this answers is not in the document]
mirooon: Ok, I though the comment was more for reminder for us but keeping this reminder in this file makes more sense as it's easy to forget.

Regarding your comment on github: [there it is](https://github.com/lifinance/contracts/pull/945#discussion_r1926645370)

docs/GlacisFacet.md
```diff
@@ -0,0 +1,99 @@
...
+  - Simply bridges tokens using glacis
+- `swapAndStartBridgeTokensViaGlacis(BridgeData memory _bridgeData, LibSwap.SwapData[] calldata _swapData, glacisData memory _glacisData)`
+  - Performs swap(s) before bridging tokens using glacis
+
+## glacis Specific Parameters
+
+The methods listed above take a variable labeled `_glacisData`. This data is specific to glacis and is represented as the following struct type:
+
+```solidity
+/// @param refundAddress Refund address
```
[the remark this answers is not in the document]
mirooon: good point!

test/solidity/Facets/GlacisFacet.t.sol
```diff
@@ -0,0 +1,384 @@
...
+        addLiquidity(
+            ADDRESS_DAI,
+            ADDRESS_SRC_TOKEN,
+            100_000 * 10 ** ERC20(ADDRESS_DAI).decimals(),
+            100_000 * 10 ** srcToken.decimals()
+        );
+
+        // Call `quoteSend` to estimate the required native fee for the transfer.
+        // This is necessary to ensure the transaction has sufficient gas for execution.
+        // The `payableAmount` parameter simulates the amount of native tokens required for the estimation.
```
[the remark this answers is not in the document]
mirooon: **Why do you use 1 ether for this parameter?**

Had the same questions as you. For me that was unclear either.

guS from Glacis team explained msgValue this way:
_Since quoteSend is a view function and therefore not payable, we receive msg.value as a parameter. When quoting, you can simulate the impact on your msg.value by passing a sample amount, such as 1 ETH, to see how it would be adjusted during an actual send._

so while we are having nativeFee we initially don't know what msg.value is 'enough'. That's why we need to provide overestimation with for example 1 ETH. So as far as I understand it goes through the full bridging logic of and tells "I only need 0.005ETH from that 1ETH" and the nativeFee is returned in `QuoteSendInfo`. IMO with 1 ETH, we’re just on the safe side of overestimation so that the function doesn’t revert.


**Unfortunately the interface does not have a description for this parameter**

Also asked for it.

Related discussion: https://lifi-protocol.slack.com/archives/C088BHRHKV3/p1736957284528519

test/solidity/Facets/GlacisFacet.t.sol
mirooon: hmm, tbh I don't know what can be improved. The base test contract shares setup function with all contracts that inherits from it and children setup only necessery parameters. 

Do you have something in mind how I could improve it?

One drawback is that when opening the file you need to scroll all the way to the bottom to see the testing parameters. There’s a trade-off between “top-down” logical ordering and quick visibility.

test/solidity/Facets/GlacisFacet.t.sol
```diff
@@ -0,0 +1,384 @@
...
+        vm.label(address(glacisFacet), "GlacisFacet");
+
+        // adjust bridgeData
+        bridgeData.bridge = "glacis";
+        bridgeData.sendingAssetId = ADDRESS_SRC_TOKEN;
+        bridgeData.minAmount = defaultSrcTokenAmount;
+        bridgeData.destinationChainId = destinationChainId;
+
+        // add liquidity for dex pair DAI-{SOURCE TOKEN}
+        // this is necessary because Glacis does not provide routes for stablecoins like USDT or USDC, forcing us to work with custom tokens that often lack liquidity on V2 dexes
```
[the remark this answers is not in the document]
mirooon: good point!

config/glacis.json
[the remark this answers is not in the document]
mirooon: Of course this is a temporary solution but I’m not fully on convinced with adding a random "important" element to the config file. I think a better approach would be to keep separate config files for test and production (e.g., config.test.json and config.prod.json).

Also IMO it’s more likely that the Glacis team will upgrade their facets rather than redeploy the entire diamond.

script/demoScripts/demoGlacisAirlift.ts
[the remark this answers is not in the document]
mirooon: Good point, I would like to follow the new pattern that will arrise from [this PR](https://github.com/lifinance/contracts/pull/950). Most of viem script is currently done interally (helped me with creating template) but it would be better to have a finalized demo script template discussed and merged first. This would help avoid duplicating efforts on the Glacis demo script.

docs/GlacisFacet.md
```diff
@@ -0,0 +1,99 @@
...
+
+Bridge doesn’t support stablecoins like **USDT** or **USDC**. They are custom tokens focused. We can find possible routes in [`routes.ts`](https://github.com/glacislabs/airlift-evm/blob/main/node-scripts/src/tests/routes.ts).
+
+The [`send`](https://github.com/glacislabs/airlift-evm/blob/main/src/facets/GlacisAirliftFacet.sol#L94) function is used to execute the cross-chain transfer. 
+Before calling [`send`](https://github.com/glacislabs/airlift-evm/blob/main/src/facets/GlacisAirliftFacet.sol#L94), the tokens must first be sent to the contract, as the function assumes the tokens are already in place. This function is preferred over [`sendAfterApproval`](https://github.com/glacislabs/airlift-evm/blob/af935df67c3fff873edea9758ef73cd46e1908c7/src/facets/GlacisAirliftFacet.sol#L112) because it eliminates the need for redundant token transfer steps, as tokens are already transferred to the contract beforehand.
```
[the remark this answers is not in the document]
mirooon: oh yes thats complitely irrelevant

config/glacis.json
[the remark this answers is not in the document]
0xDEnYO: Our scripts are not prepared to distinguish between test files and production files. Also we do not really have a "test" environment in that sense, so there is no clear logic as in when to use which. 
My main point was just that "this config file is not yet final (I believe) and we should make sure to not forget about that".

test/solidity/Facets/GlacisFacet.t.sol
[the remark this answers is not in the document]
0xDEnYO: Thanks for the explanation. 
If something is not clear to you and needs to be explained by someone else or requires extensive research, it's usually a great indicator for considering to add an extensive comment describing this topic and the underlying implementation assumptions in the code so that other developers/readers who probably have the same question as you do not need to ask anyone to understand it in all details.

test/solidity/Facets/GlacisFacet.t.sol
0xDEnYO: I personally would move these two contracts to the beginning of the file.
Or have it all integrated in one test contract like we do for other facets. 

The drawback you mentioned is what triggered me to write this comment.

mirooon: thanks for suggestion. The thing is that Solidity enforces that child contracts must be defined after parent... If I would move them it won't compile. Merging everything into one contract would reduce clarity that abtraction gives us. It's nice to keep tests modular.

Third approach would be to have seperated folder for `GlacisFacet` with `GlacisFacetTestSuite.t.sol`, `GlacisFacetWormhole.t.sol`, `GlacisFacetLink.t.sol` but IMO  it's overkill.

I would like to have opinion from @ezynda3 as well.

config/glacis.json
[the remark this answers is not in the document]
mirooon: Fully understand! My main point was that putting this message as config element it's not the best approach but it's OK! Also wanted to point out better solution as a future improvment.

Btw I have added message to the config

src/Facets/GlacisFacet.sol
```diff
@@ -0,0 +1,116 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.17;
+
```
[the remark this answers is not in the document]
mirooon: Audit will be added in the next stages

test/solidity/Facets/GlacisFacet.t.sol
```diff
@@ -0,0 +1,284 @@
...
+
+    function initiateSwapAndBridgeTxWithFacet(bool) internal override {
+        glacisFacet.swapAndStartBridgeTokensViaGlacis{
+            value: addToMessageValue
+        }(bridgeData, swapData, validGlacisData);
+    }
+
+    function testBase_CanBridgeNativeTokens() public override {
+        // facet does not support bridging of native assets
+    }
```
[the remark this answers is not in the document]
mirooon: They can be empty if facet doesn't handle native token

test/solidity/Facets/GlacisFacet.t.sol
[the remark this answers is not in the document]
mirooon: Good point, added explanation above

test/solidity/Facets/GlacisFacet.t.sol
0xDEnYO: I wasnt aware of this Solidity limitation. 
No, then feel free to leave it as-is. I dont think it makes sense to split into separate files.

script/demoScripts/demoGlacisAirlift.ts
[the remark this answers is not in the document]
0xDEnYO: agreed.
Think we can merge the codegen today and then you can continue to apply this template here

templates/facetDemoScript.template.hbs
```diff
@@ -26,13 +26,13 @@ dotenv.config()
...
-  const destinationChainId = 1 // Set destination chain id
+  const destinationChainId = 1; // Set destination chain id
 
   const { client, publicClient, walletAccount, lifiDiamondAddress, lifiDiamondContract } = await setupEnvironment(srcChain, {{constantCase name}}_FACET_ABI);
-  const signerAddress = walletAccount.address
+  const signerAddress = walletAccount.address;
 
   // === Contract addresses ===
-  const SRC_TOKEN_ADDRESS = '' as `0x${string}` // Set the source token address here.
+  const SRC_TOKEN_ADDRESS = '' as `0x${string}`; // Set the source token address here.
```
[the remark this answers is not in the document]
mirooon: It's template only that's why it's empty

templates/facetDemoScript.template.hbs
```diff
@@ -62,17 +62,17 @@
...
   //
 
-  const srcTokenName = await srcTokenContract.read.name() as string
-  const srcTokenSymbol = await srcTokenContract.read.symbol() as string
-  const srcTokenDecimals = await srcTokenContract.read.decimals() as bigint
-  const amount = parseUnits('10', srcTokenDecimals) // 10 * 1e{source token decimals}
+  const srcTokenName = await srcTokenContract.read.name() as string;
+  const srcTokenSymbol = await srcTokenContract.read.symbol() as string;
+  const srcTokenDecimals = await srcTokenContract.read.decimals() as bigint;
+  const amount = parseUnits('10', Number(srcTokenDecimals)); // 10 * 1e{source token decimals}
```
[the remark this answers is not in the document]
mirooon: as it's simple demo script there is no need for such a optimizations

templates/facetDemoScript.template.hbs
```diff
@@ -92,11 +92,11 @@
     minAmount: amount,
     hasSourceSwaps: false,
     hasDestinationCall: false,
-  }
+  };
 
   const {{camelCase name}}Data: {{titleCase name}}Facet.{{titleCase name}}DataStruct = {
     // Add your specific fields for {{titleCase name}} here.
-  }
+  };
```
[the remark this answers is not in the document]
mirooon: its empty because its only template

script/demoScripts/demoGlacis.ts
```diff
@@ -0,0 +1,197 @@
...
+  await executeTransaction(
+    () =>
+      lifiDiamondContract.write.startBridgeTokensViaGlacis(
+        [bridgeData, glacisData],
+        { value: nativeFee }
+      ),
+    'Starting bridge tokens via Glacis',
+    publicClient,
+    true
+  )
```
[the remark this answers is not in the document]
mirooon: executeTransaction has logic inside with try/catch structures or advanced error reporting with transaction logs

templates/facetDemoScript.template.hbs
```diff
@@ -62,17 +62,17 @@
...
-  const amount = parseUnits('10', srcTokenDecimals) // 10 * 1e{source token decimals}
+  const srcTokenName = await srcTokenContract.read.name() as string;
+  const srcTokenSymbol = await srcTokenContract.read.symbol() as string;
+  const srcTokenDecimals = await srcTokenContract.read.decimals() as bigint;
+  const amount = parseUnits('10', Number(srcTokenDecimals)); // 10 * 1e{source token decimals}
 
-  console.info(`\Bridge ${amount} ${srcTokenName} (${srcTokenSymbol}) from ${srcChain} --> {DESTINATION CHAIN NAME}`)
-  console.info(`Connected wallet address: ${signerAddress}`)
+  console.info(`Bridge ${amount} ${srcTokenName} (${srcTokenSymbol}) from ${srcChain} --> {DESTINATION CHAIN NAME}`);
+  console.info(`Connected wallet address: ${signerAddress}`);
```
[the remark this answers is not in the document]
mirooon: we don't need destinationChainName as a variable. Script just want to have clear log message

0xDEnYO (changes requested):

script/demoScripts/demoGlacis.ts
```diff
@@ -0,0 +1,197 @@
...
+
+dotenv.config()
+
+// #region ABIs
+
+const ERC20_ABI = erc20Artifact.abi as Narrow<typeof erc20Artifact.abi>
+const GLACIS_FACET_ABI = glacisFacetArtifact.abi as Narrow<
+  typeof glacisFacetArtifact.abi
+>
+export const AIRLIFT_ABI = [
```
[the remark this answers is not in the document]
mirooon: thanks for bringing this up! I initially had a different approach but stick to the decision to avoid using `ethers`. I realized there was another way to get ABI that was right under my nose.

I've also updated the demo script template with instructions on how to get the ABI.

src/Facets/GlacisFacet.sol
```diff
@@ -0,0 +1,117 @@
...
+    /// @param _bridgeData The core information needed for bridging
+    /// @param _glacisData Data specific to Glacis
+    function _startBridge(
+        ILiFi.BridgeData memory _bridgeData,
+        GlacisData calldata _glacisData
+    ) internal {
+        // Transfer the tokens to the Airlift contract.
+        // This step ensures that the tokens are already in place before calling the `send` function.
+        // The `send` function assumes the tokens are pre-transferred to the contract.
+        SafeERC20.safeTransfer(
```
gusjavaz, line 101: We plan to support only the `sendAfterApproval` method for better gas optimization and security. Could you test this approach to ensure compatibility with LiFi's diamond operations?

script/demoScripts/demoGlacis.ts
```diff
@@ -0,0 +1,159 @@
...
+    client,
+    publicClient,
+    walletAccount,
+    lifiDiamondAddress,
+    lifiDiamondContract,
+  } = await setupEnvironment(srcChain, GLACIS_FACET_ABI)
+  const signerAddress = walletAccount.address
+
+  // === Contract addresses ===
+  const SRC_TOKEN_ADDRESS =
```
gusjavaz, line 49: Besides W, it might be worthwhile to test with additional tokens to ensure compatibility with transporters beyond Wormhole. For example, MIM (0xB153FB3d196A8eB25522705560ac152eeEc57901 on OP) or LINK (0x88fb150bdc53a65fe94dea0c9ba0a6daf8c6e196 on BASE).

mirooon: Sure! We have LINK tests as well - https://github.com/lifinance/contracts/blob/implement-glacis-LF-11761/test/solidity/Facets/GlacisFacet.t.sol#L376-L405

src/Facets/GlacisFacet.sol
jboetticher: More specifically, we are replacing the "send" function's functionality/workflow with the "sendAfterApproval" functionality, but it will still be called "send"

0xDEnYO (approved):

.github/workflows/securityAlertsReview.yml
0xDEnYO, line 1: always makes my heart jump to see more comments being added :)

deployments/_deployments_log_file.json
```diff
@@ -29722,5 +29722,21 @@
...
+    "arbitrum": {
+      "staging": {
+        "1.0.0": [
+          {
+            "ADDRESS": "0xF82830B952Bc60b93206FA22f1cD4770cedb2840",
+            "OPTIMIZER_RUNS": "1000000",
+            "TIMESTAMP": "2025-02-26 15:22:09",
+            "CONSTRUCTOR_ARGS": "0x000000000000000000000000d9e7f6f7dc7517678127d84dbf0f0b4477de14e0",
+            "SALT": "",
+            "VERIFIED": "true"
```
0xDEnYO, line 29736: this is not true, the contract is not verified

src/Facets/GlacisFacet.sol
```diff
@@ -0,0 +1,116 @@
...
+import { ReentrancyGuard } from "../Helpers/ReentrancyGuard.sol";
+import { SwapperV2 } from "../Helpers/SwapperV2.sol";
+import { Validatable } from "../Helpers/Validatable.sol";
+import { IGlacisAirlift } from "../Interfaces/IGlacisAirlift.sol";
+
+/// @title Glacis Facet
+/// @author LI.FI (https://li.fi/)
+/// @notice Integration of the Glacis airlift (wrapper for native token bridging standards)
+/// @custom:version 1.0.0
+contract GlacisFacet is ILiFi, ReentrancyGuard, SwapperV2, Validatable {
```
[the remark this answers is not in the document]
0xDEnYO: It is unclear what you want to point out or archieve here. An audit has been completed and logged. This PR updates the auditLog.json, it also adds the report referenced in that JSON to the correct folder. And the git action that verifies the audit information is passing (not sure if you have access to results of git actions though). 
All audit requirements are fulfilled

src/Facets/GlacisFacet.sol
```diff
@@ -0,0 +1,125 @@
...
+    struct GlacisData {
+        address refundAddress;
+        uint256 nativeFee;
+    }
+
+    /// Constructor ///
+    /// @notice Initializes the GlacisFacet contract
+    /// @param _airlift The address of Glacis Airlift contract.
+    constructor(IGlacisAirlift _airlift) {
+        airlift = _airlift;
```
0xDEnYO, line 35: I personally feel OK if you add here the validation of the `_airlift` parameter not being address(0) to adhere to our new rules.
You could make one simple commit and then ask Sujith to comment on the PR confirmining that this change does not change the audit result and that a re-audit is not required for this change.

0xDEnYO (changes requested):

src/Facets/GlacisFacet.sol
mirooon: ok!

deployments/_deployments_log_file.json
mirooon: interesting... Updated.
I didn't touch it manually. Probably there is some bug in the script that mark this contract as verified even if it's not. IMO would be worth to investigate this issue.

0xDEnYO (approved):

maxklenk (approved):

PR #950, opened 2025-01-24
LF11927 - Update codegen

mirooon:
[LF-11927]

### DIFF
--- a/script/demoScripts/utils/demoScriptChainConfig.ts
+++ b/script/demoScripts/utils/demoScriptChainConfig.ts
@@ -1,0 +1,34 @@
+import { arbitrum, mainnet, Chain } from 'viem/chains'
+import networks from '../../../config/networks.json'
+import { getEnvVar } from './demoScriptHelpers'
+
+export type SupportedChain = keyof typeof networks
+
+const viemChainMap: Partial<Record<SupportedChain, Chain>> = {
+  mainnet: mainnet,
+  arbitrum: arbitrum,
+}
+
+/**
+ * Return the correct RPC environment variable
+ * (e.g. `ETH_NODE_URI_ARBITRUM` or `ETH_NODE_URI_MAINNET`)
+ * or fallback to the chain's "rpcUrl" from `networks.json`.
+ */
+export const getRpcUrl = (chain: SupportedChain) => {
+  const envKey = `ETH_NODE_URI_${chain.toUpperCase()}`
+  return getEnvVar(envKey) as string
+}
+
+/**
+ * Return the `Chain` object from viem. If you request a chain that doesn't
+ * exist in `viemChainMap`, this will throw an error.
+ */
+export const getViemChain = (chain: SupportedChain): Chain => {
+  const viemChain = viemChainMap[chain]
+  if (!viemChain) {
+    throw new Error(
+      `No viem chain object defined for chain: ${chain}. Please take a look at viemChainMap`
+    )
+  }
+  return viemChain
+}
--- a/script/demoScripts/utils/demoScriptHelpers.ts
+++ b/script/demoScripts/utils/demoScriptHelpers.ts
@@ -451,3 +451,23 @@ const getProviderForChainId = (chainId: number) => {
     throw Error(`Could not find a provider for network ${networkName}`)
   else return provider
 }
+
+export const getEnvVar = (varName: string): string => {
+  const value = process.env[varName]
+  if (!value) {
+    throw new Error(`Missing required environment variable: ${varName}`)
+  }
+  return value
+}
+
+export const normalizePrivateKey = (pk: string): `0x${string}` => {
+  if (!pk.startsWith('0x')) {
+    return `0x${pk}` as `0x${string}`
+  }
+  return pk as `0x${string}`
+}
+
+export const zeroPadAddressToBytes32 = (address: string): `0x${string}` => {
+  const hexAddress = address.replace(/^0x/, '')
+  return `0x${hexAddress.padStart(64, '0')}`
+}
--- a/templates/facetDemoScript.template.hbs
+++ b/templates/facetDemoScript.template.hbs
@@ -1,0 +1,156 @@
+import {
+  createPublicClient,
+  createWalletClient,
+  getContract,
+  http,
+  parseUnits,
+  parseEther,
+  Narrow,
+  zeroAddress
+} from 'viem'
+import { privateKeyToAccount } from 'viem/accounts'
+import { randomBytes } from 'crypto'
+import dotenv from 'dotenv'
+import erc20Artifact from '../../out/ERC20/ERC20.sol/ERC20.json'
+import {{camelCase name}}FacetArtifact from '../../out/{{titleCase name}}Facet.sol/{{titleCase name}}Facet.json'
+import config from '../../config/{{camelCase name}}.json'
+import { {{titleCase name}}Facet, ILiFi } from '../../typechain'
+import {
+  getEnvVar,
+  normalizePrivateKey,
+  zeroPadAddressToBytes32,
+} from './utils/demoScriptHelpers'
+import { getRpcUrl, getViemChain, SupportedChain } from './utils/demoScriptChainConfig'
+
+dotenv.config()
+
+const ERC20_ABI = erc20Artifact.abi as Narrow<typeof erc20Artifact.abi>
+const {{constantCase name}}_FACET_ABI = {{camelCase name}}FacetArtifact.abi as Narrow<typeof {{camelCase name}}FacetArtifact.abi>
+
+import deployments from '../../deployments/mainnet.staging.json' // change source chain
+
+async function main() {
+  // Retrieve env vars & set up clients
+  const srcChain: SupportedChain = "mainnet"; // change source chain
+  const RPC_URL = getRpcUrl(srcChain)
+  const PRIVATE_KEY = getEnvVar('PRIVATE_KEY')
+  const typedPrivateKey = normalizePrivateKey(PRIVATE_KEY)
+
+  const publicClient = createPublicClient({
+    chain: getViemChain(srcChain),
+    transport: http(RPC_URL),
+  })
+  const walletAccount = privateKeyToAccount(typedPrivateKey)
+  const walletClient = createWalletClient({
+    chain: getViemChain(srcChain),
+    transport: http(RPC_URL),
+    account: walletAccount,
+  })
+  const signerAddress = walletAccount.address
+
+  // Contract addresses / chain info
+  const LIFI_ADDRESS = deployments.LiFiDiamond as `0x${string}`
+  const SRC_TOKEN_ADDRESS = '' as `0x${string}` // // change source token addres that will be bridged
+  const EXAMPLE_ADDRESS = config.mainnet.example as `0x${string}` // change source chain
+  const destinationChainId = 10
+  const amount = parseUnits('10', 18) // 10 * 1e18
+
+  // Instantiate contracts
+  const srcTokenContract = getContract({
+    address: SRC_TOKEN_ADDRESS,
+    abi: ERC20_ABI,
+    client: { public: publicClient, wallet: walletClient }
+  })
+
+  const exampleContract = getContract({
+    address: EXAMPLE_ADDRESS,
+    abi: EXAMPLE_ABI,
+    client: { public: publicClient, wallet: walletClient }
+  })
+
+  const lifiDiamondContract = getContract({
+    address: LIFI_ADDRESS,
+    abi: {{constantCase name}}_FACET_ABI,
+    client: { public: publicClient, wallet: walletClient }
+  })
+
+  const srcTokenName = await srcTokenContract.read.name() as string
+  const srcTokenSymbol = await srcTokenContract.read.symbol() as string
+
+  console.info(`\Bridge ${amount} ${srcTokenName} (${srcTokenSymbol}) from ${srcChain} --> {DESTINATION CHAIN NAME}`) // change chain names
+  console.info(`Connected wallet address: ${signerAddress}`)
+
+  // Check token balance
+  const balance = await srcTokenContract.read.balanceOf([signerAddress])
+  console.info(`Token balance for connected wallet: ${balance}`)
+
+  if (balance === 0n) {
+    console.error(`Connected account has no ${srcTokenName} (${srcTokenSymbol}) tokens. Exiting...`)
+    process.exit(1)
+  }
+
+  // Check allowance
+  const currentAllowance = await srcTokenContract.read.allowance([
+    signerAddress,
+    LIFI_ADDRESS,
+  ]) as bigint;
+
+  console.log('Current allowance:', currentAllowance.toString())
+  
+  if (currentAllowance < amount) {
+    console.info('Allowance is insufficient. Approving required amount...')
+    try {
+      const hash = await srcTokenContract.write.approve([LIFI_ADDRESS, amount])
+      console.info(`Approval transaction sent (hash): ${hash}`)
+    } catch (error) {
+      console.error('Approval failed:', error)
+      process.exit(1)
+    }
+    console.info('Approval transaction complete. New allowance set.')
+  } else {
+    console.info('Sufficient allowance already exists. No need to approve.')
+  }
+
+  // In this part put necessary logic usually it's fetching quotes, estimating fees, signing messages etc. 
+
+
+
+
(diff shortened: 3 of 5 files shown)


templates/facetDemoScript.template.hbs
```diff
@@ -0,0 +1,167 @@
...
+    minAmount: amount,
+    hasSourceSwaps: false,
+    hasDestinationCall: false,
+  }
+
+  // Prepare bridge-specific data (replace or add required fields if needed here).
+  const {{camelCase name}}BridgeData: {{titleCase name}}Facet.{{titleCase name}}DataStruct = {
+    // Add your specific fields for {{titleCase name}} here.
+  }
+
```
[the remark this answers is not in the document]
mirooon: It's template file only

script/demoScripts/utils/demoScriptChainConfig.ts
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+ * Return the correct RPC environment variable
+ * (e.g. `ETH_NODE_URI_ARBITRUM` or `ETH_NODE_URI_MAINNET`)
+ * or fallback to the chain's "rpcUrl" from `networks.json`.
+ */
+export const getRpcUrl = (chain: SupportedChain) => {
+  const envKey = `ETH_NODE_URI_${chain.toUpperCase()}`
+  return getEnvVar(envKey) as string
+}
```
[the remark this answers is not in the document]
mirooon: In this case we don't want to use fall back

templates/facetDemoScript.template.hbs
```diff
@@ -0,0 +1,167 @@
...
+  })
+  const signerAddress = walletAccount.address
+
+  // Contract addresses / chain info
+  // LIFI_ADDRESS is the address of the LiFi Diamond contract for bridging.
+  // SRC_TOKEN_ADDRESS is the token to be bridged (set its address for the source chain).
+  // EXAMPLE_ADDRESS is a custom contract address for a contract that your facet interacts with.
+  const LIFI_ADDRESS = deployments.LiFiDiamond as `0x${string}` // Set deployments import
+  const SRC_TOKEN_ADDRESS = '' as `0x${string}` // Set the source token address here.
+  const EXAMPLE_ADDRESS = config.mainnet.example as `0x${string}` // Update for the correct chain.
```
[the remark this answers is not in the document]
mirooon: It's template file only

script/demoScripts/utils/demoScriptChainConfig.ts
0xDEnYO, line 1: What's the reasoning to move these helper functions into a different file?
In my opinion they could all live in the file `demoScriptHelpers.ts` as they all fit under this umbrella (and then are easier to locate for a developer imho.

script/demoScripts/utils/demoScriptHelpers.ts
0xDEnYO, line 1: consider adding function comments

templates/facetDemoScript.template.hbs
```diff
@@ -0,0 +1,165 @@
...
+  getEnvVar,
+  normalizePrivateKey
+} from './utils/demoScriptHelpers'
+import { getRpcUrl, getViemChain, SupportedChain } from './utils/demoScriptChainConfig'
+
+// #region ABIs
+
+const ERC20_ABI = erc20Artifact.abi as Narrow<typeof erc20Artifact.abi>
+const {{constantCase name}}_FACET_ABI = {{camelCase name}}FacetArtifact.abi as Narrow<typeof {{camelCase name}}FacetArtifact.abi>
+const EXAMPLE_ABI = [{...}] // optional define example ABI
```
0xDEnYO, line 28: not sure what is meant with example ABI?
If this is the ABI for the facet, you could name it respectively. 
Also for optional things I would suggest to make them a comment by default so they dont end up staying in the code if the developer forgets to remove them

templates/facetDemoScript.template.hbs
```diff
@@ -0,0 +1,165 @@
...
+dotenv.config()
+
+async function main() {
+  // === Set up parameters, retrieve env vars and set up clients ===
+  // `srcChain` is the source chain for the operation (change as needed).
+  // RPC URL and private key are dynamically fetched based on the chain.
+  const srcChain: SupportedChain = "mainnet"; // Set source chain
+  const RPC_URL = getRpcUrl(srcChain)
+  const PRIVATE_KEY = getEnvVar('PRIVATE_KEY')
+  const typedPrivateKey = normalizePrivateKey(PRIVATE_KEY)
```
0xDEnYO, line 41: wondering if we could even simplify this more. 
Thinking along the lines of "I only decide which is the source chain and a helper function returns everything I need (chain, rpc, normalized private key, all viem clients, LiFiDiamond contract object, etc) as we seem to have a lot of boilerplate code in every demo script

templates/facetDemoScript.template.hbs
```diff
@@ -0,0 +1,165 @@
...
+    chain: getViemChain(srcChain),
+    transport: http(RPC_URL),
+    account: walletAccount,
+  })
+  const signerAddress = walletAccount.address
+
+  // === Contract addresses ===
+  // LIFI_ADDRESS is the address of the LiFi Diamond contract for bridging.
+  // SRC_TOKEN_ADDRESS is the token to be bridged (set its address for the source chain).
+  // EXAMPLE_ADDRESS is a custom contract address for a contract that your facet interacts with.
```
0xDEnYO, line 58: Is this the "bridge router"? Example address is not a good name imho as it doesnt specify anything and doesnt give any guidance on what kind of address may be required here. Try to be more specific and also if it's optional, consider making it a comment by default

templates/facetDemoScript.template.hbs
```diff
@@ -0,0 +1,165 @@
...
+  })
+
+  const srcTokenName = await srcTokenContract.read.name() as string
+  const srcTokenSymbol = await srcTokenContract.read.symbol() as string
+  const amount = parseUnits('10', 18) // 10 * 1e18
+
+  console.info(`\Bridge ${amount} ${srcTokenName} (${srcTokenSymbol}) from ${srcChain} --> {DESTINATION CHAIN NAME}`) // Set destination chain name
+  console.info(`Connected wallet address: ${signerAddress}`)
+
+  // === Check token balance ===
```
0xDEnYO, line 89: balance and allowance checks and error handling could also be moved to a helper function as this is probably gonna be used in every script

templates/facetDemoScript.template.hbs
0xDEnYO, line 1: I have not seen this `contract.write.(...)` and `contract.read.(...)` notation before. Also couldnt find it in the viem docs. Where did you get this from?

Please make sure that a) transactions will succeed (you may need to simulate first as described here: 
https://viem.sh/docs/contract/writeContract.html) 

and b) please make sure that the await makes sure that the transaction is executed (not just sent) before the script continues. This is especially important for approvals, etc.

I think I had cases before where the script would just continue while something is still in the pipeline which lead to unexpected behaviour
https://viem.sh/docs/actions/public/waitForTransactionReceipt.html

templates/facetDemoScript.template.hbs
```diff
@@ -0,0 +1,165 @@
...
+    console.info('Sufficient allowance already exists. No need to approve.')
+  }
+
+  // === In this part put necessary logic usually it's fetching quotes, estimating fees, signing messages etc. ===
+
+
+
+
+  // === Prepare bridge data ===
+  // Generate types by running command `yarn typechain`
```
0xDEnYO, line 126: wondering if it makes sense to have some sort of helper function that checks if the types for this particular facet exist and if not, it will run the command to make sure that the required types are available

templates/facetDemoScript.template.hbs
```diff
@@ -0,0 +1,165 @@
...
+  const bridgeData: ILiFi.BridgeDataStruct = {
+    // Edit fields as needed
+    transactionId: `0x${randomBytes(32).toString('hex')}`,
+    bridge: '{{camelCase name}}',
+    integrator: 'ACME Devs',
+    referrer: zeroAddress,
+    sendingAssetId: SRC_TOKEN_ADDRESS,
+    receiver: signerAddress,
+    destinationChainId: 10, // Set destination chain id
```
0xDEnYO, line 135: would propose to have a global variable in this script that is set at the beginning and then used everywhere

templates/facetDemoScript.template.hbs
```diff
@@ -0,0 +1,165 @@
...
+  }
+
+  const {{camelCase name}}BridgeData: {{titleCase name}}Facet.{{titleCase name}}DataStruct = {
+    // Add your specific fields for {{titleCase name}} here.
+  }
+
+  // === Start bridging ===
+  try {
+    const hash = await lifiDiamondContract.write.startBridgeTokensVia{{titleCase name}}(
+      [bridgeData, {{camelCase name}}BridgeData],
```
0xDEnYO, line 148: The facet data is missing here. In most cases there will be facet data

templates/facetDemoScript.template.hbs
```diff
@@ -0,0 +1,165 @@
...
+  }
+
+  // === Start bridging ===
+  try {
+    const hash = await lifiDiamondContract.write.startBridgeTokensVia{{titleCase name}}(
+      [bridgeData, {{camelCase name}}BridgeData], 
+      // { value: fee }, // Optional fee
+    )
+
+    console.info(`Bridge transaction sent (hash): ${hash}`)
```
0xDEnYO, line 152: Again, please make sure the tx is not just broadcasted but also included in a block

templates/facetDemoScript.template.hbs
0xDEnYO, line 1: The file is well-structured and well-documented. 
Good work

0xDEnYO (changes requested):

templates/facetDemoScript.template.hbs
mirooon: most facets interact with at least one external contract, which we refer to as `_example` in the facet.template.hbs. This is also reflected in other parts of our templates. In `facetDeployScript.template.hbs` in `getConstructorArgs`  there are lines:
`
address example = json.readAddress(".Example");
return abi.encode(example);
`

In facetConfig.template.hbs there is `example` included in config file.

I thought it would make sense to also provide a placeholder for the ABI in demo script to follow the same pattern. IMO most of the facets that we will have will have to have external contract address that will implement some ABI. And example ABI placeholder can help developer where to put it.

 in other places like `facetConfig.template.hbs`, the example address isn't commented or treated as optional.

script/demoScripts/utils/demoScriptChainConfig.ts
mirooon: Ok, to double check I will leave `viemChainMap` map in [demoScriptChainConfig.ts](https://github.com/lifinance/contracts/pull/950/files/d46452e86274bb47d9d4684597695c85ffff64a5#diff-27050d030e7904c3e6101990cfb5e5839e5d19826e174df48d1bf1fa92a56c3c)

demoScriptChainConfig.ts - state
demoScriptHelpers.ts - functions

templates/facetDemoScript.template.hbs
mirooon: Isn't it the second parameter? {{camelCase name}}BridgeData ?

templates/facetDemoScript.template.hbs
mirooon: thanks for pointing it out. I used https://viem.sh/docs/contract/getContract.html to first initialize contract. Once it's initialized it's possible to directly interact with it.

script/demoScripts/utils/demoScriptHelpers.ts
```diff
@@ -451,3 +463,247 @@
...
+        process.exit(1)
+      }
+    }
+
+    return result
+  } catch (error) {
+    console.error(`${transactionDescription} failed:`, error)
+    process.exit(1)
+  }
+}
```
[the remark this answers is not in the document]
mirooon: No need to estimate gas? It's not used anywhere further in the code. 

Error details in this form is enough

script/demoScripts/utils/demoScriptChainConfig.ts
0xDEnYO: I would vouch to move these into one file (to simplify things). In a typescript project this split would make more sense but in our repo I feel it makes more sense to keep all helpers in one place so they can be found easier. 
But it's a personal opinion and I am also fine with the existing approach.

templates/facetDemoScript.template.hbs
```diff
@@ -0,0 +1,118 @@
...
+  //   abi: EXAMPLE_ABI,
+  //   client
+  // });
+  //
+
+  const srcTokenName = await srcTokenContract.read.name() as string
+  const srcTokenSymbol = await srcTokenContract.read.symbol() as string
+  const amount = parseUnits('10', 18) // 10 * 1e18
+
+  console.info(`\Bridge ${amount} ${srcTokenName} (${srcTokenSymbol}) from ${srcChain} --> {DESTINATION CHAIN NAME}`) // Set destination chain name
```
0xDEnYO, line 69: I think the comment on the end of the line is erroneous as this is a log statement and doesnt set anything

templates/facetDemoScript.template.hbs
0xDEnYO, line 1: The template is much more slim now....really like it. Thanks for putting the effort into abstracting the helper functions

0xDEnYO (changes requested):

script/demoScripts/utils/demoScriptHelpers.ts
```diff
@@ -451,3 +463,255 @@ const getProviderForChainId = (chainId: number) => {
...
+  }
+  return pk as `0x${string}`
+}
+
+/**
+ * Return the correct RPC environment variable
+ * (e.g. `ETH_NODE_URI_ARBITRUM` or `ETH_NODE_URI_MAINNET`)
+ */
+const getRpcUrl = (chain: SupportedChain) => {
+  const envKey = `ETH_NODE_URI_${chain.toUpperCase()}`
```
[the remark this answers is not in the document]
mirooon: In line 515 there is `getEnvVar` function execution that handle it

templates/facetDemoScript.template.hbs
mirooon: that's true, maybe would be better to have`from ${srcChain} --> chain id: ${destinationChainId}? 
wdyt? I am not super convinced

templates/facetDemoScript.template.hbs
mirooon: Thanks for giving valuable comments!

templates/facetDemoScript.template.hbs
0xDEnYO: ok, that's fine

templates/facetDemoScript.template.hbs
0xDEnYO: it's still there...not sure what your thumbs up means :)

templates/facetDemoScript.template.hbs
```diff
@@ -0,0 +1,118 @@
...
+  // const exampleContract = getContract({
+  //   address: EXAMPLE_ADDRESS,
+  //   abi: EXAMPLE_ABI,
+  //   client
+  // });
+  //
+
+  const srcTokenName = await srcTokenContract.read.name() as string
+  const srcTokenSymbol = await srcTokenContract.read.symbol() as string
+  const amount = parseUnits('10', 18) // 10 * 1e18
```
0xDEnYO, line 67: we should probably use the token's decimals here and not hardcode 18 decimals

ezynda3 (dismissed):

templates/facetDemoScript.template.hbs
mirooon: good catch!

0xDEnYO (approved):

templates/facetDemoScript.template.hbs
```diff
@@ -0,0 +1,119 @@
...
+  //   client
+  // });
+  //
+
+  const srcTokenName = await srcTokenContract.read.name() as string
+  const srcTokenSymbol = await srcTokenContract.read.symbol() as string
+  const srcTokenDecimals = await srcTokenContract.read.decimals() as bigint
+  const amount = parseUnits('10', srcTokenDecimals) // 10 * 1e{source token decimals}
+
+  console.info(`\Bridge ${amount} ${srcTokenName} (${srcTokenSymbol}) from ${srcChain} --> {DESTINATION CHAIN NAME}`)
```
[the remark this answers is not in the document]
mirooon: This is good. I didn't want to put variable there.

PR #955, opened 2025-01-27
Added preinstall script

mirooon:

The `scriptMaster` script relies on system-installed packages (bc and jq) for its functionality. Without these dependencies being explicitly installed, users running scriptMaster are likely to encounter various issues beacause both bc and jq are used extensively in the script. This change ensures that the required dependencies are installed on initial yarn install.

### DIFF
--- a/preinstall.sh
+++ b/preinstall.sh
@@ -1,0 +1,53 @@
+#!/bin/bash
+
+echo "Setting up the development environment..."
+
+# Function to install a package if not already installed
+install_package_linux() {
+  PACKAGE=$1
+  if ! command -v $PACKAGE &> /dev/null; then
+    echo "Installing $PACKAGE on Linux..."
+    sudo apt-get update -y
+    sudo apt-get install -y $PACKAGE
+  else
+    echo "$PACKAGE is already installed."
+  fi
+}
+
+install_package_mac() {
+  PACKAGE=$1
+  if ! command -v $PACKAGE &> /dev/null; then
+    echo "Installing $PACKAGE on macOS..."
+    brew install $PACKAGE
+  else
+    echo "$PACKAGE is already installed."
+  fi
+}
+
+# Detect the operating system
+OS=$(uname -s)
+
+if [[ "$OS" == "Linux" ]]; then
+  echo "Detected Linux. Using apt package manager."
+  install_package_linux jq
+  install_package_linux bc
+elif [[ "$OS" == "Darwin" ]]; then
+  echo "Detected macOS. Using Homebrew package manager."
+
+  # Check if Homebrew is installed
+  if ! command -v brew &> /dev/null; then
+    echo "Homebrew is not installed. Installing Homebrew..."
+    /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)"
+    # Add Homebrew to PATH for immediate use
+    eval "$($(brew --prefix)/bin/brew shellenv)"
+  fi
+
+  install_package_mac jq
+  install_package_mac bc
+else
+  echo "Unsupported operating system: $OS"
+  echo "Please install jq and bc manually."
+  exit 1
+fi
+
+echo "All necessary packages are installed."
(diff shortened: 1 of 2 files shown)


preinstall.sh
```diff
@@ -0,0 +1,53 @@
...
+    eval "$($(brew --prefix)/bin/brew shellenv)"
+  fi
+
+  install_package_mac jq
+  install_package_mac bc
+else
+  echo "Unsupported operating system: $OS"
+  echo "Please install jq and bc manually."
+  exit 1
+fi
```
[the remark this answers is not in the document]
mirooon: good point

preinstall.sh
```diff
@@ -0,0 +1,53 @@
...
+
+install_package_mac() {
+  PACKAGE=$1
+  if ! command -v $PACKAGE &> /dev/null; then
+    echo "Installing $PACKAGE on macOS..."
+    brew install $PACKAGE
+  else
+    echo "$PACKAGE is already installed."
+  fi
+}
```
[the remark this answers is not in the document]
mirooon: Good point with Check for apt-get availability but "Validate package name contains only alphanumeric chars and basic symbols" is not needed as only we decide on package names. They don't come from outside.

Good point with Failed to install as well

preinstall.sh
```diff
@@ -0,0 +1,66 @@
+#!/bin/bash
+
+echo "Setting up the development environment..."
+
```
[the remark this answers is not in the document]
mirooon: Not needed here

preinstall.sh
```diff
@@ -0,0 +1,79 @@
...
+install_package_linux() {
+  PACKAGE=$1
+  # Check for apt-get availability
+  if ! command -v apt-get &> /dev/null; then
+    echo "apt-get not found. Please install packages manually."
+    exit 1
+  fi
+  if ! command -v $PACKAGE &> /dev/null; then
+    echo "Installing $PACKAGE on Linux..."
+    sudo apt-get update -y
```
ezynda3, line 15: Thoughts: So there is an issue with this. It assumes that the Linux distro being used is either Ubuntu or Debian based. There a quite a few different variants of Linux that don't use `apt` as the package manager. The other popular variant is Fedora/Redhat which uses the `dnf` command. We may want to add this but I would say most people on Linux would be able to install this manually anyway.

mirooon: Thanks for pointing it out. Updated!

at this moment for Linux distros only apt and dnf package managers are supported. For mac only brew is handled. This should cover most of developers. Otherwise they need to install them manually.

preinstall.sh
```diff
@@ -0,0 +1,91 @@
...
+  else
+    echo "No recognized package manager found."
+    echo "Please install $PACKAGE manually."
+    exit 1
+  fi
+}
+
+install_package_mac() {
+  PACKAGE=$1
+  if ! command -v $PACKAGE &> /dev/null; then
```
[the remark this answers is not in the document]
mirooon: It's not needed, for now we use simple naming for packages

preinstall.sh
ezynda3: Looks good!

ezynda3 (dismissed):

preinstall.sh
```diff
@@ -0,0 +1,91 @@
...
+    fi
+  else
+    echo "$PACKAGE is already installed."
+  fi
+}
+
+# Detect the operating system
+OS=$(uname -s)
+
+if [[ "$OS" == "Linux" ]] && ! command -v sudo &> /dev/null; then
```
0xDEnYO, line 50: could move this into the next if-clause as you are checking $OS == "Linux" twice here

eg.

if [[ "$OS" == "Linux" ]] ; then 
  if ! command -v sudo &> /dev/null ; then
    ...error
  fi
 
  ....rest of the code

0xDEnYO (dismissed):

preinstall.sh
mirooon: true!

0xDEnYO (approved):

PR #980, opened 2025-02-07
LF-11929 - Moved core facets to global.json and made adjustments

mirooon:
[LF-11929]

Instead of maintaining duplicate arrays of core facets in both the health check and bash script we decided to keep them in a single source of truth within `global.json`.

Additionally, the GLOBAL_FILE_PATH variable should be defined in config.sh for scripts, with the value:

`GLOBAL_FILE_PATH="$(dirname "$0")/../config/global.json"`

[LF-11929]: https://lifi.atlassian.net/browse/LF-11929?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/config/global.json
+++ b/config/global.json
@@ -59,6 +59,8 @@
     "PeripheryRegistryFacet",
     "GenericSwapFacet",
     "StandardizedCallFacet",
-    "CalldataVerificationFacet"
+    "CalldataVerificationFacet",
+    "GenericSwapFacetV3",
+    "EmergencyPauseFacet"
   ]
 }
--- a/config/global.json
+++ b/config/global.json
@@ -58,7 +58,6 @@
     "WithdrawFacet",
     "PeripheryRegistryFacet",
     "GenericSwapFacet",
-    "StandardizedCallFacet",
     "CalldataVerificationFacet",
     "GenericSwapFacetV3",
     "EmergencyPauseFacet"
--- a/config/global.json
+++ b/config/global.json
@@ -48,5 +48,17 @@
     "LiFuelFeeCollector",
     "TokenWrapper",
     "LiFiDEXAggregator"
+  ],
+  "coreFacets": [
+    "DiamondCutFacet",
+    "DiamondLoupeFacet",
+    "OwnershipFacet",
+    "DexManagerFacet",
+    "AccessManagerFacet",
+    "WithdrawFacet",
+    "PeripheryRegistryFacet",
+    "GenericSwapFacet",
+    "StandardizedCallFacet",
+    "CalldataVerificationFacet"
   ]
 }
--- a/script/helperFunctions.sh
+++ b/script/helperFunctions.sh
@@ -1977,6 +1977,37 @@ function getAllNetworksArray() {
   # return ARRAY
   printf '%s\n' "${ARRAY[@]}"
 }
+
+# Function to retrieve coreFacets from global.json
+function getCoreFacetsArray() {
+  # Ensure GLOBAL_FILE_PATH is set and not empty
+  if [[ -z "$GLOBAL_FILE_PATH" ]]; then
+    echo "Error: GLOBAL_FILE_PATH is not set or empty."
+    exit 1
+  fi
+
+  local FILE="$GLOBAL_FILE_PATH"
+  local ARRAY=()
+
+  # Ensure the global file exists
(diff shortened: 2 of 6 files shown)


config/global.json
```diff
@@ -48,5 +48,17 @@
...
+  "coreFacets": [
+    "DiamondCutFacet",
+    "DiamondLoupeFacet",
+    "OwnershipFacet",
+    "DexManagerFacet",
+    "AccessManagerFacet",
+    "WithdrawFacet",
+    "PeripheryRegistryFacet",
+    "GenericSwapFacet",
+    "StandardizedCallFacet",
```
0xDEnYO, line 61: you can remove this already (as discussed in last weekly meeting)

script/helperFunctions.sh
```diff
@@ -2057,8 +2088,8 @@ function getIncludedAndSortedFacetContractsArray() {
   # get all facet contracts
   FACET_CONTRACTS=($(getIncludedFacetContractsArray "$EXCLUDE_CONFIG"))
 
-  # convert CORE_FACETS into an array
-  CORE_FACETS_ARRAY=($(echo "$CORE_FACETS" | tr ',' ' '))
+  # Get core facets from global.json
+  CORE_FACETS_ARRAY=($(getCoreFacetsArray))
```
0xDEnYO, line 2092: did you try to run this with an empty core facets array?
If I am not mistaken, we need to check the return code for the getCoreFacetsArray function.
Otherwise it might be possible that we call this helper function and it fails (but will not end the script due to the way bash scripts behave when a subshell fails). And then the subsequent code will run into unexpected behaviour.

I might be wrong but please could you check?

script/helperFunctions.sh
```diff
@@ -1977,6 +1977,37 @@ function getAllNetworksArray() {
...
+
+# Function to retrieve coreFacets from global.json
+function getCoreFacetsArray() {
+  # Ensure GLOBAL_FILE_PATH is set and not empty
+  if [[ -z "$GLOBAL_FILE_PATH" ]]; then
+    echo "Error: GLOBAL_FILE_PATH is not set or empty."
+    exit 1
+  fi
+
+  local FILE="$GLOBAL_FILE_PATH"
```
0xDEnYO, line 1989: Is there a need to introduce a new variable if this has the same value as the global variable and is never written to?

script/helperFunctions.sh
```diff
@@ -1977,6 +1977,37 @@ function getAllNetworksArray() {
...
+  local ARRAY=()
+
+  # Ensure the global file exists
+  if [[ ! -f "$FILE" ]]; then
+    echo "Error: Global configuration file $FILE not found at $FILE."
```
0xDEnYO, line 1994: Maybe you wanted to do something like this instead?

```echo "Error: Global configuration file not found at $GLOBAL_FILE_PATH."```

Also please note that we have specific functions to log error() and warning() messages in our helperFunctions.sh. 
Please use those instead.

File: config/global.json [hunk 5]
```diff
@@ -48,5 +48,17 @@
...
+  ],
+  "coreFacets": [
+    "DiamondCutFacet",
+    "DiamondLoupeFacet",
+    "OwnershipFacet",
+    "DexManagerFacet",
+    "AccessManagerFacet",
+    "WithdrawFacet",
+    "PeripheryRegistryFacet",
+    "GenericSwapFacet",
```
0xDEnYO, line 60: GenericSwapFacetV3 is missing

config/global.json
0xDEnYO, line 61: EmergencyPauseFacet is missing !!!

0xDEnYO (changes requested):

script/helperFunctions.sh
mirooon: thanks for pointing that out. I actually removed the empty array check since CORE_FACETS could also have no elements before but I getting to the point I didn't realize that the script would keep running if the subfunction fails, you are right, I didn't know about that.  I've now updated things to use the `checkFailure` function to catch any failure from `getCoreFacetsArray`. I also made a few changes to `getCoreFacetsArray`

0xDEnYO (approved):

PR #981, opened 2025-02-10
Update solc and evm version

0xDEnYO:

We used to have typescript and other compatibility issues when running our codebase on higher solidity version but that seems to have changed now. Our auditor (Sujith) has confirmed that changing to newer solc and evm version does not impose a security risk.  

**Michal:**
Cleaning ABI was introduced to fix typechain duplicate type definitions

#### Problem
Our current contract architecture and new version of solc leads to duplicate event definitions in the compiled ABI files. This happens because events are inherited multiple times through different paths in our inheritance structure. When typechain generates TS types from these ABIs, it creates duplicate type definitions, causing TS compilation errors.

#### Impact
- Local development: TS compilation fails with "Duplicate identifier" errors
- CI/CD: The `types.yaml` GitHub Action (Types Bindings) fails to generate and commit type bindings
- This affects contracts like `OwnershipFacet` and `DiamondCutFacet` where events are inherited multiple times

## Solution
Instead of modifying our contract inheritance structure, I've added an ABI cleaning step that:
1. Runs after `forge build` but before typechain type generation
2. Removes duplicate events from specified ABI files while preserving the first occurrence
3. Ensures clean type generation without duplicate definitions


.github/workflows/enforceTestCoverage.yml
```diff
@@ -62,7 +62,7 @@ jobs:
       - name: Generate Coverage Report
         if: env.solidity_changed == 'true'
         run: |
-          forge coverage --report lcov --force --evm-version 'shanghai' --ir-minimum
+          forge coverage --report lcov --force --evm-version 'cancun' --ir-minimum
```
0xDEnYO, line 65: if we update foundry.toml to cancun then we dont need an explicit flag anymore here (but I am fine with leaving it)

script/cleanABI.ts
0xDEnYO, line 1: I would personally tend to name this script "removeDuplicateEventsFromABI.ts" as this is the only thing it does and it's more specific, but I am also fine to keep it like this.

.github/workflows/enforceTestCoverage.yml
mirooon: updated

script/cleanABI.ts
mirooon: good point

script/removeDuplicateEventsFromABI.ts
```diff
@@ -0,0 +1,83 @@
...
+    })
+
+    // Update the ABI in the file
+    content.abi = cleanedABI
+    await fs.writeJson(file, content, { spaces: 2 })
+    console.log(`Cleaned duplicate events from ${file}`)
+  }
+}
+
+removeDuplicateEventsFromABI().catch(console.error)
```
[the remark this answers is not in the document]
mirooon: good catch

script/removeDuplicateEventsFromABI.ts
```diff
@@ -0,0 +1,83 @@
...
+
+  for (const file of filesToClean) {
+    if (!fs.existsSync(file)) {
+      console.log(`File ${file} not found, skipping...`)
+      continue
+    }
+
+    const content = await fs.readJson(file)
+    const abi = content.abi as any[]
+
```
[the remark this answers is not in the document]
mirooon: good catch, introduced in later commit

ezynda3 (approved):

maxklenk (approved):

PR #991, opened 2025-02-11
LF-12212 Adapt git action (ensure test coverage) to run only when needed

mirooon:
[LF-12212]

[LF-12212]: https://lifi.atlassian.net/browse/LF-12212?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/enforceTestCoverage.yml
+++ b/.github/workflows/enforceTestCoverage.yml
@@ -9,6 +9,7 @@ on:
     types: [opened, synchronize, reopened, ready_for_review]
     paths:
       - 'src/**'
+      - 'test/**'
 
 jobs:
   enforce-min-test-coverage:
--- a/.github/workflows/enforceTestCoverage.yml
+++ b/.github/workflows/enforceTestCoverage.yml
@@ -8,7 +8,7 @@ on:
   pull_request:
     types: [opened, synchronize, reopened, ready_for_review]
     paths:
-      - 'src/**'  # Only trigger if files in src/ are modified
+      - 'src/**'
 
 jobs:
   enforce-min-test-coverage:
--- a/.github/workflows/enforceTestCoverage.yml
+++ b/.github/workflows/enforceTestCoverage.yml
@@ -29,11 +29,14 @@ jobs:
       MIN_TEST_COVERAGE: ${{ secrets.MIN_TEST_COVERAGE }}
     steps:
       - uses: actions/checkout@v4.1.7
+        with:
+          fetch-depth: 0
 
       - name: Check for Solidity file changes
         id: check_files
         run: |
-          CHANGED_FILES=$(git diff --name-only origin/${{ github.base_ref }} | grep -E '^src/.*\.sol|^test/.*\.sol' || true)
+          git fetch origin ${{ github.base_ref }} --depth=1
+          CHANGED_FILES=$(git diff --name-only origin/${{ github.base_ref }}...HEAD | grep -E '^src/.*\.sol|^test/.*\.sol' || true)
           if [[ -n "$CHANGED_FILES" ]]; then
             echo "solidity_changed=true" >> "$GITHUB_ENV"
           else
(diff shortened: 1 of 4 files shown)


0xDEnYO (approved):

.github/workflows/enforceTestCoverage.yml
```diff
@@ -7,6 +7,9 @@ name: Enforce Min Test Coverage
 on:
   pull_request:
     types: [opened, synchronize, reopened, ready_for_review]
+    paths:
+      - 'src/**'
+      - 'test/**'
```
kalote, line 12: `'**.sol'`

mirooon: thanks!

.github/workflows/enforceTestCoverage.yml
```diff
@@ -7,6 +7,9 @@ name: Enforce Min Test Coverage
 on:
   pull_request:
     types: [opened, synchronize, reopened, ready_for_review]
+    paths:
+      - 'src/**.sol'
+      - 'test/**.sol'
```
kalote, line 12: why do you add the folder here? you have other .sol file that are outside of `src/` and `test/` that you don't want to match this trigger?

mirooon: we always place .sol files in the `src/` and `test/` folders as this is the standard structure required by Foundry (development and testing environment for Solidity that we use). While `.sol` files can also exist in the `script/` folder those are intended for deployment or scripts etc and we do not need to run tests for them. That's why I put specifically only `src/` and `test/`.

0xDEnYO: Didnt we say that we do not want to use the `paths` filter cause then there might be cases when the git action doesnt run (but since it is required this would cause problems)?

kalote: I think using paths is fine. It should work as expected, and if not, we will adjust.

mirooon: Was discussed that we would use paths filter and remove "required" protection. IMO it won't work without removing "required" protection

0xDEnYO: if we remove "required" then people can merge their PR even if this check fails which makes it useless

mirooon: I initially thought that marking a action as "required" meant that it must always run, but the final result could still block merging. That’s why I assumed there’s this "Expected - Waiting for status to be reported" status... makes sense

mirooon:
So to solve it and to keep action reporting and also run min test coverage script only when there is a change for `.sol` files  in `src/` and `test/` folders. I first check for direct modifications in the action logic. If no new .sol files are detected in these folders the test coverage script is skipped. Let me know your thoughts! @kalote @0xDEnYO

kalote (approved):

.github/workflows/enforceTestCoverage.yml
```diff
@@ -29,22 +29,40 @@ jobs:
...
     steps:
       - uses: actions/checkout@v4.1.7
+        with:
+          fetch-depth: 0
+
+      - name: Check for Solidity file changes
+        id: check_files
+        run: |
+          git fetch origin ${{ github.base_ref }} --depth=1
+          CHANGED_FILES=$(git diff --name-only origin/${{ github.base_ref }}...HEAD | grep -E '^src/.*\.sol|^test/.*\.sol' || true)
```
0xDEnYO, line 39: I personally feel it's nice and helpful to add more comments.....e.g. for the fact that now the git action will only check for changes in folders src/ and test/.
This could be a comment at the beginning of the file or in the code itself. 

Please try to apply extensive commenting on all work you do.

We can merge this PR for now though, think we have to touch this git action again sometime soon.

0xDEnYO (approved):

PR #1001, opened 2025-02-17
Added Olympix static analysis with alerts review

mirooon:
[LF-11765]

Explained in the notion page: https://www.notion.so/lifi/Olympix-Static-Analysis-tool-GitHub-integration-19df0ff14ac7801eaba6c7e757a61637

[LF-11765]: https://lifi.atlassian.net/browse/LF-11765?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/olympixStaticAnalysis.yml
+++ b/.github/workflows/olympixStaticAnalysis.yml
@@ -13,7 +13,7 @@ on:
       - reopened
       - ready_for_review
     paths:
-      - 'src/**.sol'
+      - 'src/**/*.sol'
 
 jobs:
   static-analysis:
@@ -29,7 +29,7 @@ jobs:
         uses: tj-actions/changed-files@v45
         with:
           files: |
-            src/**.sol
+            src/**/*.sol
 
       - name: Convert Changed Files to Args
         if: steps.changed-files.outputs.any_changed == 'true'
--- a/.github/workflows/olympixStaticAnalysis.yml
+++ b/.github/workflows/olympixStaticAnalysis.yml
@@ -1,0 +1,51 @@
+name: Olympix Static Analysis
+
+on:
+  pull_request:
+    types:
+      - opened
+      - synchronize
+      - reopened
+      - ready_for_review
+    paths:
+      - 'src/**.sol'
+      - '.github/workflows/olympixStaticAnalysis.yml'
+
+jobs:
+  static-analysis:
+    name: Static Analysis Security Check
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout Repository
+        uses: actions/checkout@v3
+
+      - name: Get Changed Solidity Files
+        id: changed-files
+        uses: tj-actions/changed-files@v45
+        with:
+          files: |
+            **/*.sol
+
+      - name: Convert Changed Files to Args
+        if: steps.changed-files.outputs.any_changed == 'true'
+        id: format-args
+        env:
+          ALL_CHANGED_FILES: ${{ steps.changed-files.outputs.all_changed_files }}
+        run: |
+          args=$(echo $ALL_CHANGED_FILES | xargs -n 1 -I {} printf -- "-p %s " "{}")
+          echo "ARGS=$args" >> $GITHUB_ENV
+
+      - name: Run Olympix Integrated Security
+        if: steps.changed-files.outputs.any_changed == 'true'
+        uses: olympix/integrated-security@main
+        env:
+          OLYMPIX_API_TOKEN: ${{ secrets.OLYMPIX_API_TOKEN }}
+        with:
+          args: --output-format sarif --output-path ./ ${{ env.ARGS }}
+
+      - name: Upload Result to GitHub Code Scanning
+        if: steps.changed-files.outputs.any_changed == 'true'
+        uses: github/codeql-action/upload-sarif@v3
+        with:
+          sarif_file: olympix.sarif
--- a/.github/workflows/olympixStaticAnalysis.yml
+++ b/.github/workflows/olympixStaticAnalysis.yml
@@ -1,5 +1,10 @@
 name: Olympix Static Analysis
 
+# - runs the olympix static analyzer on newly added or modified solidity contracts inside the src/ folder in a pull request
+# - detects potential security vulnerabilities and uploads the results to github code scanning
+# - only scans diff (added, renamed, modified) solidity files in src/ instead of the whole repository
+# - ensures security issues are identified before merging, allowing the team to review and discuss findings within the PR
+
 on:
   pull_request:
     types:
@@ -9,7 +14,6 @@ on:
       - ready_for_review
     paths:
       - 'src/**.sol'
-      - '.github/workflows/olympixStaticAnalysis.yml'
 
 jobs:
   static-analysis:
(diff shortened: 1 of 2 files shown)


.github/workflows/olympixStaticAnalysis.yml
```diff
@@ -0,0 +1,51 @@
...
+    name: Static Analysis Security Check
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout Repository
+        uses: actions/checkout@v3
+
+      - name: Get Changed Solidity Files
+        id: changed-files
+        uses: tj-actions/changed-files@v45
```
0xDEnYO, line 25: I see this action has various parameters to change its behaviour. 
Please make sure that not only "changed" but also "added", "deleted" files are covered by this action. Otherwise we will run into trouble

image

.github/workflows/olympixStaticAnalysis.yml
```diff
@@ -0,0 +1,51 @@
...
+      - '.github/workflows/olympixStaticAnalysis.yml'
+
+jobs:
+  static-analysis:
+    name: Static Analysis Security Check
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout Repository
+        uses: actions/checkout@v3
```
0xDEnYO, line 21: outdated version

.github/workflows/olympixStaticAnalysis.yml
0xDEnYO, line 1: Please see the other git actions and how they have a comment in the beginning of the file summarizing the git action and its key characteristics and limitations. Would appreciate if you could add something similar here.

.github/workflows/olympixStaticAnalysis.yml
```diff
@@ -0,0 +1,51 @@
...
+on:
+  pull_request:
+    types:
+      - opened
+      - synchronize
+      - reopened
+      - ready_for_review
+    paths:
+      - 'src/**.sol'
+      - '.github/workflows/olympixStaticAnalysis.yml'
```
0xDEnYO, line 12: why is the path of the git action itself here?

.github/workflows/securityAlertsReview.yml
```diff
@@ -0,0 +1,194 @@
+name: Security Alerts Review
+
+on:
+  push:
+    paths:
+      - 'src/**.sol'
+      - ".github/workflows/securityAlertsReview.yml"
```
0xDEnYO, line 7: why is the path of the git action itself here?

.github/workflows/securityAlertsReview.yml
0xDEnYO, line 1: same as below, pls add initial comment describing the action and its key characteristics

.github/workflows/securityAlertsReview.yml
```diff
@@ -0,0 +1,194 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Fetch PR Number
+        id: fetch_pr
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          PR_NUMBER=$(curl -s -H "Authorization: token ${GITHUB_TOKEN}" \
+            "https://api.github.com/repos/${{ github.repository }}/pulls?state=open" | jq -r '.[0].number')
```
0xDEnYO, line 29: If I am reading this right, you are pulling all open PRs and then use the first element, is that correct?
First of all, please add comments explaining your rationales and assumptions.
Second, this seems like a very risky approach to me. How do you know which one is the latest PR and if that's the PR you are looking for?

In other git actions we use `${{ github.event.pull_request.number }}` to get the PR number but that doesnt work for every trigger event type unfortunately.

Maybe something like this would be better?
https://github.com/marketplace/actions/find-current-pull-request

.github/workflows/securityAlertsReview.yml
```diff
@@ -0,0 +1,194 @@
...
+          # Log raw API response for debugging
+          echo "Raw API Response:"
+          echo "$ALERTS"
+
+          # Extract unresolved alerts (open alerts)
+          UNRESOLVED_ALERTS=$(echo "$ALERTS" | jq -c '[.[] | select(.state == "open") ]' || echo "[]")
+          # Extract dismissed alerts without comments (empty dismissed_comment)
+          DISMISSED_ALERTS=$(echo "$ALERTS" | jq -c '[.[] | select(.state == "dismissed" and (.dismissed_comment == null or .dismissed_comment == ""))]' || echo "[]")
+          # Extract dismissed alerts with comments (successful dismissals)
+          COMMENTED_ALERTS=$(echo "$ALERTS" | jq -c '[.[] | select(.state == "dismissed" and (.dismissed_comment != null and .dismissed_comment != ""))]' || echo "[]")
```
0xDEnYO, line 58: maybe rename to "RESOLVED_ALERTS" ?

.github/workflows/securityAlertsReview.yml
```diff
@@ -0,0 +1,194 @@
...
+              ALERT_FILE=$(echo "$row" | jq -r '.most_recent_instance.location.path')
+              ALERT_DESCRIPTION=$(echo "$row" | jq -r '.most_recent_instance.message.text')
+
+              COMMENT_BODY+="🟡 [View Alert]($ALERT_URL) - **File:** \`$ALERT_FILE\`\n"
+              COMMENT_BODY+="   🔹 $ALERT_DESCRIPTION\n\n"
+            done < <(echo "$DISMISSED_ALERTS" | jq -c '.[]')
+          fi
+
+          if [[ "$UNRESOLVED_COUNT" -gt 0 || "$DISMISSED_COUNT" -gt 0 ]]; then
+            COMMENT_BODY+="⚠️ **Please resolve the above issues before merging.**\n\n"
```
0xDEnYO, line 133: just an idea.....not sure if it makes sense. 
If this action finds out that there is still stuff that needs to be done, it could also put the PR back into DRAFT state. Then the creator has to address all issues and then can move it back to READY_FOR_REVIEW state. Then you would also have a clear trigger for this action to run.

0xDEnYO (changes requested):

.github/workflows/olympixStaticAnalysis.yml
mirooon: sure all tested. Will run on "added", "modified", "renamed" files.
The "changes" naming is more related to differences between feature and main branches

.github/workflows/olympixStaticAnalysis.yml
mirooon: updated

.github/workflows/olympixStaticAnalysis.yml
mirooon: removed

.github/workflows/securityAlertsReview.yml
mirooon: removed

.github/workflows/securityAlertsReview.yml
mirooon: updated

.github/workflows/securityAlertsReview.yml
mirooon: changed it to "DISMISSED_AND_COMMENTED_ALERTS". IMO "resolved" should only refer to alerts that weren't dismissed and instead they were reviewed and fixed by a developer.

mirooon: updated

.github/workflows/securityAlertsReview.yml
mirooon: updated

.github/workflows/securityAlertsReview.yml
mirooon: smart! I totally forgot that we always start with Draft status of the PR. I coded it, played around with it a bit and it works like a charm! 🕺🏻

0xDEnYO (approved):

maxklenk (approved):

PR #1039, opened 2025-03-03
fetch all pendings TXs at once

ezynda3:


script/deploy/safe/confirm-safe-tx.ts
0xDEnYO, line 1: Think there is some stuff at the beginning of the script that could be removed (e.g. the networks to be skipped, etc.). Or is this still relevant in this new version?

Also, before we merge this PR, I would like to discuss this topic of being fully API independent first cause that would mean that we have to go through all networks again and check for pending transactions. 

So let's discuss first what we actually want/need and then finalize the script.

ezynda3: Removed

0xDEnYO: you can resolve such (simple) comments once they are addressed

PR #1054, opened 2025-03-11
LF-12663 ensure codebase passes new solhint rules

mirooon:
[LF-12663]

[LF-12663]: https://lifi.atlassian.net/browse/LF-12663?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


script/deploy/facets/utils/DeployScriptBase.sol
```diff
@@ -39,7 +39,7 @@ contract DeployScriptBase is ScriptBase {
         vm.startBroadcast(deployerPrivateKey);
 
         if (isDeployed()) {
-            console.log("Contract is already deployed");
+            emit log("Contract is already deployed");
```
ezynda3, line 42: Is there any reason why we use emit rather than console.log?

mirooon: The problem is that solhint doesn’t let us allow console.log in tests/scripts and block it in production files...
That’s why I used emit log instead that we can still log values when needed without breaking this rule.

and as we use `-vvvv --json`during our script run the emitted logs will be always in `logs` field in the json result

ezynda3 (approved):

PR #1058, opened 2025-03-14
LF-12743 Create fix for gaszipfacet [GasZipFacet v2.0.3]

mirooon:
[LF-12743]

# Example bridge transactions

Arbitrum to EVM Berachain - https://www.gas.zip/scan/tx/0xcafb1cfe25f4698e38707d2df9078913557e88b50fb09c42151b2581e9693ada
Arbitrum to non-EVM Solana - https://www.gas.zip/scan/tx/0x645f7c773c30e7c48b882d4f0aecbce967e460a41f5706a5881e760fb9a8e3bf
Arbitrum to non-EVM SUI - https://www.gas.zip/scan/tx/0xf722df97dfaa10aaf90ef36fedb37af6583a42627e5d1b9ebb626900c1757fb8

Arbitrum to EVM Berachain and Polygon
https://www.gas.zip/scan/tx/0x0b106f4705312d0a4100b9076ce2018471ec84f7a6048a79ceac9c7bbda3569d
Arbitrum to EVM Base, Berachain, ZkSync
https://www.gas.zip/scan/tx/0xe86ef5623e76f0821e6f46d8693646a8359ea43586ee3fad0f4dbbb2047623fd

Added documentation:

https://www.notion.so/lifi/GasZip-V1-1c2f0ff14ac780a3b67fd8f424454316

[LF-12743]: https://lifi.atlassian.net/browse/LF-12743?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pwollemi:
Can I audit this ticket?

src/Facets/GasZipFacet.sol
```diff
@@ -107,7 +110,7 @@ contract GasZipFacet is ILiFi, ReentrancyGuard, SwapperV2, Validatable {
         if (
             _bridgeData.receiver != NON_EVM_ADDRESS &&
             _gasZipData.receiverAddress !=
-            bytes32(uint256(uint160(_bridgeData.receiver)))
+            bytes32(bytes20(uint160(_bridgeData.receiver)))
```
0xDEnYO, line 113: wondering if we should add a comment here to make it clear that we need to use this way of conversion cause GasZip requires right-padded bytes32 value. 
WDYT ?

0xDEnYO (approved):

script/demoScripts/utils/demoScriptHelpers.ts
```diff
@@ -15,6 +15,7 @@ import {
   http,
   Narrow,
   parseAbi,
+  zeroAddress,
```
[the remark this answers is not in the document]
mirooon: good catch

script/demoScripts/demoGasZip.ts
```diff
@@ -0,0 +1,87 @@
...
+
+  // === Prepare bridge data ===
+  const bridgeData: ILiFi.BridgeDataStruct = {
+    transactionId: `0x${randomBytes(32).toString('hex')}`,
+    bridge: 'gasZip',
+    integrator: 'ACME Devs',
+    referrer: zeroAddress,
+    sendingAssetId: zeroAddress, // <-- native token
+    receiver: NON_EVM_ADDRESS,
+    destinationChainId,
```
0xDEnYO, line 58: while it's not a problem for the script itself, I just want to point out that you are using GasZipChainId as "normal" chainId here which would be wrong if you were developing in the BE
The correct chainId used here should be 1151111081099710 which is our own custom chainId for Solana.

0xDEnYO (approved):

script/demoScripts/demoGasZip.ts
mirooon: hmm ok, looking at for example RelayFacet then inside of the GasZipFacet we should have mapping from our BE chains to GasZipChainId mapping. Am I right?

Is this BE mapping somewhere available?

0xDEnYO: No we do not need a mapping. 
`GasZipData` in `IGasZip.sol` has a `destinationChains` parameter which is used to signal to GasZip where to send the funds to (it can send to many chains with one call, so all chainIds are compressed into one value). 

Which actually makes me realize that your demoScript is wrong. 
You need to construct that destinationChains parameter correctly in your demo script as described here: 

https://dev.gas.zip/gas/code-examples/evm-deposit/contract-deposit

mirooon: I see that we already have `getDestinationChainsValue` function that is already in `GasZipFacet` and `GasZipPeriphery` and in both contracts this function is not used. It accepts `uint8[] calldata _chainIds` and returns `uint256 destinationChains`. GasZipData has `uint256 destinationChains` so I assume in periphery contract constructing dectinationChains parameter is already done on the backend side.

Do you know more context. If it works this way then we can remove `getDestinationChainsValue` function.

0xDEnYO: This function is a convenience function for third parties to have an easy way to calculate that parameter. 
I do believe it makes sense to keep it there.

0xDEnYO: you could use that function in your demo script by the way....

0xDEnYO (approved):

PR #1062, opened 2025-03-18
re-add missing imports in TestBase contracts

0xDEnYO:

### DIFF
--- a/test/solidity/utils/TestBase.sol
+++ b/test/solidity/utils/TestBase.sol
@@ -2,6 +2,8 @@
 pragma solidity ^0.8.17;
 
 import { Test } from "forge-std/Test.sol";
+// solhint-disable-next-line no-unused-import
+import { console } from "../utils/Console.sol";
 import { DSTest } from "ds-test/test.sol";
 import { ILiFi } from "lifi/Interfaces/ILiFi.sol";
 import { LibSwap } from "lifi/Libraries/LibSwap.sol";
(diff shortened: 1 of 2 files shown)


ezynda3 (approved):

test/solidity/utils/TestBase.sol
```diff
@@ -2,6 +2,8 @@
 pragma solidity ^0.8.17;
 
 import { Test } from "forge-std/Test.sol";
+// solhint-disable-next-line no-unused-import
```
mirooon, line 5: IMO I suggest to keep only necessary imports. So add imports only in files where it's needed.

Keeping this rule we avoid also the necessity of checking TestBase everytime.

mirooon: wdyt? @0xDEnYO

0xDEnYO: What do you mean with "necessity of checking Testbase" exactly?

0xDEnYO: The reason we did it this way is that we can easily use console in all test contracts without having to add an import (and removing it afterwards when we dont use it anymore). 

I currently do not see any negative impact from this approach but defo open for other approaches if I am missing sth

mirooon: take a look at PR https://github.com/lifinance/contracts/pull/1054/files. There we already resigned from having `LibAllowList` in TestBase and changed test files to have direct imports.

0xDEnYO: OK, understood. 
Fine for me.

PR #1066, opened 2025-03-19
Add unpause diamond script

0xDEnYO:


script/deploy/safe/add-safe-owners-and-threshold.ts
```diff
@@ -34,139 +40,233 @@ const main = defineCommand({
       type: 'string',
       description: 'Private key of the signer',
     },
+    owners: {
+      type: 'string',
+      description: 'Comma-separated list of owner addresses to add',
+    },
```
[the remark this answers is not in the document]
0xDEnYO: outdated

script/utils/viemScriptHelpers.ts
```diff
@@ -1,12 +1,20 @@
 import { Chain, defineChain, getAddress } from 'viem'
 import networksConfig from '../../config/networks.json'
 import * as dotenv from 'dotenv'
+import consola from 'consola'
+import { getDeployments } from '../demoScripts/utils/demoScriptHelpers'
+import { SupportedChain } from '../demoScripts/utils/demoScriptChainConfig'
```
ezynda3, line 6: Not important now but maybe should look into consolidating scripts/utils and shared items as something about importing types from a completely different scripts domain feels gross to me lol. Just a personal preference thing.

script/tasks/unpauseAllDiamonds.ts
```diff
@@ -0,0 +1,198 @@
...
+  const facetAddresses: Address[] = []
+  for (const facetName of facetNames) {
+    try {
+      const facetAddress = await getContractAddressForNetwork(
+        facetName,
+        networkName
+      )
+
+      if (!facetAddress)
+        throw new Error(`No ${facetName} address found in deploy log.`)
```
mirooon, line 176: I think this error is redundant as the error is already thrown for `!address` in `getContractAddressForNetwork` function

script/tasks/unpauseAllDiamonds.ts
```diff
@@ -0,0 +1,198 @@
...
+
+      facetAddresses.push(facetAddress as Address)
+    } catch (error) {
+      consola.error(
```
mirooon, line 180: I think we should differentiate between the two types of errors that getContractAddressForNetwork can throw:
1. `throw Error(`Could not deploy log for network ${network} in ${environment}`)` - this is critical, as the deployment log is expected to exist. In this case, I suggest using consola.error() and rethrowing the error
2. `throw Error('Could not find address of contract ${contractName} for network ${network} in ${environment} deploy log')` - imo this is non-critical, it just means the contract name wasn’t found in the deployment log. Here, I’d recommend logging it with consola.info() and not rethrowing the error

mirooon (changes requested):

script/tasks/unpauseAllDiamonds.ts
```diff
@@ -0,0 +1,198 @@
...
+  networkName: string,
+  blacklistFacets: string
+): Promise<Address[]> {
+  // if blacklist is empty we dont need to look for addresses
+  if (!blacklistFacets) {
+    return []
+  }
+
+  // make sure that networkName is a valid supported chain
+  if (!isValidSupportedChain(networkName)) throw Error('Network ')
```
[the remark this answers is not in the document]
0xDEnYO: updated

script/deploy/safe/safe-utils.ts
```diff
@@ -1005,3 +1005,30 @@ export async function decodeTransactionData(data: Hex): Promise<{
...
+  const chain = getViemChainForNetworkName(network)
+
+  // Get Safe information directly from the contract
+  consola.info(`Getting Safe info for ${safeAddress} on ${network}`)
+  let safeInfo
+  try {
+    // Create a public client for read operations
+    const publicClient = createPublicClient({
+      chain,
+      transport: http(chain.rpcUrls.default.http[0]),
```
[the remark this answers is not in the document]
0xDEnYO: When the `chain` object is created, it actually prefers to use the .env value and only uses the public RPC if no .env is set. So I believe this is already taken care of. Please comment or resolve @mirooon 

image

script/tasks/unpauseAllDiamonds.ts
```diff
@@ -0,0 +1,198 @@
...
+              senderAddress
+            )
+
+            if (!result.acknowledged) {
+              throw new Error(
+                `[${network.name}] MongoDB insert was not acknowledged`
+              )
+            }
+
+            consola.info('Transaction successfully stored in MongoDB')
```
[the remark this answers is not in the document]
0xDEnYO: updated

script/tasks/unpauseAllDiamonds.ts
0xDEnYO: correct, well spotted, thank you

script/tasks/unpauseAllDiamonds.ts
0xDEnYO: I kinda want to disagree. 
For me both errors are critical.

We want to disable a facet while unpausing a diamond, which itself is aa critical operation. 

If we do not find this address (doesnt matter for which reason imho) then we cannot continue as without this address we cannot prepare the calldata for the unpauseDiamond call. 
In that case I would like to know what failed so I can investigate this network and fix manually, then try again.

wdyt @mirooon

script/deploy/safe/safe-utils.ts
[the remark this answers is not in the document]
mirooon: true!

script/tasks/unpauseAllDiamonds.ts
mirooon: I was thinking what if the blacklist includes facets that don’t exist in the deploy log? In that case, we could simply ignore them. But you’re right, even if those facets don’t exist, it still means the blacklist flag is incorrect (we throw an error) and the blacklist should be fixed during the second call.

thanks for the clarification @0xDEnYO

mirooon (approved):

mirooon (approved):

script/tasks/unpauseAllDiamonds.ts
```diff
@@ -0,0 +1,200 @@
...
+    } catch (error) {
+      consola.error(
+        `Error retrieving address for facet "${facetName}" on ${networkName}:`,
+        error
+      )
+    }
+  }
+
+  return facetAddresses
+}
```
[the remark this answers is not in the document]
0xDEnYO: not required, the warning is sufficient.

script/utils/viemScriptHelpers.ts
0xDEnYO: yes we will consolidate this in a separate step

ezynda3 (approved):

PR #1078, opened 2025-03-26
Update MayanFacet to handle FastMCTP

ezynda3:


0xDEnYO (dismissed):

src/Facets/MayanFacet.sol
```diff
@@ -255,6 +255,14 @@ contract MayanFacet is ILiFi, ReentrancyGuard, SwapperV2, Validatable {
                 // 0x2072197f MayanCircle::bridgeWithFee(address,uint256,uint64,uint64,bytes32,uint32,uint8,bytes)
                 receiver := mload(add(protocolData, 0xa4))
             }
+            case 0xf58b6de8 {
```
mirooon, line 258: all good. Are we okay with keeping the old, now-redundant function signatures—like MayanCircle::bridgeWithFee() (0x94454a5d) and MayanCircle::bridgeWithLockedFee() (0x32ad465f)? As far as I can tell, these aren’t handled anymore, and the SDK no longer generates payloads for them

maxklenk: Yes, I think we should keep them. This way we avoid a breaking change of the smart contract which allows to hot-switch it on-chain and then update the backend to switch to the new function.

PR #1087, opened 2025-04-07
Deploy new network: swellchain

0xDEnYO:
https://lifi.atlassian.net/browse/LF-13079


0xDEnYO:
SAFE Threshold needs to be increased to 3 before using this network in production. 
Have not done this since Ed is on holiday and some facets are missing in the configuration.

script/deploy/_targetState.json
```diff
@@ -1311,5 +1311,83 @@
...
+        "TokenWrapper": "1.1.0",
+        "LiFiDEXAggregator": "1.6.0",
+        "Permit2Proxy": "1.0.2",
+        "GasZipPeriphery": "1.0.1",
+        "AcrossFacetV3": "1.1.0",
+        "AcrossFacetPackedV3": "1.2.0",
+        "GasZipFacet": "2.0.2"
+      }
+    }
+  },
```
mirooon, line 1366: why are there hyperevm and lens changes?

0xDEnYO: when we parse the target state file (from the Google sheet) it parses the current version which contains updates from various topics, so it is expected to have unrelated changes there.

mirooon (approved):

script/deploy/healthCheck.ts
```diff
@@ -506,7 +506,12 @@ const main = defineCommand({
...
         const safeApiUrl = networkConfig.safeApiUrl
         const configUrl = `${safeApiUrl}/v1/safes/${safeAddress}`
-        const res = await fetch(configUrl)
+        let res
+        try {
+          res = await fetch(configUrl)
+        } catch (error) {
+          logError(`Could not fetch SAFE config from configURL: ${configUrl}`)
+        }
         const safeConfig = await res.json()
```
maxklenk, line 515: How is this supposed to work? If the fetch call fails and an error is logged, the next line will also fail. Is the catch only in place for the logging or should it prevent that the whole script fails?

maxklenk (dismissed):

script/deploy/healthCheck.ts
0xDEnYO: The script should fail in this case but you are right, this could be made more explicit by exiting with an error code in the catch clause. Will update it.

script/deploy/healthCheck.ts
```diff
@@ -506,7 +506,13 @@ const main = defineCommand({
...
         const safeApiUrl = networkConfig.safeApiUrl
         const configUrl = `${safeApiUrl}/v1/safes/${safeAddress}`
-        const res = await fetch(configUrl)
+        let res
+        try {
+          res = await fetch(configUrl)
+        } catch (error) {
+          logError(`Could not fetch SAFE config from configURL: ${configUrl}`)
+          process.exit(1)
+        }
```
kalote, line 515: shouldn't we log the error too?

0xDEnYO: I've added it

kalote (dismissed):

script/deploy/healthCheck.ts
```diff
@@ -506,7 +506,15 @@ const main = defineCommand({
...
         const configUrl = `${safeApiUrl}/v1/safes/${safeAddress}`
-        const res = await fetch(configUrl)
+        let res
+        try {
+          res = await fetch(configUrl)
+        } catch (error) {
+          logError(
+            `Could not fetch SAFE config from configURL ${configUrl}: ${error}`
+          )
+          process.exit(1)
```
mirooon, line 516: If `await fetch(configUrl)` fails here the script will log error but never reaches finish(), so we lose the summary of how many errors were found. It also skips any future checks we might add below. As Max mentioned, the lines after the fetch (like safeConfig = await res.json()) will fail if res is undefined.

IMO a good fix is to wrap both the `fetch` and the `res.json()` call in the same try catch and only run the SAFE checks if `safeConfig` is not `null`. That way if the request fails we can log the error but still continue to the rest of our checks

0xDEnYO: this is outdated now as it has been replaced with new code

maxklenk (approved):

PR #1145, opened 2025-05-15
LF-13736 - DexManagerFacet -> WhitelistManagerFacet + getApprovedFunctionSignatures function and selectors array [WhitelistManagerFacet v1.0.0,IWhitelistManagerFacet v1.0.0,LibAllowList v1.0.1]

mirooon:
[LF-13736]

[LF-13736]: https://lifi.atlassian.net/browse/LF-13736?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Facets/WhitelistManagerFacet.sol
```diff
@@ -4,92 +4,78 @@ pragma solidity ^0.8.17;
...
+    function addToWhitelist(address _address) external {
         if (msg.sender != LibDiamond.contractOwner()) {
             LibAccess.enforceAccessControl();
         }
 
-        LibAllowList.addAllowedContract(_dex);
+        LibAllowList.addAllowedContract(_address);
 
-        emit DexAdded(_dex);
+        emit AddressWhitelisted(_address);
```
0xDEnYO, line 25: did you check if the BE reads/consumes any of these events (not just the functions)?

src/Libraries/LibAllowList.sol
```diff
@@ -71,13 +69,51 @@ library LibAllowList {
     /// @dev Add a selector to the allow list
     /// @param _selector the selector to add
     function addAllowedSelector(bytes4 _selector) internal {
-        _getStorage().selectorAllowList[_selector] = true;
+        AllowListStorage storage als = _getStorage();
+
+        if (als.selectorAllowList[_selector]) {
+            // Check if selector exists in the array to handle legacy state
+            // where a selector might be true in selectorAllowList mapping but missing from the array.
+            // Since this is an admin-only function called infrequently, the array iteration is acceptable.
```
0xDEnYO, line 77: If we re-sync all our selectors and start fresh, why do we need legacy handling? Wondering if this is an unnecessary gas consumption (that happens on every sync).

src/Libraries/LibAllowList.sol
0xDEnYO, line 1: I am wondering how optimal/future-proof the flat arrays with the selectors and addresses are. 
Everytime we want to add/remove something from the array, we need to iterate through the whole array. If we want to remove element 53 from an array of 70, we have 52 unnecessary iterations.

There are other concepts with an additional mapping that stores the array index of a given address, e.g.: 
mapping public (address => uint256) addressToArrayIndex;

With such a mapping you can directly remove a given element (and push the rest one down). However, we need to pay for one more storage slot (the array index for a given element). 

Just thinking loud and wondering what your opinion is here?

src/Facets/WhitelistManagerFacet.sol
mirooon: good question. Yes, I checked it with Mathias

src/Libraries/LibAllowList.sol
mirooon: very good suggestion, currently we have in sigs.json 160 selectors and I think at this point it's good to implement additional mapping approach. `we need to pay for one more storage slot` we could implement `function removeAllowedSelectorAt(bytes4 selector, uint256 selectorIndex)` where we could provide index directly via offchain script and then we entierly can get rid of index store on-chain. That would be the most gas-efficient way but still it's some overhead with external script so I would go with onchain additional mapping.

Just to mention. We’re also looping over approved contract addresses. On polygon alone we have 52 contracts, and that number could increase. So it might make sense to apply the same on-chain mapping approach there as well.

Lastly I noticed we have whitelistedAddresses split by chain. Why sigs.json is not splitted the same way per chain for consistency?

src/Facets/WhitelistManagerFacet.sol
```diff
@@ -4,92 +4,78 @@ pragma solidity ^0.8.17;
...
+/// @notice Facet contract for managing whitelisted addresses for various protocol interactions.
+/// @custom:version 1.0.3
+contract WhitelistManagerFacet is IWhitelistManagerFacet {
     /// External Methods ///
 
-    /// @notice Register the address of a DEX contract to be approved for swapping.
-    /// @param _dex The address of the DEX contract to be approved.
-    function addDex(address _dex) external {
+    /// @inheritdoc IWhitelistManagerFacet
+    function addToWhitelist(address _address) external {
```
ezynda3, line 18: Maybe change from `_address` to `_contractAddress` for better readability?

src/Libraries/LibAllowList.sol
```diff
@@ -16,6 +16,7 @@ library LibAllowList {
         mapping(address => bool) allowlist;
```
ezynda3, line 16: could change to `contractAllowList` for consistency

src/Libraries/LibAllowList.sol
ezynda3: I would say we could probably remove all variables associated with the old layout if not needed in the new layout and update the namespace so the old whitelist is no longer referenced. Thoughts?

script/deploy/healthCheck.ts
```diff
@@ -256,44 +256,61 @@ const main = defineCommand({
...
 
-      // connect with diamond to get whitelisted DEXs
-      const dexManager = getContract({
+      // connect with diamond to get whitelisted addresses
+      const whitelistManager = getContract({
         address: deployedContracts['LiFiDiamond'],
         abi: parseAbi([
-          'function approvedDexs() external view returns (address[])',
-          'function isFunctionApproved(bytes4) external returns (bool)',
+          'function approvedDexs() external view returns (address[])', // old DexManagerFacet (depricated replaced by WhitelistManagerFacet getWhitelistedAddresses)
```
ezynda3, line 268: Should probably just use the new methods and ensure the old methods no longer work.

src/Libraries/LibAllowList.sol
ezynda3: I think an additional mapping approach would be our best bet while also organizing selectors by chain to reduce the storage size. We also need to keep in mind that when mapping indexes we should store indexes as `idx+1` to avoid the edge case of the first index being confused with empty.

src/Libraries/LibAllowList.sol
mirooon: I like this approach the most @ezynda3

src/Libraries/LibAllowList.sol
mirooon: Updated. added addtional mappings.

src/Facets/WhitelistManagerFacet.sol
mirooon: updated

src/Libraries/LibAllowList.sol
mirooon: updated

script/deploy/healthCheck.ts
mirooon: updated

script/tasks/diamondSyncWhitelistedAddresses.sh
```diff
@@ -66,16 +66,16 @@
 
     RPC_URL=$(getRPCUrl "$NETWORK") || checkFailure $? "get rpc url"
 
-    # Fetch required DEX addresses from configuration
-    CFG_DEXS=$(jq -r --arg network "$NETWORK" '.[$network][]' "./config/dexs.json")
+    # Fetch required whitelisted addresses from configuration
+    CFG_WHITELISTED_ADDRESSES=$(jq -r --arg network "$NETWORK" '.[$network][]' "./config/whitelistedAddresses.json")
```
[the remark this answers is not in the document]
mirooon: it works

script/tasks/diamondSyncWhitelistedAddresses.sh
```diff
@@ -93,63 +93,63 @@ function diamondSyncDEXs {
...
-      ADDRESS_STRING=$(printf "%s," "${NEW_DEXS[@]}")
+    # Add missing whitelisted addresses
+    if [[ ! ${#NEW_WHITELISTED_ADDRESSES[@]} -eq 0 ]]; then
+      ADDRESS_STRING=$(printf "%s," "${NEW_WHITELISTED_ADDRESSES[@]}")
       PARAMS="[${ADDRESS_STRING%,}]"
 
       local ATTEMPTS=1
       while [ $ATTEMPTS -le "$MAX_ATTEMPTS_PER_SCRIPT_EXECUTION" ]; do
-        cast send "$DIAMOND_ADDRESS" "batchAddDex(address[])" "${PARAMS[@]}" --rpc-url "$RPC_URL" --private-key $(getPrivateKey "$NETWORK" "$ENVIRONMENT") --legacy >/dev/null
+        cast send "$DIAMOND_ADDRESS" "batchAddToWhitelist(address[])" "${PARAMS[@]}" --rpc-url "$RPC_URL" --private-key $(getPrivateKey "$NETWORK" "$ENVIRONMENT") --legacy >/dev/null
```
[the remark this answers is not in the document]
mirooon: it works, no need to fix

src/Libraries/LibAllowList.sol
0xDEnYO: Btw are all sigs.json the same for all chains?
>> at the moment yes but this will change in the near future

and fine with Ed's approach

src/Facets/WhitelistManagerFacet.sol
```diff
@@ -4,92 +4,78 @@ pragma solidity ^0.8.17;
...
-
-    event DexAdded(address indexed dexAddress);
-    event DexRemoved(address indexed dexAddress);
-    event FunctionSignatureApprovalChanged(
-        bytes4 indexed functionSignature,
-        bool indexed approved
-    );
-
+/// @notice Facet contract for managing whitelisted addresses for various protocol interactions.
+/// @custom:version 1.0.3
```
0xDEnYO, line 13: Feels a bit funny that we rename the contract and then continue the versioning.  Maybe only to me?

How about we start at 1.0.0 (new contract) or add a comment that it was renamed from x (V.1.0.2) to y (V1.0.3) ? 
I am also OK if we just do it this way. Just wanted to raise my concern.

src/Libraries/LibAllowList.sol
0xDEnYO: awesome

src/Interfaces/IWhitelistManagerFacet.sol
```diff
@@ -0,0 +1,80 @@
...
+    function removeFromWhitelist(address _address) external;
+
+    /// @notice Batch unregister addresses from the whitelist.
+    /// @param _addresses The addresses to be removed from the whitelist.
+    function batchRemoveFromWhitelist(address[] calldata _addresses) external;
+
+    /// @notice Adds or removes a specific function signature to/from the allowlist.
+    /// @param _signature The function signature to allow or disallow.
+    /// @param _approval Whether the function signature should be allowed.
+    function setFunctionApprovalBySignature(
```
0xDEnYO, line 40: It's a function selector, not a signature. 
Signature is "transfer(uint256)" while selector is 0x12345678

Would vouch to change this naming, too. 

Sync sigs > sync selectors

Open for other opinions as usual

src/Facets/WhitelistManagerFacet.sol
mirooon: I was also thinking about this since we’re rebranding, I initially leaned toward continuing the versioning. But you’re right it’s technically a new contract so starting from 1.0.0 makes more sense. Great catch 👍🏻

src/Interfaces/IWhitelistManagerFacet.sol
mirooon: Yes that would describe it more accurately. updated!

src/Facets/WhitelistManagerFacet.sol
```diff
@@ -0,0 +1,152 @@
...
+    /// @inheritdoc IWhitelistManagerFacet
+    function addToWhitelist(address _contractAddress) external {
+        if (msg.sender != LibDiamond.contractOwner()) {
+            LibAccess.enforceAccessControl();
+        }
+
+        LibAllowList.addAllowedContract(_contractAddress);
+
+        emit AddressWhitelisted(_contractAddress);
+    }
```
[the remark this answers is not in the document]
mirooon: good point, fixed

0xDEnYO (dismissed):

ezynda3 (dismissed):

src/Libraries/LibAllowList.sol
0xDEnYO, line 1: I am a bit worried about the _checkAddress() function.
My concerns:

a) I feel it should be named isContract() as that is what it does
and, more importantly,
b) why do we not use the existing LibAsset.isContract() implementation ?

0xDEnYO, line 1: I feel like the lib could use some more comments inside the functions :)

test/solidity/Facets/WhitelistManagerFacet.t.sol
```diff
@@ -0,0 +1,774 @@
...
+        assertEq(approved[0], address(c1));
+
+        vm.stopPrank();
+    }
+
+    function test_SucceedsIfOwnerRemovesAddress() public {
+        vm.startPrank(USER_DIAMOND_OWNER);
+
+        vm.expectEmit(true, true, true, true);
+        emit AddressWhitelisted(address(c1));
```
0xDEnYO: You could call test_SucceedsIfOwnerAddsAddress() here and move the expectEmit into that test case

test/solidity/Facets/WhitelistManagerFacet.t.sol
```diff
@@ -0,0 +1,774 @@
...
+    }
+
+    function test_SucceedsIfOwnerBatchAddsAddresses() public {
+        vm.startPrank(USER_DIAMOND_OWNER);
+
+        address[] memory addresses = new address[](3);
+        addresses[0] = address(c1);
+        addresses[1] = address(c2);
+        addresses[2] = address(c3);
+        whitelistMgr.batchAddToWhitelist(addresses);
```
0xDEnYO: same here, could move this batchAdd part into its own test case and call it from others that need it for setup

test/solidity/Facets/WhitelistManagerFacet.t.sol
```diff
@@ -0,0 +1,774 @@
...
+    }
+
+    function test_SucceedsIfOwnerApprovesFunctionSelector() public {
+        vm.startPrank(USER_DIAMOND_OWNER);
+
+        bytes4 selector = hex"faceface";
+
+        vm.expectEmit(true, true, true, true);
+        emit FunctionSelectorApprovalChanged(selector, true);
+        whitelistMgr.setFunctionApprovalBySelector(selector, true);
```
0xDEnYO, line 167: nitpick: add blank line

test/solidity/Facets/WhitelistManagerFacet.t.sol
```diff
@@ -0,0 +1,774 @@
...
+
+    function test_SucceedsIfOwnerBatchApprovesFunctionSelectors() public {
+        vm.startPrank(USER_DIAMOND_OWNER);
+
+        bytes4[] memory selectors = new bytes4[](5);
+        selectors[0] = bytes4(hex"faceface");
+        selectors[1] = bytes4(hex"deadbeef");
+        selectors[2] = bytes4(hex"deaddead");
+        selectors[3] = bytes4(hex"deadface");
+        selectors[4] = bytes4(hex"beefbeef");
```
0xDEnYO, line 181: very creative :)
So that's what you think about when you are coding...beef and dead faces...:)

test/solidity/Facets/WhitelistManagerFacet.t.sol
```diff
@@ -0,0 +1,774 @@
...
+        address[] memory remaining = whitelistMgr.getWhitelistedAddresses();
+        assertEq(remaining.length, 1);
+        assertEq(remaining[0], address(c2));
+
+        assertFalse(whitelistMgr.isAddressWhitelisted(address(c1)));
+        assertFalse(whitelistMgr.isAddressWhitelisted(address(c3)));
+
+        vm.stopPrank();
+    }
+}
```
0xDEnYO, line 785: very nice and extensive test suite !

src/Libraries/LibAllowList.sol
mirooon: I can't argue, you are totally right here I overlooked this. We don't want to repeat ourselves in the code and isContract function from LibAsset does what we actually need to check!

mirooon: ok!

test/solidity/Facets/WhitelistManagerFacet.t.sol
mirooon: good point, updated

test/solidity/Facets/WhitelistManagerFacet.t.sol
mirooon: ok, created private helper function for it

test/solidity/Facets/WhitelistManagerFacet.t.sol
mirooon: good point! updated

test/solidity/Facets/WhitelistManagerFacet.t.sol
mirooon: haha 😄 dead beef means foooood

test/solidity/Facets/WhitelistManagerFacet.t.sol
mirooon: thank you!

script/tasks/diamondSyncWhitelistedAddresses.sh
```diff
@@ -205,5 +205,5 @@ function diamondSyncDEXs {
     return 0
   fi
 
-  echo "[info] <<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<< script syncDEXs completed"
-}
+  echo "[info] <<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<< script syncWhitelistedAddresses completed"
+}
```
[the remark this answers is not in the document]
mirooon: this function is called in deployAllContracts.sh file

0xDEnYO (approved):

PR #1215, opened 2025-06-16
exclude common values: (false, true, none, verifyContract) and safe/.../out directories from pre-commit check

0xDEnYO:
When the .env secrets checker finds common values such as true, false, etc.  then it fails as it finds these strings elsewhere in the codebase. Therefore it makes sense to add exclusions for that. 

Also there were some false positives in the safe/.../out folders.
These are also excluded now. 


.husky/pre-commit
```diff
@@ -96,8 +116,16 @@ doesFileContainDotEnvSecret() {
     local VALUE=$(echo "$SECRET" | cut -d '=' -f 2- | sed -e 's/^["'\''"]*//' -e 's/["'\''"]*$//')
     local KEY=$(echo "$SECRET" | cut -d '=' -f 1)
 
-    # skip empty values
-    if [ -z "$VALUE" ]; then
+    # skip empty values and known false positives
```
mirooon, line 119: nice changes!

.github/workflows/versionControlAndAuditCheck.yml
```diff
@@ -502,7 +502,7 @@ jobs:
             else
               # Use sort and uniq to remove duplicates, then count
               AUDIT_COUNT=$(echo "$AUDIT_IDS" | sort | uniq | wc -l)
-              
+
```
mirooon: I would suggest removing these unnecessary changes to avoid triggering a security review for this PR

.husky/pre-commit
```diff
@@ -96,8 +116,16 @@ doesFileContainDotEnvSecret() {
...
+    local IS_FALSE_POSITIVE=false
+    for FALSE_POSITIVE in "${KNOWN_FALSE_POSITIVES[@]}"; do
```
ezynda3, line 121: Can be optimized:
```sh
if [ -z "$VALUE" ] || [ "$IS_FALSE_POSITIVE" = true ]; then
  continue
 fi
```

ezynda3 (approved):

PR #1230, opened 2025-06-20
Implement MongoDB Deploy Logs Scripts

ezynda3:


script/deploy/update-deployment-logs.ts
```diff
@@ -0,0 +1,883 @@
...
+    return true
+  }
+
+  // Type guard for validating JSON data structure
+  private isValidJsonStructure(data: unknown): data is IJsonDataStructure {
+    if (typeof data !== 'object' || data === null) return false
+
+    // Basic structure validation - we'll do deeper validation during iteration
+    return true
+  }
```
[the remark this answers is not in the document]
ezynda3: overkill

script/helperFunctions.sh
```diff
@@ -398,8 +480,108 @@ function getHighestDeployedContractVersionFromMasterLog() {
...
+    --optimizer-runs "$OPTIMIZER_RUNS"
+    --timestamp "$TIMESTAMP"
+    --constructor-args "$CONSTRUCTOR_ARGS"
+    --verified "$VERIFIED"
+  )
+
+  # Add optional salt parameter if provided
+  if [[ -n "$SALT" ]]; then
+    MONGO_CMD+=("--salt=$SALT")
+  fi
```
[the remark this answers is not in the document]
ezynda3: not needed

script/deploy/deploySingleContract.sh
```diff
@@ -396,12 +396,18 @@ deploySingleContract() {
 
           # update VERIFIED info in log file
           logContractDeploymentInfo "$RELAYER_NAME" "$NETWORK" "$RELAYER_TIMESTAMP" "$RELAYER_VERSION" "$RELAYER_OPTIMIZER_RUNS" "$RELAYER_CONSTRUCTOR_ARGS" "$ENVIRONMENT" "$RELAYER_ADDRESS" "$RELAYER_VERIFIED" "$SALT"
+          
+          # log to MongoDB as well
```
0xDEnYO: how about just updating the helper function (`logContractDeploymentInfo`) instead of updating every occurrence?

script/deploy/shared/deployment-utils.ts
0xDEnYO, line 1: this file is barely commented (consider using AI to add comments)

0xDEnYO, line 1: Wondering if the name is accurate. 
How about something like "mongo-log-utils" or so?
Cause the way I understand the file it's only about logging, not about deployment itself, correct?

script/helperFunctions.sh
```diff
@@ -234,12 +234,23 @@ function checkIfJSONContainsEntry {
...
 function findContractInMasterLog() {
-  # read function arguments into variables
   local CONTRACT="$1"
   local NETWORK="$2"
   local ENVIRONMENT="$3"
   local VERSION="$4"
 
+  # Try MongoDB first if enabled
+  if isMongoLoggingEnabled; then
+    echoDebug "Trying MongoDB for findContractInMasterLog: $CONTRACT $NETWORK $ENVIRONMENT $VERSION"
```
0xDEnYO: In general, we must be careful to not log anything on the happy path other than the result of the function itself, otherwise we get something like

"Trying MongoDB for findContractInMasterLog:"

But I just added something to the echoDebug function that prints its output to stdErr and not to StdOut so this could work. We can test it. But keep it in mind when updating the bash scripts (super annoying)

script/helperFunctions.sh
```diff
@@ -234,12 +234,23 @@ function checkIfJSONContainsEntry {
...
+    local MONGO_RESULT=$(queryMongoDeployment "$CONTRACT" "$NETWORK" "$ENVIRONMENT" "$VERSION")
+    if [[ $? -eq 0 && -n "$MONGO_RESULT" ]]; then
```
0xDEnYO: Shouldnt this be || instead of && ?
What if the query was without error but no entry was found?

script/helperFunctions.sh
```diff
@@ -278,11 +289,39 @@ function findContractInMasterLog() {
...
+      if ! echo "$MONGO_RESULT" | jq -e . >/dev/null 2>&1; then
+        echoDebug "MongoDB returned invalid JSON, skipping: $MONGO_RESULT"
+        echoDebug "MongoDB query failed, falling back to JSON file"
+      else
+        # Convert MongoDB result to expected format
+        local CONTRACT_NAME=$(echo "$MONGO_RESULT" | jq -r '.contractName')
+        local VERSION=$(echo "$MONGO_RESULT" | jq -r '.version')
+        local ADDRESS=$(echo "$MONGO_RESULT" | jq -r '.address')
+        
+        if [[ "$CONTRACT_NAME" != "null" && "$VERSION" != "null" ]]; then
```
0xDEnYO, line 405: also here, maybe better || instead of && ?

script/helperFunctions.sh
```diff
@@ -320,12 +359,36 @@ function findContractInMasterLogByAddress() {
...
-  local TARGET_ADDRESS=$4
+  local NETWORK="$1"
+  local ENVIRONMENT="$2"
+  local CONTRACT="$3"
+  local TARGET_ADDRESS="$4"
+
+  # Try MongoDB first if enabled
+  if isMongoLoggingEnabled; then
+    echoDebug "Trying MongoDB for getContractVersionFromMasterLog: $CONTRACT $TARGET_ADDRESS"
+    local MONGO_RESULT=$(bun script/deploy/query-deployment-logs.ts find \
```
0xDEnYO, line 461: this could potentially return many results as we have many cases where contracts have the same address across networks. I dont see that we are dealing with this case here.

script/helperFunctions.sh
```diff
@@ -320,12 +359,36 @@ function findContractInMasterLogByAddress() {
...
+      --env="$ENVIRONMENT" \
+      --address="$TARGET_ADDRESS" 2>/dev/null)
+    
+    if [[ $? -eq 0 && -n "$MONGO_RESULT" ]]; then
+      local VERSION=$(echo "$MONGO_RESULT" | jq -r '.version')
+      local CONTRACT_NAME=$(echo "$MONGO_RESULT" | jq -r '.contractName')
+      
+      # Handle special case for CelerIMFacet
```
0xDEnYO: This can be removed (CelerIM is being deprecated)

script/helperFunctions.sh
```diff
@@ -398,8 +480,90 @@ function getHighestDeployedContractVersionFromMasterLog() {
...
+}
+
+
+function queryMongoDeployment() {
+  local CONTRACT="$1"
+  local NETWORK="$2"
+  local ENVIRONMENT="$3"
+  local VERSION="$4"
+
+  if ! isMongoLoggingEnabled; then
```
0xDEnYO: Is this check meaningful here? What if logging is currently disabled but I still want to query the DB?

script/helperFunctions.sh
```diff
@@ -398,8 +480,90 @@ function getHighestDeployedContractVersionFromMasterLog() {
...
+    --version="$VERSION" 2>/dev/null
+}
+
+function checkMongoDeploymentExists() {
+  local CONTRACT="$1"
+  local NETWORK="$2"
+  local ENVIRONMENT="$3"
+  local VERSION="$4"
+
+  if ! isMongoLoggingEnabled; then
```
0xDEnYO: same as above...not sure if logging needs to be enabled (feel we are mixing concerns here without clear reason why)

script/helperFunctions.sh
```diff
@@ -398,8 +480,90 @@ function getHighestDeployedContractVersionFromMasterLog() {
...
+    --network="$NETWORK" \
+    --version="$VERSION" 2>/dev/null
+}
+
+function getLatestMongoDeployment() {
+  local CONTRACT="$1"
+  local NETWORK="$2"
+  local ENVIRONMENT="$3"
+
+  if ! isMongoLoggingEnabled; then
```
0xDEnYO: same as above...not sure if logging needs to be enabled (feel we are mixing concerns here without clear reason why)

script/helperFunctions.sh
```diff
@@ -398,8 +480,90 @@ function getHighestDeployedContractVersionFromMasterLog() {
...
+    return 1
+  fi
+
+  bun script/deploy/query-deployment-logs.ts latest \
```
0xDEnYO, line 615: are these functions returning a correct return value by just running bun script at the end?

script/helperFunctions.sh
```diff
@@ -398,8 +480,90 @@ function getHighestDeployedContractVersionFromMasterLog() {
...
+
+
+
+
+
+
+function getUnverifiedContractsFromMongo() {
+  local ENVIRONMENT="$1"
+
+  if ! isMongoLoggingEnabled; then
```
0xDEnYO: Also here I would vouch to remove

script/helperFunctions.sh
```diff
@@ -1444,20 +1608,38 @@ function verifyContract() {
   echoDebug "FULL_PATH=$FULL_PATH"
   echoDebug "CHAIN_ID=$CHAIN_ID"
 
-  # Build base verification command
-  local VERIFY_CMD="forge verify-contract --watch --chain $CHAIN_ID $ADDRESS $FULL_PATH --skip-is-verified-check"
+  # Build verification command as array for safe execution
```
0xDEnYO, line 1808: Why was this change necessary?
Did you test this properly?

script/helperFunctions.sh
ezynda3, line 246: **Re: Line 246 - && vs || logic**

The current logic `if [[ $? -eq 0 && -n "$MONGO_RESULT" ]]; then` is actually correct for this use case. Here's why:

**Function Purpose**: This is in `findContractInMasterLog` - the function should only return a result when a contract is actually found.

**Two Distinct Scenarios**:
- `$? -eq 0 && -n "$MONGO_RESULT"` = Query succeeded AND found a result → Return the result  
- `$? -eq 0 && -z "$MONGO_RESULT"` = Query succeeded BUT no result found → Fall back to JSON file
- `$? -ne 0` = Query failed → Fall back to JSON file

**Fallback Behavior**: If we used `||`, we'd return empty results as valid findings, which would prevent the fallback to the JSON file method that might actually contain the contract data.

**Expected Behavior**: When MongoDB has no record of a contract, we want to fall back to the existing JSON file system, not return "empty/not found" immediately.

The current `&&` logic ensures we only trust MongoDB results when both the query succeeds AND returns actual data. Empty successful queries should trigger the fallback mechanism.

script/helperFunctions.sh
ezynda3, line 314: **Re: Line 314 - && vs || logic**

Same reasoning as line 246 - the current logic is correct. This appears to be checking if we got valid contract name and version data from MongoDB:

`if [[ "$CONTRACT_NAME" != "null" && "$VERSION" != "null" ]]; then`

We want BOTH fields to be non-null to consider it a valid result. If either field is null, we shouldn't return partial/invalid data. The && logic ensures we only proceed when we have complete, valid data from the MongoDB query.

script/helperFunctions.sh
ezynda3: https://github.com/lifinance/contracts/pull/1230/files#r2164728290

script/helperFunctions.sh
ezynda3: https://github.com/lifinance/contracts/pull/1230/files#r2164728290

script/helperFunctions.sh
ezynda3: https://github.com/lifinance/contracts/pull/1230/files#r2164728290

script/helperFunctions.sh
ezynda3, line 1611: **Re: Line 1611 change**

This change was made to address a CodeRabbit comment about `eval` being unsafe. The verification command is now built as an array for safe execution, which prevents issues with arguments containing spaces or special characters.

Also shuffled the if logic to read better and be more maintainable.

script/helperFunctions.sh
ezynda3, line 378: Fixed: https://github.com/lifinance/contracts/commit/fa90d9e3836a5fc3ed858738241cb6fd75a3bd2c

script/helperFunctions.sh
ezynda3, line 518: Fixed: https://github.com/lifinance/contracts/commit/fa90d9e3836a5fc3ed858738241cb6fd75a3bd2c

script/helperFunctions.sh
ezynda3: This change was made to address a CodeRabbit comment about eval being unsafe. The verification command is now built as an array for safe execution, which prevents issues with arguments containing spaces or special characters.

Also shuffled the if logic to read better and be more maintainable.

Tested and it works.

script/helperFunctions.sh
ezynda3: This is actually just a naming issue. isMongoLoggingEnabled doesn't just mean "are we writing logs to MongoDB" - it means "are we using MongoDB for deployment logs" (both read and write operations).

The current logic is correct: these functions should only query MongoDB when MongoDB logging is enabled, otherwise they fall back to the JSON file system. If MongoDB logging is disabled, we want to use the existing JSON-based log system instead.

This function will eventually be removed once the local .json logs are phased out and MongoDB becomes the primary storage.

script/helperFunctions.sh
ezynda3: Same reasoning as line 246 - the current logic is correct. This appears to be checking if we got valid contract name and version data from MongoDB:

if [[ "$CONTRACT_NAME" != "null" && "$VERSION" != "null" ]]; then

We want BOTH fields to be non-null to consider it a valid result. If either field is null, we shouldn't return partial/invalid data. The && logic ensures we only proceed when we have complete, valid data from the MongoDB query.

script/helperFunctions.sh
ezynda3: The current logic if [[ $? -eq 0 && -n "$MONGO_RESULT" ]]; then is actually correct for this use case. Here's why:

Function Purpose: This is in findContractInMasterLog - the function should only return a result when a contract is actually found.

Two Distinct Scenarios:

$? -eq 0 && -n "$MONGO_RESULT" = Query succeeded AND found a result → Return the result
$? -eq 0 && -z "$MONGO_RESULT" = Query succeeded BUT no result found → Fall back to JSON file
$? -ne 0 = Query failed → Fall back to JSON file

Fallback Behavior: If we used ||, we'd return empty results as valid findings, which would prevent the fallback to the JSON file method that might actually contain the contract data.

Expected Behavior: When MongoDB has no record of a contract, we want to fall back to the existing JSON file system, not return "empty/not found" immediately.

The current && logic ensures we only trust MongoDB results when both the query succeeds AND returns actual data. Empty successful queries should trigger the fallback mechanism.

script/helperFunctions.sh
ezynda3: Fixed: https://github.com/lifinance/contracts/commit/fa90d9e3836a5fc3ed858738241cb6fd75a3bd2c

script/helperFunctions.sh
ezynda3: Fixed: https://github.com/lifinance/contracts/commit/fa90d9e3836a5fc3ed858738241cb6fd75a3bd2c

script/helperFunctions.sh
ezynda3: Fixed: https://github.com/lifinance/contracts/commit/fa90d9e3836a5fc3ed858738241cb6fd75a3bd2c

script/helperFunctions.sh
ezynda3: Fixed: https://github.com/lifinance/contracts/commit/fa90d9e3836a5fc3ed858738241cb6fd75a3bd2c

script/deploy/deploySingleContract.sh
ezynda3: Fixed: 937c7ed9

script/deploy/shared/deployment-utils.ts
ezynda3: Fixed: 0c6e69d4

ezynda3: Fixed: f4614516

script/helperFunctions.sh
0xDEnYO: ok

script/helperFunctions.sh
0xDEnYO: I understand the thinking.
But these helper functions do not contain the fallback. They either work (if flag=true) or they just return false.
I feel this is wrong, helper functions should always work. 
The logic which path to choose (mongo or json) should sit further above. And the helper functions should work regardless of that logic. Cause what if mongo logging is disabled and I want to query the DB...it's not gonna work, right?

script/helperFunctions.sh
0xDEnYO: ok

script/deploy/query-deployment-logs.ts
```diff
@@ -0,0 +1,638 @@
...
+}
+
+class DeploymentLogQuerier {
+  private client: MongoClient
+  private db: Db
+  private collection: Collection<IDeploymentRecord>
+
+  public constructor(
+    private config: IConfig,
+    private environment: 'staging' | 'production'
```
mirooon: I suggest to use `EnvironmentEnum` from `script/common/types.ts` file

package.json
```diff
@@ -84,6 +84,15 @@
     "remove-from-diamond": "bun script/tasks/cleanUpProdDiamond.ts",
     "test": "forge test --evm-version 'cancun'",
     "test:fix": "npm run lint:fix; npm run format:fix; npm run test",
+    "mongo-logs:sync": "bun script/deploy/update-deployment-logs.ts sync",
+    "mongo-logs:add": "bun script/deploy/update-deployment-logs.ts add",
+    "mongo-logs:update": "bun script/deploy/update-deployment-logs.ts update",
+    "mongo-logs:query": "bun script/deploy/query-deployment-logs.ts",
+    "mongo-logs:latest": "bun script/deploy/query-deployment-logs.ts latest",
+    "mongo-logs:list": "bun script/deploy/query-deployment-logs.ts list",
```
mirooon: I noticed some inconsistency across commands. When I run query latest or query find, I get the full raw document from the DB, which includes all metadata fields (like _id, updatedAt, etc). But when I run query list, the output is more human-readable but also omits some fields (like contractNetworkKey, constructorArgs, etc). feels a bit inconsistent.

wdyt?

Example output for reference:

bun run mongo-logs:query latest --env production --network arbitrum --contract GasZipFacet
$ bun script/deploy/query-deployment-logs.ts latest --env production --network arbitrum --contract GasZipFacet
ℹ Connected to MongoDB collection: production                                                                                                                                                                                                    8:58:00 AM
{
  "_id": "68555f4e2359b4efc476e78f",
  "address": "0xF5c923a087fb3c554579e2DD10AB6E37E0f6F849",
  "contractName": "GasZipFacet",
  "network": "arbitrum",
  "version": "2.0.0",
  "constructorArgs": "0x0000000000000000000000002a37d63eadfe4b4682a3c28c1c2cd4f109cc2762",
  "contractNetworkKey": "GasZipFacet-arbitrum",
  "contractVersionKey": "GasZipFacet-2.0.0",
  "createdAt": "2025-06-20T13:17:01.935Z",
  "optimizerRuns": "1000000",
  "salt": null,
  "timestamp": "2024-11-11T14:19:32.000Z",
  "updatedAt": "2025-06-20T13:17:01.935Z",
  "verified": true
}
bun run mongo-logs:query find --address 0xF5c923a087fb3c554579e2DD10AB6E37E0f6F849        
$ bun script/deploy/query-deployment-logs.ts find --address "0xF5c923a087fb3c554579e2DD10AB6E37E0f6F849"
ℹ Connected to MongoDB collection: production                                                                                                                                                                                                    8:58:09 AM
{
  "_id": "68555f4e2359b4efc476e78d",
  "address": "0xF5c923a087fb3c554579e2DD10AB6E37E0f6F849",
  "contractName": "GasZipFacet",
  "network": "bsc",
  "version": "2.0.0",
  "constructorArgs": "0x0000000000000000000000002a37d63eadfe4b4682a3c28c1c2cd4f109cc2762",
  "contractNetworkKey": "GasZipFacet-bsc",
  "contractVersionKey": "GasZipFacet-2.0.0",
  "createdAt": "2025-06-20T13:17:01.935Z",
  "optimizerRuns": "1000000",
  "salt": null,
  "timestamp": "2024-11-11T14:22:00.000Z",
  "updatedAt": "2025-06-20T13:17:01.935Z",
  "verified": true
}
bun run mongo-logs:query list                                                             
$ bun script/deploy/query-deployment-logs.ts list
ℹ Connected to MongoDB collection: production                                                                                                                                                                                                    8:58:25 AM
✔ Found 50 deployment(s) (Page 1 of 44, Total: 2159):                                                                                                                                                                                            8:58:25 AM

Contract: GlacisFacet
Network: viction
Version: 1.0.0
Address: 0x4b67f2bB8855c5d222dF8159B57C54b795403452
Timestamp: 2025-06-19T18:53:49.000Z
Verified: true
Optimizer Runs: 1000000
Salt: 20250619
---

Contract: GasZipPeriphery
Network: katana
Version: 1.0.1
Address: 0x6BCAF079647788ceC14f10D6601Ecb34d0b79c6E
Timestamp: 2025-06-13T12:33:29.000Z
Verified: true
Optimizer Runs: 1000000

---

...


ℹ Use --page 2 to see more results

script/deploy/query-deployment-logs.ts
```diff
@@ -0,0 +1,638 @@
...
+const config: IConfig = {
+  mongoUri: process.env.MONGODB_URI || 'mongodb://localhost:27017',
+  batchSize: 100,
+  databaseName: 'contract-deployments',
+}
+
+class DeploymentLogQuerier {
+  private client: MongoClient
+  private db: Db
+  private collection: Collection<IDeploymentRecord>
```
mirooon: I am getting this TS lint issues:

Property 'collection' has no initializer and is not definitely assigned in the constructor.

same for db

you can just do:

`private collection!: Collection<IDeploymentRecord>`

script/deploy/query-deployment-logs.ts
```diff
@@ -0,0 +1,638 @@
...
+      consola.info(`Connected to MongoDB collection: ${collectionName}`)
+    } catch (error) {
+      consola.error('Failed to connect to MongoDB:', error)
+      throw error
+    }
+  }
+
+  public async getLatestDeployment(
+    contractName: string,
+    network: string
```
mirooon, line 76: I suggest to use `SupportedChain` from `script/common/types.ts` file

mirooon: same below

script/deploy/shared/mongo-log-utils.ts
```diff
@@ -0,0 +1,558 @@
...
+  version: string
+  /** Contract address on the blockchain */
+  address: string
+  /** Number of optimizer runs used during compilation */
+  optimizerRuns: string
+  /** Timestamp when the contract was deployed */
+  timestamp: Date
+  /** Constructor arguments used during deployment */
+  constructorArgs: string
+  /** Optional salt value used for CREATE2 deployments */
```
mirooon: shouldn't be CREATE3?

script/deploy/shared/mongo-log-utils.ts
```diff
@@ -0,0 +1,558 @@
...
+   * @param version - Version of the contract
+   * @param _environment - Deployment environment (currently unused but kept for consistency)
+   * @returns Function that transforms raw data to IDeploymentRecord or null if invalid
+   */
+  public static transformRawToDeployment =
+    (
+      contractName: string,
+      network: string,
+      version: string,
+      _environment: string
```
mirooon: unused arg

script/deploy/query-deployment-logs.ts
```diff
@@ -0,0 +1,638 @@
...
+      .toArray()
+  }
+
+  public async disconnect(): Promise<void> {
+    await this.client.close()
+  }
+}
+
+// Helper function to format deployment record for display
+function formatDeployment(deployment: IDeploymentRecord): string {
```
mirooon, line 171: I think it's worth to add createdAt and updatedAt then? wdyt? or name it as summary

script/deploy/query-deployment-logs.ts
```diff
@@ -0,0 +1,638 @@
...
+      const deployment = await querier.getLatestDeployment(
+        args.contract,
+        args.network
+      )
+
+      if (deployment) outputJSON(deployment)
+      else process.exit(1)
+    } catch (error) {
+      consola.error('Query failed:', error)
+      process.exit(1)
```
mirooon, line 254: i think if the program stops right away and does not run any finally blocks, so await querier.disconnect() is skipped and the database connection stays open. I would call  await querier.disconnect() also before process.exit(1). Same for other cases with process.exit(). wdyt?

script/helperFunctions.sh
ezynda3: Fixed: f65988fb083c0e01c40e64448c2fd745ad3f2d10

script/helperFunctions.sh
ezynda3: Fixed: f65988fb083c0e01c40e64448c2fd745ad3f2d10

script/helperFunctions.sh
ezynda3: Fixed: f65988fb083c0e01c40e64448c2fd745ad3f2d10

script/helperFunctions.sh
ezynda3: Fixed: f65988fb083c0e01c40e64448c2fd745ad3f2d10

package.json
ezynda3: Fixed: 1a138bcf5da1d0d35538b534f8a7c1dbf7ede340

script/deploy/query-deployment-logs.ts
ezynda3: fixed: 3298710023497f4a30595870fb38125c2f17787d

script/deploy/query-deployment-logs.ts
ezynda3: Fixed: 3298710023497f4a30595870fb38125c2f17787d

script/deploy/query-deployment-logs.ts
ezynda3: Fixed: 3298710023497f4a30595870fb38125c2f17787d

script/deploy/query-deployment-logs.ts
ezynda3: Fixed: 3298710023497f4a30595870fb38125c2f17787d

script/deploy/shared/mongo-log-utils.ts
ezynda3: Fixed: 8c00d3af37b2a5cd19d88f7a4988b05702a0b84e

script/deploy/shared/mongo-log-utils.ts
ezynda3: fixed: 8c00d3af37b2a5cd19d88f7a4988b05702a0b84e

0xDEnYO (approved):

mirooon (approved):

PR #1238, opened 2025-06-26
Redeploy GenericSwapFacetV3 with new native address on hyperevm

0xDEnYO:


deployments/_deployments_log_file.json
```diff
@@ -28150,6 +28150,16 @@
...
           }
+        ],
+        "1.0.2": [
+          {
+            "ADDRESS": "0xa12D48dBf4248d702D5b41e271742914D7d8A321",
+            "OPTIMIZER_RUNS": "1000000",
+            "TIMESTAMP": "2025-06-26 09:08:01",
+            "CONSTRUCTOR_ARGS": "0x0000000000000000000000000000000000000000000000000000000000000000",
+            "SALT": "20250624",
+            "VERIFIED": true
```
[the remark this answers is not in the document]
0xDEnYO: fixed

mirooon (approved):

config/networks.json
```diff
@@ -444,7 +444,7 @@
   "hyperevm": {
     "name": "hyperevm",
     "chainId": 999,
-    "nativeAddress": "0x2222222222222222222222222222222222222222",
+    "nativeAddress": "0x0000000000000000000000000000000000000000",
```
mirooon, line 447: Maybe this is not the right PR because it’s only for redeployment, but I wanted to share a thought. We use `NATIVE_ADDRESS` in `GenericSwapFacetV3`, and we also use `NATIVE_ASSETID` in `LibAsset` lib, which is actually `NULL_ADDRESS` which equals `address(0)`.

This means we might have two different sources for the native asset address: one from `network.json`, and one hardcoded as `address(0)` in `LibAsset`. I’m not sure if this could cause problems later but imo it’s something we should think about

0xDEnYO: yeah it's a good point. I wonder why this has not caused any problems yet...

mirooon (approved):

PR #1302, opened 2025-07-25
make gaszipfacet a core facet

0xDEnYO:

### DIFF
--- a/script/deploy/facets/UpdateCoreFacets.s.sol
+++ b/script/deploy/facets/UpdateCoreFacets.s.sol
@@ -27,6 +27,7 @@ contract DeployScript is UpdateScriptBase {
             path,
             ".PeripheryRegistryFacet"
         );
+        address gasZip = _getConfigContractAddress(path, ".GasZipFacet");
         address genSwapAddress = _getConfigContractAddress(
             path,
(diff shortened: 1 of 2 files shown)


script/deploy/facets/UpdateCoreFacets.s.sol
```diff
@@ -110,6 +114,14 @@ contract DeployScript is UpdateScriptBase {
...
         }
 
+        // GasZipFacet
+        selectors = getSelectors("GasZipFacet", exclude);
+        if (loupeExists) {
+            buildDiamondCut(selectors, gasZipAddress);
+        } else {
+            buildInitialCut(selectors, gasZipAddress);
+        }
+
```
mirooon: in this file there is a lot of repetitive code, following DRY principle what do you think about reducing logic into single reusable function? I believe that in this file we can have a function with list of names and for the next core facet just add new element. Or what could be even better, core facet could be taken from global.json to keep list of core facets in one place. wdyt about it?

0xDEnYO: yes, great suggestion. Very good point. Didnt think about that at all.
Updated.

PR #1321, opened 2025-08-07
(feat): RelayDepositoryFacet [RelayDepositoryFacet v1.0.0,IRelayDepository v1.0.0]

0xDEnYO:
https://lifi.atlassian.net/browse/LF-14766

### DIFF
--- a/deployments/arbitrum.diamond.staging.json
+++ b/deployments/arbitrum.diamond.staging.json
@@ -22,7 +22,6 @@
         "Version": "1.0.0"
       },
       "0xa574407A8bA37E83841eCa2084b1DAC51d0f0b8A": {
-      "0xa574407A8bA37E83841eCa2084b1DAC51d0f0b8A": {
         "Name": "",
         "Version": ""
       },


src/Facets/RelayDepositoryFacet.sol
```diff
@@ -0,0 +1,148 @@
...
+
+            IRelayDepository(_relayDepositoryData.depository).depositErc20(
+                msg.sender,
+                _bridgeData.sendingAssetId,
+                _bridgeData.minAmount,
+                _relayDepositoryData.orderId
+            );
+        }
+
+        emit RelayDepositoryDeposit(
```
mirooon: what's the purpose of RelayDepositoryDeposit event?
depository contract already emits RelayErc20Deposit or RelayNativeDeposit

src/Facets/RelayDepositoryFacet.sol
```diff
@@ -0,0 +1,148 @@
...
+    address public immutable RELAY_DEPOSITORY;
+
+    /// Types ///
+
+    /// @dev Relay Depository specific parameters
+    /// @param orderId Unique identifier for this deposit order
+    /// @param depository The address of the Relay Depository contract (must match immutable)
+    struct RelayDepositoryData {
+        bytes32 orderId;
+        address depository;
```
mirooon: from this conversation:

image
image

I think we can remove `depository` from `RelayDepositoryData` and use `RELAY_DEPOSITORY` to call deposits functions in `_startBridge` function

src/Facets/RelayDepositoryFacet.sol
```diff
@@ -0,0 +1,148 @@
...
+        bytes32 indexed orderId,
+        address indexed depository
+    );
+
+    /// Constructor ///
+
+    /// @param _relayDepository The address of the Relay Depository contract
+    constructor(address _relayDepository) {
+        if (_relayDepository == address(0)) {
+            revert InvalidCallData();
```
mirooon, line 45: line not covered in the tests

script/demoScripts/demoRelayDepository.ts
```diff
@@ -0,0 +1,154 @@
...
+
+  let bridgeData: ILiFi.BridgeDataStruct = {
+    transactionId: utils.randomBytes(32),
+    bridge: 'RelayDepository',
+    integrator: 'ACME Devs',
+    referrer: '0x0000000000000000000000000000000000000000',
+    sendingAssetId: '0x0000000000000000000000000000000000000000', // Native ETH
+    receiver: address,
+    minAmount: ethers.utils.parseEther('0.0001'),
+    destinationChainId: 0, // We are not bridging, just depositing, hence no destination chain id
```
ezynda3, line 55: nit: maybe would make sense to use the same chainId as the source chain?

src/Facets/RelayDepositoryFacet.sol
ezynda3: We can remove this.

src/Facets/RelayDepositoryFacet.sol
ezynda3: This can be removed I think

script/demoScripts/demoRelayDepository.ts
0xDEnYO: cant, this will fail when we validate the bridgeData

src/Facets/RelayDepositoryFacet.sol
0xDEnYO: test added

src/Facets/RelayDepositoryFacet.sol
0xDEnYO: ok, this was muddled up.
The parameter was here to be able to determine the depositorAddress (instead of using msg.sender as this wont work when using gasless transactions for example).
I've updated this, pls have a look again.

src/Facets/RelayDepositoryFacet.sol
0xDEnYO: correct, I removed it

src/Facets/RelayDepositoryFacet.sol
```diff
@@ -0,0 +1,135 @@
...
+    /// Internal Methods ///
+
+    /// @dev Contains the business logic for depositing into Relay Depository
+    /// @param _bridgeData The core information needed for bridging
+    /// @param _relayDepositoryData Data specific to Relay Depository
+    function _startBridge(
+        ILiFi.BridgeData memory _bridgeData,
+        RelayDepositoryData calldata _relayDepositoryData
+    ) internal {
+        if (LibAsset.isNativeAsset(_bridgeData.sendingAssetId)) {
```
mirooon, line 119: I couldn’t find this explicitly documented, but if the depositor defaults to the receiver, we should revert when depositorAddress == address(0). Otherwise, the RELAY_DEPOSITORY contract would credit msg.sender (https://etherscan.io/address/0x4cd00e387622c35bddb9b4c962c136462338bc31#code) which in our case would be the diamond contract

0xDEnYO: very good catch @mirooon 
Thanks a lot. 
Added (incl tests)

mirooon (dismissed):

docs/RelayDepositoryFacet.md
```diff
@@ -0,0 +1,155 @@
+# RelayDepositoryFacet
+
+## How it works
+
+The RelayDepositoryFacet enables direct deposits of assets into Relay Protocol V2 Depositories. This facet supports both native tokens and ERC20 tokens, with optional swap functionality before depositing.
+
+The facet is configured with a specific Relay Depository address during deployment via its constructor parameter. All deposit operations must use this configured depository address.
+
+**Note**: If a deposit amount is higher than the amount associated with the orderId, the overpaid amount will be forwarded to the destination chain (to the receiverAddress specified in the off-chain data associated with orderId)
+
```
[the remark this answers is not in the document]
0xDEnYO: Since unit tests do not show what happens with the funds after depositing, these tests would not add any benefits

ezynda3 (approved):

PR #1332, opened 2025-08-22
add warning when whitelisting transferFrom selector

0xDEnYO:

### DIFF
--- a/script/tasks/diamondSyncSigs.sh
+++ b/script/tasks/diamondSyncSigs.sh
@@ -34,7 +34,7 @@ function diamondSyncSigs {
     echo ""
 
     if [[ "$NETWORK" != "All (non-excluded) Networks" ]]; then
-      checkRequiredVariablesInDotEnv $NETWORK
+      checkRequiredVariablesInDotEnv "$NETWORK"
     fi
   fi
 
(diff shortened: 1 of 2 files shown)


script/tasks/diamondSyncSigs.sh
```diff
@@ -110,6 +110,23 @@ function diamondSyncSigs {
       # Fallback to cast send
       if [[ $ATTEMPTS == 1 ]]; then
         CFG_SIGS=($(jq -r '.[] | @sh' "./config/sigs.json" | tr -d \' | tr '[:upper:]' '[:lower:]'))
+
+        # Check for dangerous function selector: transferFrom (0x23b872dd)
+        for sig in "${CFG_SIGS[@]}"; do
+          if [[ "$sig" == "0x23b872dd" ]]; then
```
mirooon: maybe would be add to add comment that if we change selectors here then the list of selectors should be changed also in diamondSyncSigs.ts.

or even better we could store dangerous function selectors in global.json to keep the list in one place?

script/tasks/diamondSyncSigs.ts
```diff
@@ -82,6 +82,26 @@ const main = defineCommand({
 
     // Check if function signatures are approved
     const { sigs } = await import(`../../config/sigs.json`)
+
+    // Security check: Prevent whitelisting of dangerous function selectors
```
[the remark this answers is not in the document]
0xDEnYO: are you sure? Cause the script has been working fine like this

script/tasks/diamondSyncSigs.sh
0xDEnYO: very good point, updated

script/tasks/diamondSyncSigs.ts
```diff
@@ -82,6 +82,24 @@ const main = defineCommand({
 
     // Check if function signatures are approved
     const { sigs } = await import(`../../config/sigs.json`)
+    const { blacklistedFunctionSelectors } = await import(
+      `../../config/global.json`
+    )
+
+    // Security check: Prevent whitelisting of dangerous function selectors
+    for (const sig of sigs)
+      if (blacklistedFunctionSelectors.includes(sig.toLowerCase())) {
```
mirooon: imo we should also lowerCase selectors from blacklistedFunctionSelectors. I know we usually operate on lowercased selector but just in case. Also to keep consistency with bash logic

0xDEnYO: good point, updated

PR #1351, opened 2025-09-02
Simple balance checking util

ezynda3:

### DIFF
--- a/script/balances.ts
+++ b/script/balances.ts
@@ -1,0 +1,312 @@
+#!/usr/bin/env bun
+
+import { defineCommand, runMain } from 'citty'
+import { consola } from 'consola'
+import type { TronWeb } from 'tronweb'
+import {
+  createPublicClient,
+  http,
+  formatEther,
+  isAddress as isValidEvmAddress,
+} from 'viem'
+
+import networksConfig from '../config/networks.json'
+
+import { initTronWeb } from './troncast/utils/tronweb'
+import { node_url } from './utils/network'
+
+// Load networks configuration
+
+interface INetworkConfig {
+  name: string
+  chainId: number
+  nativeAddress: string
+  nativeCurrency: string
+  wrappedNativeAddress: string
+  status: string
+  type: string
+  rpcUrl: string
+  verificationType: string
+  explorerUrl: string
+  explorerApiUrl: string
+  multicallAddress: string
+  safeAddress: string
+  gasZipChainId: number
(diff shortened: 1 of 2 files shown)


script/balances.ts
```diff
@@ -0,0 +1,312 @@
...
+): Promise<IBalanceResult> {
+  try {
+    // Try to get RPC URL from environment variable first, fallback to networks.json
+    const rpcUrl = node_url(network.name) || network.rpcUrl
+
+    const client = createPublicClient({
+      transport: http(rpcUrl, {
+        timeout: 10_000,
+        retryCount: 2,
+        retryDelay: 1000,
```
0xDEnYO, line 98: interesting...should we add this to our other helper functions, too?

script/balances.ts
```diff
@@ -0,0 +1,312 @@
...
+  const successCount = results.filter((r) => !r.error).length
+  const errorCount = results.filter((r) => r.error).length
+  const nonZeroBalances = results.filter(
+    (r) => !r.error && parseFloat(r.formattedBalance) > 0
+  )
+
+  console.log('\nSummary:')
+  console.log(`  Total chains checked: ${results.length}`)
+  console.log(`  Successful queries: ${successCount}`)
+  console.log(`  Failed queries: ${errorCount}`)
```
0xDEnYO, line 175: maybe we should also print which networks where unsuccessful ?

script/balances.ts
0xDEnYO, line 1: If I understand this script correctly, it will just distinguish between 0 and non-0. 
However, in most cases we will run low on funds on networks and the balance will be non-0 but still too low which the script will not detect. What are you thoughts on this?

ezynda3: It just shows the actual balances 0 or not.

script/balances.ts
ezynda3: Added a ticket https://lifi.atlassian.net/jira/software/c/projects/LF/boards/5?selectedIssue=LF-15417

script/balances.ts
ezynda3: image

script/balances.ts
ezynda3: Updated

script/balances.ts
0xDEnYO: awesome, thanks. 
You can resolve such comments directly btw

0xDEnYO (approved):

PR #1360, opened 2025-09-04
LF-13997 LDA 2.0 - Scripts

mirooon:
[LF-13997]

[LF-13997]: https://lifi.atlassian.net/browse/LF-13997?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


deployments/arbitrum.lda.staging.json
0xDEnYO, line 1: Can this file be removed? I think we agreed that all contracts will be in network.json or network.staging.json (so no dedicated LDA log file, only a dedicated diamond log file)

script/deploy/deployFacetAndAddToLDADiamond.sh
0xDEnYO, line 1: do we need a dedicated script for this? 
I think we could reuse/easily adapt the existing one.

script/deploy/deploySingleContract.sh
```diff
@@ -15,7 +15,7 @@ deploySingleContract() {
   local ENVIRONMENT="$3"
   local VERSION="$4"
   local EXIT_ON_ERROR="$5"
-
+  local IS_LDA_CONTRACT="$6"
```
0xDEnYO: why not just check if ```CONTRACT == LiFiDEXAggregator```?
Would have the same effect as the additional flag but keep it more simple

script/deploy/deployAllLDAContracts.sh
```diff
@@ -0,0 +1,407 @@
...
+    echo "[info] <<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<< STAGE 4 completed"
+  fi
+
+  # Stage 5: Deploy non-core LDA facets and add to diamond
+  if [[ $START_STAGE -le 5 ]]; then
+    echo ""
+    echo "[info] >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>> STAGE 5: Deploy non-core LDA facets and add to diamond"
+
+    # Get all LDA facet contract names from the LDA Facets directory
+    local LDA_FACETS_PATH="src/Periphery/LDA/Facets/"
```
0xDEnYO: are there any core facets that are in this directory? 
I think all the core facets are in src/Periphery/Facets/

script/deploy/deployAllLDAContracts.sh
0xDEnYO, line 1: we dont need all the stage handling etc. here. 
Also no deployment of create3factory
Please try to keep it simple.

The goal is to have a single script that runs and gets the job done:

- deploy the LDA diamond (using existing DiamondCutFacet)
- add existing core facets
- deploy LDA facets 
- add LDA facets
- make sure LDA setup is correct and complete
- transfer ownership to multisig, not to the timelock (!!!!) 
- create LDA diamond log

deployments/arbitrum.lda.staging.json
mirooon: updated

script/deploy/deployAllLDAContracts.sh
mirooon: No, all LDA's facets are in src/Periphery/LDA/Facets/ and core facets are in src/Facets.
what is wrong with it?

script/deploy/deployAllLDAContracts.sh
```diff
@@ -0,0 +1,279 @@
...
+  
+  # Get all LDA facet contract names from the LDA Facets directory
+  local LDA_FACETS_PATH="src/Periphery/LDA/Facets/"
+  echo "[info] Getting LDA facets from directory: $LDA_FACETS_PATH"
+  
+  # Deploy all non-core LDA facets and add to diamond
+  for FACET_NAME in $(getContractNamesInFolder "$LDA_FACETS_PATH"); do
+    echo "[info] Processing LDA facet: $FACET_NAME"
+    
+    echo "[info] Deploying and adding LDA facet: $FACET_NAME"
```
0xDEnYO, line 179: could remove one of these logs, they are very similar

script/deploy/deploySingleContract.sh
```diff
@@ -2,20 +2,20 @@
 
 # deploys a single contract
 # should be called like this:
-# $(deploySingleContract "Executor" "BSC" "staging" "1.0.0" true)
+# $(deploySingleContract "Executor" "BSC" "staging" "1.0.0" true false)
```
0xDEnYO: the last "false" needs to be removed

script/deploy/zksync/utils/UpdateScriptBase.sol
```diff
@@ -1,213 +1,16 @@
...
 
-import { ScriptBase } from "./ScriptBase.sol";
 import { stdJson } from "forge-std/StdJson.sol";
-import { DiamondCutFacet } from "lifi/Facets/DiamondCutFacet.sol";
-import { DiamondLoupeFacet } from "lifi/Facets/DiamondLoupeFacet.sol";
-import { LibDiamond } from "lifi/Libraries/LibDiamond.sol";
+import { BaseUpdateScript } from "./BaseUpdateScript.sol";
 
-contract UpdateScriptBase is ScriptBase {
+contract UpdateScriptBase is BaseUpdateScript {
```
0xDEnYO: We have UpdateScriptBase and BaseUpdateScript
I find this naming pretty confusing tbh

So far (almost) all our Base contracts end with Base (TestBase, ScriptBase, etc.).
Maybe you can find a naming that makes the relationship clearer?

script/deploy/facets/utils/BaseUpdateScript.sol
```diff
@@ -0,0 +1,360 @@
...
+    }
+
+    function update(
+        string memory name
+    )
+        internal
+        virtual
+        returns (address[] memory facets, bytes memory cutData)
+    {
+        address facet = json.readAddress(string.concat(".", name));
```
0xDEnYO: should we use getConfigContractAddress to make sure the contract is deployed?

script/deploy/ldaHealthCheck.ts
```diff
@@ -0,0 +1,411 @@
...
+
+import { defineCommand, runMain } from 'citty'
+import { consola } from 'consola'
+
+import type { SupportedChain } from '../common/types.js'
+
+const errors: string[] = []
+
+// Helper function to get RPC URL from networks.json
+const getRpcUrl = (
```
0xDEnYO: candidate for central utility/helper function

script/deploy/ldaHealthCheck.ts
```diff
@@ -0,0 +1,411 @@
...
+    // Validate environment
+    if (environment !== 'staging' && environment !== 'production') {
+      consola.error(
+        `Invalid environment: ${environment}. Must be 'staging' or 'production'.`
+      )
+      process.exit(1)
+    }
+
+    // Load main deployment file (contains LiFiDEXAggregatorDiamond address and shared infrastructure)
+    // For staging, try environment-specific file first, then fallback to main
```
0xDEnYO: Not sure if I get the idea behind this logic. 
So if staging is not deployed (and the healthcheck runs for staging environment), you want to check if production is deployed and then pass the healthcheck?

Imho that wouldnt make much sense as you want to verify the deployment of this particular environment, not just any of the two. 
Could you please comment and let me know your reasoning?

script/deploy/ldaHealthCheck.ts
```diff
@@ -0,0 +1,411 @@
...
+      if (!expectedMultisigAddress) {
+        if (environment === 'production') {
+          logError(
+            `No multisig address (safeAddress) found in networks.json for network ${network}`
+          )
+        } else {
+          consola.warn(
+            `No multisig address (safeAddress) found in networks.json for network ${network}`
+          )
+          consola.info('For staging environment, this is acceptable')
```
0xDEnYO: for staging we can skip the whole step (can save the owner's call to the diamond and some of the code / branches below)

script/tasks/ldaDiamondUpdateFacet.sh
0xDEnYO, line 1: Do we have to have two "diamondUpdateFacet.sh" scripts (one for diamond, one for LDA)? Dont they behave entirely similar, only that their target is different (which could be decided via a parameter)?

If we can, I would prefer to merge these to reduce duplicate code

script/helperFunctions.sh
```diff
@@ -1130,6 +1130,66 @@ function saveDiamondPeriphery() {
     printf %s "$result" >"$DIAMOND_FILE"
   done
 }
+
+# LDA-specific diamond save functions
+function saveLDADiamondFacets() {
```
0xDEnYO: same here. Do we need a dedicated one or could we combine the logic in saveDiamondFacets by just adding a parameter?

just trying to keep our codebase as small as possible

script/helperFunctions.sh
```diff
@@ -4381,6 +4457,7 @@ function updateDiamondLogs() {
   # read function arguments into variable
   local ENVIRONMENT=$1
   local NETWORK=$2
+  local DIAMOND_TYPE=${3:-"LiFiDiamond"}  # Optional parameter, defaults to regular diamond
```
0xDEnYO: What if we update both diamond logs (regular and LDA) when we call this function for a particular network?
Those two actions could also be executed in parallel.

script/deploy/zksync/utils/UpdateScriptBase.sol
mirooon: good point, I've merged BaseUpdateScript to UpdateScriptBase and removed BaseUpdateScript completely.

fixed in: https://github.com/lifinance/contracts/pull/1360/commits/c3f6b6e6b605e2e78521c71e54b42460ddfadf3d

feel free to resolve

script/deploy/ldaHealthCheck.ts
mirooon: I agree, that logic was flawed. I’ve updated it so that now each environment handles its own files independently, without mixing between them

script/tasks/ldaDiamondUpdateFacet.sh
mirooon: good point!

script/helperFunctions.sh
mirooon: good point

script/helperFunctions.sh
mirooon: I don't understand you point. We will we call it in parallel? Even if they run in parallel they touch different files right?

0xDEnYO: My point was that when I say "update diamond logs for network mainnet" then it should update both the diamond log and the LDA log file.
And these two file updates can be done in parallel.

Another PR incoming to parallelize the whole diamond log update function itself (https://github.com/lifinance/contracts/pull/1367)

mirooon: Addressed this issue in new PR: https://github.com/lifinance/contracts/pull/1370

PR #1370, opened 2025-09-09
LF-13997 LDA 2.0 Scripts update diamond logs parallelized

mirooon:
[LF-13997]

[LF-13997]: https://lifi.atlassian.net/browse/LF-13997?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/script/helperFunctions.sh
+++ b/script/helperFunctions.sh
@@ -896,7 +896,8 @@ function saveDiamondFacets() {
   if [[ -n "$FACETS_DIR" ]]; then
     mkdir -p "$FACETS_DIR"
   else
-    # fallback for safety, though the caller should always provide it now
+    # fallback for safety, for single-threaded operation where FACETS_DIR is not provided
+    # for parallel operation, the caller should always provide it
     FACETS_DIR=$(mktemp -d)
     trap 'rm -rf "$FACETS_DIR" 2>/dev/null' EXIT
   fi
@@ -4392,7 +4393,6 @@ function updateDiamondLogForNetwork() {
     echo "[$NETWORK] Trying to get facets for diamond $DIAMOND_ADDRESS now - attempt $((attempts + 1))"
     # try to execute call
     local KNOWN_FACET_ADDRESSES=$(cast call "$DIAMOND_ADDRESS" "facetAddresses() returns (address[])" --rpc-url "$RPC_URL") 2>/dev/null
-    echoDebug "KNOWN_FACET_ADDRESSES=$KNOWN_FACET_ADDRESSES"
     # check the return code the last call
     if [ $? -eq 0 ]; then
       break # exit the loop if the operation was successful
@@ -4499,7 +4499,7 @@ function updateDiamondLogs() {
   local NETWORK=$2
 
   # Define all diamond contract names to process
-  local DIAMOND_CONTRACT_NAMES=("LiFiDiamond" "LiFiDiamondImmutable" "LiFiDEXAggregatorDiamond")
+  local DIAMOND_CONTRACT_NAMES=("LiFiDiamond" "LiFiDEXAggregatorDiamond")
 
   # if no network was passed to this function, update all networks
   if [[ -z $NETWORK ]]; then
@@ -4534,7 +4534,6 @@ function updateDiamondLogs() {
     for ENVIRONMENT in "${ENVIRONMENTS[@]}"; do
       echo " -----------------------"
       echo " current ENVIRONMENT: $ENVIRONMENT"
-      echo " current ENVIRONMENT: $ENVIRONMENT"
 
       # Process all diamond contract names for this network/environment in parallel
       for DIAMOND_CONTRACT_NAME in "${DIAMOND_CONTRACT_NAMES[@]}"; do
@@ -4552,7 +4551,6 @@ function updateDiamondLogs() {
   done
 
   # Wait for all background jobs to complete and capture exit codes
-  echo "Waiting for all diamond log updates to complete..."
   echo "Waiting for all diamond log updates to complete..."
   local failed_jobs=()
   local job_count=${#pids[@]}


script/helperFunctions.sh
```diff
@@ -914,41 +885,21 @@ function saveDiamondFacets() {
...
-  local TEMP_DIR
-  TEMP_DIR=$(mktemp -d)
-  trap 'rm -rf "$TEMP_DIR" 2>/dev/null' EXIT
-  local FACETS_DIR="$TEMP_DIR/facets"
-  mkdir -p "$FACETS_DIR"
+  # ensure the directory for intermediate files exists
+  if [[ -n "$FACETS_DIR" ]]; then
+    mkdir -p "$FACETS_DIR"
+  else
+    # fallback for safety, though the caller should always provide it now
```
0xDEnYO: I dont fully understand your rationale behind this intermediate files directory. 
Your comments say the caller should always provide it but in diamondFacetUpdate you call with ""
image

Why is it not safe to just always use the same (i.e. some) tmp directory?

script/helperFunctions.sh
```diff
@@ -4429,7 +4392,7 @@ function updateDiamondLogForNetwork() {
     echo "[$NETWORK] Trying to get facets for diamond $DIAMOND_ADDRESS now - attempt $((attempts + 1))"
     # try to execute call
     local KNOWN_FACET_ADDRESSES=$(cast call "$DIAMOND_ADDRESS" "facetAddresses() returns (address[])" --rpc-url "$RPC_URL") 2>/dev/null
-
+    echoDebug "KNOWN_FACET_ADDRESSES=$KNOWN_FACET_ADDRESSES"
```
0xDEnYO: this helper function is going to produce a lot more output when we update all networks at once (also with the   ```echoDebug "in function updateDiamondLogForNetwork"``` above). 
I felt the existing output was sufficient. 
What was your rationale to add more to it?
Also, the logs are missing the [network] prefix which works fine when you update a single network but will get super messy when updating all of them

script/helperFunctions.sh
```diff
@@ -4482,24 +4477,29 @@ function updateDiamondLogForNetwork() {
...
 }
 
 function updateDiamondLogs() {
   # read function arguments into variable
   local ENVIRONMENT=$1
   local NETWORK=$2
-  local DIAMOND_TYPE=${3:-"LiFiDiamond"}  # Default to LiFiDiamond, can be "LiFiDEXAggregatorDiamond" for LDA
+
+  # Define all diamond contract names to process
+  local DIAMOND_CONTRACT_NAMES=("LiFiDiamond" "LiFiDiamondImmutable" "LiFiDEXAggregatorDiamond")
```
0xDEnYO: LiFiDiamondImmutable is outdated and needs to be removed from the codebase. This will/shall be done in a separate (low prio) cleanup PR when we have time for it. However, we should not actively include it in new code.

script/helperFunctions.sh
```diff
@@ -4525,51 +4526,64 @@ function updateDiamondLogs() {
...
-  # loop through all networks
+  # Loop through all networks and environments
   for NETWORK in "${NETWORKS[@]}"; do
     echo ""
     echo "current Network: $NETWORK"
 
     for ENVIRONMENT in "${ENVIRONMENTS[@]}"; do
       echo " -----------------------"
       echo " current ENVIRONMENT: $ENVIRONMENT"
+      echo " current ENVIRONMENT: $ENVIRONMENT"
```
0xDEnYO: line duplicate

script/helperFunctions.sh
```diff
@@ -4525,51 +4526,64 @@ function updateDiamondLogs() {
...
+        job_info+=("$NETWORK:$ENVIRONMENT:$DIAMOND_CONTRACT_NAME")
+        job_index=$((job_index + 1))
+      done
     done
   done
 
   # Wait for all background jobs to complete and capture exit codes
-  echo "Waiting for all $DIAMOND_TYPE log updates to complete..."
+  echo "Waiting for all diamond log updates to complete..."
+  echo "Waiting for all diamond log updates to complete..."
```
0xDEnYO: line duplicate

script/helperFunctions.sh
mirooon: agree, line removed

script/helperFunctions.sh
mirooon: regarding intermediate directory it gives a dedicated space for each job that is conflict free from the others. That facets and peripheries have dedicated folders and before them there is a unique folder with which name is constructed with diamond contract name.

regarding comment that's true the caller should always provide this but in case it's used in multi-threaded environment. The screenshot that you posted work in diamondUpdateFacet.sh on single thread so it's safe. Updated comment for more clarity.

Why is it not safe to just always use the same (i.e. some) tmp directory?
What do you mean exactly? What do you mean exactly? In our current setup whenever the main function creates jobs, it also creates a dedicated TEMP_DIR for them. This directory is removed once all jobs are finished. When jobs run concurrently each has its own isolated temporary space. so at the and the jobs they don't touch their files at all

fixed comment in: https://github.com/lifinance/contracts/pull/1370/commits/a775644ea403c66e1b1f2e40dffabf05bf171dce

0xDEnYO: my question was just "why does this need to be a function parameter"?
It seems like an internal logic thing, nothing that needs to be controlled by the caller of the function. 
You could derive it from the time or whatever to make it unique, I just dont see why it needs to be a parameter

mirooon: imo TEMP_DIR has to be created in the caller as there we are deleting it after all jobs are finished. To create subfolders for jobs we need TEMP_DIR. Of course we could put the logic that put data inside of TEMP_DIR without having a parameter but if TEMP_DIR path change then we would need also to remember that child has to change path as well. If you have some ready to go solution in mind please let me know

of course we could resign from global temp dir as it is right now but centralized clean up is better

0xDEnYO: got it, thanks for the explanation.

PR #1376, opened 2025-09-11
LF-13736 - WhitelistManagerFacet, LibAllowList added granular mapping

mirooon:
[LF-13736]

This PR addresses a audit finding https://github.com/burrasec/LiFi_2025.08.08_WhitelistManagerFacet/issues/1 and discussion https://lifi-protocol.slack.com/archives/C094C46JSDP/p1755576664131549 by transitioning from separate contract and selector whitelists to a more secure, granular model that approves specific (contract, selector) pairs. To maintain full backward compatibility with our many existing deployed contracts, I've implemented a dual write system within the LibAllowList library. This library uses a mechanisms to keep the new granular mapping and the old global arrays synchronized ensuring older facets continue to function without interruption as we migrate.

[LF-13736]: https://lifi.atlassian.net/browse/LF-13736?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,5 +1,6 @@
 #!/usr/bin/env sh
 . "$(dirname -- "$0")/_/husky.sh"
+
 echo ""
 echo "Running 'forge build' and 'typechain generation'..."
 
@@ -7,13 +8,40 @@ echo "Running 'forge build' and 'typechain generation'..."
 echo "Building contracts with forge..."
 forge build --skip test
 
-# If the build fails, abort the commit
 if [ $? -ne 0 ]; then
-  echo "\nForge build failed. Aborting commit.\n"
+  printf '\n%s\n\n' "Forge build failed. Aborting commit."
+  exit 1
+fi
+
+echo "Forge build completed successfully!"
+echo ""
+
+# Run typechain generation (now that artifacts are available)
+echo "Generating TypeChain types..."
+bun typechain:incremental
+
+if [ $? -ne 0 ]; then
+  printf '\n%s\n\n' "TypeChain generation failed. Aborting commit."
+  exit 1
+fi
+
+echo "TypeChain generation completed successfully!"
+echo ""
+
+# Run TypeScript compilation to catch any type errors from outdated typings
+echo "Running TypeScript compilation to check for type errors..."
+bunx tsc-files --noEmit
+
+if [ $? -ne 0 ]; then
+  printf '\n%s\n' "TypeScript compilation failed. This may indicate outdated TypeChain types."
+  printf '%s\n\n' "Please run 'bun typechain' manually and fix any type errors before committing."
   exit 1
 fi
 
 echo ""
+echo "TypeScript compilation passed - all types are up to date!"
+echo ""
+
 echo ""
 echo "Running 'bun lint-staged' now:"
 bun lint-staged
@@ -60,16 +88,16 @@ else
 fi
 
 printRed() {
-  local MESSAGE=$1
+  local MESSAGE="$1"
   printf '\033[31m%s\033[0m\n' "$MESSAGE"
 }
 printYellow() {
-  local MESSAGE=$1
+  local MESSAGE="$1"
   printf '\033[33m%s\033[0m\n' "$MESSAGE"
 }
 
 printAdvise() {
-  local ACTION=$1
+  local ACTION="$1"
 
   echo ""
   printf '\033[91m%s\033[0m\n' "NEXT STEPS"
@@ -84,8 +112,8 @@ printAdvise() {
 
 # checks if the given line number or the line above contain a comment to disable the check
 isKnownFalsePositiveMatch() {
-  local FILE=$1
-  local LINE_NUMBER=$2
+  local FILE="$1"
+  local LINE_NUMBER="$2"
 
   if [ "$LINE_NUMBER" -gt 1 ]; then
     local PREV_LINE=$(sed "$((LINE_NUMBER - 1))q;d" "$FILE")
@@ -112,7 +140,7 @@ isKnownFalsePositiveMatch() {
 
 # checks if a file contains any of the secrets in .env
 doesFileContainDotEnvSecret() {
-  local FILE=$1
+  local FILE="$1"
   # iterate through all secrets
   for SECRET in $ENV_SECRETS; do
     # extract key and value
@@ -152,8 +180,7 @@ doesFileContainDotEnvSecret() {
 
 # checks if a file contains a potential private key (=> a 64 hex string)
 doesFileContainPotentialPrivateKey() {
-  local FILE=$1
-  local KEY=$2
+  local FILE="$1"
 
   # check file content for matches with private key regEx
   local MATCHES=$(grep -E -nH "$ETH_PRIVATE_KEY_PATTERN" "$FILE")
@@ -201,7 +228,7 @@ processGitDiff() {
     fi
 
     # Check for potential private keys
-    RESULT_PRIVKEY=$(doesFileContainPotentialPrivateKey "$FILE" "$KEY")
+    RESULT_PRIVKEY=$(doesFileContainPotentialPrivateKey "$FILE")
     if [[ -n $RESULT_PRIVKEY ]]; then
       printYellow "$RESULT_PRIVKEY"
       echo ""


src/Facets/WhitelistManagerFacet.sol
```diff
@@ -15,69 +15,47 @@ contract WhitelistManagerFacet is IWhitelistManagerFacet {
     /// External Methods ///
```
mirooon, line 15: WhitelistManagerFacet things around like it's test and interface are still in progress

test/solidity/Facets/PioneerFacet.t.sol
```diff
@@ -3,22 +3,14 @@ pragma solidity ^0.8.17;
...
-import { TestBaseFacet } from "../utils/TestBaseFacet.sol";
-import { LibAllowList } from "lifi/Libraries/LibAllowList.sol";
 import { PioneerFacet } from "lifi/Facets/PioneerFacet.sol";
 import { InvalidCallData } from "lifi/Errors/GenericErrors.sol";
+import { TestBaseFacet } from "../utils/TestBaseFacet.sol";
+import { TestWhitelistManagerBase } from "../utils/TestWhitelistManagerBase.sol";
 
 // Stub PioneerFacet Contract
-contract TestPioneerFacet is PioneerFacet {
+contract TestPioneerFacet is PioneerFacet, TestWhitelistManagerBase {
```
mirooon, line 12: I realized that instead of creating separate Test<Contract>s, the WhitelistManagerFacet can be included inside TestBase. At that level, the addAllowedContractSelector function will already be available for use in the tests

src/Interfaces/IWhitelistManagerFacet.sol
```diff
@@ -6,58 +6,52 @@ pragma solidity ^0.8.17;
 /// @notice Interface for WhitelistManagerFacet facet for managing approved contracts and function selectors.
```
mirooon, line 6: WhitelistManagerFacet things around like it's test and interface are still in progress

.gitignore
```diff
@@ -67,3 +67,4 @@ test/solidity/TestPlayground.t.sol
 
 AGENTS.md
 CLAUDE.md
+config/whitelist.staging.json
```
0xDEnYO: nitpick: think this would fit better into category "autogenerated files"

deployments/_deployments_log_file.json
```diff
@@ -44096,12 +44321,12 @@
       }
     },
     "base": {
-      "production": {
+      "staging": {
```
0xDEnYO: a production address is being removed from the log. 
This might be the same issue that I described here: 
https://lifi-protocol.slack.com/archives/C088UJW2DPZ/p1761708856389909

I guess this was not intended?

script/deploy/facets/utils/UpdateScriptBase.sol
```diff
@@ -259,28 +267,48 @@ contract UpdateScriptBase is ScriptBase {
...
+            executor = json.readAddress(".devWallet");
+        }
+
+        // get function selectors that should be approved for executor
         bytes memory rawConfig = json.parseRaw(
             ".approvedSelectorsForDeployerWallet"
         );
 
+        emit log("rawConfig: ");
+        emit log_bytes(rawConfig);
```
0xDEnYO: Do we need to print the rawConfig?
We are printing the amount of function selectors and every single selector (line 306)....I dont think we need to print the raw data, too as it just clogs up the output.

script/helperFunctions.sh
```diff
@@ -1932,27 +1932,36 @@ function getBytecodeFromArtifact() {
...
   local ENVIRONMENT="$2"
 
-  local FILEPATH_WHITELISTED_ADDRESSES="config/whitelistedAddresses.json"
-  local FILEPATH_GLOBAL_CONFIG="config/global.json"
+  # Determine the correct whitelist file based on environment
+  local WHITELIST_FILE
+  if [[ "$ENVIRONMENT" == "production" ]]; then
+    WHITELIST_FILE="config/whitelist.json"
+  else
+    WHITELIST_FILE="config/whitelist.staging.json"
```
0xDEnYO, line 1945: I am not sure if my comment is valid. 
Just keep in mind that we cannot assume that whitelist.staging.json is present since we do not push it to github. It depends on from where this helper function is called.

src/Facets/WhitelistManagerFacet.sol
```diff
@@ -15,67 +15,44 @@ contract WhitelistManagerFacet is IWhitelistManagerFacet {
...
         if (msg.sender != LibDiamond.contractOwner()) {
             LibAccess.enforceAccessControl();
         }
-        uint256 length = _selectors.length;
-        for (uint256 i = 0; i < length; ++i) {
-            _setFunctionWhitelist(_selectors[i], _whitelisted);
+        if (_contracts.length != _selectors.length) {
+            revert InvalidConfig();
         }
+        for (uint256 i = 0; i < _contracts.length; ++i) {
```
0xDEnYO: nitpick: use gas saving pattern to increase loop counter in unchecked block

src/Interfaces/IWhitelistManagerFacet.sol
```diff
@@ -4,61 +4,105 @@ pragma solidity ^0.8.17;
...
 
-    /// @notice Adds or removes a specific function selector to/from the whitelist.
+    /// @notice Sets the whitelist status for a specific contract and selector pair.
+    /// @param _contract The contract address to whitelist or unwhitelist.
     /// @param _selector The function selector to whitelist or unwhitelist.
-    /// @param _whitelisted Whether the function selector should be whitelisted.
-    function setFunctionWhitelistBySelector(
+    /// @param _whitelisted Whether the contract and selector pair should be whitelisted.
+    /// @dev MARKER SELECTOR: To maintain backward compatibility and whitelist a contract
+    ///      without knowing its exact function selectors, use 0xffffffff as a marker selector.
```
0xDEnYO: is this really correct?

Is the 0xffffffff the selector that we use to whitelist contracts as approveTo (only) ?
If so, then we should describe that as such. Whitelisting a contract without knowing its function selectors sounds like we would be whitelisting all its function selectors (or at least it's unclear what is the state of this really)

if you update, also update function comment below

src/Libraries/LibAllowList.sol
```diff
@@ -6,147 +6,308 @@ import { LibAsset } from "./LibAsset.sol";
...
-/// @notice Library for managing and accessing the contract address allow list
+/// @notice Manages a dual-model allow list to support a secure, granular permissions system
+/// while maintaining backward compatibility with a non-granular, global system.
+/// @dev This library is the single source of truth for all whitelist state changes.
+/// It ensures that both the new granular mapping and the global arrays used by older
+/// contracts are kept perfectly synchronized. The long-term goal is to migrate all usage
+/// to the new granular system. New development should exclusively use the "Primary Interface" functions.
+///
+/// Special Marker Selector:
+/// - Contracts with no specific function selectors use 0xffffffff as a marker selector.
```
0xDEnYO: also here, please lets be more clear. 
contracts that are whitelisted as approveTo only (but no functions can be called on these contracts) 

(sth like this)

src/Libraries/LibAllowList.sol
```diff
@@ -6,147 +6,308 @@ import { LibAsset } from "./LibAsset.sol";
...
+        mapping(address => mapping(bytes4 => uint256)) selectorIndices;
+        /// @dev Flag to indicate completion of a one-time data migration.
         bool migrated;
     }
 
-    /// @dev Adds a contract address to the allow list
-    /// @param _contract the contract address to add
-    function addAllowedContract(address _contract) internal {
-        // validate contract address is not zero address
+    // All new contracts and facets should exclusively use these functions. They operate
```
0xDEnYO: I think this comment is a bit out of place here. 
The contracts and facets will only use function `contractSelectorIsAllowed` , so I would expect this comment rather above that function.

There is only one (admin-only) path/contract that manipulates (adds to & removes from) the whitelist.

script/deploy/facets/UpdateWhitelistManagerFacet.s.sol
```diff
@@ -44,80 +47,994 @@ contract DeployScript is UpdateScriptBase {
...
+            uint256 contractIndex = 0;
+            contractIndex = _populatePeripheryContracts(
+                whitelistJson,
+                contracts,
+                selectors,
+                contractIndex
+            );
+        } catch {
+            // Network not found, throw error
+            revert(
```
0xDEnYO, line 330: so if for one network we do not have a periphery section then the whole update script fails?
I guess we will have one for each network anyway (except `localanvil`) but I wonder if reverting is the correct strategy here. And did you make sure that we catch the localanvil case?

Actually we could also remove the localanvil I guess...

Just checking to make sure we dont have a non-executable script but I guess you tested it already and it works anyway?

script/tasks/diamondSyncWhitelist.sh
```diff
@@ -0,0 +1,739 @@
...
+  echo ""
+  echo "[info] >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>> running script syncWhitelist now...."
+
+  # Load environment variables
+  source .env
+
+  # Load configuration & helper functions
+  source script/helperFunctions.sh
+
+  # Configuration flag - set to true to allow token contracts in DEX lists
```
0xDEnYO: pls add a warning that this is a serious security risk and is strongly discouraged.
Enabling this flag should only be considered after discussion and agreement with the whole SC team.

script/tasks/diamondSyncWhitelist.sh
```diff
@@ -0,0 +1,739 @@
...
+          echo "[$NETWORK] Please check whitelist.json or whitelist.staging.json and remove address zero"
+          echo "[$NETWORK] Discuss with backend team if this address is needed"
+          echo ""
+        } >> "$FAILED_LOG_FILE"
+        return 1
+      fi
+      
+      # Check if address has code
+      CHECKSUMMED=$(cast --to-checksum-address "$ADDRESS")
+      CODE=$(cast code "$CHECKSUMMED" --rpc-url "$RPC_URL")
```
0xDEnYO, line 409: pls make sure that we run this check only on addresses that will be newly added and not on every address (that is already whitelisted). 

Also, we can consider parallelizing these calls for faster script runtime.

But maybe that is an improvement for later, we should focus now on getting this topic out of the door.

script/tasks/diamondSyncWhitelist.sh
```diff
@@ -0,0 +1,739 @@
...
+      fi
+      
+      # Parse selectors (comma-separated)
+      if [[ -n "$SELECTORS_STR" && "$SELECTORS_STR" != "" ]]; then
+        # Split selectors by comma (use tr for portability)
+        SELECTORS=($(echo "$SELECTORS_STR" | tr ',' ' '))
+        
+        for SELECTOR in "${SELECTORS[@]}"; do
+          if [[ -n "$SELECTOR" && "$SELECTOR" != "" ]]; then
+            PAIR_KEY="$(echo "$ADDRESS" | tr '[:upper:]' '[:lower:]')|$SELECTOR"
```
0xDEnYO: why do you lowercase the address here? 

(same below, row 421)

script/deploy/healthCheck.ts
```diff
@@ -321,350 +372,360 @@ const main = defineCommand({
...
-
-          // Get Safe info directly from the contract
-          const safeInfo = await getSafeInfoFromContract(
-            publicClient,
-            safeAddress
-          )
+        //          ╭─────────────────────────────────────────────────────────╮
+        //          │                   SAFE Configuration                    │
+        //          ╰─────────────────────────────────────────────────────────╯
+        consola.box('Checking SAFE configuration...')
```
0xDEnYO: safes are only used on PROD, not on staging. since you added handling for both environments, dont you need to add it here, too?

script/deploy/facets/utils/UpdateScriptBase.sol
mirooon: removed

script/deploy/facets/UpdateWhitelistManagerFacet.s.sol
mirooon: Do we actually use localanvil? I remember trying to deploy the full diamond there but it didn’t work for me.

I assume we should ALWAYS have PERIPHERIES for every chain, while having DEXES is not necessarily required. for example on completely new chains where no DEXES exist yet

script/deploy/healthCheck.ts
mirooon: updated

script/tasks/diamondSyncWhitelist.sh
mirooon: good point added

script/tasks/diamondSyncWhitelist.sh
mirooon: good point, created ticket for it: https://lifi.atlassian.net/browse/LF-16406

script/tasks/diamondSyncWhitelist.sh
mirooon: I usually normalize addresses to lowercase if it comes from different sources when I do comparisons. whitelist.json can have not lowercased addresses, I am not sure what exactly rpc call returns but imo it's safe to lowercase for comparison. At the end for evm casting is irrelevant so it's safe

deployments/_deployments_log_file.json
mirooon: good catch. fixes. I have no idea why did it happen

script/helperFunctions.sh
mirooon: yeah, I don’t see a case where we would run `addPeripheryToWhitelistJson`, for example, from a gh action especially in the staging environment. I think we’re good here.

src/Facets/WhitelistManagerFacet.sol
mirooon: good point, updated

src/Interfaces/IWhitelistManagerFacet.sol
mirooon: good point. Updated to:

```
    /// @dev MARKER SELECTOR: For backward compatibility with legacy address only checks,
    ///      use 0xffffffff as an "empty/marker" selector when whitelisting a contract.
    ///      This makes isAddressWhitelisted(_contract) return true but does not allow any
    ///      granular calls; in the contract-selector system, actual selectors must still
    ///      be explicitly whitelisted for real usage.
```

src/Libraries/LibAllowList.sol
mirooon: good point, fixed it

mirooon: btw. I don't want to mention approveTo as it's complitely seperated contract

feel free to resolve

src/Libraries/LibAllowList.sol
mirooon: removed this comment

script/tasks/diamondSyncWhitelist.sh
0xDEnYO: yes for comparing I understand and agree
It just looks like a bit of back and forth here. 

e.g. 
- in 364 you iterate through required pairs
- in 371 you lowercase (and cut the 0x which arguably you dont have to, can also compare to 0x000... so that you can reuse that lowercased address later)
- in 387 you checksum the address to use it for cast call
- in 401 (aka 421 for else branch) you lowercase again and subsequently write to NEW_PAIRS
- in 468 you iterate through NEW_PAIRS and checksum the address again in 474

script/deploy/facets/UpdateWhitelistManagerFacet.s.sol
0xDEnYO: what if we deploy to a new network?

We will deploy the core facets and then add them to the diamond (i.e. calling this script here). 
In that case there is no periphery yet for that new chain as we will only deploy it in the steps after. 

I think we can remove localanvil but should keep that separate.

src/Interfaces/IWhitelistManagerFacet.sol
0xDEnYO: I am sorry to be fussy about this but I really struggle with the "empty/marker" wording as it doesnt really say what this is used for.
There is only one very specific use case for this function selector:  whitelisting a contract to be used as `approveTo` (only) in LibSwap.SwapData but not allowing any function calls to this contract.

Can we call it the "ApproveTo-Only Selector"?


e.g (for the explanation).
Some DEXs have specific contracts that need to be approved to while another (router) contract has to be called to initiate the swap. We use the 0xffffffff function selector for whitelisting such `approveTo`-only contracts while not  allowing any function calls to it.

script/deploy/facets/UpdateWhitelistManagerFacet.s.sol
mirooon: when we deploy on new network then the `UpdateWhitelistManagerFacet.s.sol` script is not called. Only `DeployWhitelistMangerFacet.s.sol `script is called, then we deploy periphery contracts (including `WhitelistManagerFacet`), then we update `whitelist.json` and then we call whitelisting script.

so for new chain we dont call `UpdateWhitelistManagerFacet.s.sol `
if `WhitelistManagerFacet` is deployed then we call `UpdateWhitelistManagerFacet.s.sol` so then we require PERIPHERIES to be present for selected chain

src/Interfaces/IWhitelistManagerFacet.sol
mirooon: ok, I've changed it: https://github.com/lifinance/contracts/pull/1376/commits/a873a30afb71928cd92360f469772757a2fcd4fb

script/tasks/diamondSyncWhitelist.sh
mirooon: got it! fixed in: https://github.com/lifinance/contracts/pull/1376/commits/fdc93ad8387401be6457cd2e98e3cefe425d3679

0xDEnYO (approved):

PR #1395, opened 2025-09-24
create FeeForwarder contract (draft)

0xDEnYO:
https://lifi.atlassian.net/browse/LF-15657


script/deploy/facets/DeployFeeForwarder.s.sol
```diff
@@ -0,0 +1,35 @@
+// SPDX-License-Identifier: LGPL-3.0-only
```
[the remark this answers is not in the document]
0xDEnYO: not correct, our conventions.md file clearly states that all .sol files should use LGPL. Please save this as a learning and also make sure that PRs meet all conventions described in that file.

script/deploy/zksync/DeployFeeForwarder.zksync.s.sol
```diff
@@ -0,0 +1,35 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.17;
+
+import { DeployScriptBase } from "./utils/DeployScriptBase.sol";
```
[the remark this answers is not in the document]
0xDEnYO: all other zksync deploy scripts use the same path, it seems correct to me

src/Periphery/FeeForwarder.sol
```diff
@@ -0,0 +1,109 @@
+// SPDX-License-Identifier: LGPL-3.0-only
+pragma solidity ^0.8.17;
```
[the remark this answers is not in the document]
0xDEnYO: in which style guide is this defined? Most our contracts do not have a blank line there?

src/Periphery/FeeForwarder.sol
```diff
@@ -0,0 +1,109 @@
...
+                distribution.amount
+            );
+
+            unchecked {
+                ++i;
+            }
+        }
+
+        emit FeesForwarded(token, distributions);
+    }
```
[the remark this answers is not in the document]
0xDEnYO: correct, updated

src/Periphery/FeeForwarder.sol
```diff
@@ -0,0 +1,109 @@
...
+        if (remainingNativeBalance != 0) {
+            LibAsset.transferAsset(
+                address(0),
+                payable(msg.sender),
+                remainingNativeBalance
+            );
+        }
+
+        emit FeesForwarded(address(0), distributions);
+    }
```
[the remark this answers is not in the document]
0xDEnYO: this is intentional. A comment has been added to explain the decision.

src/Periphery/FeeForwarder.sol
```diff
@@ -0,0 +1,109 @@
+// SPDX-License-Identifier: LGPL-3.0-only
+pragma solidity ^0.8.17;
+
+import { WithdrawablePeriphery } from "../Helpers/WithdrawablePeriphery.sol";
+import { LibAsset } from "../Libraries/LibAsset.sol";
+import { InvalidCallData } from "../Errors/GenericErrors.sol";
+
```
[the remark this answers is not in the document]
0xDEnYO: We dont need InvalidReceiver. I dont understand why you think we need it

script/deploy/tron/deploy-and-register-periphery.ts
```diff
@@ -591,7 +593,133 @@ async function deployAndRegisterPeripheryImpl(options: {
...
+      consola.error(` Failed to deploy FeeForwarder:`, error.message)
+      deploymentResults.push({
+        contract: 'FeeForwarder',
+        address: 'FAILED',
+        txId: 'FAILED',
+        cost: 0,
+        version: '0.0.0',
+      })
+    }
+
```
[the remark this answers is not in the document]
0xDEnYO: good catch...updated

test/solidity/Periphery/FeeForwarder.t.sol
```diff
@@ -0,0 +1,370 @@
...
+            memory distributions = new FeeForwarder.FeeDistribution[](1);
+        distributions[0] = FeeForwarder.FeeDistribution({
+            recipient: address(0xb0b),
+            amount: 1 ether
+        });
+
+        // Act & Assert - Native token transfers will fail naturally when trying to transfer, saving gas
+        vm.expectRevert();
+        feeForwarder.forwardERC20Fees(address(0), distributions);
+    }
```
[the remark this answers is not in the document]
0xDEnYO: good catch. added

test/solidity/Periphery/FeeForwarder.t.sol
```diff
@@ -0,0 +1,370 @@
...
+            memory distributions = new FeeForwarder.FeeDistribution[](1);
+        distributions[0] = FeeForwarder.FeeDistribution({
+            recipient: address(0xb0b),
+            amount: 1 ether
+        });
+
+        // Act & Assert - Transaction will revert due to insufficient funds when trying to transfer ETH
+        vm.expectRevert();
+        feeForwarder.forwardNativeFees{ value: 0.5 ether }(distributions);
+    }
```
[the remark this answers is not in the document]
0xDEnYO: good catch. Added

src/Periphery/FeeForwarder.sol
```diff
@@ -0,0 +1,121 @@
...
+        // also we do not check sufficient msg.value / native balance to save gas
+        // the tx will revert anyway in this case
+
+        // forward all native fee amounts to the recipients
+        for (uint256 i; i < _distributions.length; ) {
+            FeeDistribution calldata distribution = _distributions[i];
+
+            // we do intentionally not check for amount == 0 to save gas
+
+            LibAsset.transferAsset(
```
mirooon: what about using directly using `transferNativeAsset` function? then we can save gas and skip checking `isNativeAsset(assetId)` in `transferAsset` function

src/Periphery/FeeForwarder.sol
```diff
@@ -0,0 +1,121 @@
...
+                ++i;
+            }
+        }
+
+        // return any remaining native tokens to the caller
+        // since the contract is designed to not hold any funds and does not collect any dust
+        // we can safely return the remaining native balance to the caller
+        uint256 remainingNativeBalance = address(this).balance;
+        if (remainingNativeBalance != 0) {
+            LibAsset.transferAsset(
```
mirooon: same here we could use `transferNativeAsset`

src/Periphery/FeeForwarder.sol
```diff
@@ -0,0 +1,121 @@
...
+                address(0),
+                payable(msg.sender),
```
mirooon, line 106: _"This contract is highly gas-optimized and designed to be called by the LiFiDiamond after fees have been calculated for a swap or bridge execution"_ - as described in this case msg.sender it's our LiFiDiamond but we probably don't want to do any transfer back to our diamond.

test/solidity/Periphery/FeeForwarder.t.sol
```diff
@@ -0,0 +1,370 @@
...
+    FeeForwarder private feeForwarder;
+    ERC20 private feeToken;
+
+    event FeesForwarded(
+        address indexed token,
+        FeeForwarder.FeeDistribution[] distributions
+    );
+
+    function setUp() public {
+        feeForwarder = new FeeForwarder(address(this));
```
mirooon: I suggest using dedicated unique address for the withdraw wallet instead of address(this) to make the tests clearer

test/solidity/Periphery/FeeForwarder.t.sol
```diff
@@ -0,0 +1,370 @@
...
+
+        // Act & Assert
+        vm.expectEmit(true, false, false, true, address(feeForwarder));
+        emit FeesForwarded(address(feeToken), distributions);
+
+        feeForwarder.forwardERC20Fees(address(feeToken), distributions);
+
+        assertEq(feeToken.balanceOf(address(0xb33f)), 1 ether);
+        assertEq(feeToken.balanceOf(address(0xb0b)), 0.5 ether);
+        assertEq(feeToken.balanceOf(address(this)), 1000 ether - total);
```
mirooon: While reading the tests, I wasn’t immediately sure what 1000 ether referred to. It might be clearer to define it as a variable eg `initialBalance`
imo we usually prank USER_SENDER in all our tests. I think it would be good to align with that convention here as well and since FeeForwarded is expected to be called by the diamond it might be better to prank the diamond contract here instead.

test/solidity/Periphery/FeeForwarder.t.sol
```diff
@@ -0,0 +1,370 @@
...
+    }
+
+    function test_ForwardNativeFeesWithEmptyDistributions() public {
+        // Arrange
+        FeeForwarder.FeeDistribution[]
+            memory distributions = new FeeForwarder.FeeDistribution[](0);
+        uint256 balanceBefore = address(this).balance;
+
+        // Act & Assert - Empty arrays should not revert and should not emit events (gas optimization)
+        feeForwarder.forwardNativeFees{ value: 1 ether }(distributions);
```
mirooon, line 262: Not sure if this works as expected, but you could use recorded logs (https://getfoundry.sh/reference/cheatcodes/get-recorded-logs/) to improve the test. for example by checking that the log length is 0.

src/Periphery/FeeForwarder.sol
0xDEnYO: good point. I thought about it. But that function is (still) private. 
In some other PR we changed this (I think in the WhitelistManager stuff) so I thought its better to not have concurrent PRs updating the same Lib....

but I agree that `transferNativeAsset()` should be called. 
I will add it

src/Periphery/FeeForwarder.sol
0xDEnYO: updated

src/Periphery/FeeForwarder.sol
0xDEnYO: we do want to transfer back to our diamond as this is just the end of a pre-bridge swap step....

so we do: 
1) feeCollection  (diamond sends money to feeForwarder which forwards fees and sends remaining ETH back)
2) bridging (remaining funds will be swapped/bridged/whatever)

test/solidity/Periphery/FeeForwarder.t.sol
0xDEnYO: A "FeesForwarded" event will be emitted even if no tokens are transferred so log length would never be 0. 
The comment here was wrong. I updated it.

test/solidity/Periphery/FeeForwarder.t.sol
0xDEnYO: Yes, agreed with the hardcoded values. I added variables for it. 

For the USER_SENDER vs calling via diamond question....if you look at other periphery test files (e.g. FeeCollector, Executor, etc.) we call the contracts all directly. 
Since the test file is also a contract, it doesnt really make a difference to call via diamond, other than that it complicates the test file.
The files are unit test files (making sure the functions work as intended), what you describe goes more into integration test direction (which we dont do).

Do you see any specific benefit of testing via diamond or a specific scenario we should make sure works correctly?

test/solidity/Periphery/FeeForwarder.t.sol
0xDEnYO: good point. We actually have WITHDRAW_WALLET in TestBase.sol so we can also use this instead.
Updated

test/solidity/Periphery/FeeForwarder.t.sol
mirooon: I’m fine leaving it, but using the diamond would make it closer to the real world setup since the function gets called by the diamond directly. That way the tests are easier to read and show the actual use case. Plus, the diamond receive function is already covered, so we wouldn’t need to add it here.

not a big deal though. I’m good with it as is just thought it might help readability. Feel free to resolve.

src/Periphery/FeeForwarder.sol
mirooon: makes sense now. Can we add this explanation to the md file with the flow? Maybe also note that after this call, the rest of the native tokens still need to be handled showing what could be possible scenerios

audit/auditLog.json
```diff
@@ -499,6 +506,15 @@
...
     },
+    "FeeForwarder": {
+      "1.0.0": ["audit20251015"]
+    },
+    "FeeForwarderV2": {
+      "1.0.0": ["audit20250508"]
+    },
+    "FeeForwarderV3": {
+      "1.0.0": ["audit20250508"]
+    },
```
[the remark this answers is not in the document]
mirooon: good point

audit/auditLog.json
```diff
@@ -392,19 +392,26 @@
...
-    "audit20251001": {
-      "auditCompletedOn": "01.10.2025",
+    "audit20251015": {
+      "auditCompletedOn": "15.10.2025",
       "auditedBy": "Sujith Somraaj (individual security researcher)",
       "auditorGitHandle": "sujithsomraaj",
-      "auditReportPath": "./audit/reports/2025.10.01_EcoFacet(v1.0.0).pdf",
-      "auditCommitHash": "8d0e66a69c765e83c13339d7a77071ce2f26968a"
+      "auditReportPath": "./audit/reports/2025.10.15_FeeForwarder(v1.0.0).pdf",
+      "auditCommitHash": "5022c0aca332e2a066f1398e3456632679a836f9"
```
ezynda3, line 414: This actually removes the EcoFacet audit entry...

ezynda3 (changes requested):

audit/auditLog.json
mirooon: it doesn't. It's still in lines 395-401.

ezynda3 (approved):

PR #1422, opened 2025-10-17
add polymer cctp facet + scripts

RnkSngh:

### DIFF
--- a/script/demoScripts/PolymerCCTP.s.sol
+++ b/script/demoScripts/PolymerCCTP.s.sol
@@ -12,7 +12,7 @@ contract CallPolymerCCTPFacet is Script {
     function run() external payable {
         uint256 deployerPrivateKey = vm.envUint("PRIVATE_KEY");
         address diamondAddress = vm.envAddress("DIAMOND_ADDRESS");
-        uint32 destinationDomain = uint32(vm.envUint("DESTINATION_DOMAIN"));
+        uint32 destinationChainId = uint32(vm.envUint("DESTINATION_CHAIN_ID"));
         address receiver = vm.addr(deployerPrivateKey);
         uint256 amount = uint256(1000);
         uint256 polymerTokenFee = uint256(10);
@@ -41,7 +41,7 @@ contract CallPolymerCCTPFacet is Script {
             sendingAssetId: usdcAddress,
             receiver: receiver,
             minAmount: amount,
-            destinationChainId: uint256(destinationDomain), // Using domain as chain ID to avoid SLOAD to read from a mapping - this seems like it can just directly be passed via calldata.
+            destinationChainId: destinationChainId, // Using domain as chain ID to avoid SLOAD to read from a mapping - this seems like it can just directly be passed via calldata.
             hasSourceSwaps: false,
             hasDestinationCall: false
         });
@@ -56,7 +56,7 @@ contract CallPolymerCCTPFacet is Script {
 
         console2.log("Calling startBridgeTokensViaPolymerCCTP...");
         console2.log("Amount:", amount);
-        console2.log("Destination Domain:", destinationDomain);
+        console2.log("Destination Domain:", destinationChainId);
         console2.log("Receiver:", receiver);
 
         // Call the bridge function
--- a/script/deploy/facets/DeployDiamondWithPolymerCCTPFacet.s.sol
+++ b/script/deploy/facets/DeployDiamondWithPolymerCCTPFacet.s.sol
@@ -36,8 +36,9 @@ contract DeployDiamondWithPolymerCCTPFacet is Script {
 
         // Add PolymerCCTPFacet to diamond
         console2.log("Adding PolymerCCTPFacet to diamond...");
-        bytes4[] memory selectors = new bytes4[](1);
+        bytes4[] memory selectors = new bytes4[](2);
         selectors[0] = PolymerCCTPFacet.startBridgeTokensViaPolymerCCTP.selector;
+        selectors[1] = PolymerCCTPFacet.initPolymerCCTP.selector;
 
         LibDiamond.FacetCut[] memory cuts = new LibDiamond.FacetCut[](1);
         cuts[0] = LibDiamond.FacetCut({
--- a/src/Facets/PolymerCCTPFacet.sol
+++ b/src/Facets/PolymerCCTPFacet.sol
@@ -251,6 +251,18 @@ contract PolymerCCTPFacet is ILiFi, ReentrancyGuard, SwapperV2, Validatable, LiF
         if (chainId == 84532) {
             return 6; // Base Sepolia
         }
+        if (chainId == 421614) {
+            return 3; // Arbitrum Sepolia 
+        }
+        if (chainId == 43113) {
+            return 1; // Avalanche Fuji 
+        }
+        if (chainId == 80002) {
+            return 7; // Polygon Amoy
+        }
+        if (chainId == 1301) {
+            return 10; // Unichain Sepolia 
+        }
 
         revert InvalidCallData();
     }
--- a/testnet-addresses.json
+++ b/testnet-addresses.json
@@ -1,30 +1,44 @@
 {
   "11155111": {
-    "usdc": "0x5fd84259d66Cd46123540766Be93DFE6D43130D7",
-    "polymerFeeRecipient": "0x81b5c1c0343ff0087C04F543D0f36dC4745b999F",
+    "usdc": "0x1c7D4B196Cb0C7B01d743Fbc6116a902379C7238",
+    "polymerFeeRecipient": "0x3d1ed7307Ee6dEB2e1cEA9D96C33574fA2a35Bb6",
     "tokenMessenger": "0x8FE6B999Dc680CcFDD5Bf7EB0974218be2542DAA",
-    "diamondProxy": "0x4d09Dd99652e20E6795D08d25ecC03e9600836a4"
+    "diamondProxy": "0x45D672abaa635a6e4F1C0a4071Fe4C8cd95d6E1B"
   },
   "11155420": {
     "usdc": "0x5fd84259d66Cd46123540766Be93DFE6D43130D7",
-    "polymerFeeRecipient": "0x81b5c1c0343ff0087C04F543D0f36dC4745b999F",
+    "polymerFeeRecipient": "0x3d1ed7307Ee6dEB2e1cEA9D96C33574fA2a35Bb6",
     "tokenMessenger": "0x8FE6B999Dc680CcFDD5Bf7EB0974218be2542DAA",
-    "diamondProxy": "0x4d09Dd99652e20E6795D08d25ecC03e9600836a4"
+    "diamondProxy": "0x827A6FeB2E1B060f30606314B6D8167a10AF193f"
   },
   "421614": {
     "usdc": "0x75faf114eafb1BDbe2F0316DF893fd58CE46AA4d",
-    "polymerFeeRecipient": "0x81b5c1c0343ff0087C04F543D0f36dC4745b999F",
+    "polymerFeeRecipient": "0x3d1ed7307Ee6dEB2e1cEA9D96C33574fA2a35Bb6",
     "tokenMessenger": "0x8FE6B999Dc680CcFDD5Bf7EB0974218be2542DAA",
-    "diamondProxy": "0x06B90798BAc18a7927e62a194523131502560911"
+    "diamondProxy": "0x15B7f6B56CB47270184C65F649D597c5dA7083fF"
   },
   "84532": {
     "usdc": "0x036CbD53842c5426634e7929541eC2318f3dCF7e",
-    "polymerFeeRecipient": "0x81b5c1c0343ff0087C04F543D0f36dC4745b999F",
+    "polymerFeeRecipient": "0x3d1ed7307Ee6dEB2e1cEA9D96C33574fA2a35Bb6",
     "tokenMessenger": "0x8FE6B999Dc680CcFDD5Bf7EB0974218be2542DAA",
-    "diamondProxy": "0x4a0258A6627e9D8cef54B52Bde9814B13570Cbb7"
+    "diamondProxy": "0x8A4A0308738CbA78C95Fc85fb0D84ebeABEfe3d8"
   },
-  "2": {
-    "usdc": "4zMMC9srt5Ri5X14GAgXhaHii3GnPAEERYPJgZJDncDU",
-    "diamondProxy": "2sFuM27ehoddDAjGbrZTbYo25LfGbdfUgvqRT4XYehdW"
+  "43113": {
+    "usdc": "0x5425890298aed601595a70AB815c96711a31Bc65",
+    "tokenMessenger": "0x8FE6B999Dc680CcFDD5Bf7EB0974218be2542DAA",
+    "polymerFeeRecipient": "0x3d1ed7307Ee6dEB2e1cEA9D96C33574fA2a35Bb6",
+    "diamondProxy": "0x56cBeC86A95Ea3B17Dcd993e9AdEfD777d4AeF22"
+  },
+  "80002": {
+    "usdc": "0x41E94Eb019C0762f9Bfcf9Fb1E58725BfB0e7582",
+    "tokenMessenger": "0x8FE6B999Dc680CcFDD5Bf7EB0974218be2542DAA",
+    "polymerFeeRecipient": "0x3d1ed7307Ee6dEB2e1cEA9D96C33574fA2a35Bb6",
+    "diamondProxy": "0xFCfdBF8023cD204A7890beD02E4809d9345Fe125"
+  },
+  "1301": {
+    "usdc": "0x31d0220469e10c4E71834a79b1f276d740d3768F",
+    "tokenMessenger": "0x8FE6B999Dc680CcFDD5Bf7EB0974218be2542DAA",
+    "polymerFeeRecipient": "0x3d1ed7307Ee6dEB2e1cEA9D96C33574fA2a35Bb6",
+    "diamondProxy": "0x63C0d99458E2d2153cF95ad8b800e191633F0d69"
   }
 }
--- a/script/demoScripts/demoPolymer.ts
+++ b/script/demoScripts/demoPolymer.ts
@@ -493,7 +493,7 @@ async function main() {
 
   // Set from/to addresses based on chain types
   const fromUserAddress = isFromSolana ? solanaAddress : evmAddress
-  const toUserAddress = "2sFuM27ehoddDAjGbrZTbYo25LfGbdfUgvqRT4XYehdW" // isToSolana ? solanaAddress : evmAddress
+  const toUserAddress =  isToSolana ? solanaAddress : evmAddress
 
   console.log(`From Address (${isFromSolana ? 'Solana' : 'EVM'}): ${fromUserAddress}`)
   console.log(`To Address (${isToSolana ? 'Solana' : 'EVM'}): ${toUserAddress}`)
--- a/script/deploy/DeployPolymerContracts.sh
+++ b/script/deploy/DeployPolymerContracts.sh
@@ -130,3 +130,6 @@ jq --arg chainId "$CHAIN_ID" \
 echo -e "${GREEN}Updated $ADDRESSES_FILE${NC}"
 echo -e "${GREEN}Deployment complete!${NC}"
 echo -e "${BLUE}To call the facet, run: ./script/deploy/CallPolymerCCTP.sh <RPC_URL> <PRIVATE_KEY> <DESTINATION_DOMAIN> $ADDRESSES_FILE${NC}"
+
+echo -e "${BLUE}Initializing polymer facet ${NC}"
+cast send "$DIAMOND_ADDRESS" "initPolymerCCTP()" --rpc-url "$RPC_URL" --private-key "$PRIVATE_KEY"


script/demoScripts/PolymerCCTP.s.sol
```diff
@@ -0,0 +1,69 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.17;
```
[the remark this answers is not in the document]
RnkSngh: The other s.sol files in this repo (e.g. DeployAcrossFacet.s.sol, DeployAccessManagerFacet.s.sol) all use Unlicenced with pragma ^0.8.17 so i'll neglect this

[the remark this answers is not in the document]
RnkSngh: The other s.sol files in this repo (e.g. DeployAcrossFacet.s.sol, DeployAccessManagerFacet.s.sol) all use Unlicenced with pragma ^0.8.17 so i'll neglect this

script/demoScripts/PolymerCCTP.s.sol
```diff
@@ -0,0 +1,69 @@
...
+            bridge: "PolymerCCTP",
+            integrator: "LiFi",
+            referrer: address(0),
+            sendingAssetId: usdcAddress,
+            receiver: receiver,
+            minAmount: amount,
+            destinationChainId: uint256(destinationDomain), // Using domain as chain ID for simplicity
+            hasSourceSwaps: false,
+            hasDestinationCall: false
+        });
```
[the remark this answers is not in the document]
RnkSngh: this is incorrect - the block.chainId would be the source chain id, and not the destination chain id.

Ultimately, the issue is not in this script, but instead of whether we should use the destinationChainId in this `ILifi.BridgeData` to represent the true chainId or to just represent the domainId. This decision was based on the existing `MayanFacet.sol` in this repo, - doing it this way saves some gas since we pass in less calldata instead of passing in the extra domainId (or, avoid a SLOAD from storing the chain -> domainID mapping in each contract). 

I'm going to leave as-is for now since it seems like we're following existing patterns, but happy to revisit upon a comment from a human.

script/demoScripts/demoPolymer.sh
```diff
@@ -0,0 +1,211 @@
...
+
+if [ "$TO_TOKEN" = "null" ] || [ -z "$TO_TOKEN" ]; then
+  echo "Error: USDC address not found for chain ID $TO_CHAIN_ID in $ADDRESSES_FILE"
+  exit 1
+fi
+
+# Derive wallet address from private key
+USER_ADDRESS=$(cast wallet address "$PRIVATE_KEY")
+echo "Wallet Address: $USER_ADDRESS"
+echo ""
```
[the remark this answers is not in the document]
RnkSngh: added

src/Interfaces/IMessageTransmitter.sol
```diff
@@ -0,0 +1,16 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.20;
+
```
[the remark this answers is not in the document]
RnkSngh: this is an external interface

src/Facets/PolymerCCTPFacet.sol
```diff
@@ -0,0 +1,100 @@
...
+            _bridgeData.minAmount,
+            uint32(_bridgeData.destinationChainId),
+            _bridgeData.receiver == NON_EVM_ADDRESS
+                ? _polymerData.nonEvmAddress
+                : bytes32(uint256(uint160(_bridgeData.receiver))),
+            usdc,
+            bytes32(0), // Unrestricted caller
+            _polymerData.maxCCTPFee, // maxFee - 0 means no fee limit
+            _polymerData.minFinalityThreshold // minFinalityThreshold - use default
+        );
```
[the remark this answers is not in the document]
RnkSngh: keccaking the receiving address is nonsense; but i can implement the NON_EVM_ADDRESS validation

src/Interfaces/IPolymerCCTP.sol
```diff
@@ -0,0 +1,22 @@
...
+pragma solidity ^0.8.17;
+
+import {ILiFi} from "../Interfaces/ILiFi.sol";
+
+struct PolymerCCTPData {
+    uint256 polymerTokenFee;
+    uint256 maxCCTPFee;
+    bytes32 nonEvmAddress; // Should only be nonzero if submitting to a nonEvm chain
+    uint32 minFinalityThreshold;
+}
```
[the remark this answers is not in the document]
RnkSngh: this doesn't make any sense

src/Interfaces/ITokenMessenger.sol
```diff
@@ -0,0 +1,22 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.20;
+
```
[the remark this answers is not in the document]
RnkSngh: this is an external interface so we can neglect this

src/Interfaces/IPolymerCCTP.sol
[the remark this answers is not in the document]
RnkSngh: not sure what guidelines are being referred to here, but looking at existing mayan facet we can just use the receiver address passed in from the brdigeData  - I don't see why we need to pass in the receiver seperately

script/demoScripts/demoPolymer.sh
```diff
@@ -0,0 +1,268 @@
...
+  exit 1
+fi
+
+# Extract the first route and its first step
+FIRST_STEP=$(echo "$ROUTES_RESPONSE" | jq -r '.routes[1].steps[0]')
+
+if [ "$FIRST_STEP" = "null" ]; then
+  echo "No routes found"
+  exit 1
+fi
```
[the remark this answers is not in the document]
RnkSngh: this is to test fast path so we want [1]

src/Facets/PolymerCCTPFacet.sol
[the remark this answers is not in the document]
RnkSngh: implemented

src/Facets/PolymerCCTPFacet.sol
```diff
@@ -0,0 +1,161 @@
...
+/// @notice Provides functionality for bridging USDC through Polymer CCTP
+/// @custom:version 1.0.0
+contract PolymerCCTPFacet is IPolymerCCTPFacet, ILiFi, ReentrancyGuard, SwapperV2, Validatable, LiFiData {
+    ITokenMessenger public immutable TOKEN_MESSENGER;
+    address public immutable USDC;
+    address payable public immutable POLYMER_FEE_RECEIVER;
+
+    modifier validatePolymerData(ILiFi.BridgeData memory _bridgeData, PolymerCCTPData calldata _polymerData) {
+        // Note; we can't use the validateBridgeData because it doesn't support the NON_EVM_ADDRES PATH 
+        if (LibUtil.isZeroAddress(_bridgeData.receiver) && _bridgeData.receiver != NON_EVM_ADDRESS) {
```
0xDEnYO: The second part of this check can be removed as it will always be true if the first part is true (it cannot be address(0) AND == NON_EVM_ADDRESS)

src/Facets/PolymerCCTPFacet.sol
```diff
@@ -0,0 +1,161 @@
...
+        }
+
+        if (_polymerData.polymerTokenFee >= _bridgeData.minAmount) {
+            revert FeeCannotBeLessThanAmount();
+        }
+        _;
+    }
+
+    constructor(address _tokenMessenger, address _usdc, address _polymerFeeReceiver) {
+        // TODO: Do we want to have fee collector here?
```
0xDEnYO: no, we dont need to since FeeCollector will be called using LibSwap.SwapData[]

src/Facets/PolymerCCTPFacet.sol
```diff
@@ -0,0 +1,161 @@
...
+
+        if (_tokenMessenger == address(0) || _usdc == address(0) || _polymerFeeReceiver == address(0)) {
+            revert InvalidAddress();
```
0xDEnYO: nitpick: we usually use InvalidConfig() for invalid constructor parameters

src/Facets/PolymerCCTPFacet.sol
```diff
@@ -0,0 +1,161 @@
...
+            revert InvalidReceiver();
+        }
+        if (_bridgeData.minAmount == 0) {
+            revert InvalidAmount();
+        }
+        if (_bridgeData.destinationChainId == block.chainid) {
+            revert CannotBridgeToSameNetwork();
+        }
+
+        if (_bridgeData.sendingAssetId != USDC) {
```
0xDEnYO: this check is a duplicate to `onlyAllowSourceToken` modifier

one of them is sufficient.
Would suggest to keep the modifier as it also uses our InvalidAddress() error from GenericErrors.sol so you can remove that InvalidAddress error from the interface, too.

src/Facets/PolymerCCTPFacet.sol
```diff
@@ -0,0 +1,161 @@
...
+        nonReentrant
+        refundExcessNative(payable(msg.sender))
+        validatePolymerData(_bridgeData, _polymerData)
+        onlyAllowSourceToken(_bridgeData, USDC)
+        doesNotContainSourceSwaps(_bridgeData)
+        doesNotContainDestinationCalls(_bridgeData)
+    {
+
+        // TODO: Do we need this check if it's always going to be usdc?
+        LibAsset.depositAsset(_bridgeData.sendingAssetId, _bridgeData.minAmount);
```
0xDEnYO: I think we can safely use 
```solidity 
LibAsset.transferFromERC20(USDC, msg.sender, address(this), _bridgeData.minAmount + _polymerData.polymerTokenFee)
```

Considerations:
- we ensured `minAmount != 0` (in modifier)
- we ensured `bridgeData.sendingAssetId is == USDC` (in modifier)
- bridgeData.minAmount is typically the amount that actually goes into the bridge
- we can then use bridgeData.minAmount below without any further calculations

src/Facets/PolymerCCTPFacet.sol
```diff
@@ -0,0 +1,161 @@
...
+    /// @param _bridgeData The core information needed for bridging
+    /// @param _polymerData Data specific to PolymerCCTP
+    function _startBridge(
+        ILiFi.BridgeData memory _bridgeData,
+        PolymerCCTPData calldata _polymerData
+    ) internal {
+        LibAsset.transferERC20(USDC, POLYMER_FEE_RECEIVER, _polymerData.polymerTokenFee);
+
+        // TODO we don't need to use safe approve here?
+        IERC20(USDC).approve(address(TOKEN_MESSENGER), _bridgeData.minAmount - _polymerData.polymerTokenFee);
```
0xDEnYO: We would be comfortable with a standing max approval here to save gas.
Our diamond is designed to not hold any funds, so even if TOKEN_MESSENGER turns malicious, it cannot steal any funds. And in case there would be any funds sitting in our contract, there would be other ways to grab them, too. So having a max approval here doesnt introduce any new risks but saves gas. 

A comment to explain this rationale would be good for auditors and bug bounty hunters.

src/Facets/PolymerCCTPFacet.sol
```diff
@@ -0,0 +1,161 @@
...
+    /// @dev Contains the business logic for the bridge via PolymerCCTP
+    /// @param _bridgeData The core information needed for bridging
+    /// @param _polymerData Data specific to PolymerCCTP
+    function _startBridge(
+        ILiFi.BridgeData memory _bridgeData,
+        PolymerCCTPData calldata _polymerData
+    ) internal {
+        LibAsset.transferERC20(USDC, POLYMER_FEE_RECEIVER, _polymerData.polymerTokenFee);
+
+        // TODO we don't need to use safe approve here?
```
0xDEnYO: I think it's safe with USDC as it has a clean implementation but I would suggest to leave the comment in there for the auditors to confirm as well.

src/Facets/PolymerCCTPFacet.sol
```diff
@@ -0,0 +1,161 @@
...
+            _bridgeData.receiver == NON_EVM_ADDRESS
+                ? _polymerData.nonEvmAddress
+                : bytes32(uint256(uint160(_bridgeData.receiver))),
+            USDC,
+            bytes32(0), // Unrestricted caller
+            _polymerData.maxCCTPFee, // maxFee - 0 means no fee limit
+            _polymerData.minFinalityThreshold // minFinalityThreshold - use default
+        );
+
+        emit PolymerCCTPFeeSent(_bridgeData.minAmount, _polymerData.polymerTokenFee, _polymerData.minFinalityThreshold);
```
0xDEnYO, line 167: I understand that this event is required for your off-chain logic to pick up the transaction. This would be good to explain in a comment.

src/Interfaces/IMessageTransmitter.sol
0xDEnYO, line 1: where is this being used?

src/Interfaces/IPolymerCCTP.sol
```diff
@@ -0,0 +1,32 @@
...
+
+import {ILiFi} from "../Interfaces/ILiFi.sol";
+
+struct PolymerCCTPData {
+    // Token fee taken in USDC by the facet (optional; may be zero)
+    uint256 polymerTokenFee;
+    // CCTP fee caps/params
+    uint256 maxCCTPFee;
+    bytes32 nonEvmAddress; // Should only be nonzero if submitting to a nonEvm chain
+    uint32 minFinalityThreshold;
```
0xDEnYO: please add comment to explain the parameter

test/solidity/Facets/PolymerCCTPFacet.t.sol
```diff
@@ -0,0 +1,320 @@
...
+
+// Mock DEX for swapping
+contract MockDEX {
+    function swap(address tokenIn, address tokenOut, uint256 amountIn, uint256 amountOut) external {
+        ERC20(tokenIn).transferFrom(msg.sender, address(this), amountIn);
+        ERC20(tokenOut).transfer(msg.sender, amountOut);
+    }
+}
+
+contract PolymerCCTPFacetTest is Test {
```
0xDEnYO, line 84: We have TestBaseFacet.sol with standard tests for all bridge facets that could be leveraged here. 
We could also help you with this stuff once the contract is finalized.

Unit test line coverage must be 100%.

testnet-addresses.json
0xDEnYO, line 1: we would usually store these values in 
config/polymercctp.json

but we can also take this over once you have the contract finalized.

src/Facets/PolymerCCTPFacet.sol
```diff
@@ -0,0 +1,161 @@
...
+    function _startBridge(
+        ILiFi.BridgeData memory _bridgeData,
+        PolymerCCTPData calldata _polymerData
+    ) internal {
+        LibAsset.transferERC20(USDC, POLYMER_FEE_RECEIVER, _polymerData.polymerTokenFee);
+
+        // TODO we don't need to use safe approve here?
+        IERC20(USDC).approve(address(TOKEN_MESSENGER), _bridgeData.minAmount - _polymerData.polymerTokenFee);
+
+        // Need tocheck: can we just use destinationChainID as the normal chain id? and can we just mpass in min Amount as the amountT?
```
0xDEnYO: since we support nonEVM chains here and LI.FI uses custom chain IDs for nonEVM chains, we need to add a mapping for lifiChainId > uint32(chainID) (recognized by TokenMessenger contract)

Our custom chainIDs can be found in LiFiData.sol

If new non-EVM chains are being added later on, we will update the facet/mapping and redeploy it. 

You can check RelayFacet
https://github.com/lifinance/contracts/blob/main/src/Facets/RelayFacet.sol#L224
or AcrossFacetV4 
https://github.com/lifinance/contracts/blob/main/src/Facets/AcrossFacetV4.sol#L252

to see how such a mapping could look

src/Facets/PolymerCCTPFacet.sol
```diff
@@ -0,0 +1,161 @@
...
+            bytes32(0), // Unrestricted caller
+            _polymerData.maxCCTPFee, // maxFee - 0 means no fee limit
+            _polymerData.minFinalityThreshold // minFinalityThreshold - use default
+        );
+
+        emit PolymerCCTPFeeSent(_bridgeData.minAmount, _polymerData.polymerTokenFee, _polymerData.minFinalityThreshold);
+
+        // Emit Li.Fi standard event
+        // TODO: Check - do we need to emit this event?
+        emit LiFiTransferStarted(BridgeData(
```
0xDEnYO, line 170: please add (somewhere)

        if (_bridgeData.receiver == NON_EVM_ADDRESS) {
            emit BridgeToNonEVMChainBytes32(
                _bridgeData.transactionId,
                _bridgeData.destinationChainId,
                _polymerData.nonEvmAddress
            );
        }
        
        
Consider to do the check (_bridgeData.receiver == NON_EVM_ADDRESS) only once, e.g. store in variable. In that context it also makes sense to move some of the parameter validations from the modifier into _startBridge function to save gas as otherwise we will do this check several times

src/Facets/PolymerCCTPFacet.sol
```diff
@@ -0,0 +1,161 @@
...
+        }
+        if (_bridgeData.destinationChainId == block.chainid) {
+            revert CannotBridgeToSameNetwork();
+        }
+
+        if (_bridgeData.sendingAssetId != USDC) {
+            revert InvalidSendingAsset(_bridgeData.sendingAssetId, USDC);
+        }
+
+        if (_polymerData.polymerTokenFee >= _bridgeData.minAmount) {
```
0xDEnYO: this check and the associated custom error can potentially also be removed if we change the way we deposit (minAmount + feeAmount).

src/Facets/PolymerCCTPFacet.sol
RnkSngh: Ah good catch - I meant this to validate that the `polymerData.nonEvmAddress` is *nonzero* for the case where an address different than the `NON_EVM_ADDRESS` is passed into the `_bridgeData.receiver` - assuming this is worth doing so I'm going to update this with the intended check

src/Facets/PolymerCCTPFacet.sol
RnkSngh: We had fee deduction as you suggested initially, but changed it to subtract the token fee from the amount from some feedback from @maxklenk - can you confirm that this is indeed the finalized way of passing in amount?

src/Facets/PolymerCCTPFacet.sol
RnkSngh: that makes sense - so to confirm, implementing your suggestion means removing the approve altogether here and assuming it will be done once outside of this function?

maxklenk: I agree that a maxApproval should be done once. We have some helpers in other facets to trigger those approvals once. @0xDEnYO can we reuse them or should we have a dedicated function in here?

src/Facets/PolymerCCTPFacet.sol
maxklenk: I think `_bridgeData.minAmount` should the full token amount needed for bridging (including the fee), as this is how we handle it for other bridges (they usually deduct the fee in their contracts). This is especially important for the swapAndBridge flow, cause the minAmount will be the swap result and there will be no additional USDC we can pay the fee with.

src/Facets/PolymerCCTPFacet.sol
0xDEnYO: since there is only one token (USDC) we can have a simple init function that we can call after the facet was added to the diamond 

something like this:

```solidity


    /// @notice Sets a max approval from diamond to TokenMessenger
    /// It is safe to set a max approval since the diamond is designed to not hold any funds 
    /// (that could otherwise be stolen if TokenMessenger turns malicious)
    /// We also dont need to store the initialization status of this facet since it will not break from being 
    /// initialized multiple times (plus it's an admin-only function)
    function initPolymerCCTP() external {
        LibDiamond.enforceIsContractOwner();

        IERC20(USDC).approve(address(TOKEN_MESSENGER), type(uint256).max);
    } 
    
````

File: src/Facets/PolymerCCTPFacet.sol (hunk 13 above)
Comment (0xDEnYO) [reply src/Facets/PolymerCCTPFacet.sol]: discussed this with Max internally and we will go forward with what Max suggested. 
We will treat the _bridgeData.minAmount as bridgeAmount + feeAmount
Which also means we will keep the code where we need to subtract the feeAmount from the (potentially updated) bridgeAmount. 

I will delete my other review comments

File: src/Facets/PolymerCCTPFacet.sol (hunk 21 above)
Comment (0xDEnYO) [reply src/Facets/PolymerCCTPFacet.sol]: If your fee is always proportional then we can keep this check (and you can resolve the comment). 
If there are potential cases where a very little amount is bridged and the (min) fee is higher then we need to remove it.

File: script/demoScripts/CallPolymerCCTP.sh
Comment (0xDEnYO) [inline script/demoScripts/CallPolymerCCTP.sh:1]: If you could add a typescript demoScript we would really appreciate it, it helps our backend team to integrate faster.

AI/Cursor is usually pretty good at creating those if you instruct it to look at existing demoScripts, use helper functions etc.

File: src/Facets/PolymerCCTPFacet.sol (hunk 21 above)
Comment (RnkSngh) [reply src/Facets/PolymerCCTPFacet.sol]: in the case where this check is invalidated, it would still revert even if we didn't have this if statement due to underflow. So if we really want to optimize for gas and not for readability; i think we can remove? I removed this check since it seems like we want to be more gas-oriented -

File: src/Facets/PolymerCCTPFacet.sol (hunk 20 above)
Comment (RnkSngh) [reply src/Facets/PolymerCCTPFacet.sol:170]: sounds good - I was being mindful of following CEI patterns but I also agree it makes more sense & might be cleaner, especially since we should only ever be interacting with usdc. 

I moved logic around for gas ops - assuming the compiler optimizer will also do some work as well to implement things like copy/pasting internal methods to avoid popping/pushing to stack, etc, it should be pretty lean now

File: src/Facets/PolymerCCTPFacet.sol (hunk 10 above)
Comment (RnkSngh) [reply src/Facets/PolymerCCTPFacet.sol]: removed comment!

File: src/Facets/PolymerCCTPFacet.sol (hunk 11 above)
Comment (RnkSngh) [reply src/Facets/PolymerCCTPFacet.sol]: updated to use your canonical error

File: src/Facets/PolymerCCTPFacet.sol (hunk 12 above)
Comment (RnkSngh) [reply src/Facets/PolymerCCTPFacet.sol]: updated to use the onlyAllowSourceToken

File: src/Facets/PolymerCCTPFacet.sol (hunk 13 above)
Comment (RnkSngh) [reply src/Facets/PolymerCCTPFacet.sol]: sounds good - will keep as-is then, thanks for resolving

File: src/Facets/PolymerCCTPFacet.sol (hunk 14 above)
Comment (RnkSngh) [reply src/Facets/PolymerCCTPFacet.sol]: thanks for the snippet - I added, w some minor adjustments to the natspec + also added todo comments for things to check during audits

File: src/Facets/PolymerCCTPFacet.sol (hunk 15 above)
Comment (RnkSngh) [reply src/Facets/PolymerCCTPFacet.sol]: left a comment in the code

File: src/Facets/PolymerCCTPFacet.sol (hunk 16 above)
Comment (RnkSngh) [reply src/Facets/PolymerCCTPFacet.sol:167]: added, thanks!

File: src/Interfaces/IMessageTransmitter.sol
Comment (RnkSngh) [reply src/Interfaces/IMessageTransmitter.sol:1]: it's not - I was confusing this with tokenMessenger. This unused interface should now be removed ~

File: src/Interfaces/IPolymerCCTP.sol (hunk 17 above)
Comment (RnkSngh) [reply src/Interfaces/IPolymerCCTP.sol]: added; also updated the ITokenMessenger to include[ circle's natspec](https://github.com/circlefin/evm-cctp-contracts/blob/4061786a5726bc05f99fcdb53b0985599f0dbaf7/src/v2/TokenMessengerV2.sol#L32)

File: src/Facets/PolymerCCTPFacet.sol (hunk 19 above)
Comment (RnkSngh) [reply src/Facets/PolymerCCTPFacet.sol]: ah gotcha; makes sense and seems to be a good use of the diamond pattern to avoid SLOADs. 

I added all the domain ids - [currently supported by cctp](https://developers.circle.com/stablecoins/usdc-contract-addresses) ; defaulting to canonical chain ids for chains that aren't in your custom LifiData.sol chain ids. The only chain which is supported by cctp and which overlaps with your LifiData.sol mapping is Solana - does that sound correct? 

I also took the liberty of adding testnet chains to this mapping method; I know you all don't have contracts on testnets but Polymer does, and that would greatly simplify our integration with cctp for testing . Including these testnet chains in the mapping shouldn't impact mainnet gas costs for calling this method for the happy path since the testnet chains are the last cases and the function should return in mainnet chains for valid inputs. But let me know if you if you still have good reasons for not wanting the testnet chains in there.

File: script/demoScripts/CallPolymerCCTP.sh
Comment (RnkSngh) [reply script/demoScripts/CallPolymerCCTP.sh:1]: added - should be at scripts/demoScripts/demoPolymer.ts now

File: src/Facets/PolymerCCTPFacet.sol (hunk 21 above)
Comment (0xDEnYO) [reply src/Facets/PolymerCCTPFacet.sol]: yes, agreed.

File: script/deploy/facets/DeployDiamondWithPolymerCCTPFacet.s.sol [hunk 22]
```diff
@@ -0,0 +1,54 @@
...
+pragma solidity ^0.8.17;
+
+import {Script, console2} from "forge-std/Script.sol";
+import {LiFiDiamond} from "lifi/LiFiDiamond.sol";
+import {DiamondCutFacet} from "lifi/Facets/DiamondCutFacet.sol";
+import {PolymerCCTPFacet} from "lifi/Facets/PolymerCCTPFacet.sol";
+import {DeployScriptBase} from "./utils/DeployScriptBase.sol";
+import {LibDiamond} from "lifi/Libraries/LibDiamond.sol";
+
+contract DeployDiamondWithPolymerCCTPFacet is Script {
```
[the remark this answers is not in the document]
0xDEnYO: We will have to adapt / split this script to be coherent with the rest of our deploy scripts (as suggested by coderabbit).
It also mixes deploying a diamond etc which it shouldnt. 
Feel free to check out the other Deploy... and Update... scripts to derive the patterns.

However, we can also take this over as it's easier for us to do that.

src/Facets/PolymerCCTPFacet.sol
```diff
@@ -0,0 +1,246 @@
...
+/// @notice Provides functionality for bridging USDC through Polymer CCTP
+/// @custom:version 1.0.0
+contract PolymerCCTPFacet is IPolymerCCTPFacet, ILiFi, ReentrancyGuard, SwapperV2, Validatable, LiFiData {
+    ITokenMessenger public immutable TOKEN_MESSENGER;
+    address public immutable USDC;
+    address payable public immutable POLYMER_FEE_RECEIVER;
+
+    // Note: we can't use the validateBridgeData because it doesn't support the NON_EVM_ADDRESS PATH
+    modifier validatePolymerData(ILiFi.BridgeData memory _bridgeData, PolymerCCTPData calldata _polymerData) {
+        // Note: There are more checks within the _startBridge method so we don't strictly follow CEI, but this should be fine given checks should ensure that we only ever interact with USDC
```
0xDEnYO, line 47: correct. Also the diamond does not keep any state so the CEI pattern is less relevant for our protocol.

src/Facets/PolymerCCTPFacet.sol
0xDEnYO: we can still apply this though to save some gas (and remove the comment)
```solidity
LibAsset.transferFromERC20(USDC, msg.sender, address(this), _bridgeData.minAmount)
```

src/Facets/PolymerCCTPFacet.sol
```diff
@@ -0,0 +1,246 @@
...
+/// @title PolymerCCTPFacet
+/// @author LI.FI (https://li.fi)
+/// @notice Provides functionality for bridging USDC through Polymer CCTP
+/// @custom:version 1.0.0
+contract PolymerCCTPFacet is IPolymerCCTPFacet, ILiFi, ReentrancyGuard, SwapperV2, Validatable, LiFiData {
+    ITokenMessenger public immutable TOKEN_MESSENGER;
+    address public immutable USDC;
+    address payable public immutable POLYMER_FEE_RECEIVER;
+
+    // Note: we can't use the validateBridgeData because it doesn't support the NON_EVM_ADDRESS PATH
```
0xDEnYO: we could use validateBridgeData cause it just makes sure that _receiver is non-zero which is valid for both paths. 
However, the current implementation is slightly more gas-efficient so I am fine to leave it as is.
But we should remove this comment.

src/Facets/PolymerCCTPFacet.sol
```diff
@@ -0,0 +1,246 @@
...
+
+        // This case first for gas ops since it will likely be triggered more often
+        if (_bridgeData.receiver != NON_EVM_ADDRESS) {
+            // _bridgeData.receiver != NON_EVM_ADDRESS -> mint to _bridgeData.receiver
+            if (_bridgeData.receiver == address(0)) {
+                revert InvalidReceiver();
+            }
+
+            TOKEN_MESSENGER.depositForBurn(
+                _bridgeData.minAmount - _polymerData.polymerTokenFee,
```
0xDEnYO: nitpick: store in variable and use variable below (event emission) to potentially save gas

src/Facets/PolymerCCTPFacet.sol
0xDEnYO: looks fantastic. Thanks !

src/Facets/PolymerCCTPFacet.sol
```diff
@@ -0,0 +1,246 @@
...
+                _bridgeData.destinationChainId,
+                _bridgeData.hasSourceSwaps,
+                _bridgeData.hasDestinationCall
+            ));
+    }
+
+    /// @notice Get CCTP domain ID for destination chain
+    /// @param chainId LIFI specific chain id
+    /// @return CCTP domain ID recognized by TokenMessenger
+    function _chainIdToDomainId(uint256 chainId) internal pure returns (uint32) {
```
0xDEnYO, line 187: are they ordered by expected volume/likelihood?
If not, maybe we can do some optimizations? e.g. Base is probably more likely than avalanche ?

src/Interfaces/IPolymerCCTP.sol
```diff
@@ -0,0 +1,30 @@
...
+/// @custom:version 1.0.0
+interface IPolymerCCTPFacet {
+    error InvalidSendingAsset(address actual, address expected);
+
+    event PolymerCCTPFeeSent(uint256 bridgeAmount, uint256 polymerFee, uint32 minFinalityThreshold);
+
+    /// @notice Start bridging tokens via Polymer CCTP
+    /// @param _bridgeData Core LiFi bridge data
+    /// @param _polymerData Polymer/CCTP specific data (see struct)
+    function startBridgeTokensViaPolymerCCTP(ILiFi.BridgeData memory _bridgeData, PolymerCCTPData calldata _polymerData)
```
0xDEnYO: In our current structure we do not define an interface for each facet. 
FacetData and custom errors are defined in the facet itself. 
We would prefer to stay coherent and would probably remove this interface when integrating this into our codebase, unless you have an explicit need for it?

src/Interfaces/ITokenMessenger.sol
```diff
@@ -0,0 +1,85 @@
...
+     * @param amount amount of tokens to burn
+     * @param destinationDomain destination domain to receive message on
+     * @param mintRecipient address of mint recipient on destination domain, as bytes32
+     * @param burnToken token to burn `amount` of, on local domain
+     * @param destinationCaller authorized caller on the destination domain, as bytes32. If equal to bytes32(0),
+     * any address can broadcast the message.
+     * @param maxFee maximum fee to pay on the destination domain, specified in units of burnToken
+     * @param hookData hook data to append to burn message for interpretation on destination domain
+     */
+    function depositForBurnWithHook(
```
0xDEnYO: if we do not use this function anywhere, we also dont need to keep it in this interface. We usually only add functions that are relevant for us.

src/Facets/PolymerCCTPFacet.sol
RnkSngh: I also considered ordering by volume / likelihood as well but ultimately it's very chain dependent and hard to reason about. 

e.g. your suggestion of putting base first might be more optimal for most chains (assuming base is indeed the popular chain) but it would be strictly worse for bridges originating from base itself, since base should _never_ occur as a destination domain if it's the source domain, but always be checked against in your suggestion. So it's not enough to know that base is a popular chain; we'd instead also need to be sure that base is more often used as a `to` bridge than a `from` bridge, etc. 

I'm happy to implement any ordering you suggest, but I certainly don't know enough to suggest an ordering myself ~

src/Interfaces/ITokenMessenger.sol
RnkSngh: fair enough - removed!

src/Interfaces/IPolymerCCTP.sol
RnkSngh: sure thing - removed  ~

src/Facets/PolymerCCTPFacet.sol
RnkSngh: done; indeed saves ~60 gas, thanks for the suggestion

src/Facets/PolymerCCTPFacet.sol
RnkSngh: i modified the comment to clarify that it's for gas op instead of that it wouldn't work ~

src/Facets/PolymerCCTPFacet.sol
RnkSngh: FYI - your suggestion actually raised runtime gas by 1; I still implemented it though since it's likely just the fact that we're using a mock implementation in the unit test I ran? It could also be due to some obscure compiler op happening ~

src/Facets/PolymerCCTPFacet.sol
```diff
@@ -0,0 +1,257 @@
...
+        external
+        payable
+        nonReentrant
+        refundExcessNative(payable(msg.sender))
+        validatePolymerData(_bridgeData, _polymerData)
+        onlyAllowSourceToken(_bridgeData, USDC)
+        doesNotContainSourceSwaps(_bridgeData)
+        doesNotContainDestinationCalls(_bridgeData)
+    {
+        // TODO: Do we need this check if it's always going to be usdc?
```
maxklenk, line 94: Remove TODO comment

src/Facets/PolymerCCTPFacet.sol
0xDEnYO: I think at the end of the day the gas usage is not so huge here despite this being a lot of if branches. I would be OK to leave as-is. 

We could optimize this a little by ordering incrementally by chainId and implementing a simple tree searching algo. 

@maxklenk what is your take on this?

src/Facets/PolymerCCTPFacet.sol
0xDEnYO: lol, not sure how to explain this but it should be the cleanest impl

src/Facets/PolymerCCTPFacet.sol
maxklenk: I think the overhead is not big and predicting the volume for a custom order is hard to guess and maintain.
I saw nested ifs that represent a tree structure to get to all cases with a constant amount of if checks, we can try that if you want.

PR #1447, opened 2025-11-04
WhitelistManagerFacet add bytes zero check, fixed errors and optimized UpdateWhitestlistManagerFacet script

mirooon:

- Added zero bytes check for selector in `LibAllowList.sol` lib
- Fixed errors and optimized `UpdateWhitestlistManagerFacet.s.sol` script
- Added staging deployments for `bsc`, `optimism` and `gnosis` with `WhitelistManagerFacet`
- Added other small adjustments

### DIFF
--- a/script/deploy/facets/UpdateWhitelistManagerFacet.s.sol
+++ b/script/deploy/facets/UpdateWhitelistManagerFacet.s.sol
@@ -113,17 +113,20 @@ contract DeployScript is UpdateScriptBase {
         view
         returns (address[] memory allContracts, bytes4[][] memory allSelectors)
     {
+        // Parse JSON once
+        JSONParserLib.Item memory root = JSONParserLib.parse(whitelistJson);
+
         // 1. Parse the ".DEXS" section
         (
             address[] memory dexContracts,
             bytes4[][] memory dexSelectors
-        ) = _parseDexsSection(whitelistJson);
+        ) = _parseDexsSection(root);
 
         // 2. Parse the ".PERIPHERY" section
         (
             address[] memory peripheryContracts,
             bytes4[][] memory peripherySelectors
-        ) = _parsePeripherySection(whitelistJson);
+        ) = _parsePeripherySection(root);
 
         // 3. Merge the results from both sections
         uint256 totalContracts = dexContracts.length +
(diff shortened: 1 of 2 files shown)


script/deploy/facets/UpdateWhitelistManagerFacet.s.sol
```diff
@@ -23,6 +25,7 @@ contract DeployScript is UpdateScriptBase {
     }
 
     function getCallData() internal override returns (bytes memory) {
+        vm.pauseGasMetering();
```
0xDEnYO, line 28: just to learn...why is this here?

script/deploy/facets/UpdateWhitelistManagerFacet.s.sol
```diff
@@ -269,212 +299,182 @@ contract DeployScript is UpdateScriptBase {
...
-            contractIndex = _populatePeripheryContracts(
-                whitelistJson,
-                contracts,
-                selectors,
-                contractIndex
-            );
-        } catch {
-            // Network not found, throw error
+        // Parse JSON once using JSONParserLib
+        JSONParserLib.Item memory root = JSONParserLib.parse(whitelistJson);
```
0xDEnYO: we are doing this operation two times (once for periphery and once for DEXs. 
Wondering about the runtime/load. Maybe makes sense to push this to a storage variable and reuse?
But I may be wrong there...

script/deploy/facets/UpdateWhitelistManagerFacet.s.sol
mirooon: So this is one of the reason why it was showing failed: `0x, <empty revert data>` error. Under the hood we were getting `OutOfGas` error because forge by default using script also calculates the gas of this heavy operations of json parsing. But actually It doesnt make sense to calculate the gas for it because what matters is the final onchain call and how much gas it takes, not the heavy data preparation that happens offchain.

0xDEnYO: got it, thanks for sharing

script/deploy/facets/UpdateWhitelistManagerFacet.s.sol
mirooon: good catch, fixed! https://github.com/lifinance/contracts/pull/1447/commits/b8839f828b9834ac71469e20011eb3d3e5e4a847

PR #1456, opened 2025-11-10
added PolymerCCTPFacet + deploy scripts, docs, etc. [PolymerCCTPFacet v1.0.0, ITokenMessenger v1.0.0]

0xDEnYO:
https://lifi.atlassian.net/browse/LF-14910

### DIFF
--- a/script/demoScripts/demoPolymerCCTP.ts
+++ b/script/demoScripts/demoPolymerCCTP.ts
@@ -313,7 +313,6 @@ async function main() {
   consola.info('')
 
   // Polymer API uses its own chain ID mapping (Solana = 2, not LiFi's 1151111081099710)
-  // Polymer API uses its own chain ID mapping (Solana = 2, not LiFi's 1151111081099710)
   const destinationChainIdPolymer = BRIDGE_TO_SOLANA
     ? POLYMER_CHAIN_ID_SOLANA
     : DST_CHAIN_ID


config/polymercctp.json
```diff
@@ -0,0 +1,105 @@
+{
+  "mainnet": {
+    "tokenMessenger": "0x28b5a0e9C621a5BadaA536219b3a228C8168cf5d",
```
RnkSngh: might be worth clarifying - either through renaming this property or through in-line comments somewherhe - that this is v2 addresses

docs/PolymerCCTPFacet.md
```diff
@@ -0,0 +1,101 @@
...
+    D{LiFiDiamond}-- DELEGATECALL -->PolymerCCTPFacet;
+    PolymerCCTPFacet -- CALL --> TM(TokenMessenger)
+```
+
+## Public Methods
+
+- `function startBridgeTokensViaPolymerCCTP(BridgeData memory _bridgeData, PolymerCCTPData calldata _polymerData)`
+  - Simply bridges USDC using Polymer CCTP
+- `function swapAndStartBridgeTokensViaPolymerCCTP(BridgeData memory _bridgeData, SwapData[] calldata _swapData, PolymerCCTPData calldata _polymerData)`
+  - Performs swap(s) before bridging USDC using Polymer CCTP
```
RnkSngh, line 20:
```suggestion
  - Performs swap(s) before bridging USDC using Polymer CCTP, useful for adding hooks on dest chain. 
```

docs/PolymerCCTPFacet.md
```diff
@@ -0,0 +1,101 @@
...
+
+## Polymer CCTP Specific Parameters
+
+The methods listed above take a variable labeled `_polymerData`. This data is specific to Polymer CCTP and is represented as the following struct type:
+
+```solidity
+struct PolymerCCTPData {
+  // Token fee taken in USDC by the facet (optional; may be zero)
```
RnkSngh:
```suggestion
  // Token fee taken in USDC by the facet (optional; should be zero for slow path )
```

docs/PolymerCCTPFacet.md
```diff
@@ -0,0 +1,101 @@
...
+  uint256 polymerTokenFee;
+  // maximum fee to pay on the destination domain, specified in units of burnToken
```
RnkSngh:
```suggestion
  // maximum fee to paid on the destination domain through the difference between burned tokens on src chain and minted token on dest chain, specified in units of burnToken
```

src/Facets/PolymerCCTPFacet.sol
```diff
@@ -0,0 +1,320 @@
...
+
+        TOKEN_MESSENGER = ITokenMessenger(_tokenMessenger);
+        USDC = _usdc;
+        POLYMER_FEE_RECEIVER = payable(_polymerFeeReceiver);
+    }
+
+    /// @notice Sets a max approval from lifiDiamond to TokenMessenger
+    /// It is safe to set a max approval since the diamond is designed to not hold any funds (that could otherwise be stolen if TokenMessenger turns malicious)
+    /// We also don't need to store the initialization status of this facet since it will not break from being initialized multiple times (plus it's an admin-only function)
+    function initPolymerCCTP() external virtual {
```
RnkSngh: might be worth moving to the constructor if we only intend to do this once

src/Facets/PolymerCCTPFacet.sol
```diff
@@ -0,0 +1,320 @@
...
+    /// @param _polymerData Data specific to PolymerCCTP
+    /// @notice Requires caller to approve the LifiDiamondProxy of the bridge amount + polymerFee
+    function startBridgeTokensViaPolymerCCTP(
+        ILiFi.BridgeData memory _bridgeData,
+        PolymerCCTPData calldata _polymerData
+    )
+        external
+        payable
+        nonReentrant
+        refundExcessNative(payable(msg.sender))
```
RnkSngh: do we need these this (or the swaAndStart) to be payable? if not we should consider removing the payable and in turn the `refundExcessNative` for gas saving

src/Facets/PolymerCCTPFacet.sol
```diff
@@ -0,0 +1,320 @@
...
+            return 13; // Sonic
+        }
+        if (chainId == 480) {
+            return 14; // World Chain
+        }
+        if (chainId == 1329) {
+            return 16; // Sei
+        }
+        if (chainId == 56) {
+            return 17; // BNB Smart Chain
```
RnkSngh: indeed we should remove this one since from the v2 docs it looks like from [the v2 docs that BNB only supports USYC ](https://developers.circle.com/cctp/evm-smart-contracts#messagev2-testnet)-  my bad!

script/demoScripts/demoPolymerCCTP.ts
```diff
@@ -0,0 +1,203 @@
...
+      false
+    )
+    consola.info('✅ Balance and allowance verified')
+  }
+
+  // Execute transaction
+  if (SEND_TX) {
+    consola.info('🚀 Executing bridge transaction...')
+
+    const executeTxData = await polymerCCTPFacet.populateTransaction
```
RnkSngh: this script looks like it doesn't use the polymer cctp service?

0xDEnYO: added now (still need to add the correct API URL)

src/Facets/PolymerCCTPFacet.sol
0xDEnYO: done

src/Facets/PolymerCCTPFacet.sol
0xDEnYO: We cannot do that. If we move it to the constructor, the approval will be made from the facet contract to the TokenMessenger (but we need it from the diamond).

docs/PolymerCCTPFacet.md
0xDEnYO: added

docs/PolymerCCTPFacet.md
0xDEnYO: added

docs/PolymerCCTPFacet.md
0xDEnYO: Hm not sure if I can agree here. This function's main purpose (compared to the startBridge... function is to add swap steps and fee collection on src chain).

config/polymercctp.json
0xDEnYO: updated

src/Facets/PolymerCCTPFacet.sol
0xDEnYO: If we swap from native to USDC and then bridge USDC the transaction will have msg.value. If we remove the modifier, this would fail. We need to leave it there to allow such routes (and the refundExcessNative in case of accidental overpays).

test/solidity/Facets/PolymerCCTPFacet.t.sol
```diff
@@ -0,0 +1,531 @@
...
+        vm.expectEmit(true, true, true, true, _facetTestContractAddress);
+        emit PolymerCCTPFacet.PolymerCCTPFeeSent(
+            bridgeData.minAmount,
+            validPolymerData.polymerTokenFee,
+            validPolymerData.minFinalityThreshold
+        );
+
+        ILiFi.BridgeData
+            memory adjustedBridgeData = getBridgeDataWithAdjustedAmount();
+        vm.expectEmit(true, true, true, true, _facetTestContractAddress);
```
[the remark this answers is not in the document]
0xDEnYO: we cannot change the amount that is passed into this event as this is required as-is for off-chain logic to pick up the deposit

src/Facets/PolymerCCTPFacet.sol
```diff
@@ -0,0 +1,320 @@
...
+                _polymerData.nonEVMReceiver
+            );
+        }
+
+        // Emitting a specific event for Polymer CCTP bridges greatly simplifies off-chain relayer filtering, despite it containing redundant information as emitted in the token messenger's depositForBurn event
+        emit PolymerCCTPFeeSent(
+            _bridgeData.minAmount,
+            _polymerData.polymerTokenFee,
+            _polymerData.minFinalityThreshold
+        );
```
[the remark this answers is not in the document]
0xDEnYO: we cannot change the amount that is passed into this event as this is required as-is for off-chain logic to pick up the deposit

src/Facets/PolymerCCTPFacet.sol
RnkSngh: ah; you're totally right! disregard~

src/Facets/PolymerCCTPFacet.sol
RnkSngh: ah gotcha; makes sense, thanks for the explanation ~

docs/PolymerCCTPFacet.md
RnkSngh: You all requested this method for hooks as well. It's not mutually exclusive with using it for a swap step - the suggestion might be helpful context for pointing to where to implement hooks later down the line.

But ultimately it's just a comment; feel free to neglect or implement as you see fit ~

src/Facets/PolymerCCTPFacet.sol
RnkSngh: @0xDEnYO It's worth leaving a note here to check during audit that there's no other method in any facet on your proxy where erc20 approve can be called on USDC (or, more generally, any erc20 address). 

Otherwise there could be a DOS vector where a griefer could re-set this max approve through approving USDC for 0 amount, which would block all subsequent bridge calls ~

docs/PolymerCCTPFacet.md
0xDEnYO: destination hooks/calls (when implemented later) can be used on both functions (startBridge..., swapAndStartBridge..)
So this comment is misleading in the sense that it suggest that one of the functions supports it while the other doesnt. 
We can (and will) also update the docs when we add hooks, so for now I would leave as is and resolve this.

src/Facets/PolymerCCTPFacet.sol
0xDEnYO: very smart comment, thank you. We will look into this during audit.

script/demoScripts/demoPolymerCCTP.ts
```diff
@@ -0,0 +1,398 @@
...
+    : DST_CHAIN_ID
+
+  const { quote: polymerQuote, polymerTokenFee } = await getPolymerQuote(
+    sourceChainId,
+    destinationChainIdPolymer,
+    sendingAssetId,
+    sendingAssetId, // Same token (USDC) on both sides
+    fromAmount,
+    receiverAddress
+  )
```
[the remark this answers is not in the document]
0xDEnYO: check if this is solved now

script/demoScripts/demoPolymerCCTP.ts
```diff
@@ -0,0 +1,398 @@
...
+// Using testnet chain IDs for devnet API
+const DST_CHAIN_ID_EVM =
+  SRC_CHAIN === 'arbitrum'
+    ? LIFI_CHAIN_ID_OPTIMISM_SEPOLIA
+    : LIFI_CHAIN_ID_ARBITRUM_SEPOLIA // Optimism Sepolia or Arbitrum Sepolia
+const DST_CHAIN_ID = BRIDGE_TO_SOLANA ? LIFI_CHAIN_ID_SOLANA : DST_CHAIN_ID_EVM
+
+// Token addresses
+const sendingAssetId =
+  SRC_CHAIN === 'arbitrum' ? ADDRESS_USDC_ARB : ADDRESS_USDC_OPT
```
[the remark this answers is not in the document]
0xDEnYO: check if this is solved now

src/Facets/PolymerCCTPFacet.sol
```diff
@@ -0,0 +1,317 @@
...
+    /// @param _polymerData Data specific to PolymerCCTP
+    /// @notice Requires caller to approve the LifiDiamondProxy of the bridge amount + polymerFee
+    function startBridgeTokensViaPolymerCCTP(
+        ILiFi.BridgeData memory _bridgeData,
+        PolymerCCTPData calldata _polymerData
+    )
+        external
+        payable
+        nonReentrant
+        refundExcessNative(payable(msg.sender))
```
maxklenk: While I see that the swapAndBridge has to accept native, I don't think this is the case for the direct `startBridgeTokensViaPolymerCCTP` function. It should only receive USDC.

src/Facets/PolymerCCTPFacet.sol
```diff
@@ -0,0 +1,317 @@
...
+        // Should only be nonzero if submitting to a nonEVM chain
+        bytes32 nonEVMReceiver;
+        // the minimum finality at which a burn message will be attested to, will be passed directly to tokenMessenger.depositForBurn method.
+        // 1000 = fast path, 2000 = standard path
+        uint32 minFinalityThreshold;
+    }
+
+    /// Errors ///
+
+    error InvalidSendingAsset(address actual, address expected);
```
ezynda3: I think we already have a `InvalidSendingToken` error we can use from `GenericErrors.sol`

script/demoScripts/demoPolymerCCTP.ts
```diff
@@ -0,0 +1,387 @@
+#!/usr/bin/env bunx tsx
+
+import { consola } from 'consola'
+import { config } from 'dotenv'
+import { BigNumber, constants, utils } from 'ethers'
```
ezynda3: Any reason to use `ethers` vs `viem`?

ezynda3 (dismissed):

src/Facets/PolymerCCTPFacet.sol
0xDEnYO: good point, removed

src/Facets/PolymerCCTPFacet.sol
0xDEnYO: the error was actually unused, I removed it. 
Good catch.

script/demoScripts/demoPolymerCCTP.ts
0xDEnYO: updated

audit/auditLog.json
```diff
@@ -440,6 +440,13 @@
...
       "auditReportPath": "./audit/reports/2025.11.19_LiFiIntentEscrowFacet(v1.0.0).pdf",
       "auditCommitHash": "4e71699761d075f43f5d9cd66c4e1be082fa08ee"
+    },
+    "audit20251201": {
+      "auditCompletedOn": "01.12.2025",
+      "auditedBy": "Sujith Somraaj (individual security researcher)",
+      "auditorGitHandle": "sujithsomraaj",
+      "auditReportPath": "./audit/reports/2025.12.01_PolymerCCTPFacet(v1.0.0).pdf",
+      "auditCommitHash": "605f7e14af2dd96fe8f50de253c51c7979245584"
     }
```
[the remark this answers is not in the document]
0xDEnYO: fixed

ezynda3 (approved):

script/demoScripts/demoPolymerCCTP.ts
```diff
@@ -0,0 +1,485 @@
...
+    }`
+  )
+  consola.info(`⚡ Mode: ${USE_FAST_MODE ? 'Fast' : 'Standard'}`)
+  consola.info('')
+
+  // Polymer API uses its own chain ID mapping (Solana = 2, not LiFi's 1151111081099710)
+  // Polymer API uses its own chain ID mapping (Solana = 2, not LiFi's 1151111081099710)
+  const destinationChainIdPolymer = BRIDGE_TO_SOLANA
+    ? POLYMER_CHAIN_ID_SOLANA
+    : DST_CHAIN_ID
```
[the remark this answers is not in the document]
0xDEnYO: removed

PR #1461, opened 2025-11-12
Grant access to deployer on migrate

ezynda3:
- Changed escaped double quotes to single quotes with double quotes inside
- Fixes string literal quote style to comply with solhint rules


src/Facets/WhitelistManagerFacet.sol
```diff
@@ -211,6 +212,14 @@ contract WhitelistManagerFacet is IWhitelistManagerFacet {
 
         // Mark as migrated
         als.migrated = true;
+
```
0xDEnYO: adding this in the contract would mean we need to reaudit.
Can we find a way to avoid this?
I already deployed a bunch of the contracts. If this is the only spot we can add this then I would maybe say we should rather run a script manually on the chains as it's less effort than reaudit and redeployment.
If I am not mistaken we should be able to add this logic in the Update.. script, no?
e.g. if we call the `approveDeployerWallet` function after the migration?

The following is AI-generated, please review carefully:
```solidity
contract DeployScript is UpdateScriptBase {
    using stdJson for string;

    function run() public returns (address[] memory facets, bytes memory cutData) {
        (facets, cutData) = update("WhitelistManagerFacet");

        // After diamond cut succeeds, grant deployer wallet access control permissions
        // This allows the deployer wallet to call batchSetContractSelectorWhitelist
        if (!noBroadcast && cut.length > 0) {
            approveDeployerWallet();
        }

        return (facets, cutData);
    }
```

0xDEnYO: sth like this (AI-generated, pls check thoroughly)

```solidity
  // --- 3. Parse Whitelist ---
        (address[] memory contracts, bytes4[][] memory selectors) = _parseWhitelistJson(whitelistJson);

        // --- 4. Add WhitelistManagerFacet's own functions to whitelist ---
        (contracts, selectors) = _addFacetToWhitelist(contracts, selectors);

        // --- 5. ABI-Encode Call Data ---
        bytes memory callData =
            abi.encodeWithSelector(WhitelistManagerFacet.migrate.selector, selectorsToRemove, contracts, selectors);
        vm.resumeGasMetering();
        return callData;
    }

    /// @notice Adds WhitelistManagerFacet's own functions to the whitelist arrays.
    /// @dev This ensures batchSetContractSelectorWhitelist is whitelisted during migration.
    ///      Note: The update script must be executed by the diamond owner (not the deployer wallet
    ///      if ownership has been transferred to a Safe/Timelock in production).
    function _addFacetToWhitelist(address[] memory contracts, bytes4[][] memory selectors)
        internal
        returns (address[] memory, bytes4[][] memory)
    {
        // Get WhitelistManagerFacet address from deployment JSON
        address whitelistManagerFacet = json.readAddress(".WhitelistManagerFacet");

        // Create new arrays with one extra slot
        address[] memory contractsWithFacet = new address[](contracts.length + 1);
        bytes4[][] memory selectorsWithFacet = new bytes4[][](contracts.length + 1);

        // Copy existing contracts and selectors
        for (uint256 i = 0; i < contracts.length; i++) {
            contractsWithFacet[i] = contracts[i];
            selectorsWithFacet[i] = selectors[i];
        }

        // Add WhitelistManagerFacet's batchSetContractSelectorWhitelist function
        contractsWithFacet[contracts.length] = whitelistManagerFacet;
        bytes4[] memory facetSelectors = new bytes4[](1);
        facetSelectors[0] = WhitelistManagerFacet.batchSetContractSelectorWhitelist.selector;
        selectorsWithFacet[contracts.length] = facetSelectors;

        return (contractsWithFacet, selectorsWithFacet);
    }

```

ezynda3: So what we can do is create a throwaway migrate contract. It has the same logic that would be in `migrate()` but is just deployed and never used again. Would live in the deploy scripts dir.

script/deploy/facets/UpdateWhitelistManagerFacet.s.sol
```diff
@@ -4,17 +4,29 @@ pragma solidity ^0.8.17;
...
     function run()
         public
         returns (address[] memory facets, bytes memory cutData)
     {
-        return update("WhitelistManagerFacet");
+        // Deploy the migration contract
+        vm.startBroadcast(deployerPrivateKey);
+        migrateContract = address(new MigrateWhitelistManager());
+        vm.stopBroadcast();
+
```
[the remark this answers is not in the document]
ezynda3: We only do this in the case of building calldata for SAFE transactions. This contract needs to be deployed regardless.

PR #1508, opened 2025-12-03
run update periphery before synching whitelist

0xDEnYO:
This makes sure that the whitelist party maintained by the SC team are always up to date when running the sync whitelist command

### DIFF
--- a/script/tasks/diamondSyncWhitelist.sh
+++ b/script/tasks/diamondSyncWhitelist.sh
@@ -13,7 +13,7 @@ function diamondSyncWhitelist {
   # Update whitelist periphery and composer entries before syncing
   echo ""
   echo "[info] Updating whitelist periphery and composer entries..."
-  bunx tsx ./script/tasks/updateWhitelistPeriphery.ts || checkFailure $? "update whitelist periphery"
+  bunx tsx script/tasks/updateWhitelistPeriphery.ts || checkFailure $? "update whitelist periphery"
   echo "[info] Whitelist periphery update completed"
   echo ""
 


script/tasks/diamondSyncWhitelist.sh
```diff
@@ -10,6 +10,13 @@ function diamondSyncWhitelist {
   # Load configuration & helper functions
   source script/helperFunctions.sh
 
+  # Update whitelist periphery and composer entries before syncing
+  echo ""
+  echo "[info] Updating whitelist periphery and composer entries..."
+  bunx tsx ./script/tasks/updateWhitelistPeriphery.ts || checkFailure $? "update whitelist periphery"
```
mirooon: I would put it below `ENVIRONMENT` variable definition and pass `ENVIRONMENT` param to `updateWhitelistPeriphery.ts `. Like we do in `deployAllContracts.sh` script

0xDEnYO: @mirooon 
if we do not pass a parameter, then both whitelist files get updated. 
I feel this is the correct and desired handling here, no?

(in the back of my mind there is also the point that we should soon update the whitelist on staging everytime we update on production so that we keep the staging setup in sync as Victor ran into problems there several times already)

script/tasks/diamondSyncWhitelist.sh
```diff
@@ -10,6 +10,13 @@ function diamondSyncWhitelist {
...
+  echo "[info] Whitelist periphery update completed"
+  echo ""
```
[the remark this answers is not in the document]
0xDEnYO: updated

PR #1517, opened 2025-12-05
enhance Patcher contract with SafeTransferLib integration (fix the issues caused by relying on the custom approve function) [Patcher v1.0.1]

mirooon:

some tokens do not follow the standard ERC-20 return value semantics. Specifically, their approve functions return no data, while the IERC interface that Patcher is using solidity compiler assume a bool return value. Tthe compiler inserts an implicit check that the call returns at least 32 bytes. For tokens like USDT, this results in a revert even though the approval itself would normally succeed.

To address this, I integrated SafeTransferLib. Its approval logic explicitly handles tokens return bool or empty data. This guarantees compatibility with nonstandard tokens


src/Periphery/Patcher.sol
```diff
@@ -254,6 +257,13 @@ contract Patcher {
 
     /// Private Methods ///
 
+    /// @notice Helper function to reset the approval for a token
+    /// @param token The ERC20 token to reset the approval for
+    /// @param target The contract to reset the approval for
+    function _resetApprove(address token, address target) private {
```
0xDEnYO, line 263: a helper function with one line of code seems a little over-engineered to me as you could also just use `token.safeApprove(..)` directly. Why did you decide to go for a helper function?

mirooon: good question. When I used `token.safeApprove(..)` as you suggested I was getting stack too deep errors so I decided to go with this helper function

0xDEnYO: ok, then let's just leave it there. 
(generally: it helps the reviewer to add this info in a comment)

PR #1525, opened 2025-12-10
Update mongo scripts

ezynda3:


script/deploy/shared/mongo-log-utils.ts
```diff
@@ -28,14 +28,16 @@ export interface IDeploymentRecord {
...
-  /** Solidity compiler version used during deployment */
-  solcVersion?: string
-  /** EVM version used during deployment */
-  evmVersion?: string
+  /** Solidity compiler version used during deployment (empty string if not specified) */
+  solcVersion: string
+  /** EVM version used during deployment (empty string if not specified) */
+  evmVersion: string
+  /** ZK Solidity compiler version for zkSync deployments (empty string if not specified) */
+  zkSolcVersion: string
```
0xDEnYO, line 40: should we make this optional (otherwise we need to write empty values for all non-zkEVM networks which creates a lot of no-op data)?

script/deploy/shared/mongo-log-utils.ts
```diff
@@ -460,6 +378,29 @@ export function createDeploymentKey(record: IDeploymentRecord): string {
...
 
+/**
+ * Raw deployment data structure from JSON files
+ */
+export interface IRawDeploymentData {
+  ADDRESS: string
+  OPTIMIZER_RUNS: string
+  TIMESTAMP: string
+  CONSTRUCTOR_ARGS: string
+  SALT?: string
```
0xDEnYO: here it's still optional while here not anymore:
https://github.com/lifinance/contracts/pull/1525/files#diff-98223741c7a624b8e6f24bafc6b718918f20c424292958397e0bc8de2d76467eR32

script/deploy/shared/mongo-log-utils.ts
```diff
@@ -460,6 +378,29 @@ export function createDeploymentKey(record: IDeploymentRecord): string {
...
+  VERIFIED: string
+  SOLC_VERSION?: string
```
0xDEnYO: why are these two optional?

script/deploy/shared/mongo-log-utils.ts
```diff
@@ -487,6 +428,9 @@ export class RecordTransformer {
         CONSTRUCTOR_ARGS: string
         VERIFIED: string
         SALT?: string
+        SOLC_VERSION?: string
```
0xDEnYO, line 495: here all three are optional. I dont understand why the optionality varies

script/deploy/shared/mongo-log-utils.ts
```diff
@@ -498,8 +442,12 @@ export class RecordTransformer {
           optimizerRuns: validData.OPTIMIZER_RUNS,
           timestamp: new Date(validData.TIMESTAMP),
           constructorArgs: validData.CONSTRUCTOR_ARGS,
-          salt: validData.SALT || undefined,
+          // JSON is source of truth - all optional fields default to empty string
+          salt: validData.SALT || '',
           verified: validData.VERIFIED === 'true',
+          solcVersion: validData.SOLC_VERSION || '',
```
0xDEnYO, line 512: is there a check that ensures that we do have all required values for both cases (evm/zkevm/tron?)

so in EVM we want solc and evm version and in zkevm we want all three (or am I wrong?). Example:
https://abscan.org/address/0x9BacA4D8aB4328D167C4503a6c2951c70175CaA9#code (all three values exist here (and are required for manual verification)

script/deploy/shared/deployment-cache.ts
```diff
@@ -0,0 +1,411 @@
...
+    }
+  }
+
+  /**
+   * Gets the file path for the cache data file
+   * @param environment - The deployment environment
+   * @returns Full path to the cache data file
+   * @private
+   */
+  private getCacheFilePath(environment: 'staging' | 'production'): string {
```
0xDEnYO: introduce a type for `'staging' | 'production'` that can be reused in all scripts / mongoDB stuff

script/deploy/validate-deployment-logs.ts
```diff
@@ -0,0 +1,549 @@
...
+  } catch {
+    return defaultValue
+  }
+}
+
+/**
+ * Configuration for validation
+ */
+const config: IConfig = {
+  mongoUri: getEnvVarWithDefault('MONGODB_URI', 'mongodb://localhost:27017'),
```
0xDEnYO: why do we default to a locally running mongoDB? we do not have this set up as far as I am aware?

ezynda3: removed

script/deploy/shared/deployment-cache.ts
ezynda3: Introduced a new type in 874b2566

script/deploy/shared/mongo-log-utils.ts
ezynda3: Added checks in 874b2566

script/deploy/shared/mongo-log-utils.ts
ezynda3: Made all values required everywhere for consistency. Any missing values must be explicitly `""` 874b2566

script/deploy/shared/mongo-log-utils.ts
ezynda3: Made all values required everywhere for consistency

script/deploy/shared/mongo-log-utils.ts
ezynda3: Made all values required everywhere for consistency

script/deploy/shared/mongo-log-utils.ts
ezynda3: Made all values required everywhere for consistency. This matches with with the original JSON and makes it easier to normalize in my opinion. Mongo simply removes the the value from the struct if its not set which can cause headaches.

script/deploy/validate-deployment-logs.ts
```diff
@@ -0,0 +1,536 @@
...
+    } finally {
+      await validator.disconnect()
+    }
+  },
+})
+
+// Run the CLI
+runMain(validateCommand)
+
+export { DeploymentValidator, type IValidationResults }
```
mirooon, line 536: do we want remove function ENABLE_MONGODB_LOGGING and write to mongo by default?

script/deploy/shared/mongo-log-utils.ts
```diff
@@ -460,6 +435,36 @@ export function createDeploymentKey(record: IDeploymentRecord): string {
   return `${record.contractName}-${record.network}-${record.version}-${record.address}`
 }
 
+/**
+ * Deployment environment type - used throughout the codebase
+ */
+export type DeploymentEnvironment = 'staging' | 'production'
```
mirooon, line 441: what about `EnvironmentEnum` from `common/types.ts`?

script/deploy/validate-deployment-logs.ts
mirooon: if we would have mongodb loggin always enabled then we could also check MONGODB_URI env var if exist at the beginning of scriptMaster?

script/deploy/shared/deployment-logger.ts
```diff
@@ -0,0 +1,426 @@
+/**
+ * Deployment Logger - MongoDB-First Logging System
+ *
+ * This module provides a unified interface for logging contract deployments
+ * with MongoDB as the primary source of truth. It handles:
+ * - Writing deployment records to MongoDB
+ * - Cache invalidation for fresh reads
+ * - Optional local JSON file updates (for gradual migration)
```
mirooon, line 8: do we have any database backups? Losing the db would be a huge mess. I know it’s an edge case, but I would prefer keeping local JSON file updates mandatory rather than optional

Also, shouldn’t the deployments folder be tracked in this repo instead of being listed in .gitignore?

ezynda3: The DB is literally a copy of `_deployments_log_file.json` so in that sense we have a backup.

script/deploy/validate-deployment-logs.ts
ezynda3: This will be removed in the final migration PR https://github.com/lifinance/contracts/pull/1539 I want to merge this PR first

script/deploy/shared/mongo-log-utils.ts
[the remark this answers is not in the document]
ezynda3: We will backfill using the update script for sure

script/deploy/shared/deployment-logger.ts
ezynda3: Only the local .cache is gitignored. This is intended.

script/deploy/shared/mongo-log-utils.ts
ezynda3: Updated here https://github.com/lifinance/contracts/pull/1539/commits/dd2375b69c12835c942c922e329f2074d46a6e83

script/deploy/shared/mongo-log-utils.ts
mirooon: for zkevm we do not need evm version right? we don't specify it as we use zksolc @ezynda3 @0xDEnYO
I am commenting just to double check. In updated code I don't see validation for evm version for zkevm and that's good

mirooon: I am good to approve it

ezynda3: @mirooon correct

PR #1531, opened 2025-12-11
add converter parameter to TokenWrapper for WETH9-compatible interfac… [TokenWrapper v1.2.0]

ezynda3:
…e support


archive/config/tokenwrapper.json
```diff
@@ -1,48 +1,183 @@
 {
   "comment": "Note: this file is generated by a backend script, do not change it manually!",
```
0xDEnYO, line 2: is this comment really true? I doubt it

archive/config/tokenwrapper.json
```diff
@@ -1,48 +1,183 @@
...
+  },
+  "abstract": {
+    "wrappedToken": "0x3439153EB7AF838Ad19d56E1571FBD09333C2809",
+    "converter": "0x0000000000000000000000000000000000000000"
+  },
+  "berachain": {
+    "wrappedToken": "0x6969696969696969696969696969696969696969",
+    "converter": "0x0000000000000000000000000000000000000000"
+  },
+  "optimism": {
```
0xDEnYO: please sort alphabetically as in other config files, too

archive/config/tokenwrapper.json
```diff
@@ -1,48 +1,183 @@
...
+  },
+  "swellchain": {
+    "wrappedToken": "0x4200000000000000000000000000000000000006",
+    "converter": "0x0000000000000000000000000000000000000000"
+  },
+  "aurora": {
+    "wrappedToken": "0xc9bdeed33cd01541e1eed10f90519d2c06fe3feb",
+    "converter": "0x0000000000000000000000000000000000000000"
+  },
+  "harmony": {
```
0xDEnYO: nitpick: could be removed entirely

script/deploy/facets/DeployTokenWrapper.s.sol
```diff
@@ -29,6 +29,19 @@ contract DeployScript is DeployScriptBase {
             string.concat(".", network, ".wrappedNativeAddress")
         );
 
+        // try to get converter address, default to address(0) if not found
```
0xDEnYO: this means if there is an error in the JSON we deploy with address(0) despite wanting to deploy something with a different address. I dont think this is in our interest. 
Since the config file has a `converterAddress` value for supported network, we dont really need a try/catch here imho.

Instead, we can just read the address from the config file and if it's `!= address(0)` then I believe we should make sure it's a contract address. Wdyt?

script/deploy/resources/deployRequirements.json
```diff
@@ -695,5 +695,24 @@
         "allowToDeployWithZeroAddress": "false"
       }
     }
+  },
+  "TokenWrapper": {
+    "configData": {
+      "_wrappedToken": {
+        "configFileName": "tokenwrapper.json",
+        "keyInConfigFile": ".<NETWORK>.wrappedToken",
+        "allowToDeployWithZeroAddress": "true"
```
0xDEnYO: why would we want to allow to deploy the contract with zero address arguments for `_wrappedToken` or `_owner`? This doesnt make any sense to me

And with regards to the converter: 

Option 1: 
- we require to have a value for _converter , but we allow address(0) values
- we will need a address(0) value for each network in the config file
- we remove the try/catch logic from deploy scripts (since that doesnt match with the intended behaviour)

Option 2 (my preference):
- we do not mention `_converter` in deployRequirements at all
- in config file:
-   we only add converter address for networks where it's non-zero
-   we add a comment explaining converterAddress handling that at the top of the file
- in deploy scripts we apply fallback to address(0) logic

script/deploy/tron/deploy-and-register-periphery.ts
```diff
@@ -802,11 +802,28 @@ async function deployAndRegisterPeripheryImpl(options: {
 
             const refundWalletHex = globalConfig.refundWallet
 
-            const constructorArgs = [wrappedNativeHex, refundWalletHex]
+            // Try to get converter address, default to zero address if not found
```
0xDEnYO: same here, we can just pull from config instead of assigning a default value (since we have an entry for tron in the config)

script/deploy/zksync/DeployTokenWrapper.zksync.s.sol
```diff
@@ -29,6 +29,19 @@ contract DeployScript is DeployScriptBase {
...
 
+        // try to get converter address, default to address(0) if not found
+        address converterAddress;
+        try
+            vm.parseJsonAddress(
+                vm.readFile(path),
+                string.concat(".", network, ".converterAddress")
+            )
+        returns (address addr) {
+            converterAddress = addr;
```
0xDEnYO, line 41: I guess here we dont verify if isContract since it's zkEVM but we could add a comment that explains that

src/Periphery/TokenWrapper.sol
```diff
@@ -18,10 +18,13 @@ interface IWrapper {
 /// @title TokenWrapper
 /// @author LI.FI (https://li.fi)
 /// @notice Provides functionality for wrapping and unwrapping tokens
-/// @custom:version 1.1.0
+/// @custom:version 1.2.0
 contract TokenWrapper is WithdrawablePeriphery {
     uint256 private constant MAX_INT = 2 ** 256 - 1;
```
0xDEnYO: unused, can be removed

src/Periphery/TokenWrapper.sol
```diff
@@ -30,18 +33,37 @@ contract TokenWrapper is WithdrawablePeriphery {
...
         address _wrappedToken,
+        address _converter,
         address _owner
     ) WithdrawablePeriphery(_owner) {
-        wrappedToken = _wrappedToken;
-        IERC20(wrappedToken).approve(address(this), MAX_INT);
+        WRAPPED_TOKEN = _wrappedToken;
+        CONVERTER = _converter;
+        USE_CONVERTER = _converter != address(0);
+        WRAPPER_ADDRESS = USE_CONVERTER ? _converter : _wrappedToken;
```
0xDEnYO: nitpick: feel this variable is unnecessary. We could just use CONVERTER (and in case we do not have a converter we also assign the _wrappedToken address to it). 

the variable CONVERTER is only used once in this whole contract (line 80) and that occurrence could arguably be replaced with WRAPPER_ADDRESS.

src/Periphery/TokenWrapper.sol
```diff
@@ -30,18 +33,37 @@ contract TokenWrapper is WithdrawablePeriphery {
...
     }
 
     /// External Methods ///
 
     /// @notice Wraps the native token
     function deposit() external payable {
-        IWrapper(wrappedToken).deposit{ value: msg.value }();
-        IERC20(wrappedToken).transfer(msg.sender, msg.value);
+        if (USE_CONVERTER) {
+            uint256 balanceBefore = IERC20(WRAPPED_TOKEN).balanceOf(
```
0xDEnYO: I understand the scenario of having different decimals. 
Getting balance before/after is one way to approach this which comes with some caveats:
- we accept any return value without any validation
- we need to do balance checks and calculations

Consider the following approach: having a SWAP_RATIO_MULTIPLIER that we pass into the constructor to become an immutable parameter. 
This will become a hardcoded value (gas-efficient) and it would ensure that we receive what we expect. 
When we deposit we multiply by it, when we withdraw we divide by it.

src/Periphery/TokenWrapper.sol
```diff
@@ -30,17 +31,21 @@
     // solhint-disable-next-line no-empty-blocks
     constructor(
         address _wrappedToken,
+        address _converter,
```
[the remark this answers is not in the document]
0xDEnYO: I do not see the validation of the constructor parameters. Why was this marked as fixed? @ezynda3

src/Periphery/TokenWrapper.sol
```diff
@@ -50,10 +68,22 @@ contract TokenWrapper is WithdrawablePeriphery {
         // to have `wad` equal to the minimum between the balance and the
         // given allowance, in our specific usecase allowance is always
         // nearly MAX_UINT256. Using the balance only is a gas optimisation.
-        uint256 wad = IERC20(wrappedToken).balanceOf(msg.sender);
-        IERC20(wrappedToken).transferFrom(msg.sender, address(this), wad);
-        IWrapper(wrappedToken).withdraw(wad);
-        SafeTransferLib.safeTransferETH(msg.sender, wad);
+        uint256 wad = IERC20(WRAPPED_TOKEN).balanceOf(msg.sender);
+        IERC20(WRAPPED_TOKEN).transferFrom(msg.sender, address(this), wad);
```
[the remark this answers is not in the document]
0xDEnYO: This comment was not resolved. We should use safeTransfer for all token transfers

src/Periphery/TokenWrapper.sol
```diff
@@ -50,10 +72,22 @@ contract TokenWrapper is WithdrawablePeriphery {
...
+
+        if (USE_CONVERTER) {
+            // Approve converter to spend wrappedToken
+            LibAsset.maxApproveERC20(IERC20(WRAPPED_TOKEN), CONVERTER, wad);
```
0xDEnYO: I think this should be moved into the constructor to do it once (and for all) and save gas on all subsequent calls

src/Periphery/TokenWrapper.sol
```diff
@@ -50,10 +72,22 @@ contract TokenWrapper is WithdrawablePeriphery {
         // to have `wad` equal to the minimum between the balance and the
         // given allowance, in our specific usecase allowance is always
         // nearly MAX_UINT256. Using the balance only is a gas optimisation.
-        uint256 wad = IERC20(wrappedToken).balanceOf(msg.sender);
-        IERC20(wrappedToken).transferFrom(msg.sender, address(this), wad);
-        IWrapper(wrappedToken).withdraw(wad);
-        SafeTransferLib.safeTransferETH(msg.sender, wad);
+        uint256 wad = IERC20(WRAPPED_TOKEN).balanceOf(msg.sender);
```
0xDEnYO: nitpick: I really dislike "wad" as a variable name since it's not really self-explanatory. Can we use `amount`?

test/solidity/utils/TestWrappedConverter.sol
```diff
@@ -0,0 +1,44 @@
+// SPDX-License-Identifier: Unlicense
```
0xDEnYO: wrong license

test/solidity/utils/TestConverterWithDecimals.sol
```diff
@@ -0,0 +1,56 @@
+// SPDX-License-Identifier: Unlicense
```
0xDEnYO: wrong license

test/solidity/Periphery/TokenWrapper.t.sol
0xDEnYO, line 1: wrong license

test/solidity/Periphery/TokenWrapper.t.sol
```diff
@@ -51,4 +96,121 @@ contract TokenWrapperTest is DSTest {
         tokenWrapper.withdraw();
         assert(address(this).balance - initialBalance == 1 ether);
     }
+
+    // ========== CONVERTER TESTS ==========
+
+    function testCanDepositWithConverter() public {
```
0xDEnYO: all tests are not named according to our naming convention.....🫩
Please also update existing (untouched) tests to match our conventions.

test/solidity/Periphery/TokenWrapper.t.sol
```diff
@@ -51,4 +96,121 @@ contract TokenWrapperTest is DSTest {
...
+    function testCanDepositWithDecimalConverter() public {
+        // Initial state: user has no 6-decimal tokens
+        assertEq(token6Decimals.balanceOf(address(this)), 0);
+
+        // User deposits 1 ETH (18 decimals)
+        // Should receive 1 USDT (6 decimals) after conversion
+        tokenWrapperWithDecimalConverter.deposit{ value: 1 ether }();
+
+        // User should now have 1 USDT (1e6, not 1e18)
+        assertEq(token6Decimals.balanceOf(address(this)), 1_000_000);
```
0xDEnYO: the 1_000_000 could be a very meaningful constant in this test contract

test/solidity/utils/TestBasicToken.sol
```diff
@@ -0,0 +1,22 @@
+// SPDX-License-Identifier: Unlicense
```
0xDEnYO: wrong license

archive/config/tokenwrapper.json
0xDEnYO, line 1: why is this file in archive folder?

archive/config/tokenwrapper.json
```diff
@@ -1,48 +1,183 @@
...
+  },
+  "scroll": {
+    "wrappedToken": "0x5300000000000000000000000000000000000004",
+    "converter": "0x0000000000000000000000000000000000000000"
+  },
+  "soneium": {
+    "wrappedToken": "0x4200000000000000000000000000000000000006",
+    "converter": "0x0000000000000000000000000000000000000000"
+  },
+  "sonic": {
```
0xDEnYO: seems like the chain which actually has the converter (stable) is missing here?

src/Periphery/TokenWrapper.sol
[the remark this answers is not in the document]
ezynda3: Fixed: 3ffe919e. SafeTransferLib used throughout

test/solidity/utils/TestWrappedConverter.sol
ezynda3: Fixed: 3ffe919e. License changed to LGPL-3.0-only

test/solidity/utils/TestConverterWithDecimals.sol
ezynda3: Fixed: 3ffe919e. License changed to LGPL-3.0-only

test/solidity/Periphery/TokenWrapper.t.sol
ezynda3: Fixed: 3ffe919e. License changed to LGPL-3.0-only

test/solidity/utils/TestBasicToken.sol
ezynda3: Fixed: 3ffe919e. License changed to LGPL-3.0-only

test/solidity/Periphery/TokenWrapper.t.sol
ezynda3: Fixed: 3ffe919e. All tests use test_ prefix

archive/config/tokenwrapper.json
ezynda3: Fixed dcb8a6d1: File deleted, uses networks.json

archive/config/tokenwrapper.json
ezynda3: Fixed dcb8a6d1: File deleted with comment

archive/config/tokenwrapper.json
ezynda3: Fixed dcb8a6d1: Deleted, networks.json already sorted

archive/config/tokenwrapper.json
ezynda3: Fixed dcb8a6d1: File removed as suggested

archive/config/tokenwrapper.json
ezynda3: Fixed dcb8a6d1: Added stable to networks.json

script/deploy/facets/DeployTokenWrapper.s.sol
ezynda3: Fixed dcb8a6d1, d364b8b2: Added converter contract validation

script/deploy/zksync/DeployTokenWrapper.zksync.s.sol
ezynda3: Fixed dcb8a6d1: Added zkEVM limitation comment

script/deploy/resources/deployRequirements.json
ezynda3: Fixed dcb8a6d1: Implemented Option 2 approach

src/Periphery/TokenWrapper.sol
ezynda3: Fixed dcb8a6d1: Removed unused MAX_INT constant

src/Periphery/TokenWrapper.sol
ezynda3: Fixed dcb8a6d1: Removed WRAPPER_ADDRESS, use CONVERTER

src/Periphery/TokenWrapper.sol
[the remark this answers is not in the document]
ezynda3: Fixed dcb8a6d1, 8843234d: Added constructor parameter validation

src/Periphery/TokenWrapper.sol
ezynda3: Fixed dcb8a6d1: Moved approval to constructor

src/Periphery/TokenWrapper.sol
ezynda3: Fixed dcb8a6d1: Renamed wad to amount

test/solidity/Periphery/TokenWrapper.t.sol
ezynda3: Fixed dcb8a6d1: Created ONE_USDT constant

src/Periphery/TokenWrapper.sol
ezynda3: I think we can simplify even further by just fetching wrapped token decimals inside the constructor then calculating the ratio and saving it as immutable. Saves us from having to add and maintain another config var. Thoughts?

config/networks.json
```diff
@@ -1069,6 +1069,26 @@
     "deployedWithSolcVersion": "0.8.17",
     "devNotes": "Contract verification is not working. In contact with the team to find a resolution. "
   },
+  "stable": {
+    "name": "stable",
+    "chainId": 101010,
+    "nativeAddress": "0x0000000000000000000000000000000000000000",
+    "nativeCurrency": "gUSDT",
+    "wrappedNativeAddress": "0x779ded0c9e1022225f8e0630b35a9b54be713736",
+    "converterAddress": "0xded1660192d4d82e7c0b628ba556861edbb5cada",
```
0xDEnYO: if you add this here, we also need to adjust the type in `script/common/types.ts`

However, I am not sure if this is the right place. I feel a config file for tokenWrapper.json (only containing the converter entry for stable) would be the cleaner approach. But that's just an opinion

src/Periphery/TokenWrapper.sol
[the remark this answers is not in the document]
0xDEnYO: we use InvalidConfig() error for this

src/Periphery/TokenWrapper.sol
0xDEnYO: yes I like it

config/networks.json
ezynda3: So we actually archived the tokenwrapper.json file in favor of networks.json (that's why it appeared in the archive folder). Not sure we should bring it back if we don't use like 90% of it anyway...
@mirooon what do you think?

mirooon: @ezynda3 I think it's more tokenWrapper-specific, I would keep it in tokenWrapper.json

0xDEnYO: @ezynda3 to clarify on this: 
I was not suggesting to unarchive the file as-is. 
It is absolutely correct to pull the wrappedNative address from networks.json to avoid duplicate content.
But the tokenWrapper.json file can just contain the converter address (and only for networks for which one exists), so it would currently be only one entry: stable

ezynda3: created a separate config for converter addresses tokenconverters.json

src/Periphery/TokenWrapper.sol
ezynda3: updated to use the method we discussed. bc81b9c1

src/Periphery/TokenWrapper.sol
```diff
@@ -1,59 +1,94 @@
...
     // solhint-disable-next-line no-empty-blocks
     constructor(
         address _wrappedToken,
+        address _converter,
         address _owner
     ) WithdrawablePeriphery(_owner) {
-        wrappedToken = _wrappedToken;
-        IERC20(wrappedToken).approve(address(this), MAX_INT);
+        if (_wrappedToken == address(0)) revert NoTransferToNullAddress();
+        if (_owner == address(0)) revert NoTransferToNullAddress();
```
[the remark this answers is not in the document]
ezynda3: @coderabbitai check and resolve

src/Interfaces/IWrapper.sol
0xDEnYO, line 1: wondering if IWETH would not be a better name for this interface?
Which by the way is also available in our dependencies, so I guess we could remove this file actually. wdyt?

src/Periphery/TokenWrapper.sol
```diff
@@ -1,59 +1,94 @@
...
+        if (USE_CONVERTER) {
+            if (!LibAsset.isContract(_converter)) revert InvalidContract();
+            CONVERTER = _converter;
+            // Approve converter once for all future withdrawals (gas optimization)
+            LibAsset.maxApproveERC20(
+                IERC20(_wrappedToken),
+                _converter,
+                type(uint256).max
+            );
+            // Calculate swap ratio based on decimal difference between native (18) and wrapped token
```
0xDEnYO: the comment doesnt match the code. You are not really calculating a difference between decimals. 
Instead, you are just using the wrappedToken's decimals as the multiplier

src/Periphery/TokenWrapper.sol
```diff
@@ -1,59 +1,94 @@
...
 
-    /// @notice Wraps the native token
+    /// @notice Wraps the native token and transfers wrapped tokens to caller
+    /// @dev If converter is set, uses it to convert native to wrapped tokens using precalculated ratio
+    /// @dev If no converter, wraps native 1:1 and transfers msg.value of wrapped tokens
     function deposit() external payable {
-        IWrapper(wrappedToken).deposit{ value: msg.value }();
-        IERC20(wrappedToken).transfer(msg.sender, msg.value);
+        IWrapper(CONVERTER).deposit{ value: msg.value }();
+        uint256 wrappedAmount = (msg.value * SWAP_RATIO_MULTIPLIER) / 1 ether;
```
0xDEnYO: would recommend to use a speaking constant instead of `1 ether`, e.g.:
 BASE_DENOMINATOR
 
 same below

0xDEnYO: since the case of a converter is a very rare case, I would move this calculation behind `if(USE_CONVERTER)...` so that we dont always have to make this calculation

```solidity
uint256 amount = msg.value;
IWrapper(CONVERTER).deposit{ value: amount }();
if(USE_CONVERTER)
  wrappedAmount = (amount * SWAP_RATIO_MULTIPLIER) / BASE_DENOMINATOR;
SafeTransferLib.safeTransfer(WRAPPED_TOKEN, msg.sender, amount);

```

same in withdraw() function

ezynda3: added

ezynda3: added guard

src/Periphery/TokenWrapper.sol
ezynda3: updated comment

src/Interfaces/IWrapper.sol
ezynda3: Using solady WETH

src/Periphery/TokenWrapper.sol
```diff
@@ -1,59 +1,108 @@
...
-    /// @notice Wraps the native token
+    /// @notice Wraps the native token and transfers wrapped tokens to caller
+    /// @dev If converter is set, uses it to convert native to wrapped tokens using precalculated ratio
+    /// @dev If no converter, wraps native 1:1 and transfers msg.value of wrapped tokens
     function deposit() external payable {
-        IWrapper(wrappedToken).deposit{ value: msg.value }();
-        IERC20(wrappedToken).transfer(msg.sender, msg.value);
+        uint256 amount = msg.value;
+        WETH(payable(CONVERTER)).deposit{ value: amount }();
+        uint256 wrappedAmount;
```
0xDEnYO: we could just use `amount` here instead of introducing another variable and assiging amount to it in else branch. 

Same in withdraw function below.

src/Periphery/TokenWrapper.sol
```diff
@@ -1,59 +1,104 @@
...
+            amount
+        );
+
+        WETH(payable(CONVERTER)).withdraw(amount);
+
+        if (USE_CONVERTER) {
+            amount = (amount * BASE_DENOMINATOR) / SWAP_RATIO_MULTIPLIER;
+        }
+
+        SafeTransferLib.safeTransferETH(msg.sender, amount);
```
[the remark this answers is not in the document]
ezynda3: @coderabbitai please check the tests and see whether or not this is still true...

PR #1534, opened 2025-12-12
improved runtime of pre-commit significantly

0xDEnYO:
Sometimes the pre-commit check was taking forever, especially on larger git diffs. 
This PR introduces some optimizations and parallelizes tasks more efficiently which significantly reduces the runtime of the checks. 

This is vibe-coded, please review carefully.

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -86,27 +86,125 @@ print_status() {
   fi
 }
 
-# Helper function to filter verbose output
+# Helper function to colorize error output
+colorize_output() {
+  local tool=$1
+  local content=$2
+
+  case "$tool" in
+    lint-staged)
+      # Colorize lint-staged output: file paths in cyan, errors in red, warnings in yellow
+      if [ -z "$content" ]; then
+        echo "$content"
+      else
+        # Use $'...' syntax for proper escape sequence interpretation
+        echo "$content" | while IFS= read -r line || [ -n "$line" ]; do
+          # Colorize file paths (lines that are primarily file paths ending in .sol/.ts/.js/.tsx)
+          if echo "$line" | grep -qE '\.(sol|ts|js|tsx)$'; then
+            printf $'\033[1;36m%s\033[0m\n' "$line"
+            continue
+          fi
+
+          # Colorize line numbers (format: number:number)
+          line=$(echo "$line" | sed $'s|\\([0-9][0-9]*:[0-9][0-9]*\\)|\033[33m\\1\033[0m|g')
+
+          # Colorize error keyword (standalone word)
+          line=$(echo "$line" | sed $'s| error | \033[1;31merror\033[0m |g')
+          line=$(echo "$line" | sed $'s|^error |\033[1;31merror\033[0m |')
+          line=$(echo "$line" | sed $'s| error$| \033[1;31merror\033[0m|')
+          line=$(echo "$line" | sed $'s| Error | \033[1;31mError\033[0m |g')
+          line=$(echo "$line" | sed $'s|^Error |\033[1;31mError\033[0m |')
+          line=$(echo "$line" | sed $'s| Error$| \033[1;31mError\033[0m|')
+
+          # Colorize warning keyword
+          line=$(echo "$line" | sed $'s| warning | \033[1;33mwarning\033[0m |g')
+          line=$(echo "$line" | sed $'s|^warning |\033[1;33mwarning\033[0m |')
+          line=$(echo "$line" | sed $'s| warning$| \033[1;33mwarning\033[0m|')
+          line=$(echo "$line" | sed $'s| Warning | \033[1;33mWarning\033[0m |g')
+          line=$(echo "$line" | sed $'s|^Warning |\033[1;33mWarning\033[0m |')
+          line=$(echo "$line" | sed $'s| Warning$| \033[1;33mWarning\033[0m|')
+
+          # Colorize problem/problems
+          line=$(echo "$line" | sed $'s|problem|\033[1;31m&\033[0m|g')
+
+          # Colorize symbols
+          line=$(echo "$line" | sed $'s|✖|\033[1;31m✖\033[0m|g')
+          line=$(echo "$line" | sed $'s|⚠|\033[1;33m⚠\033[0m|g')
+
+          printf '%s\n' "$line"
+        done
+      fi
+      ;;
+    forge)
+      # Colorize forge output: file paths in cyan, errors in red
+      if [ -z "$content" ]; then
+        echo "$content"
+      else
+        echo "$content" | sed \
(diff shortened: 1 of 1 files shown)


.husky/pre-commit
```diff
@@ -160,114 +202,201 @@ doesFileContainDotEnvSecret() {
...
+  print_status "success" "$TYPECHAIN_SUMMARY"
+fi
 
-    echo ""
-    WARNING="Warning: Secret value(s) from .env found. This code cannot be committed."
-    printf '\033[36m%s\033[0m\n' "$WARNING"
-    printAdvise "abort"
-    echo ""
+# TypeScript compilation check (only if both TS/JS files changed AND typechain was regenerated)
+if [ "$HAS_TS_JS_FILES" = "yes" ] && [ "$HAS_SOL_FILES" = "yes" ]; then
```
mirooon, line 508: I think here we don't have to have condition [ "$HAS_SOL_FILES" = "yes" ] as we only need to run tsc files on ts/js files

.husky/pre-commit
```diff
@@ -160,114 +202,201 @@ doesFileContainDotEnvSecret() {
...
+  if ! wait_and_check "$LINT_STAGED_PID" "$LINT_STAGED_EXIT" "$LINT_STAGED_OUTPUT" "Lint-staged failed. Aborting commit." "lint-staged"; then
+    exit 1
   fi
+  print_status "success" "Lint-staged completed"
+fi
 
-  exit 0
-}
+print_status "info" "Checking for secrets and private keys..."
+run_secret_checks
```
mirooon: I would run it before all heavy work

.husky/pre-commit
```diff
@@ -160,114 +202,201 @@ doesFileContainDotEnvSecret() {
...
+  (forge build --skip test > "$FORGE_OUTPUT" 2>&1; echo $? > "$FORGE_EXIT") &
+  FORGE_PID=$!
+else
+  print_status "skip" "Skipping forge build (no Solidity files changed)"
+  echo 0 > "$FORGE_EXIT"
+fi
 
-  # process all files in git diff and search for secrets
-  local RESULT=$(processGitDiff)
+if [ "$HAS_TYPE_FILES" = "yes" ]; then
```
mirooon: I’m not sure we gain much from this and it adds extra complexity. lint-staged is already scoped in package.json to the file extensions it should scan. If we introduce new linting rules for additional extensions, we’d have to remember to update both package.json and the pre-commit script

.husky/pre-commit
```diff
@@ -1,74 +1,47 @@
...
+HAS_TS_JS_FILES="no"
+HAS_TYPE_FILES="no"
 
-if [ $? -ne 0 ]; then
-  printf '\n%s\n' "TypeScript compilation failed. This may indicate outdated TypeChain types."
-  printf '%s\n\n' "Please run 'bun typechain' manually and fix any type errors before committing."
-  exit 1
-fi
+echo "$STAGED_FILES" | grep -qE '\.sol$' && HAS_SOL_FILES="yes" && HAS_TYPE_FILES="yes"
+echo "$STAGED_FILES" | grep -qE '\.(ts|js|tsx)$' && HAS_TS_JS_FILES="yes" && HAS_TYPE_FILES="yes"
```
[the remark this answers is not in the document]
0xDEnYO: updated

.husky/pre-commit
0xDEnYO: good point, adjusted

.husky/pre-commit
0xDEnYO: yep, good catch

.husky/pre-commit
0xDEnYO: fixed

.husky/pre-commit
```diff
@@ -1,74 +1,47 @@
...
-echo " > logs will only be shown once task is completed"
-echo ""
+FORGE_OUTPUT="$TEMP_DIR/forge.out"
+FORGE_EXIT="$TEMP_DIR/forge.exit"
+TYPECHAIN_OUTPUT="$TEMP_DIR/typechain.out"
+TYPECHAIN_EXIT="$TEMP_DIR/typechain.exit"
+TSC_OUTPUT="$TEMP_DIR/tsc.out"
+TSC_EXIT="$TEMP_DIR/tsc.exit"
+LINT_STAGED_OUTPUT="$TEMP_DIR/lint-staged.out"
+LINT_STAGED_EXIT="$TEMP_DIR/lint-staged.exit"
```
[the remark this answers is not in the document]
0xDEnYO: fallback added

script/playgroundHelpers.sh
```diff
@@ -719,15 +719,14 @@ function analyzeFailingTx() {
...
 
   local NETWORK="$1"
   local TX_HASH="$2"
 
   # Validate required parameters
   if [[ -z "$NETWORK" || -z "$TX_HASH" ]]; then
     error "Usage: analyzeFailingTx NETWORK TX_HASH"
-    error "Example: analyzeFailingTx arbitrum 0xedc3d7580e0b333f7c232649b0506aa3e811b0f5060d84e75a91b0dec68b4cc9"
     return 1
   fi
```
mirooon, line 732: I did some testing and noticed that for example when I put a blankline between parameters in sol file the old script treats the commit as empty and blocks it. new script also blocks the commit in this case but it doesn’t provide any error message explaining why the commit is rejected. please take a look

```
miron@MacBook-Pro-Micha contracts % git commit -m "test"
  Checking for secrets and private keys...
✓ Secret check passed

━━━ Pre-commit Checks ━━━
  Building contracts with forge...
  Running lint-staged (formatting & linting)...
✓ Forge build completed
  Generating TypeChain types...
✓ Successfully generated 1145 typings!
[STARTED] bash -c 'if [ "$#" -gt 0 ]; then output=$(bunx solhint --noPrompt "$@" 2>&1); ret=$?; if echo "$output" | grep -q "No files to lint!"; then exit 0; else echo "$output"; exit $ret; fi; else exit 0; fi # If no lintable files, exit silently to avoid commit errors' _
[COMPLETED] bash -c 'if [ "$#" -gt 0 ]; then output=$(bunx solhint --noPrompt "$@" 2>&1); ret=$?; if echo "$output" | grep -q "No files to lint!"; then exit 0; else echo "$output"; exit $ret; fi; else exit 0; fi # If no lintable files, exit silently to avoid commit errors' _
[STARTED] Reverting to original state because of errors...
[COMPLETED] Reverting to original state because of errors...
error: "lint-staged" exited with code 1

Lint-staged failed. Aborting commit.

husky - pre-commit hook exited with code 1 (error)
```

I don't know how it's with different errors but having a reason why it was rejected would be good

0xDEnYO: added, please check

mirooon (approved):

kalote (approved):

PR #1539, opened 2025-12-17
refactor: migrate deployment logging from JSON to MongoDB-only

ezynda3:

MongoDB migration for deployment logging system.

This PR completes the migration from dual JSON/MongoDB logging to MongoDB-only for the master deployment log file (`_deployments_log_file.json`). MongoDB is already populated and validated, so this removes the redundant JSON operations.

## Changes

- **`logContractDeploymentInfo()`**: Removed JSON file operations and lock file mechanism. Now writes directly to MongoDB only. Deployments fail if MongoDB write fails (no silent continue).

- **Query functions updated to MongoDB-only** (removed JSON fallbacks):
  - `findContractInMasterLog()`
  - `findContractInMasterLogByAddress()`
  - `getContractVersionFromMasterLog()`
  - `getHighestDeployedContractVersionFromMasterLog()`
  - `getConstructorArgsFromMasterLog()`

- **`isMongoLoggingEnabled()`**: Simplified to only check if `MONGODB_URI` is set (removed `ENABLE_MONGODB_LOGGING` check).

- **`cleanupStaleLocks()`**: Removed master log lock file from cleanup (MongoDB handles concurrency).

- **`.env.example`**: Removed `ENABLE_MONGODB_LOGGING` variable, updated comments.

- **Removed dead code**: `logContractDeploymentInfo_BACKUP` function.

## What's NOT changing

- Network-specific JSON files (`.json`, `.diamond.json`, `.staging.` variants) remain unchanged
- TypeScript MongoDB tools (`update-deployment-logs.ts`, `query-deployment-logs.ts`) unchanged
- `_deployments_log_file.json` remains for reference but won't be updated

## Testing

- All shell scripts pass syntax check (`bash -n`)
- All query functions tested against live MongoDB data
- Error handling verified (missing `MONGODB_URI`, non-existent contracts)

> Note: This PR only modifies shell scripts and environment configuration. No smart contract changes.


script/helperFunctions.sh
```diff
@@ -320,131 +195,70 @@ function findContractInMasterLog() {
...
+      local JSON_ENTRY="{\"$ADDRESS\": {\"Name\": \"$CONTRACT_NAME\", \"Version\": \"$VERSION\"}}"
+      echo "$JSON_ENTRY"
+      return 0
+    fi
+  fi
 
-  echo "[info] address not found"
+  echo "[info] address not found in MongoDB"
   exit 1
 }
```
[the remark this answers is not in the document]
ezynda3: In this case we want to exit because we need these vars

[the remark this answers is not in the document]
ezynda3: But I think you may be right because the calling scripts will handle the errors and exit right?

script/deploy/shared/deployment-logger.ts
```diff
@@ -0,0 +1,426 @@
...
+          $set: {
+            ...record,
+            updatedAt: now,
+          },
+          $setOnInsert: {
+            createdAt: now,
+          },
+        },
+        { upsert: true }
+      )
```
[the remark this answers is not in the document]
ezynda3: @coderabbitai check and resolve please

script/deploy/shared/deployment-cache.ts
```diff
@@ -0,0 +1,415 @@
...
+   * Automatically refreshes cache if stale
+   *
+   * @param environment - The deployment environment
+   * @param options - Options for cache behavior
+   * @returns Array of deployment records
+   *
+   * @example
+   * ```typescript
+   * const cache = new DeploymentCache(config)
+   * const records = await cache.get('production')
```
mirooon, line 219: do we call somewhere cache.get? I see it's not used anywhere.
Morover to we need this type of caching. I think reads and writes to database are not frequent

script/deploy/query-deployment-logs.ts
```diff
@@ -10,35 +10,15 @@
...
-  databaseName: string
-}
+import {
+  type IDeploymentRecord,
+  type IConfig,
+  ValidationUtils,
+} from './shared/mongo-log-utils'
 
 const config: IConfig = {
   mongoUri: process.env.MONGODB_URI || 'mongodb://localhost:27017',
```
mirooon: getEnvVar('MONGODB_URI') ?

script/helperFunctions.sh
```diff
@@ -320,256 +195,149 @@ function findContractInMasterLog() {
...
+    local VERSION=$(echo "$MONGO_RESULT" | jq -r '.version')
+    local CONTRACT_NAME=$(echo "$MONGO_RESULT" | jq -r '.contractName')
 
-      # extract address
-      ADDRESS=$(echo "$ENTRY" | awk -F'"' '/"ADDRESS":/{print $4}')
+    # Handle CelerIMFacet variants
+    local EXPECTED_CONTRACT="$CONTRACT"
+    if [[ "$CONTRACT" == *"CelerIMFacet"* ]]; then
+      EXPECTED_CONTRACT="CelerIMFacet"
+    fi
```
mirooon: I think we can remove it

script/deploy/validate-deployment-logs.ts
```diff
@@ -0,0 +1,523 @@
...
+ */
+const config: IConfig = {
+  mongoUri: getEnvVar('MONGODB_URI'), // Required - no default value
+  batchSize: 100,
+  databaseName: 'contract-deployments',
+}
+
+const logFilePath = path.join(
+  process.cwd(),
+  'deployments/_deployments_log_file.json'
```
mirooon, line 48: I still don’t fully understand how we’re supposed to manage the logs. MongoDB will be our source of truth, while _deployments_log_file.json will serve as a local copy. These two need to stay as closely in sync as possible, which means we’ll need a clear strategy for keeping _deployments_log_file.json aligned with MongoDB.

When exactly do we update _deployments_log_file.json, and what role does the .cache directory play in this process?

For now we have that 
1. during deployment the logContractDeploymentInfo() is called
2. mongodb updated automatically -> update-deployment-logs.ts add writes to MongoDB
3. JSON file not updated  -> _deployments_log_file.json is not written to (so what's the purpose of validate-deployment-logs.ts?)
4. .cache directory as far as I see is never used

Regarding sync direction I see only JSON -> MongoDB only but I dont see MongoDB -> JSON. I think it's needed to fetch the latest state from mongo no? I still don't catch the purpose. Could you elaborate. Maybe we could have some session or documentation as you mentioned?

ezynda3: We will not be writing to the deployment_log after this at all. The only purpose of this script for now is to push the most recent version of json log to mongo before we stop writing to it altogether. We will eventually not even need this script.

ezynda3: This will be removed later. We only need it to make mongo reflect exactly what is in the json

script/helperFunctions.sh
ezynda3: removed

script/deploy/query-deployment-logs.ts
ezynda3: fixed

script/deploy/shared/deployment-cache.ts
ezynda3: updated read operations to use cache by default

PR #1551, opened 2025-12-22
SMAR-76 - Finalize WhitelistManager migration [WhitelistManagerFacet v1.1.0, SwapperV2 v1.2.0, IWhitelistManagerFacet v1.1.0]

mirooon:
[SMAR-76]

[SMAR-76]: https://lifi.atlassian.net/browse/SMAR-76?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Helpers/SwapperV2.sol
```diff
@@ -189,23 +194,34 @@ contract SwapperV2 is ILiFi {
...
+            address callTo = currentSwap.callTo;
+            address approveTo = currentSwap.approveTo;
 
             if (
-                !((LibAsset.isNativeAsset(currentSwap.sendingAssetId) ||
-                    LibAllowList.contractIsAllowed(currentSwap.approveTo)) &&
-                    LibAllowList.contractIsAllowed(currentSwap.callTo) &&
-                    LibAllowList.selectorIsAllowed(
-                        bytes4(currentSwap.callData[:4])
+                !LibAllowList.contractSelectorIsAllowed(
```
mirooon, line 205: @0xDEnYO from ticket: _"update SwapperV2._executeSwaps to use WhitelistManagerFacet (and contract:selector)"_

imo direct usage of LibAllowList is more gas efficient. Is this what you had in mind?

src/Facets/WhitelistManagerFacet.sol
```diff
@@ -10,7 +10,7 @@ import { CannotAuthoriseSelf, InvalidConfig } from "../Errors/GenericErrors.sol"
 /// @title WhitelistManagerFacet
 /// @author LI.FI (https://li.fi)
 /// @notice Facet contract for managing whitelisted addresses for various protocol interactions.
-/// @custom:version 1.0.0
+/// @custom:version 2.0.0
```
0xDEnYO: you are just removing a function, no breaking change. Hence I would argue this is v1.1.0

src/Helpers/SwapperV2.sol
0xDEnYO: yes

src/Facets/WhitelistManagerFacet.sol
mirooon: imo it's a breaking change as here we remove migrate function and getters function so if we call from script for example migrate function it will brake so imo it's a breaking change. Even if we were using it for team operations

ezynda3: I would consider it a breaking change but in general could still be expressed as `v1.1.0` imo

PR #1573, opened 2026-01-05
Fix RelayFacet storage [RelayFacet v1.0.2]

0xDEnYO:
https://lifi.atlassian.net/browse/SMAR-119

### DIFF
--- a/src/Facets/RelayFacet.sol
+++ b/src/Facets/RelayFacet.sol
@@ -138,7 +138,9 @@ contract RelayFacet is
     /// @notice Returns whether a requestId has already been consumed (replay protection)
     /// @param requestId Relay API request ID
     /// @return True if already consumed
-    function consumedIds(bytes32 requestId) external view returns (bool) {
+    function isRequestIdConsumed(
+        bytes32 requestId
+    ) external view returns (bool) {
         return
             getStorage().consumedIds[requestId] ||
             DEPRECATED_consumedIds[requestId];
@@ -203,6 +205,10 @@ contract RelayFacet is
         ILiFi.BridgeData memory _bridgeData,
         RelayData calldata _relayData
     ) internal {
+        // mark requestId as consumed
+        Storage storage s = getStorage();
+        s.consumedIds[_relayData.requestId] = true;
+
         // check if sendingAsset is native or ERC20
         if (LibAsset.isNativeAsset(_bridgeData.sendingAssetId)) {
             // Native
@@ -234,8 +240,6 @@ contract RelayFacet is
                 revert(LibUtil.getRevertMsg(reason));
             }
         }
-
-        getStorage().consumedIds[_relayData.requestId] = true;
 
         // Emit special event if bridging to non-EVM chain
         if (_bridgeData.receiver == NON_EVM_ADDRESS) {


src/Facets/RelayFacet.sol
```diff
@@ -107,12 +129,19 @@ contract RelayFacet is
...
     }
 
     /// External Methods ///
 
+    /// @notice Returns whether a requestId has already been consumed (replay protection)
+    /// @param requestId Relay API request ID
+    /// @return True if already consumed
+    function consumedIds(bytes32 requestId) external view returns (bool) {
+        return getStorage().consumedIds[requestId];
+    }
```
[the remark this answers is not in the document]
0xDEnYO: good catch, updated. Please check if resolved

src/Facets/RelayFacet.sol
```diff
@@ -107,12 +129,21 @@ contract RelayFacet is
...
+        RELAY_RECEIVER = _relayReceiver;
+        RELAY_SOLVER = _relaySolver;
     }
 
     /// External Methods ///
 
+    /// @notice Returns whether a requestId has already been consumed (replay protection)
+    /// @param requestId Relay API request ID
+    /// @return True if already consumed
+    function consumedIds(bytes32 requestId) external view returns (bool) {
```
mirooon: wdyt about `isRequestIdConsumed` function name?

src/Facets/RelayFacet.sol
```diff
@@ -61,7 +80,10 @@ contract RelayFacet is
         RelayData calldata _relayData
     ) {
         // Ensure that the id isn't already consumed
-        if (consumedIds[_relayData.requestId]) {
+        if (
```
mirooon, line 83: I assume you wanted to use `consumedIds` function here or remove `consumedIds` function

src/Facets/RelayFacet.sol
```diff
@@ -204,7 +235,7 @@ contract RelayFacet is
             }
         }
 
-        consumedIds[_relayData.requestId] = true;
+        getStorage().consumedIds[_relayData.requestId] = true;
```
mirooon: requestId should be marked before external call to keep CEI pattern

src/Facets/RelayFacet.sol
0xDEnYO: like it, updated

src/Facets/RelayFacet.sol
0xDEnYO: I dont understand your comment to be honest. 
The consumedIds mapping was renamed (which was correctly applied here, too) and we also check the new storage struct if the given requestId is consumed already. 

Could you please elaborate what your review comment aims to point out?

src/Facets/RelayFacet.sol
0xDEnYO: yep, OK. Good point. Updated

src/Facets/RelayFacet.sol
mirooon: @0xDEnYO if condidtion has exactly the same logic as isRequestIdConsumed (previously consumedIds function). I think it's worth to reuse it to keep DRY rule

PR #1607, opened 2026-01-29
add scheduleBatch handling and decoding to execute-pending-timelock

0xDEnYO:

### DIFF
--- a/script/deploy/safe/execute-pending-timelock-tx.ts
+++ b/script/deploy/safe/execute-pending-timelock-tx.ts
@@ -59,20 +59,12 @@ interface IDeploymentData {
   [key: string]: string | undefined
 }
 
-/** Result of fetching pending operations for a network (used when pre-checking all networks in parallel). Clients are not stored so we open a fresh RPC when processing. */
+/** Result of pre-checking a network (MongoDB only, no RPC). Used to decide which networks to process; processNetwork always opens a fresh RPC and fetches pending ops on-chain. */
 interface IPendingFetchResult {
   network: INetworkConfig
-  readyOperations: ITimelockOperation[]
-  totalPendingCount: number
-  notScheduledOperations: Array<{
-    operationId: string
-    transactionId: string
-    safeTxHash: string
-    executionHash?: string
-  }>
-  timelockAddress?: Address
-  deploymentData?: IDeploymentData
-  /** Set when prefetch failed; callers must not treat as "no ready ops" without checking. */
+  /** Number of pending timelock txs in MongoDB (on-chain ready count unknown until processNetwork runs). */
+  pendingInMongoCount: number
+  /** Set when prefetch failed; callers must not treat as "no pending" without checking. */
   fetchError?: unknown
 }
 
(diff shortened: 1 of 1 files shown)


script/deploy/safe/execute-pending-timelock-tx.ts
```diff
@@ -670,24 +868,77 @@ async function getPendingOperations(
...
+            predecessor,
+            salt
+          )
+          const firstValue = valuesArr[0]
+          const firstPayload = payloadsArr[0]
+          if (firstValue === undefined || firstPayload === undefined) continue
+          target = targetsArr[0] as Address
+          value = firstValue
+          innerData = firstPayload
+          isBatch = true
```
mirooon: this condition and also isBatch variable lead to code code complexity. creates a lot of code because of this. What if we would threat one single transaction as batch of one? this would be cleaner and less error prone and also will remove isBatch checks later. The changes in confirm-safe-tx would be also needed then

script/deploy/safe/execute-pending-timelock-tx.ts
```diff
@@ -419,14 +525,73 @@ async function fetchPendingTimelockTransactions(
...
+        { quiet: true }
+      )
+
+    return {
+      network,
+      readyOperations,
+      totalPendingCount,
+      notScheduledOperations,
+      publicClient,
+      walletClient,
```
mirooon: function creates a publicClient and walletClient via setupEnvironment for every network and returns them. so if we pre‑fetch them for many networks and then go into interactive mode those clients stay alive the whole time, so code keeps rpcs connections open while the operator is thinking choosing option

script/deploy/safe/execute-pending-timelock-tx.ts
0xDEnYO: outdated/addressed

script/deploy/safe/execute-pending-timelock-tx.ts
0xDEnYO: yes, that is correct. I am not sure what you are suggesting here exactly or what is the issue. Could you please rephrase/elaborate @mirooon

script/deploy/safe/execute-pending-timelock-tx.ts
```diff
@@ -419,14 +554,73 @@ async function fetchPendingTimelockTransactions(
...
+      timelockAddress,
+      deploymentData,
+    }
+  } catch {
+    return empty
+  }
```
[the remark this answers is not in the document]
0xDEnYO: addressed, pls check if sufficiently resolvec

script/deploy/safe/execute-pending-timelock-tx.ts
```diff
@@ -670,24 +898,77 @@ async function getPendingOperations(
...
+          ) {
+            consola.warn(
+              `[${networkName}] Transaction ${tx._id} scheduleBatch has empty arrays; skipping.`
+            )
+            continue
+          }
+          targets = targetsArr
+          values = valuesArr
+          payloads = payloadsArr
+          predecessor = pred
```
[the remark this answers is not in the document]
0xDEnYO: addressed, pls check if sufficiently resolvec

script/deploy/safe/execute-pending-timelock-tx.ts
mirooon: the issue is that we can have rpc timeouts. We open all rpc clients at the start for all networks, then just keep them open while the dev is for example checking execution payload in the interactive menu. If the user takes more than a minute or two to pick an option, the rpc provider can close the connection because it’s inactive. Then later we try to use that same client to send a tx but the connection is already dead so there can be hang ups and connection reset error. Also we have 60+ chains so we open 60+ rpcs at once. That’s pretty heavy and on weaker machines it can be painful. We haven't had that problem so far, I would check if we already have similar fetching in the code so we can move forward with this code but just that we are aware of this issue. lmk and I can approve this PR and we can monitor

0xDEnYO: got it. I wasnt aware of this issue. Thanks for pointing it out. I updated it to not keep the RPC connection. Please could you check if this solves the issue sufficiently in your opinion?

mirooon: not exactly. I played with AI a bit and the suggestion is:

The specific lines that cause the "timeout/crash" risk are still present. Function is still returning the live clients in the pre-fetch phase and attempting to reuse them in the execution phase.

Here are the two specific blocks you need to change to fix it:

1. Fix fetchPendingForNetwork
Current status: You are returning the publicClient and walletClient. This keeps the connection open. Fix: Delete those lines from the return object.

2. Fix processNetwork
Current status: You have an if block that checks if preFetched.publicClient exists and tries to reuse it. Fix: Delete that entire if block. You must force the script to create a fresh connection every time it enters processNetwork.

Replace lines 463–508 with this:

By making these two changes, the fetch phase becomes stateless (just data), and the process phase becomes fresh (new connection), eliminating the timeout risk.

0xDEnYO: ok, I made the fetch-state stateless now without any RPC connections (just reading from MongoDB).
Then when we process networks we open a fresh connection. 

How does that sound? :)

mirooon: nice! great approach 👍🏻

PR #1618, opened 2026-02-06
Improved multisig execution and error handling. Added helper functions

mirooon:

- Added helper functions.
- Improved the GS013 error description.
- Added decoding for grantRole.
- Also added Execute with Deployer when all signatures are already collected. This came from a real case where I accidentally signed with an old deployer. After removing the wrong signature from the DB, I couldn’t execute the tx with the correct deployer anymore. Now, if a transaction already has enough signatures (tx.canExecute / hasEnoughSignatures), the script offers both Execute and Execute with Deployer in both flows.

### DIFF
--- a/script/playgroundHelpers.sh
+++ b/script/playgroundHelpers.sh
@@ -2085,6 +2085,7 @@ export -f removeDeployerWhitelistPermission
 export -f manageTimelockCanceller
 export -f manageSafeOwner
 export -f proposeTimelockAdminTransfer
+export -f grantTimelockRole
 
 # Utility functions
 export -f validateDependencies


script/deploy/safe/confirm-safe-tx.ts
```diff
@@ -196,6 +196,17 @@ const processTxs = async (
           '   Possible causes: invalid signature format or incorrect signer.'
         )
       }
+      if (errorMsg.includes('GS013')) {
+        consola.error(
+          '   GS013 means the inner call (e.g. Timelock.schedule) failed and Safe was executed with safeTxGas=0.'
+        )
+        consola.error(
+          `   Likely cause: safeAddress for this network (${safeAddress} from config/networks.json for "${network}") does not have TIMELOCK_PROPOSER_ROLE on the LiFiTimelockController.`
```
0xDEnYO: I am not sure if we should do this. 
GS013 is a very generic error code that pretty much just says "the underlying tx reverted" (for whatever reason). 

We only ever had this case with the missing proposer role once and it's really rare case that just existed cause we switched safes in the meantime (which we dont do regularly). 

I think if we want to improve this whole flow, we would be better off adding transaction simulation and extraction of potential reverts or error reasons from that simulation (which is a bigger topic of course).

I feel OK with the first error message here, just not with the 2nd and 3rd.


In order to catch this case that the safe doesnt have proposer role we could also just add a check for it. But to be honest, I dont think it's necessary as our deployment scripts usually ensure this.

script/playgroundHelpers.sh
```diff
@@ -947,6 +947,177 @@ function _createTimelockCancellerProposal() {
...
+# _grantTimelockProposalRole: If Safe lacks PROPOSER_ROLE, grant it. Uses Safe if Safe is timelock admin,
+# otherwise uses deployer wallet (private key from getPrivateKey) if deployer is timelock admin.
+# Usage: _grantTimelockProposalRole NETWORK TIMELOCK_ADDRESS SAFE_ADDRESS RPC_URL [ENVIRONMENT]
+#   NETWORK          - Network name
+#   TIMELOCK_ADDRESS - LiFiTimelockController address
+#   SAFE_ADDRESS     - Safe multisig address (from config/networks.json)
+#   RPC_URL          - RPC URL for the network
+#   ENVIRONMENT      - Optional, default "production"
+# Returns: 0 on success (already has role, or proposal sent, or deployer sent), 1 on failure
+function _grantTimelockProposalRole() {
```
0xDEnYO: If you create helper functions like this (which is good), I would suggest to always make them as generic as possible so we can use them for different purposes. 
E.g. here you can add a parameter "ROLE" and then use this function to grant any role on the timelock controller. 

Second, you are using "SAFE_ADDRESS" as parameter. But we could also use this function to grant a role to a wallet (e.g. deployer wallet). So also here I would just use sth more generic like "GRANT_ROLE_TO".

script/deploy/safe/confirm-safe-tx.ts
mirooon: ok got your point. I agree. updated

script/playgroundHelpers.sh
mirooon: ok, yep thats good idea

script/deploy/safe/confirm-safe-tx.ts
mirooon: ok got your point. I agree. updated

script/playgroundHelpers.sh
```diff
@@ -947,6 +947,189 @@ function _createTimelockCancellerProposal() {
...
+    echo "[$NETWORK] $GRANT_ROLE_TO already has $ROLE_NAME on timelock $TIMELOCK_ADDRESS"
+    return 0
+  fi
+
+  # 2) Who can grant? Only TIMELOCK_ADMIN_ROLE can grant. Check network Safe first, then deployer.
+  local SAFE_ADDRESS
+  SAFE_ADDRESS=$(getValueFromJSONFile "./config/networks.json" "$NETWORK.safeAddress" 2>/dev/null || true)
+  if [[ $? -ne 0 || -z "$SAFE_ADDRESS" || "$SAFE_ADDRESS" == "null" ]]; then
+    SAFE_ADDRESS=""
+  fi
```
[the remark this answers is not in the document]
mirooon: good point

script/playgroundHelpers.sh
```diff
@@ -947,6 +947,177 @@ function _createTimelockCancellerProposal() {
...
+  echo "[$NETWORK] Deployer is timelock admin. Sending grantRole(PROPOSER_ROLE, Safe) from deployer..."
+  local DEPLOYER_PRIVATE_KEY
+  DEPLOYER_PRIVATE_KEY=$(getPrivateKey "$NETWORK" "$ENVIRONMENT") || {
+    error "[$NETWORK] Failed to get deployer private key"
+    return 1
+  }
+  if ! universalSendRaw "$NETWORK" "$ENVIRONMENT" "$TIMELOCK_ADDRESS" "$CALLDATA" "$DEPLOYER_PRIVATE_KEY"; then
+    error "[$NETWORK] Failed to send grantRole from deployer"
+    return 1
+  fi
```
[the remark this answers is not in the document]
mirooon: good point

script/playgroundHelpers.sh
```diff
@@ -947,6 +947,189 @@ function _createTimelockCancellerProposal() {
...
+  CALLDATA=$(cast calldata "grantRole(bytes32,address)" "$TIMELOCK_ADMIN_ROLE" "$NEW_SAFE_ADDRESS" 2>&1)
+  if [[ $? -ne 0 || -z "$CALLDATA" || "$CALLDATA" =~ ^Error ]]; then
+    error "[$NETWORK] Failed to build grantRole calldata: $CALLDATA"
+    return 1
+  fi
+  if [[ ! "$CALLDATA" =~ ^0x[0-9a-fA-F]+$ ]] || [[ ${#CALLDATA} -lt 10 ]]; then
+    error "[$NETWORK] Invalid grantRole calldata: $CALLDATA"
+    return 1
+  fi
+
```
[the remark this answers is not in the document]
mirooon: not important for now. Please remind about this next time when changed

script/playgroundHelpers.sh
```diff
@@ -1901,12 +2084,11 @@ export -f removeAccessManagerPermission
 export -f removeDeployerWhitelistPermission
 export -f manageTimelockCanceller
 export -f manageSafeOwner
+export -f proposeTimelockAdminTransfer
```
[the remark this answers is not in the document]
mirooon: good point!

PR #1628, opened 2026-02-11
Add analyze bytecode cursor command

0xDEnYO:

None — internal SC-tooling PR (no production code, no Solidity changes). Skill / command only.

## What this PR adds

A new Cursor / Claude Code command, `/analyze-unverified-contract <address> <network>` (or `/analyze-unverified-contract <block-explorer-url>`), plus 5 supporting bash helpers in `script/playgroundHelpers.sh`. The command takes an unverified contract and produces a single structured markdown report (`report-unverified-<network>-<short_addr>.md`) covering: address validation, RPC resolution, proxy detection, Heimdall disassembly + optional decompilation, function-selector extraction, multi-source selector → signature resolution, and an interface-hint summary.

Use cases: investigating unverified contracts during audits / partner reviews / onchain incident triage / bounty reports; chained from `/analyze-tx` when the target of a transaction turns out to be unverified.

## Why this shape

1. **One command, one report.** Reviewers want a single file they can open, not 6 artifacts to chase. The selectors file, decompile output, etc. are referenced by path from the report.
2. **Agent-agnostic layout.** Canonical file lives at `.agents/commands/analyze-unverified-contract.md`; symlinks at `.cursor/commands/` and `.claude/skills/analyze-unverified-contract/SKILL.md` — same pattern as the other commands (`analyze-tx`, `add-network`, `deprecate-contract`, etc.).
3. **Bundled bash helpers, not Python.** Following the repo's TypeScript/Bash-only rule. The 5 helpers (`resolveContractTarget`, `extractSelectorsFromOpcodes`, `decodeSelector4byte`, `decodeSelectors4byte`, `generateUnverifiedContractReportSkeleton`) all live in `script/playgroundHelpers.sh` next to existing `decodeSelector(s)4byte` and are sourced by anyone using the command.
4. **Proxy detection up front.** EIP-1967 standard, EIP-1967 beacon, and EIP-1167 minimal proxies are detected before Heimdall runs — otherwise we'd disassemble the proxy shell and miss the actual logic. Selected based on prevalence in practice. EIP-1822 (UUPS pre-1967), GnosisSafe (heuristic), and EIP-2535 Diamond (no single impl) explicitly out of scope and documented as such in the function header.
5. **4-tier selector resolution** (known table → local `out/*.json` → `cast 4byte` → 4byte.directory) — local sources first so we don't hammer 4byte's rate limit; cheap before expensive.
6. **Skill-creator review applied.** After the initial draft + @mirooon review, ran a skill-creator audit on the SKILL.md and addressed the high-value gaps: sharpened the description for triggering, bundled the 3 main shell pipelines as helpers, added a Conventions stanza + Failure modes table, de-duplicated overlapping sections.

## Test plan

- [ ] `source script/helperFunctions.sh && source script/playgroundHelpers.sh` succeeds without errors.
- [ ] `declare -F resolveContractTarget extractSelectorsFromOpcodes generateUnverifiedContractReportSkeleton decodeSelector4byte decodeSelectors4byte` lists all five.
- [ ] `resolveContractTarget` rejects invalid addresses with a non-zero return code.
- [ ] `extractSelectorsFromOpcodes` accepts both `PUSH4 0xABCD…` and `PUSH4 abcd…` Heimdall output formats; filters out `0xffffffff`, `0x4e487b71`, `0x08c379a0`; normalizes lowercase; deduplicates.
- [ ] `generateUnverifiedContractReportSkeleton 0x... <network>` emits a 7-section markdown skeleton.
- [ ] Symlinks resolve from `.cursor/commands/` and `.claude/skills/analyze-unverified-contract/` to the canonical `.agents/commands/` file (`ls -la` shows `->`).
- [ ] **End-to-end (one volunteer + an RPC URL)**: pick an unverified contract on `base` (e.g. a random no-source address from basescan), run `/analyze-unverified-contract <addr> base`, and confirm the produced `report-unverified-base-<short>.md` has all 7 sections filled with plausible content.
- [ ] **Proxy case**: pick a known UUPS proxy, confirm `resolveContractTarget` reports `kind=eip1967` and `target=<impl>`.

- [ ] I have added tests that cover the functionality / test the bug — bash helpers are smoke-tested manually; no automated tests for skills/commands in this repo today
- [x] For new facets: I have checked all points from this list: https://www.notion.so/lifi/New-Facet-Contract-Checklist-157f0ff14ac78095a2b8f999d655622e — N/A (no facets)
- [x] I have updated any required documentation — the SKILL.md itself is documentation; references to other docs (analyze-tx fallback link) updated

- [x] I have checked that any arbitrary calls to external contracts are validated and or restricted — N/A, no Solidity in this PR; the only external calls are `cast storage` / `cast code` / `cast call beacon.implementation()` (read-only, no state writes) plus the 4byte.directory HTTP GET.
- [x] I have checked that any privileged calls (i.e. storage modifications) are validated and or restricted — N/A, no state-modifying code.
- [x] I have ensured that any new contracts have had AT A MINIMUM 1 preliminary audit conducted on `<date>` by `<company/auditor>` — N/A, no contracts. AuditNotRequired label applied.


script/playgroundHelpers.sh
```diff
@@ -2055,6 +2055,110 @@ function transferStagingDiamondOwnership() {
...
+
+  local FOURBYTE_URL="${FOURBYTE_URL:-https://www.4byte.directory/api/v1/signatures}"
+  local RESPONSE
+  RESPONSE=$(curl -sS -L "${FOURBYTE_URL}/?hex_signature=${SELECTOR}" 2>/dev/null || echo '{"count":0}')
+  local COUNT
+  COUNT=$(echo "$RESPONSE" | jq -r '.count // 0')
+  if [[ "$COUNT" == "0" || "$COUNT" == "null" ]]; then
+    printf "%s\t%s\n" "$SELECTOR" "(no match)"
+    return 0
+  fi
```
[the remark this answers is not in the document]
0xDEnYO: addressed, pls check

.cursor/commands/analyze-unverified-contract.md
```diff
@@ -0,0 +1,134 @@
...
+- **Decompile** (bytecode → pseudo-Solidity + ABI): `heimdall decompile <TARGET> -r <RPC_URL> -o <OUTPUT> -d`. Use `--include-sol` and/or `--include-yul` for full output; `--skip-resolving` to skip selector resolution.
+- **Outputs**: e.g. `opcodes-<network>-<short_addr>.txt/` (Heimdall may create a directory; opcodes are in `disassembled.asm` inside) or `opcodes.txt`; optional `decompiled-<short_addr>/`.
+- **More insight**: `heimdall dump <TARGET> -r <RPC_URL>` (storage slots); `heimdall cfg <TARGET> -r <RPC_URL> -o <OUTPUT>` (control-flow graph). Use when summarizing structure or proxy/storage patterns.
+
+---
+
+## 6. Extract function selectors from opcodes
+
+- **Pattern**: `PUSH4 <8 hex digits>` (e.g. `PUSH4 8388464e`).
+- **Exclude**: `0xffffffff` (sentinel), `0x4e487b71` (Error(string)).
```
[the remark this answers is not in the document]
0xDEnYO: addressed, pls check

.cursor/commands/analyze-unverified-contract.md
```diff
@@ -0,0 +1,134 @@
...
+
+- **Skip if** only disassemble/decompile: Heimdall accepts `<TARGET>` = address and fetches via RPC; no need to run `cast code` first.
+- **When needed**: To save bytecode to a file (e.g. for other tools or offline use): `cast code "<ADDRESS>" --rpc-url "<RPC_URL>" > bytecode-<short_addr>.bin`
+
+---
+
+## 4. Heimdall installation check
+
+- Run `command -v heimdall` (or `which heimdall`).
+- If **not found**: Tell the user: "Heimdall is not installed. Install with: `curl -L http://get.heimdall.rs | bash`, then in a new terminal run `bifrost`." Offer to continue after they install. Do **not** proceed with disassemble/decompile until heimdall is available.
```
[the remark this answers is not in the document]
0xDEnYO: addressed, pls check

.cursor/commands/analyze-unverified-contract.md
```diff
@@ -0,0 +1,134 @@
+---
```
mirooon: nice! Imo its worth to add as a fallback for unverified contracts in `analyze-tx.md` command. wdyt?

.cursor/commands/analyze-unverified-contract.md
```diff
@@ -0,0 +1,134 @@
...
+- **When needed**: To save bytecode to a file (e.g. for other tools or offline use): `cast code "<ADDRESS>" --rpc-url "<RPC_URL>" > bytecode-<short_addr>.bin`
+
+---
+
+## 4. Heimdall installation check
+
+- Run `command -v heimdall` (or `which heimdall`).
+- If **not found**: Tell the user: "Heimdall is not installed. Install with: `curl -L https://get.heimdall.rs | bash`, then in a new terminal run `bifrost`." Offer to continue after they install (reference the same HTTPS install URL if they ask how to install). Do **not** proceed with disassemble/decompile until heimdall is available.
+
+---
```
mirooon: I would add here proxy check to make sure we are not analyzing proxy contract but the implementation contracts. Something like this:

```
- Before decompiling, check if the contract is an EIP-1967 proxy by reading its implementation slot:
  `cast storage <ADDRESS> 0x360894a13ba1a3210667c828492db98dca3e2076cc3735a920a3ca505d382bbc --rpc-url <RPC_URL>`
- If the result is a non-zero address, **this is a proxy**. Use this new implementation address for the Heimdall target in the steps below to extract the actual logic.
```

.cursor/commands/analyze-unverified-contract.md
0xDEnYO: Good call — done. Added a fallback pointer in `analyze-tx.md` §5 (Calldata Decoding for non-LiFiDiamond root tx) in commit `a8cd01ea`. When the target is unverified (no `out/` artifact + `cast 4byte` returns nothing), it points at `/analyze-unverified-contract` so the two commands chain naturally.

.cursor/commands/analyze-unverified-contract.md
0xDEnYO: Done — and went a bit further than EIP-1967. Added a proxy-check step before disassemble/decompile in commit `a8cd01ea`, then refactored into a bundled bash helper `resolveContractTarget` (in `script/playgroundHelpers.sh`) and extended coverage in `ccb718ce`:

- **EIP-1967 standard** implementation slot (`0x360894…382bbc`)
- **EIP-1967 beacon** slot (`0xa3f0ad…133d50`) → `beacon.implementation()`
- **EIP-1167 minimal proxy** / OpenZeppelin Clones — bytecode prefix match

The helper prints `target=… proxy=… kind=direct|eip1967|eip1967-beacon|eip1167`. When `kind != direct`, the skill retargets Heimdall at the resolved implementation and records both addresses in §2 of the report.

Explicitly **not** covered (documented in the function header with rationale): EIP-1822 (UUPS pre-1967 — legacy, rare), GnosisSafe (slot-0 detection is heuristic, prone to false positives on any contract storing an address at slot 0), and EIP-2535 Diamond (no single implementation — needs a different analysis path via `facets()`).

script/playgroundHelpers.sh
```diff
@@ -2055,6 +2055,312 @@ function transferStagingDiamondOwnership() {
...
+  if [[ -z "$INPUT" ]]; then
+    error "extractSelectorsFromOpcodes: OPCODES_FILE is required"
+    return 1
+  fi
+  if [[ ! -f "$INPUT" ]]; then
+    error "extractSelectorsFromOpcodes: file not found: $INPUT"
+    return 1
+  fi
+  # The Heimdall output uses `PUSH4 0xABCDEF12` or `PUSH4 abcdef12` depending on version — accept both.
+  grep -oE 'PUSH4 (0x)?[[:xdigit:]]{8}' "$INPUT" |
```
mirooon, line 2311: **Selector extraction can make §5 of the generated report actively misleading, not just incomplete.**

`grep -oE 'PUSH4 ...'` matches every `PUSH4` constant, but `PUSH4` is not synonymous with "this contract's selector". Sources of false positives:

1. **Interface IDs checked inside `supportsInterface`** — e.g. `0x80ac58cd` (ERC-721). The contract is *asking about* the interface, not exposing it.
2. **Selectors of OTHER contracts the code calls** — a router that calls `token.transfer(...)` has `PUSH4 0xa9059cbb` in its bytecode, identical to a contract that implements `transfer`. The report would claim a USDC consumer "implements ERC-20".
3. **Custom-error selectors** — `error Foo();` compiles to a bytes4 that appears in revert paths but is not callable.
4. **Bitmask / packed-field constants** that happen to be 4 bytes.

**Fix**: anchor the grep on the dispatcher fingerprint `DUP1 PUSH4 <sel> EQ PUSH2/3 <dest> JUMPI` — only those are real dispatched selectors. Heimdall's own selector resolver does this; that's why its lists are tighter than ours.

**Until the helper is fixed**, the SKILL.md should at minimum:
- Relabel §4 of the report as "PUSH4 constants — candidate selectors" (not "selectors exposed");
- Require multi-selector + dispatcher-pattern confirmation in §5 before naming an interface, otherwise mark it as "uses or implements" rather than "implements".

This is the only known failure mode that produces a *wrong* report rather than a *thin* one, so I'd treat it as blocking before this skill gets used in a bounty triage.

.agents/commands/analyze-unverified-contract.md
```diff
@@ -0,0 +1,112 @@
...
+### 3. Detect proxies and set TARGET
+
+```bash
+eval "$(resolveContractTarget "$ADDRESS" "$RPC_URL")"
+# Now $target, $proxy, $kind are set.
+```
+
+If `$kind` is anything other than `direct`, **set `TARGET=$target`** (the resolved implementation); otherwise `TARGET=$ADDRESS`. Record both `$proxy` and `$target` in §2 of the report.
+
+Proxies hold dispatch bytecode only — running Heimdall on the proxy itself yields an empty selector list. The helper covers EIP-1967 standard, EIP-1967 beacon, and EIP-1167 (Clones); for the less-common patterns (EIP-1822 UUPS pre-1967, GnosisSafe, EIP-2535 Diamond) you'll see `kind=direct` and zero selectors in Step 5 — that's your signal to investigate manually (slot 0, `facets()` view, etc.).
```
mirooon: **Diamond detection is one `cast call` away — worth doing rather than falling back to "investigate manually".**

```bash
cast call $ADDRESS 'facets()(tuple(address,bytes4[])[])' --rpc-url $RPC_URL
```

Non-revert + non-empty return ⇒ EIP-2535 diamond. Then:
- Extend `resolveContractTarget` with `kind=diamond target=<facet1,facet2,…>`.
- Run Steps 4–7 per facet (each facet becomes its own `TARGET` for Heimdall).
- The `facets()` return already gives the selector → facet mapping, so §4 and §5 of the report come for free — Heimdall only contributes the structure summary per facet.

Worth prioritising because **this repo is a diamond protocol** — diamonds in the wild are exactly what `/analyze-unverified-contract` will hit most often (LI.FI investigations, partner reviews, cross-protocol audits). Saying "investigate manually" for the most common case in scope undersells the skill. Could be a follow-up PR if you'd rather not grow this one, but it's the obvious next step.

mirooon (commented):
Two structural notes from a re-read — both worth addressing before this becomes the canonical "how to look at an unverified contract" skill in the repo. Inline below.

.agents/commands/analyze-unverified-contract.md
0xDEnYO: Done in [`aa340d74`](https://github.com/lifinance/contracts/pull/1628/commits/aa340d74). Extended `resolveContractTarget` with a 4th detection step:

```bash
cast call "$ADDRESS" 'facets()(tuple(address,bytes4[])[])' --rpc-url "$RPC_URL"
```

Non-revert + non-empty ⇒ `kind=diamond`, `target=<facet1,facet2,…>` (comma-separated lowercase). Address regex (`0x + 40 hex`) is collision-safe against the bytes4 selectors (8 hex) inside the inner arrays.

Step 3 of the command doc now branches on `$kind`: for `diamond`, the user iterates Steps 4–7 once per facet (artifacts written as `opcodes-$NETWORK-$SHORT-<facet_short>.txt`), then aggregates per-facet selector tables into §4 of the report. `facets()` return remains the authoritative selector → facet mapping (called out in the PUSH4-caveat note below — opcode extraction is secondary signal only on diamonds).

script/playgroundHelpers.sh
0xDEnYO: Good catch — agreed, this is the more dangerous failure mode (selectors mid-report look authoritative). Addressed in [`aa340d74`](https://github.com/lifinance/contracts/pull/1628/commits/aa340d74) at the doc layer rather than by tightening the regex, since reliably distinguishing dispatch-table `PUSH4`s from interface-ID / external-call `PUSH4`s from opcodes alone needs CFG-level reasoning (`EQ` + `JUMPI` follow-up, jump-table prologue detection) — too brittle for a bash filter, and we'd rather over-collect than drop a real selector.

Three changes:

1. **`extractSelectorsFromOpcodes` docstring** — explicit "false positives" block listing the three sources (interface IDs in `supportsInterface`, selectors of called contracts, keccak literals), and the heuristic for a reliable signal (selector appears in the dispatch jump table near the entrypoint with `EQ` + `JUMPI`).
2. **Command doc Step 5** — same caveat, plus a diamond-specific note: when `kind=diamond`, use the `facets()` return as authoritative and treat opcode extraction as secondary signal.
3. **Report skeleton §5** — added vetting guidance inline so the agent filling the report sees it at the right moment (after selectors are resolved, before claiming an interface is implemented).

This way the extraction stays broad/loss-free; the vetting work happens once, in the right section of the report.

mirooon (approved):

PR #1705, opened 2026-04-09
test: add fee forwarding after swap test case

maxklenk:
Test how flexible we can collect fees in different tokens/order in our processes.


0xDEnYO:
Closing this spike. Its question — *can `FeeForwarder` compose as a swap step inside a GenericSwap multiswap to collect integrator fees after the swap?* — is answered (yes), and the branch is now ~700 commits behind main.

The lasting value is the integration-coverage gap (FeeForwarder exercised inside a multiswap, which the unit tests in `FeeForwarder.t.sol` don't cover). Replacing this with a fresh, properly-asserted integration test against current main rather than rebasing this draft.

0xDEnYO:
Replaced by #1978 — a properly-asserted integration test against current main.

PR #1744, opened 2026-04-27
Fix deployAllContracts gaps surfaced by smoke test

gvladika:
# Which Linear task belongs to this PR?

[EXSC-239](https://linear.app/lifi-linear/issue/EXSC-239/make-lifi-deployment-part-of-ci)

# Why did I implement it this way?

Setting up the CI deploy smoke test (follow-up PR branched off this one) required `deployAllContracts.sh` to actually run end-to-end against a fresh local anvil chain. Doing that surfaced a series of gaps where the script depended on interactive flow, drifted from sibling scripts, or assumed prior state. Each commit fixes one issue in isolation:

- **`9828ee41` — wire DiamondLoupeFacet before UpdateCoreFacets.** Stage 3 deployed the diamond and immediately called `UpdateCoreFacets`, which reverts with `DiamondLoupeFacetNotFound()` on a freshly minted diamond because the loupe selectors aren't wired yet. Added the missing `UpdateDiamondLoupeFacet` step between diamond deploy and core-facet cut.
- **`a13635f5` — skip GasZipFacet in UpdateCoreFacets when gasZipChainId is 0.** `deployCoreFacets.sh` and `healthCheck.ts` already skip GasZipFacet for networks with no GasZip support; `UpdateCoreFacets.s.sol` did not, so it reverted on `vm.parseJsonAddress` for any such network.
- **`e243d498` — source updateFacetConfig.sh in deployAllContracts.** Stage 4 calls `updateFacetConfig`, but the function lives in `script/tasks/updateFacetConfig.sh` and `deployAllContracts.sh` did not source it. The interactive flow worked because `scriptMaster.sh` auto-sources every `script/tasks/*.sh`; direct callers got `command not found`.
- **`aec814b8` — add localanvil staging target state.** `deployPeripheryContracts` and the non-core facets loop skip every contract that has no entry in `_targetState.json` for the running `(network, environment)` tuple. Adds a minimal `localanvil.staging.LiFiDiamond` block so the periphery deploy/register paths actually run end-to-end.
- **`bce803a8` — invoke updateWhitelistPeriphery via bun, not bunx tsx.** `bunx tsx` fetches tsx into `/private/tmp/bunx-501-tsx@latest` and runs from there, where Node's resolver cannot see the project's `node_modules`, so the script failed with `ERR_MODULE_NOT_FOUND` for `citty`. `bun` runs `.ts` natively and resolves from the project root, matching how `healthCheck.ts` is invoked.
[...]

PR #1747, opened 2026-04-28
Add deploy smoke-test GitHub Actions workflow [DeBridgeDlnFacet v1.0.1]

gvladika:

[EXSC-239](https://linear.app/lifi-linear/issue/EXSC-239/make-lifi-deployment-part-of-ci)

Stacks on top of #1744. That PR adds `script/deploy/_smokeDeploy.sh` and the deploy-pipeline fixes that make a clean run possible; this PR wires the runner.

The workflow runs `_smokeDeploy.sh` followed by `healthCheck.ts` against a fresh anvil chain and an empty Mongo on every PR targeting `main`. Triggers chosen with input from the team:

- `pull_request` to `main` (skips drafts; matches `forge.yml`).
- `workflow_dispatch` for manual re-runs from the Actions UI.

Example of successful run: https://github.com/lifinance/contracts/actions/runs/25042796452/job/73350029593?pr=1747

Implementation notes:

- Every `uses:` is pinned to a 40-char SHA per `.claude/rules/500-github-actions.md`. The three core actions (`actions/checkout@692973e3`, `foundry-rs/foundry-toolchain@8b0419c6`, `oven-sh/setup-bun@3d267786`) reuse the SHAs already pinned in `.github/workflows/diamondEmergencyPauseStaging.yml`. `actions/upload-artifact@ea165f8d` is the latest stable v4 (v4.6.2).
- Mongo runs as a `services:` container (`mongo:7`) with a `mongosh ping` health probe. Empty DB per run is correct — that's the "no prior deploys" state the deploy bash expects, and it's what made the local smoke test work.
- Anvil starts in the background under the standard Foundry test mnemonic (already in `.env.example`); a bounded readiness loop polls `cast chain-id` for up to 30 s before continuing.
- `gum` is installed from the official Charm GitHub deb release pinned to `v0.14.5`. The deb is verified against the publisher's SHA-256 (`fd1a3471…b2ba`) before `dpkg -i`; mismatch fails the run under `set -euo pipefail`.
- `.env` is built from `.env.example` with three CI overrides: `VERIFY_CONTRACTS=false`, `ETH_NODE_URI_LOCALANVIL=http://localhost:8545`, `MONGODB_URI=mongodb://localhost:27017/contract-deployments`. `PRIVATE_KEY=$PRIVATE_KEY_ANVIL` is appended as a literal so `set -a; source .env` expands it at source time.
- On step failure, `smoke-deploy.log` and `deployments/localanvil*.json` upload as the `deploy-smoke-test-logs` artifact (7-day retention) so a failed run is debuggable from the PR UI without a re-run.
- Inline shell uses `set -euo pipefail` per `.claude/rules/300-bash.md`. `actionlint` and `zizmor` both pass cleanly; `persist-credentials: false` is set on checkout (smoke test does not push back).
- Permissions: `contents: read` only.
- Timeout: 20 min (~2x the observed local upper bound).
- No secrets used. Anvil keys are the public Foundry test values; Mongo is throwaway.

End-to-end verification: ran on a real GitHub runner via a temporary `push: branches: [add-deployment-action]` trigger, confirmed exit 0, then reverted the temporary trigger so the merged state contains only `pull_request` to `main` + `workflow_dispatch`.


.github/workflows/deploy-smoke-test.yml
```diff
@@ -0,0 +1,119 @@
...
+        # gum is required by helper functions sourced via deployAllContracts;
+        # the smoke driver mocks `gum choose` for the start-stage prompt but
+        # falls through to the real binary for any other invocation.
+        run: |
+          set -euo pipefail
+          GUM_VERSION=0.14.5
+          curl -fsSLo /tmp/gum.deb "https://github.com/charmbracelet/gum/releases/download/v${GUM_VERSION}/gum_${GUM_VERSION}_amd64.deb"
+          sudo dpkg -i /tmp/gum.deb
+          gum --version
+
```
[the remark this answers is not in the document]
gvladika: Fixed

.github/workflows/deploy-smoke-test.yml
```diff
@@ -0,0 +1,124 @@
...
+jobs:
+  deploy-smoke-test:
+    if: ${{ github.event.pull_request.draft == false }}
+    runs-on: ubuntu-latest
+    timeout-minutes: 20
+
+    services:
+      mongo:
+        image: mongo:7
```
mirooon: can we also stick to hash version like `@sha256:<digest>` ?

gvladika: Done

.github/workflows/deploy-smoke-test.yml
```diff
@@ -0,0 +1,124 @@
...
+      - name: Install dependencies
+        run: bun install --frozen-lockfile
+
+      - name: Prepare .env
+        run: |
+          set -euo pipefail
+          cp .env.example .env
+          # Override the example default that turns on contract verification.
+          sed -i 's/^VERIFY_CONTRACTS=.*/VERIFY_CONTRACTS=false/' .env
+          # Append CI-only values. _smokeDeploy.sh sources .env with `set -a`,
```
mirooon: _smokeDeploy.sh I am wondering if maybe its wort renaming?                                                                     
The leading underscore convention signals "internal/not for direct use" but if this script is also meant to be runnable locally then we could actually remove underrscore

gvladika: Good point, done

mirooon (approved):

Matt-LiFi (approved):

PR #1764, opened 2026-05-07
fix(safe): exclude testnets from add-safe-owners --all-networks

0xDEnYO:
## Summary
`--all-networks` in `add-safe-owners-and-threshold.ts` now skips `networks.json` entries with `type: "testnet"`, so chains like Arbitrum Sepolia / Optimism Sepolia are not included in bulk runs.

Single-network usage (`--network <key>`) is unchanged for explicit testnet runs if needed.

## Testing
- `bunx eslint script/deploy/safe/add-safe-owners-and-threshold.ts`
- `bunx tsc-files --noEmit script/deploy/safe/add-safe-owners-and-threshold.ts`

Made with [Cursor](https://cursor.com)


gvladika (approved):
Lgtm

PR #1769, opened 2026-05-08
chore(ci): pin foundry version

gvladika:

https://linear.app/lifi-linear/issue/EXSC-273/s1-standardize-foundry-version-across-the-team-pin-170

Pin foundry via a single `.foundry-version` file consumed everywhere:

**How it works in practice:** the pinned version is set by editing one line in `.foundry-version` (currently `1.7.1`); CI installs it via the composite action and the husky pre-commit hook re-verifies it locally, so bumping the pin is a one-line PR that propagates to every workflow and every dev's laptop with no other touch-points.

- Composite action `.github/actions/setup-foundry` reads the file, installs that exact version, and verifies the result.
- Seven workflows now call `./.github/actions/setup-foundry` instead of `foundry-rs/foundry-toolchain` directly.
- Drop the stale `version: v1.0.0` override in `healthCheckForNewNetworkDeployment.yml`.
- `script/utils/verify-foundry-version.sh` runs the same check in CI and in `.husky/pre-commit`, so dev laptops and runners hit the same gate.
- README documents `foundryup --install "$(cat .foundry-version)"`.
- `forge.yml` retry block: replace `Wandalen/wretry.action` with a short bash loop that runs `forge test` once and then `forge test --rerun` on retries (re-runs only failed tests instead of the full suite each attempt). Same retry budget, dramatically faster on transient flake.
- New global rule (`.agents/rules/000-global-standards.md`): require verifying external CLI flags against `--help` or source before writing them into code, docs, or error-message strings — added after an earlier `foundryup -v <version>` propagation mistake was caught by review.

Refs EXSC-273.

The currently pinned foundry version is `1.7.1`.


.github/actions/setup-foundry/action.yml
```diff
@@ -0,0 +1,32 @@
...
+        if [ -z "$version" ]; then
+          echo "::error::.foundry-version is empty"
+          exit 1
+        fi
+        echo "version=${version}" >> "$GITHUB_OUTPUT"
+
+    - name: Install Foundry
+      uses: foundry-rs/foundry-toolchain@8b0419c685ef46cb79ec93fbdc131174afceb730 # v1.6
+      with:
+        version: ${{ steps.pin.outputs.version }}
```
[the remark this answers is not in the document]
gvladika: Fixed

README.md
```diff
@@ -177,7 +177,16 @@ bun i
 forge install
 ```
 
-3. Set up environment variables:
+3. Install the pinned Foundry version:
+
+```bash
+foundryup -v "$(cat .foundry-version)"
+foundryup -u "$(cat .foundry-version)"
+```
```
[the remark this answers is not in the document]
gvladika: Fixed

script/utils/verify-foundry-version.sh
```diff
@@ -0,0 +1,51 @@
...
+if [ -z "$EXPECTED" ]; then
+  printf '\033[31m✗ .foundry-version is empty\033[0m\n' >&2
+  exit 1
+fi
+
+if ! command -v forge >/dev/null 2>&1; then
+  printf '\033[31m✗ forge not on PATH\033[0m\n' >&2
+  printf '   install foundry %s: foundryup -v %s && foundryup -u %s\n' "$EXPECTED" "$EXPECTED" "$EXPECTED" >&2
+  exit 1
+fi
```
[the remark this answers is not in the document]
gvladika: Fixed

script/utils/verify-foundry-version.sh
```diff
@@ -0,0 +1,51 @@
...
+
+if [ "$ACTUAL" != "$EXPECTED" ]; then
+  printf '\033[31m✗ foundry version mismatch\033[0m\n' >&2
+  printf '   expected: %s\n' "$EXPECTED" >&2
+  printf '   actual:   %s\n' "$ACTUAL" >&2
+  printf '   fix:      foundryup -v %s && foundryup -u %s\n' "$EXPECTED" "$EXPECTED" >&2
+  exit 1
+fi
```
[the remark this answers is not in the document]
gvladika: Fixed

.github/actions/setup-foundry/action.yml
```diff
@@ -0,0 +1,32 @@
+# Composite action: install pinned foundry + verify
```
mirooon: what do you think about mentioning this action in `500-github-actions.md` rules?

gvladika: Good point, https://github.com/lifinance/contracts/pull/1769/changes/fb9773572112bb03998969101b137f27cc065a21

mirooon (dismissed):

mirooon (dismissed):

.husky/pre-commit
```diff
@@ -484,6 +484,11 @@ TSC_PID=""
 
 print_section "Pre-commit Checks"
 
+# Verify foundry binary matches .foundry-version
+if ! bash "$GIT_ROOT/script/utils/verify-foundry-version.sh" --quiet; then
```
0xDEnYO: hmm...if we need to switch to another foundry-version (e.g. tempo-specific, nightly, fallback for whatever reason) then we cannot commit anymore. Is this restriction really necessary? What does it protect us from?

gvladika: There is no way to force devs to use specific foundry version on local machines. Eg. you switch once to tempo-fork, and then forget to switch it back, so all upcoming work is done with that version. That's not necessarily a big deal, but preferably we should all use the same (latest stable) version of foundry whenever possible. 

This pre-commit check will flag if your local version drifted. In case you really need to use another version you can simply commit with `--no-verify` flag, or even better, you do the work with custom version, and then this checks "forces" you to switch back to target version before committing.

gvladika: Changed to warning instead of failure: https://github.com/lifinance/contracts/pull/1769/changes/6e3757de7f2da7da8a03b6b0752de19263c53552

0xDEnYO (approved):

maxklenk (approved):

PR #1776, opened 2026-05-13
chore: add creating-user-stories skill

0xDEnYO:
## Summary

- Adds the `creating-user-stories` Claude Code skill — companion to `sc-design-review` ([#1753](https://github.com/lifinance/contracts/pull/1753)). Where sc-design-review consumes a PRD and produces a contract design, this skill produces the intermediate persona-grouped stories + open-questions catalogue that feeds it.
- **Intentionally domain-agnostic** — the skill body avoids smart-contract-specific terminology so other teams can adopt it. Examples and patterns use neutral language (pricing, auth layers, etc.). The one SC-specific artifact (`references/deep-review-persona-pass.md` persona set) is flagged in the skill as needing re-casting when adopted by other domains.
- 4 files, 563 lines, no Solidity changes — skill files only under `.claude/skills/creating-user-stories/`.

## What's in the skill

Main `SKILL.md` (360 lines) covers:

- **One-action-one-benefit grammar** for stories: `[verb] [object], so that [why]` — strict, with a fix-pattern table for common smells.
- **Readability format** for every story body and every open-question body: question-as-title, plain sentences, labeled (a)/(b)/(c) options with the recommendation referencing the chosen letter, Pros/Cons only where the trade-off is real.
- **Two-page output** — a Stories page (one readable sentence per story, no acceptance hints or implementation detail in the story body) and a companion Open Questions page.
- **Mandatory lane tagging** on every open Q (⚠️ PRODUCT DECISION NEEDED / [TECH] / [META]) with a strict criterion: if the user-visible outcome is the same regardless of which option is picked → [TECH]; otherwise ⚠️ PRODUCT.
- **Cross-doc alignment is non-negotiable** — phantom ❓Q-refs erode the entire catalogue. Mandatory verification after every edit (phantom / orphan / cross-ref / counts).
- **Counts footer as audit trail** — never silently delete a question; every removal documented so a year later someone reading \"why no Q3.3?\" gets a one-line answer.
- **Three-lane lightweight final review** (spec coverage / product decision / tech implementation) as the default before locking the catalogue. Deep multi-persona review reserved for high-stakes audit prep.

Reference files (progressive disclosure — loaded on demand):

- `references/question-maintenance.md` (64 lines) — audit-phase patterns: lifecycle dispositions (merge / reframe / elevate-to-story / remove / resolve), reframing when an upstream decision changes a question's context, layered-concern decomposition for cross-cutting topics, and recognizing questions that should be removed rather than answered.
[...]

PR #1777, opened 2026-05-13
creating-user-stories: promote research phase to first-class workflow step

0xDEnYO:
## Summary

Follow-up to #1776 — based on an evaluation pass that ran the skill cold against the [Programmable Vault Wrapper PRD](https://www.notion.so/lifi/LI-FI-Programmable-Vault-Wrapper-PRD-352f0ff14ac780e9b426c99364cfe814) and compared the output to the published [Earn PRD Review](https://www.notion.so/lifi/Earn-PRD-Review-35af0ff14ac7805b98ece561ec483cde) catalogue.

**Finding it addresses**: ~20–30% of the published catalogue's value lives in research-derived stories and open Qs (Synthetix MultiRewards, Coinbase/Morpho boost, Kiln upgrade-key inversion, ERC-4626 inflation attack). The current skill mentions this as a single prose line in workflow step 7. A cold operator reads that and skips it.

**Targets** \`chore/add-creating-user-stories-skill\` so #1776 stays the single landing point.

## What changed

- **\`SKILL.md\`** — forward-pointer in workflow step 7; new \"Research phase\" section between the two workflow blocks (~50 lines). Carries trigger condition, comparable-set roles (blueprint / marquee-customer-reality / competitor / field-wide / adjacent-but-distinct), recipe, COPY/IMPROVE/AVOID frame, output contract, anti-patterns.
- **\`references/research-phase.md\`** — net-new (~250 lines). Subagent prompt template, synthesis template, and three worked examples spanning SC (vault wrapper, real \`raw/\` realised output cited), DevOps (on-call paging tool), and API (feature-flag service). The multi-domain examples make the domain-agnosticism claim concrete rather than asserted.

## Genericity strategy

Three moves that compound — discussed in detail in the evaluation that produced this diff:

1. **Trigger framed by *consequence*** (\"feeds high-stakes downstream work\"), not domain.
2. **Comparable set defined by *roles*, not products** — the domain instantiates the role.
3. **Multi-domain worked examples in the reference file**, not the skill body — \`SKILL.md\` stays abstract; the reference carries one realised example per domain.

**Load-bearing through-line**: COPY / IMPROVE / AVOID. Ports unchanged across domains.

## Out of scope (separate follow-ups from the same evaluation)

- Expand the mandatory verification checklist (Q-link target page resolution, bidirectional cross-link check, sub-point existence, ID-gap audit trail). \`evaluation/improvements.md\` #I2.
- Codify the provenance-tag set (\`*(new — surfaces a PRD contradiction)*\`, \`*(provisional)*\`, \`*(blocked by QX.Y)*\`). #I3.
- Re-cast the worked examples in SKILL.md body toward truly neutral domains. #I4.

## Test plan

Skills aren't compile-checked. Worth verifying:

- [ ] Read the new \"Research phase\" section cold — does it feel actionable to a non-SC team?
- [ ] Walk through Example B (DevOps on-call) and Example C (feature-flag API) — do they read as \"real\" or as stubs?
- [ ] Confirm the trigger condition (\"feeds high-stakes downstream work\") is tight enough that it doesn't fire on every quick spec.
- [ ] Confirm the COPY/IMPROVE/AVOID frame is the same in skill body and all three worked examples.

### DIFF
--- a/.claude/skills/creating-user-stories/SKILL.md
+++ b/.claude/skills/creating-user-stories/SKILL.md
@@ -296,8 +296,57 @@ When the catalogue is past initial extraction and you're refining or auditing it
 4. Number within each persona in lifecycle order (deploy → configure → operate → emergency).
 5. For each item, draft the one-line story.
 6. Pass over: what's missing? Standard categories specs forget — key rotation, observability, emergency response, edge cases. Add them.
-7. Comparable-product sweep: what do analogous products in the space do? Add as `*(research-derived)*`.
+7. Comparable-product sweep — load the **Research phase** below if the catalogue feeds high-stakes downstream work. Research-derived items end up tagged `*(research-derived)*` with a `*Source*: [link]` pointer.
 8. Open-questions extraction: every ambiguity, every "decide later" → a Q.
+
+## Research phase — when the catalogue feeds high-stakes downstream work
+
+**Trigger** — the catalogue feeds an audit, an estimate, a third-party review, a budget decision, or any artifact someone signs off on. Skip for internal-only catalogues used for a single team's quick alignment.
+
+Without this phase, a catalogue typically underproduces by ~20–30% on coverage of failure modes, edge cases, prior-art-anti-patterns, and "things to invert from what the obvious blueprint did." That gap doesn't show up at draft time — it shows up at audit / review when reviewers ask "what about X?" and X was visible in any comparable product the operator didn't read.
+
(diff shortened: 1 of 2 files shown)


0xDEnYO:
Pivoting: the research phase belongs in `sc-design-review` ([#1753](https://github.com/lifinance/contracts/pull/1753)), not in `creating-user-stories`. Keeping the user-stories skill focused on user-story work. Follow-up PR against #1753's branch incoming.

PR #1789, opened 2026-05-14
ci: path-gate forge + smoke, cache Foundry across all workflows

0xDEnYO:
## Summary
- **Path-gate the two slow required checks** (`Run Unit Tests`, `Deploy Smoke Test`) so the heavy job only runs when files that can actually influence the outcome change. Docs-only PRs no longer wait minutes for an unconditional rebuild.
- **Required-check safety**: each workflow gets a tiny `scope-validator` aggregator job that always reports a final status — green when the heavy job either succeeded *or* was skipped because nothing relevant changed. This avoids the GitHub footgun where `on.paths` filters make required checks hang forever in "Expected — Waiting for status to be reported".
- **Cache Foundry everywhere** via a new composite action `.github/actions/setup-foundry-cached` (installs Foundry at a single pinned SHA and restores `~/.foundry`, `~/.svm`, `cache/`, `out/`). All 7 foundry-using workflows now use it — `forge build`/`forge test`/`forge coverage` reuse compiled artifacts between runs.
- **Toolchain unification**: previously workflows pinned three different `foundry-toolchain` SHAs (`c7450ba…`, `de808b1…`, `8b0419c…`). All now go through the composite action's single pin (`8b0419c…` aka v1.6).
- **Rename** `.github/workflows/forge.yml` → `forge-unit-tests.yml` to match the smoke workflow's naming style.

## Path scope (conservative)

**Forge unit tests trigger on:**
`foundry.toml`, `remappings.txt`, `src/**`, `test/**`, `lib/**`, `.gitmodules`, the workflow file itself, the composite action.

**Smoke test triggers on:**
`foundry.toml`, `remappings.txt`, `src/**`, `lib/**`, `.gitmodules`, `config/**`, `deployments/**`, `script/deploy/**`, `script/utils/**`, `script/tasks/**`, `script/resources/**`, `.env.example`, `package.json`, `bun.lockb`, the workflow file itself, the composite action.

`test/**` is intentionally NOT in the smoke scope. Docs / `*.md` / `tasks/` / `templates/` / `audit/` etc. are not in either.

## Follow-up (admin action required)
On `main` branch protection, **swap the required status checks**:
- `run-unit-tests` → `scope-validator` (from `Run Unit Tests` workflow)
- `deploy-smoke-test` → `scope-validator` (from `Deploy Smoke Test` workflow)

If protection isn't updated, currently-open PRs will keep waiting on the old job names. Do this in the same change window as the merge of this PR.

## Test plan
- [ ] PR touching only a `*.md` → both heavy jobs skip, both `scope-validator` jobs report green within ~10s.
- [ ] PR touching `src/Facets/*.sol` → forge runs; smoke runs.
- [ ] PR touching only `test/**` → forge runs; smoke skips (scope-validator green).
- [ ] PR bumping a submodule under `lib/` → both run.
- [ ] PR editing `.github/workflows/forge-unit-tests.yml` → forge runs (self-test).
- [ ] Verify second run of `forge test` on the same commit hits the cache and shaves build time.
- [ ] After admin updates branch protection: confirm a docs-only PR is mergeable.


0xDEnYO:
Closing in favor of:
- [#1790](https://github.com/lifinance/contracts/pull/1790) — adds Foundry caching to the `setup-foundry` composite action introduced in [#1769](https://github.com/lifinance/contracts/pull/1769) (stacked PR).
- [#1791](https://github.com/lifinance/contracts/pull/1791) — adds path-gating + required-check aggregator to the forge and smoke workflows.

Splits the scope of this PR cleanly so each layer lands on top of #1769's foundry pinning work without duplicating it.

PR #1791, opened 2026-05-14
ci: path-gate forge unit tests + deploy smoke test + min test coverage

0xDEnYO:
## Summary
`Run Unit Tests`, `Deploy Smoke Test`, and `Enforce Min Test Coverage` are all required-for-merge but run unconditionally today, so a docs-only PR still waits minutes for a full rebuild + anvil pipeline + coverage run. This PR adds path-gating to all three, without breaking branch protection.

**Pattern (per workflow):**
1. `detect-changes:` runs first — uses `dorny/paths-filter` to compute whether files that can actually influence outcome were touched (conservative include list).
2. Heavy job gates on the output. Skipped if nothing relevant changed.
3. `required-check:` runs last with `if: always()` and reports green when the heavy job either succeeded or was skipped. **This is the job that branch protection should require** — it sidesteps the GitHub footgun where top-level `on.paths` filters make required checks hang forever in *"Expected — Waiting for status to be reported"*.

Also renames `.github/workflows/forge.yml` → `forge-unit-tests.yml` to match the smoke workflow's naming style.

## Path scope (conservative)

**Forge unit tests:**
`foundry.toml`, `remappings.txt`, `src/**`, `test/**`, `lib/**`, `.gitmodules`, the workflow file itself.

**Smoke test:**
`foundry.toml`, `remappings.txt`, `src/**`, `lib/**`, `.gitmodules`, `config/**`, `deployments/**`, `script/deploy/**`, `script/utils/**`, `script/tasks/**`, `script/resources/**`, `.env.example`, `package.json`, `bun.lockb`, the workflow file itself.

`test/**` is intentionally NOT in the smoke scope. Docs/markdown/audit/templates/tasks are not in either.

**Min test coverage:**
`foundry.toml`, `remappings.txt`, `src/**`, `test/**`, `lib/**`, `.gitmodules`, the workflow file itself. Same scope as forge unit tests since `forge coverage` runs the same compilation + test surface. Broader than the previous in-job `.sol`-only filter — `foundry.toml`/`lib/**`/`remappings.txt` changes can shift coverage and were previously skipped silently. The redundant in-job `git diff` step is removed in favour of the outer gate.

## Scope of this PR
Only path-gating + the rename. Foundry-toolchain unification + caching is being handled separately in [#1769](https://github.com/lifinance/contracts/pull/1769) (pin foundry version) and [#1790](https://github.com/lifinance/contracts/pull/1790) (add caching to that composite). Whichever of those merges first, this PR will need a trivial rebase to swap its `foundry-rs/foundry-toolchain@…` step for `./.github/actions/setup-foundry`.

## Follow-up (admin action required)
On `main` branch protection, **swap the required status checks**:
- `run-unit-tests` → `required-check` (from `Run Unit Tests` workflow)
- `deploy-smoke-test` → `required-check` (from `Deploy Smoke Test` workflow)
- `enforce-min-test-coverage` → `coverage-required` (from `Enforce Min Test Coverage` workflow)

If protection isn't updated, currently-open PRs will keep waiting on the old job names. Do this in the same change window as the merge of this PR.

## Test plan
- [ ] PR touching only a `*.md` → all three heavy jobs skip, all three `required-check` / `coverage-required` jobs report green within ~10s.
- [ ] PR touching `src/Facets/*.sol` → forge runs; smoke runs; coverage runs (and posts PR comment).
- [ ] PR touching only `test/**` → forge runs; coverage runs; smoke skips.
- [ ] PR touching only `foundry.toml` → forge, smoke, and coverage all run (validates broader coverage scope vs. previous `.sol`-only inner filter).
- [ ] PR bumping a submodule under `lib/` → all three run.
- [ ] PR editing `.github/workflows/forge-unit-tests.yml` → forge runs (self-test). Same for the other two workflow files.
- [ ] After admin updates branch protection: confirm a docs-only PR is mergeable.

### DIFF
--- a/.github/workflows/forge-unit-tests.yml
+++ b/.github/workflows/forge-unit-tests.yml
@@ -44,9 +44,17 @@ jobs:
               - '.github/workflows/forge-unit-tests.yml'
 
   run-unit-tests:
-    # will only run once the PR is in "Ready for Review" state AND relevant files changed
+    # Runs on PRs (once out of draft) and on push/workflow_dispatch.
+    # For non-PR events `github.event.pull_request` is null, so the draft check
+    # is gated behind an event-name guard to avoid `null == false` evaluating to
+    # false and silently skipping the job after a merge to main.
     needs: detect-changes
-    if: ${{ github.event.pull_request.draft == false && needs.detect-changes.outputs.forge == 'true' }}
+    if: >-
+      ${{
+        needs.detect-changes.outputs.forge == 'true' &&
+        (github.event_name != 'pull_request' ||
+         github.event.pull_request.draft == false)
+      }}
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
@@ -102,12 +110,20 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Validate upstream result
+        # Validate `detect-changes` succeeded first. If it failed, `run-unit-tests`
+        # would be `skipped` due to the dependency failure — and treating that
+        # `skipped` as success would mask the infra failure from branch protection.
         run: |
           set -euo pipefail
-          result="${{ needs.run-unit-tests.result }}"
-          if [[ "$result" == "success" || "$result" == "skipped" ]]; then
-            echo -e "\033[32munit-tests-required OK (run-unit-tests=$result)\033[0m"
+          detect_result="${{ needs.detect-changes.result }}"
+          test_result="${{ needs.run-unit-tests.result }}"
+          if [[ "$detect_result" != "success" ]]; then
+            echo -e "\033[31munit-tests-required FAILED (detect-changes=$detect_result)\033[0m" >&2
+            exit 1
+          fi
+          if [[ "$test_result" == "success" || "$test_result" == "skipped" ]]; then
+            echo -e "\033[32munit-tests-required OK (run-unit-tests=$test_result)\033[0m"
             exit 0
           fi
-          echo -e "\033[31munit-tests-required FAILED (run-unit-tests=$result)\033[0m" >&2
+          echo -e "\033[31munit-tests-required FAILED (run-unit-tests=$test_result)\033[0m" >&2
           exit 1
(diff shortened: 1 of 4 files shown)


.github/workflows/forge-unit-tests.yml
```diff
@@ -0,0 +1,129 @@
+# - Run (Foundry) Unit Test Suite
```
mirooon, line 1: README.md file has hardcoded link to old `forge.yml` but file change its name to `forge-unit-tests.yml.`

.github/workflows/forge-unit-tests.yml
```diff
@@ -0,0 +1,129 @@
...
+    outputs:
+      forge: ${{ steps.filter.outputs.forge }}
+    steps:
+      - uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
+      - name: Detect relevant changes
+        id: filter
+        uses: dorny/paths-filter@de90cc6fb38fc0963ad72b210f1f284cd68cea36 # v3.0.2
+        with:
+          filters: |
+            forge:
```
mirooon, line 37: I think we should keep the package.json because otherwise changes to scripts can silently break CI.

For example, if someone modifies the fetch-rpcs script and removes dependencies like mongodb, dotenv, or consola (all imported by script/mongoDb/fetch-rpcs.ts), bun install could still succeed, but bun fetch-rpcs would fail at runtime during imports.

In that case, if the filter marks forge=false, the tests get skipped and the PR still merges green, even though the script is broken.

.github/workflows/forge-unit-tests.yml
0xDEnYO: Good catch — fixed in f16bbbc0. Badge now points at `forge-unit-tests.yml`.

.github/workflows/forge-unit-tests.yml
0xDEnYO: hmm, but this git action only runs the solidity tests. Its purpose is not to check if scripts are broken.
I see your point, but in that case we should imho rather have a dedicated action that checks scripts after they have been touched, wdyt?

.github/workflows/forge-unit-tests.yml
0xDEnYO: updated

.github/workflows/forge-unit-tests.yml
mirooon: yes agree, your are right. Created ticket for it

mirooon (approved):

maxklenk (approved):

PR #1792, opened 2026-05-14
chore(skills): add /pr-ready local CodeRabbit pre-flight

0xDEnYO:
## Summary

Adds a `/pr-ready` command (skill in Claude Code, command in Cursor) that runs **CodeRabbit's local CLI** against the current branch and resolves findings **before** a PR is opened or updated. The goal is to eliminate the wait-and-fix loop where CodeRabbit findings only surface after CI runs — by the time a PR lands in review, it should be on par with what cloud CodeRabbit would have asked for.

- New command: `.agents/commands/pr-ready.md` (source of truth) + symlinks for Cursor and Claude Code
- Rule update: `.agents/rules/099-finish.md` makes `/pr-ready` the mandatory final step for both humans and agents — no `gh pr create`, no "Ready for Review" flip, no new commits on an open PR until the local review is clean (or remaining items are explicitly documented in the PR body)
- README + PR template updated so humans see the same step
- Cloud CodeRabbit stays enabled as the safety net

## Per-developer setup (one-time)

```bash
curl -fsSL https://cli.coderabbit.ai/install.sh | sh
coderabbit auth login
coderabbit --version
```

Auth token lives in `~/.config/coderabbit/`. **No shared secrets in the repo.**

## Workflow it enforces

1. Tests/lints pass locally
2. Run `/pr-ready` → executes `coderabbit review --base origin/main --plain`
3. Triage findings into Apply / Discuss / Reject buckets
4. Re-run until clean
5. `gh pr create` (or push to existing PR)

## Why a command, not a pre-push hook

Pre-push fires on every push including WIP/draft pushes — too eager. `/pr-ready` triggers at the actual "ready for review" moment, which is when the cost of the review is justified. A pre-push warning could be added later as belt-and-suspenders if desired.

## Test plan

- [ ] Install the CLI locally per the README in the skill (`curl -fsSL https://cli.coderabbit.ai/install.sh | sh && coderabbit auth login`)
- [ ] On a small feature branch, run `/pr-ready` in Claude Code and confirm it invokes `coderabbit review --base origin/main --plain` and groups findings into Apply / Discuss / Reject
- [ ] Same flow in Cursor via the `/pr-ready` slash command — confirm the symlink-based command resolves the same content
- [ ] Verify `.claude/skills/pr-ready/SKILL.md` and `.cursor/commands/pr-ready.md` both resolve to `.agents/commands/pr-ready.md` (symlink integrity)
- [ ] Confirm `.agents/rules/099-finish.md` checklist surfaces `/pr-ready` in an agent's final summary
- [ ] Confirm `.github/pull_request_template.md` shows the new `/pr-ready` checkbox on new PRs


0xDEnYO:
Good catch — addressed in 772b1b99. Two-part fix matching what you described:

**Skill respects existing installs.** If `coderabbit` is already on `PATH`, `preinstall.sh` leaves it alone (devs are free to run a newer/different version). The pinning only kicks in when *we* are the ones downloading.

**When we install: pin + verify.** I bypassed upstream's `curl … | sh` entirely (it fetches `latest` and performs no integrity check) and went one level deeper:

1. Resolve platform → `darwin-arm64` / `darwin-x64` / `linux-arm64` / `linux-x64`.
2. Download `https://cli.coderabbit.ai/releases/<PINNED>/coderabbit-<slug>.zip` directly.
3. Verify SHA-256 against a hardcoded per-platform constant in `preinstall.sh`. Mismatch → **abort** (no extract, no install) with recovery instructions.
4. Extract the single `coderabbit` binary into `~/.local/bin` and `chmod +x`.

Initial pin: **v0.4.5** (current latest), SHA-256 baked in for all four platforms. Bumping the pin is documented in `.agents/commands/pr-ready.md` → *Bumping the CodeRabbit pin* — single source of truth, deliberate (not opportunistic) updates, one commit per bump.

If you have time, the bits I'd most like a second pair of eyes on:
- the platform-slug mapping in `install_coderabbit_cli` (matches upstream's `x64` / `arm64` convention, not `x86_64` / `aarch64`)
- the soft-fail policy — every error returns 0 so `bun install` never breaks; the trade-off is a dev with a corrupt/blocked download proceeds without the CLI. I documented the manual recovery path but happy to make it hard-fail if you'd rather.

mirooon (dismissed):

.agents/commands/pr-ready.md
```diff
@@ -0,0 +1,290 @@
+---
+name: pr-ready
+description: Run CodeRabbit locally against the current branch and resolve findings before opening (or updating) a PR. The mandatory final step before `gh pr create` or flipping a draft PR to Ready-for-Review. Triggered automatically by the pre-PR gate hook, or invoke directly with `/pr-ready`.
```
gvladika, line 3: The skill says "Triggered automatically by the pre-PR gate hook" and the rule says "no gh pr create… until the local review is clean", but the PR doesn't register pr-ready-gate.py as a PreToolUse hook anywhere. There's no .claude/settings.json change registering it.

.agents/rules/099-finish.md
```diff
@@ -11,4 +11,5 @@ alwaysApply: true
...
 - **Testing**: After Solidity changes → `forge test` (or note suites remaining); after TS → lint/tests with Bun; after Bash → check execution flags/sourcing. State explicitly if anything not run.
 - **Linting**: Run the relevant linter on **all files you created or edited** (e.g. `bunx eslint` for TS/JS, or the project’s lint command) and fix all reported issues before finalizing. Do not claim the code is free of lint errors unless the linter has been run on those files and exited successfully.
-- **Summary format**: Start with applied rules (filename/anchor), include tests/lints run, call out follow-ups/gaps.
+- **PR-ready (mandatory final step before opening or updating a PR)**: Run `/pr-ready` to execute a local CodeRabbit review against the current branch and resolve all actionable findings before pushing. This applies to humans and agents alike. Do **not** run `gh pr create`, mark a draft as Ready for Review, or push new commits to an open PR until the local review is clean (or remaining items are explicitly documented in the PR body). See `.agents/commands/pr-ready.md` for the full workflow. Do not bypass with `--no-verify` or similar escape hatches.
```
gvladika: The rule (.agents/rules/099-finish.md) says: don't push new commits to an open PR until /pr-ready is clean. But the gate only intercepts gh pr create and gh pr ready — git push is not in the regex. So:
  - For draft PRs: dev pushes commits freely, eventually runs gh pr ready → gate fires → must /pr-ready first. Works.
  - For Ready-for-Review PRs: dev pushes new commits and the gate is silent. The rule depends on the dev / agent being self-disciplined.

gvladika (dismissed):

.agents/commands/pr-ready.md
0xDEnYO: Good catch — fixed in 3f522d52. Added a `PreToolUse` hook entry on `Bash` to `.claude/settings.json` that invokes `python3 .claude/scripts/pr-ready-gate.py`. The gate is now actually wired in for agents working out of this repo.

.agents/rules/099-finish.md
0xDEnYO: Right — fixed in 3f522d52. The gate now also intercepts `git push`, but only when the current branch has an OPEN, non-draft PR (resolved via `gh pr view --json isDraft,state`, fail-open if `gh` is missing/unauth/no PR). Pushes on branches with no PR or on draft PRs stay frictionless so WIP iteration is unaffected. Bypass extends naturally: `PR_READY_OK=1 git push …`. Rule + skill docs updated.

gvladika (approved):

maxklenk (approved):

PR #1801, opened 2026-05-14
fix(scripts): repair fundNewWalletOnAllChains for partial-fund runs

gvladika:

_None — internal script cleanup._

`script/tasks/fundNewWalletOnAllChains.ts` had three blockers that surfaced when funding the dev wallet on the subset of chains where its balance was zero:

1. **`--doNotFundChains` was parsed but never applied** to the target network list, so the script always funded every Gas.zip-supported active mainnet. Now intersects the parsed chainIds against the supported-mainnet list, logs both the excluded set and any chainIds that didn't match anything, and bails early if the filter leaves zero targets.

2. **No env-var fallback for the funding key.** Other task scripts (`proposeDeBridgeDlnChainIdMappings.ts`, `proposeMegaETHBridgeRegistrations.ts`) load the key via `getPrivateKey('PRIVATE_KEY_PRODUCTION', argOverride)` from `script/deploy/safe/safe-utils.ts`. Switched to that helper, which also strips the `0x` prefix automatically — eliminating the `invalid private key: expected ui8a of size 32` failure when the env var includes the prefix. `--privKeyFundingWallet` is now optional and overrides the env var when set.

3. **`BigNumber.from("0.20")` rejected fractional USD inputs.** Replaced the ethers `BigNumber` math (and the unused `getNativeAmountRequired` helper) with native `bigint` plus viem's `parseUnits` / `formatUnits`, using a `10^6` USD scale so sub-cent amounts work. This also removes the ethers dependency from the script.

Header usage example updated to `bunx tsx ...` per `.agents/rules/200-typescript.md`.

### DIFF
--- a/script/tasks/fundNewWalletOnAllChains.ts
+++ b/script/tasks/fundNewWalletOnAllChains.ts
@@ -4,6 +4,7 @@ import { BigNumber, type BigNumberish } from 'ethers'
 import { createPublicClient, createWalletClient, http, parseAbi } from 'viem'
 import { privateKeyToAccount } from 'viem/accounts'
 
+import { getPrivateKey } from '../deploy/safe/safe-utils'
 import {
   getAllActiveNetworks,
   getViemChainForNetworkName,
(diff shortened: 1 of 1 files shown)


gvladika:
Closing — superseded by switch to direct cast send transfers for funding the dev wallet across chains.

PR #1802, opened 2026-05-14
feat: fund DevWallet alongside PauserWallet in deployment Stage 9

mirooon:

Stage 9 of `deployAllContracts.sh` previously only funded the PauserWallet. The DevWallet is used to execute timelock proposals on production and to run staging deployments, so it also needs a native-token balance on every new network before execution scripts can run. Rather than adding a separate stage, I extended Stage 9 to fund both wallets in a single step, reusing the already-fetched RPC URL and private key. The private-key lookup was also hoisted above both funding sub-steps to avoid a redundant call.

### DIFF
--- a/script/deploy/deployAllContracts.sh
+++ b/script/deploy/deployAllContracts.sh
@@ -348,36 +348,39 @@ deployAllContracts() {
     fi
 
     # --- Fund DevWallet ---
-    # DevWallet is used for executing timelock proposals on production and for staging deployments.
-    # It must have a native balance on every new network before execution scripts can run.
-    local DEV_WALLET_ADDRESS
-    DEV_WALLET_ADDRESS=$(getValueFromJSONFile "./config/global.json" "devWallet")
-    if [[ $? -ne 0 ]]; then
-      error "failed to read devWallet address from ./config/global.json"
-      exit 1
-    fi
-    if [[ -z "$DEV_WALLET_ADDRESS" || "$DEV_WALLET_ADDRESS" == "null" ]]; then
-      error "DevWallet address not found. Cannot fund DevWallet"
-      exit 1
-    fi
-
-    BALANCE=$(cast balance "$DEV_WALLET_ADDRESS" --rpc-url "$RPC_URL")
-    checkFailure $? "get DevWallet balance for $DEV_WALLET_ADDRESS on $NETWORK"
-    echo "DevWallet Balance: $BALANCE"
-
-    if [[ "$BALANCE" == "0" ]]; then
-      echo "DevWallet balance is 0. How much wei would you like to send to $DEV_WALLET_ADDRESS?"
-      read -r FUNDING_AMOUNT || FUNDING_AMOUNT=""
-
-      # Validate that FUNDING_AMOUNT is a non-empty numeric value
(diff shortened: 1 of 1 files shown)


script/deploy/deployAllContracts.sh
```diff
@@ -335,18 +342,44 @@ deployAllContracts() {
...
+      exit 1
+    fi
+
+    BALANCE=$(cast balance "$DEV_WALLET_ADDRESS" --rpc-url "$RPC_URL")
+    checkFailure $? "get DevWallet balance for $DEV_WALLET_ADDRESS on $NETWORK"
+    echo "DevWallet Balance: $BALANCE"
+
+    if [[ "$BALANCE" == "0" ]]; then
+      echo "DevWallet balance is 0. How much wei would you like to send to $DEV_WALLET_ADDRESS?"
+      read -r FUNDING_AMOUNT || FUNDING_AMOUNT=""
```
gvladika: For simplicity, what do you think about hard-coding some small amount instead of asking (same for pauser)? Something like 0.0002e18 shall work everywhere I think

script/deploy/deployAllContracts.sh
```diff
@@ -335,18 +342,44 @@ deployAllContracts() {
...
-      fi
-
       echo "Funding PauserWallet $PAUSER_WALLET_ADDRESS with $FUNDING_AMOUNT wei"
       universalCast "sendValue" "$NETWORK" "$ENVIRONMENT" "$PAUSER_WALLET_ADDRESS" "$FUNDING_AMOUNT" "$PRIVATE_KEY_TO_USE"
       checkFailure $? "fund PauserWallet $PAUSER_WALLET_ADDRESS on $NETWORK"
     fi
 
+    # --- Fund DevWallet ---
+    # DevWallet is used for executing timelock proposals on production and for staging deployments.
+    # It must have a native balance on every new network before execution scripts can run.
```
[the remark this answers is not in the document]
mirooon: very very good point

script/deploy/deployAllContracts.sh
mirooon: hmm, I think it may have been hardcoded before, and that amount probably wasn’t sufficient for some chains. Daniel likely has the most experience integrating new chains, so it’s worth checking with him. @0xDEnYO wdyt?

0xDEnYO: it wasnt hardcoded...I just left it intentional open cause chains differ in how much gas they use.
I didnt want to run in a situation where I believe that I funded the pauserWallet but then it doesnt have enough. 

Do you think it's sufficiently safe to hardcode that amount (across all the variety of chains we see (and yet have to see)?

gvladika: Thinking more about it, it's probably not safe to hard-code it. Some chains can get pretty congested and in those occasions funds requirement to issue TX raise. One option is to calculate this amount per chain on the fly like: `gasPrice * 300_000 (example gas needed) * safetyMultiplier`. But honestly, we only run this once in a while when deploying new chains, so keeping this step manual is not too bad.

mirooon: added default value as discussed on slack

gvladika (approved):

PR #1803, opened 2026-05-14
chore(skills): add create-pr skill

mirooon:
https://linear.app/lifi-linear/issue/EXP-458/build-create-pr-skill-that-consumes-pr-ready-as-pre-flight

Adds a `create-pr` Claude Code skill that automates branch → commit → push → PR. The skill:

- reads `.github/pull_request_template.md` verbatim and fills only specific named fields (Linear link, rationale, ticked checkboxes);
- resolves a Linear link from conversation context, falling back to the Linear MCP, and asks the user in chat (with a skip option) if neither finds a match — never fabricates;
- **verifies** before ticking the author checklist: walks `git diff main...HEAD` for the self-review tick; inspects `git log main..HEAD --oneline` for the single-concern tick;
- runs the test suite (`forge test` / `bun test:ts`) — the gap `.husky/pre-commit` deliberately leaves (the hook uses `forge build --skip test` and doesn't invoke `bun test:ts`). Lint / typecheck / build / solhint / secret scan are already enforced by pre-commit + `.agents/hooks/post-edit-validate.sh`, so the skill does not duplicate them;
- delegates to the `/pr-ready` sibling skill for local CodeRabbit pre-flight (mandatory per `.agents/rules/099-finish.md` once #1792 lands), and appends a `## /pr-ready deferred findings` section to the body when items are deferred/rejected;
- handles edge cases at step 1: no changes / detached HEAD / on `main` with local changes / branch already has an open PR;
- shows a pre-flight summary (self-review findings, check statuses, full filled-in body, `/pr-ready` commits added) and waits for explicit `y/n` before pushing;
- leaves the reviewer checklist unchecked and offers to post the PR to `#dev-sc-review` via `/post-pr-for-review`.

**Aikido findings reviewed**: two findings on the skill file (template-read at line 72; Linear MCP at line 76) — both reviewed as false positives. The PR template is a checked-in public file with no secrets; `.husky/pre-commit` scans for secrets + private keys before any commit lands (step 4); the step 9 pre-flight summary requires explicit `y/n` before push. Linear is the team's own authenticated ticket tracker via MCP — not a third-party data sink — and branch/commit text becomes publicly visible at `git push` to this public repo seconds later anyway.


.claude/skills/create-pr/SKILL.md
```diff
@@ -0,0 +1,203 @@
...
+- **Already on a feature branch with an open PR**: offer to just commit + push + update the existing PR instead of opening a new one.
+- **On `main`/`master` with local changes**: require a new branch (step 3); never commit to a protected branch.
+
+### 2. Determine branch name
+
+- If user supplied one, use it exactly.
+- Else derive from the primary change: `feat/<short-slug>`, `fix/<short-slug>`, `chore/<short-slug>`.
+- Never commit directly to `main` or `master` — create a new branch.
+
+### 3. Create branch (if needed)
```
0xDEnYO: should add that linear ticket ID is in branch name (ideally)

.claude/skills/create-pr/SKILL.md
```diff
@@ -0,0 +1,203 @@
...
+
+```bash
+git commit -m "<subject line>"
+```
+
+### 5. Build PR body from template (locally, before pushing)
+
+Read `.github/pull_request_template.md` verbatim. Fill in:
+
+- **Linear task link**: resolve in this order. Always include the issue ID (e.g. `EXSC-327`) somewhere in the body — Linear's GitHub integration uses it to auto-create the bidirectional link (PR appears in the ticket's "Links" sidebar; ticket appears in the PR's "Linked issues" panel). No extra Linear MCP call is needed for cross-linking.
```
0xDEnYO: If I am not mistaken: 

Linear only auto-updates tickets if a PR either
a) has the ticket ID in the branch name 
or 
b) says "Fixes EXSC-123" in the description (the word fixes is important)

.claude/skills/create-pr/SKILL.md
```diff
@@ -0,0 +1,203 @@
...
+`bun test:ts`). Per `.agents/rules/099-finish.md`, run them now:
+
+- **Solidity changes**: `forge test` (or `forge test --match-path` if scope is clear).
+- **TypeScript / JS changes**: `bun test:ts`.
+- **Docs / markdown / skill files only**: skip; note `N/A` in the summary.
+
+If anything fails, **stop and surface the failure** — do not push without explicit user
+override.
+
+### 8. Run `/pr-ready`
```
0xDEnYO: Maybe it makes sense to run this before running the test suite (if changes come up that will be fixed, it's important to run the test suite after the changes have been fixed).
Maybe makes sense to describe the flow/order a bit further.

.claude/skills/create-pr/SKILL.md
```diff
@@ -0,0 +1,203 @@
...
+
+```
+About to create PR on branch `<branch-name>`:
+
+Files committed (<N>):
+  • <file1>
+  • <file2>
+  ...
+
+Commit: "<commit message>"
```
0xDEnYO: which commit message is that? Of the last commit?

.claude/skills/create-pr/SKILL.md
```diff
@@ -0,0 +1,203 @@
...
+After the PR is created, ask:
+> PR created: <url>. Want me to post it to #dev-sc-review? (`/post-pr-for-review`)
+
+## Failure modes
+
+- **Uncommitted secrets / sensitive files**: warn and skip those files.
+- **Branch already has open PR**: surface URL, offer to push new commits instead.
+- **Push rejected (non-fast-forward)**: report error; do not force-push without explicit user instruction.
+- **No template found**: use minimal body with title + rationale only.
+- **Linear task unknown**: leave the section blank with an HTML comment (`<!-- No Linear task -->`) rather than fabricating a link.
```
0xDEnYO: I think we should switch to having a linear task for everything since it's nice and easy to create those now. So if there is no linear task, then maybe we should create one automatically? 
wdyt?

.claude/skills/create-pr/SKILL.md
mirooon: you are right: linear's gh integration only auto-links when the ID is either in the branch name or paired with a magic keyword in the title/body. The keyword set is broader than just `Fixes` though: `Fixes` / `Closes` / `Resolves` auto-close the ticket on merge, while Ref / References links without closing (useful for partial work). The skill now requires one of those routes and renders the body line as `Fixes <ID>` by default

fixed in: https://github.com/lifinance/contracts/commit/f9260149

.claude/skills/create-pr/SKILL.md
mirooon: yes! good point

fixed in: https://github.com/lifinance/contracts/commit/537a4bb7

.claude/skills/create-pr/SKILL.md
mirooon: good point! fixed in: https://github.com/lifinance/contracts/commit/e50050a7

.claude/skills/create-pr/SKILL.md
mirooon: good catch. The summary now shows a  commits since main list straight from git log main..HEAD --oneline, with the step-4 commit and any pr-ready: commits visible in order. I also folded the separate "/pr-ready commits added" sub-section into that unified list so there's only one source of truth.

expanded in: https://github.com/lifinance/contracts/commit/a87fb8ae

.claude/skills/create-pr/SKILL.md
mirooon: agree! added in: https://github.com/lifinance/contracts/commit/5b629293

To make sure that we still have human readable titles rather than title dirived from diffs I didn't go with fully automatic approach. The skill now offers to create one when none is found, with three options: e (default — show the proposed title and let you edit before creating), y (accept the proposed title as-is), s (skip). On accept it creates the ticket in EXSC to make sure that we still have human readable titles rather than title dirived from diffs

.claude/skills/create-pr/SKILL.md
```diff
@@ -0,0 +1,221 @@
...
+     - **Auto-accept** the top hit only if its title shares ≥3 meaningful tokens with the branch/commit AND its status is active (`statusType: started` or `unstarted`).
+     - **Ambiguous** (top hit doesn't pass the threshold, or several candidates look plausible): show the top 3 in chat with `ID — Title — status`, and ask which to link (with a "skip" option). Use `AskUserQuestion` if available; otherwise plain chat.
+  4. **Not found** — offer to create a new Linear ticket before falling back. Default action is **edit** so the user always sees the proposed title before anything is created:
+     - Propose in chat: `No Linear ticket found. Create one in EXSC?`
+       - `e` (default) — show the proposed title (`<derived from branch + commit subject>`) and let the user adjust before creating.
+       - `y` — create immediately with the proposed title as-is.
+       - `s` — skip; proceed without a ticket.
+     - On `y` / `e` (after the user confirms the edited title): call `mcp__claude_ai_Linear__save_issue` with `team: "SmartContract"`, the (edited) title, and a short body summarizing the change + a placeholder for the PR URL. Use the returned ID:
```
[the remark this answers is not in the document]
mirooon: good point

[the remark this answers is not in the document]
mirooon: The fair part of the flag was the git branch -m: the original prompt only asked about the ticket, and the rename was  described in a sub-bullet underneath, so a y arguably consented to ticket creation alone rather than both side-effects. Fixed in d369d75c — the prompt now lists both actions side-by-side with their concrete inputs (team + proposed title; current branch → proposed new branch), so a single y/e covers both unambiguously, and s explicitly leaves the branch as-is.

.claude/skills/create-pr/SKILL.md
```diff
@@ -0,0 +1,221 @@
...
+- **Linear task link or ID** (optional). If known from context, include in PR body.
+
+## Related conventions
+
+This skill creates PRs. For **editing** an existing PR's title/body, follow CLAUDE.md →
+"Creating and editing PRs via gh" (uses `gh api -X PATCH` with a JSON payload).
+
+The `/pr-ready` skill (`.agents/commands/pr-ready.md`) is a mandatory pre-flight per
+`.agents/rules/099-finish.md`: it runs CodeRabbit's local CLI against the branch and
+resolves findings before the PR is opened. This skill integrates `/pr-ready` as step 8
```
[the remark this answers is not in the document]
mirooon: godo point!

0xDEnYO (approved):

.cursor/commands/create-pr.md
```diff
@@ -0,0 +1 @@
+../../.agents/commands/create-pr.md
```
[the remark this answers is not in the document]
mirooon: The file is already a symlink:
  lrwxr-xr-x  .cursor/commands/create-pr.md -> ../../.agents/commands/create-pr.md
  git ls-files -s → 120000 ... .cursor/commands/create-pr.md   (120000 = symlink mode)
  It's tracked correctly as a symlink (mode 120000), exactly like its siblings (add-network.md, pr-ready.md, etc.). CodeRabbit reviews the diff blob content, where a Git symlink appears as a blob
  whose text is the link target (../../.agents/commands/create-pr.md). It misread that target-path blob as a plain text file. The convention in add-new-rule.md is already satisfied.

.agents/commands/create-pr.md
```diff
@@ -0,0 +1,237 @@
...
+```bash
+git push -u origin <branch-name>
+```
+
+### 11. Create PR
+
+Write body to a temp file and create via `gh`:
+
+```bash
+gh pr create --title "<title>" --body "$(cat /tmp/pr-body.md)" --base main --head <branch-name>
```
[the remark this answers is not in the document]
mirooon: Two reasons it doesn't hold up:
  - Race condition concern is irrelevant here. This is skill documentation for an interactive, single-user AI workflow — not a concurrent CI script. /tmp/pr-body.md is a placeholder path in prose,
  not a hot loop spawning parallel writers.
  - It contradicts the repo's own established convention. .agents/context.md:45 (the project's CLAUDE.md-loaded instructions) already standardizes on jq ... /tmp/pr-body.md > /tmp/payload.json.
  Changing only create-pr.md to mktemp would make the skill inconsistent with the documented house pattern.
  
  The one genuinely reasonable sub-point is that gh pr create supports --body-file directly, which is marginally cleaner than --body "$(cat ...)". But that's cosmetic and, again, would diverge from
  context.md. Not worth acting on in isolation.

PR #1106, opened 2025-04-17
harmonized ownership of periphery contracts (>> withdrawWallet)

0xDEnYO:
# Which Jira task belongs to this PR?
https://lifi.atlassian.net/browse/LF-13279

# Why did I implement it this way?

# Checklist before requesting a review

- [ ] I have performed a self-review of my code
- [ ] This pull request is as small as possible and only tackles one problem
- [ ] I have added tests that cover the functionality / test the bug
- [ ] For new facets: I have checked all points from this list: https://www.notion.so/lifi/New-Facet-Contract-Checklist-157f0ff14ac78095a2b8f999d655622e
- [ ] I have updated any required documentation

# Checklist for reviewer (DO NOT DEPLOY and contracts BEFORE CHECKING THIS!!!)

- [ ] I have checked that any arbitrary calls to external contracts are validated and or restricted
- [ ] I have checked that any privileged calls (i.e. storage modifications) are validated and or restricted
- [ ] I have ensured that any new contracts have had AT A MINIMUM 1 preliminary audit conducted on <date> by <company/auditor>

PR #1815, opened 2026-05-18
chore(deploy): harmonize Periphery ownership to refundWallet

0xDEnYO:
## Summary

Aligns the deploy scripts (and the healthCheck script) of the three Periphery contracts whose owner key was inconsistent with the agreed policy.

## Policy

All Periphery contracts that hold owner / withdraw authority use **`refundWallet`**, with two intentional exceptions:

- **FeeCollector** — keeps its dedicated `feeCollectorOwner` (on the way to deprecation, intentionally left as-is).
- **FeeForwarder** — keeps `withdrawWallet` (treasury reporting context).

## Changes

| Contract | Before | After |
|---|---|---|
| ERC20Proxy | `deployerAddress` (ephemeral — has produced a stale-EOA-owned mainnet deployment) | `refundWallet` |
| GasZipPeriphery | per-network `safeAddress` | `refundWallet` |
| Permit2Proxy | per-network `safeAddress` | `refundWallet` |

All six deploy script files (3 EVM + 3 zksync mirrors) updated to read `.refundWallet` from `config/global.json`, matching the canonical pattern used by `DeployExecutor.s.sol`, `DeployTokenWrapper.s.sol`, the receiver family, etc.

`script/deploy/healthCheck.ts` updated: the EVM + Tron ERC20Proxy ownership checks now expect `refundWallet` instead of `deployerWallet`. (`GasZipPeriphery` and `Permit2Proxy` have no ownership checks in healthCheck.) Note that the current mainnet ERC20Proxy is owned by a stale historical deployer EOA (`0x11f1…c00c`, not `deployerWallet`), so healthCheck was already failing on it today — this PR aligns the expectation; the migration will close the gap on-chain.

## Out of scope

Existing on-chain deployments keep their current owners. A migration script (`transferOwnership` per chain) is a separate follow-up — particularly important for mainnet `ERC20Proxy`.

## Context

Supersedes [#1106](https://github.com/lifinance/contracts/pull/1106), which became unrebaseable after main's `config.sh → .env` deploy-idiom migration. Full audit of the periphery surface was done before this PR — only the three contracts above were misaligned.

## Test plan

- [x] `forge build` clean.
- [x] `tsc --noEmit` clean for `healthCheck.ts`.
- [x] Local CodeRabbit pre-flight: `No findings ✔` on both commits.
- [ ] Next deployment of any of the 3 contracts on any chain verifies the constructor arg matches `refundWallet`.
- [ ] HealthCheck on a freshly redeployed network reports `ERC20Proxy owner is correct`.


script/deploy/facets/DeployERC20Proxy.s.sol
```diff
@@ -16,12 +19,21 @@ contract DeployScript is DeployScriptBase {
...
+            "/config/global.json"
+        );
+
+        // read file into json variable
+        string memory globalConfigJson = vm.readFile(globalConfigPath);
+
+        // extract refundWallet address
+        address refundWalletAddress = globalConfigJson.readAddress(
+            ".refundWallet"
+        );
```
mirooon, line 35: I think it's worth catching in this PR since you're already touching exactly the right spot.

`stdJson.readAddress(".refundWallet")` returns `address(0)` silently if the key is ever missing or renamed in `global.json`. And `TransferrableOwnership.constructor` doesn't check for zero (`src/Helpers/TransferrableOwnership.sol:25-27`). For the three contracts in this PR:

- **ERC20Proxy** — no inline check (`src/Periphery/ERC20Proxy.sol:19`) → would deploy bricked
- **GasZipPeriphery** — has the check, reverts `InvalidConfig` (L38-44) → safe
- **Permit2Proxy** — no check (L46-61) → would deploy bricked

"Bricked" being: `owner = address(0)`, so no `withdrawToken`, no `setAuthorizedCaller`, and `transferOwnership` can't save you either since `onlyOwner` wants `msg.sender == address(0)`. No recovery.

The repo already has the right place to assert this — `script/deploy/resources/deployRequirements.json`. Every other periphery contract with an owner arg has a `_owner` entry there with `allowToDeployWithZeroAddress: "false"` (ReceiverStargateV2, all the AcrossV3/V4 + Chainflip receivers, TokenWrapper, RelayerCelerIM, FeeCollector, FeeForwarder). ERC20Proxy has no entry at all, Permit2Proxy and GasZipPeriphery have entries but only declare the *other* constructor args — `_owner` is missing for all three.

Could you add the `_owner` entries pointing at `.refundWallet` with `allowToDeployWithZeroAddress: "false"`? At minimum for ERC20Proxy and Permit2Proxy where it actually matters — GasZipPeriphery's is just consistency since the contract already self-protects.

Tangent: probably also worth adding `if (_owner == address(0)) revert InvalidConfig();` inline to ERC20Proxy and Permit2Proxy at some point — that'd kill the class properly. But that's a source change, definitely not this PR. Can spin off a ticket if you want.

mirooon: tiny addendum — same gap exists for `Executor`. Constructor has no inline `_owner == 0` check (there's even a literal `// TODO: add constructor parameter validation on next contract update`) and no `_owner` entry in `deployRequirements.json` (only `contractAddresses.ERC20Proxy`). Out of scope for this PR, but worth a follow-up alongside the migration ticket.

script/deploy/healthCheck.ts
```diff
@@ -788,11 +788,11 @@ const main = defineCommand({
           client: publicClient,
         })
         const erc20ProxyOwner = await erc20ProxyContract.read.owner()
-        if (getAddress(erc20ProxyOwner) !== getAddress(deployerWallet))
+        if (getAddress(erc20ProxyOwner) !== getAddress(refundWallet))
           logError(
             `ERC20Proxy owner is ${getAddress(
               erc20ProxyOwner
-            )}, expected ${getAddress(deployerWallet)}`
+            )}, expected ${getAddress(refundWallet)}`
```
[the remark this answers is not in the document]
0xDEnYO: the healthcheck only runs for new networks so it's fine to change it now

script/deploy/facets/DeployERC20Proxy.s.sol
0xDEnYO: Good catch — addressed in d693c7c0. Added `_owner` → `global.json .refundWallet` with `allowToDeployWithZeroAddress: "false"` for ERC20Proxy (new entry), Permit2Proxy and GasZipPeriphery. Inline `address(0)` reverts in ERC20Proxy/Permit2Proxy left as the follow-up ticket as you suggested.

0xDEnYO: Pulled this in too — addressed in 96018c51. Same one-line entry, same risk class, didn't see a reason to defer it. The inline-revert + the source-side `TODO` cleanup still belong on the follow-up ticket.

mirooon (approved):

gvladika (approved):

PR #1823, opened 2026-05-19
fix(workflows): stop cancelled version-control runs from showing as failures

gvladika:

n/a — CI hygiene, no Linear task.

PRs were regularly showing red `version-control` rows in the Checks UI that turned green on a manual re-run. The "failures" are cancelled runs: GitHub renders cancellations as red rows next to the eventual green pass.

Survey of the last 100 `VersionControlAndAuditVerification` runs: 9 cancellations, 2 real failures (legitimate missing-audit-log errors, not addressed here).

Walking through each of the 9 cancellations and locating the killer run on the same SHA:

- **8 of 9** are killed by a `pull_request_review` event arriving within seconds of a `synchronize` (or another `pull_request`) on the same SHA. This check is a pure function of the PR diff, so review events can't change the outcome.
- **1 of 9** is two `pull_request` events on the same SHA, 14s apart — the title-bot's `edited` event re-triggering after this workflow's own `gh pr edit` updates the title.

All 9 are same-SHA, so two minimal trigger removals cover them:

- **Drop the `pull_request_review` trigger.** Removes the 8 review-driven cancellations.
- **Drop `edited` from the `pull_request` trigger types.** Breaks the title-bot's self-trigger loop. Human title edits don't change the contract diff, and the next `synchronize` catches anything content-relevant.

Concurrency key is left as PR-number-only with `cancel-in-progress: true`. An earlier draft of this PR added the head SHA to the key, but on closer inspection it wasn't load-bearing for any observed cancellation (all 9 are same-SHA) and would have opened a race in the downstream `audit-verification` job, which writes the `AuditRequired`/`AuditNotRequired`/`AuditCompleted` labels and has no concurrency block of its own. Thanks to @mirooon for catching this.

### DIFF
--- a/.github/workflows/versionControlAndAuditCheck.yml
+++ b/.github/workflows/versionControlAndAuditCheck.yml
@@ -25,7 +25,14 @@ name: VersionControlAndAuditVerification
 
 on:
   pull_request:
-    types: [opened, edited, synchronize, review_requested, ready_for_review]
+    types: [opened, synchronize, review_requested, ready_for_review]
+    # `pull_request_review` is intentionally not a trigger: this check is a pure
+    # function of the PR diff, so review events can't change the outcome and
+    # only create same-SHA duplicate runs that show up as cancelled (red) rows.
+    # `edited` is also intentionally omitted: this workflow updates the PR title
+    # itself via `gh pr edit`, which re-triggers an `edited` event on the same
+    # SHA and cancels the run in progress. Human title edits don't change the
+    # contract diff, and the next `synchronize` would catch any real change.
 
 permissions:
   contents: read # required to fetch repository contents
@@ -36,13 +43,8 @@ jobs:
   version-control:
     if: ${{ github.event.pull_request.draft == false && github.event.pull_request.base.ref == 'main' }}
     runs-on: ubuntu-latest
-    # Key concurrency by PR + head SHA so only redundant runs on the *same* commit
-    # cancel each other. A new commit (synchronize) gets its own group and runs to
-    # completion instead of being cancelled by — or cancelling — runs on the prior SHA.
-    # Without the SHA in the key, GitHub renders the cancelled run as a red "fail"
-    # row in the Checks UI even though nothing actually broke.
     concurrency:
-      group: sc-core-dev-approval-${{ github.event.pull_request.number }}-${{ github.event.pull_request.head.sha }}
+      group: sc-core-dev-approval-${{ github.event.pull_request.number }}
       cancel-in-progress: true
     env:
       CONTINUE: false # makes sure that variable is correctly initialized in all cases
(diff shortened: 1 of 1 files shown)


.github/workflows/versionControlAndAuditCheck.yml
```diff
@@ -38,8 +36,13 @@ jobs:
   version-control:
     if: ${{ github.event.pull_request.draft == false && github.event.pull_request.base.ref == 'main' }}
     runs-on: ubuntu-latest
+    # Key concurrency by PR + head SHA so only redundant runs on the *same* commit
```
mirooon: Walking through each of the 9 cancellations and finding the killer run on the same SHA, the SHA-in-key change actually doesn't help any of them — they're all same-SHA:

- **8 of 9** are eliminated just by dropping `pull_request_review` above (6 cancelled-by-review; 2 `pull_request` events killed by a later `pull_request_review` on the same SHA).
- **1 of 9** (sha `f0b11f7f`) is two `pull_request` events on the **same SHA**, 14s apart — the title-bot's `edited` event re-triggering after this workflow's own title update. Same SHA → same group either way, `cancel-in-progress` still kills the older run.

So the SHA-key change isn't load-bearing for the observed cancellations. It also opens a small new race: `audit-verification` has `needs: version-control` but no concurrency block of its own. Previously a new commit on the same PR cascaded a cancel into the old commit's `audit-verification`; now both run to completion and both mutate the PR's audit labels (`AuditRequired` / `AuditNotRequired` / `AuditCompleted`) — last writer wins. Usually harmless since consecutive commits touch the same contracts, but a fresh race window that didn't exist before.

FWIW the cross-SHA cancellation this change prevents was also never user-visible: the PR Checks UI tracks the head commit, so a cancelled run on a superseded commit only shows on that commit's own page.

Alternative — drop `pull_request_review` _and_ `edited` from the triggers, and keep the original PR-number-only concurrency key:

```yaml
on:
  pull_request:
    types: [opened, synchronize, review_requested, ready_for_review]

# ...
concurrency:
  group: sc-core-dev-approval-${{ github.event.pull_request.number }}
  cancel-in-progress: true
```

Removing `edited` breaks the title-bot self-trigger loop at the source. 9 of 9 cancellations addressed, no new races, no extra CI cost. Only thing you give up: re-running the version check when a _human_ edits the PR title — the next `synchronize` catches it anyway. Worth a sanity check that nothing else relies on the `edited` re-run though.

gvladika: Agreed on all points — applied your alternative in df45451. Dropped `pull_request_review` and `edited` from the triggers and reverted the concurrency key to PR-number-only, so `audit-verification` doesn't get a new label-write race. Thanks for the careful walkthrough.

mirooon (approved):

Matt-LiFi (approved):

PR #1828, opened 2026-05-20
chore(emergency-pause): RPC pacing/retries + staging EmergencyPauseFacet redeploy

mirooon:

N/A — operational follow-up to #1824 (emergency-pause staging hardening).

## Problem

The previous `script/utils/diamondEMERGENCYPauseStagingGitHub.sh` and `.github/workflows/diamondEmergencyPauseStaging.yml` had multiple gaps that would compromise the emergency-pause flow during a real incident:

**`diamondEMERGENCYPauseStagingGitHub.sh`:**

1. **No RPC retry/back-off.** RPCs can throttle reads under load — paid tiers included, and especially under the kind of incident-response load this script runs under. A single transient throttle was treated as terminal.
2. **Loose pre-pause classifier (`0x*` glob).** Any string starting with `0x` was accepted as an unpaused address — garbage like `0xnot-actually-an-address` would be misclassified as "diamond not paused" and the script would proceed to send `pauseDiamond()`.
3. **No fail-closed on unknown state.** If the pre-pause `owner()` read returned neither a clean address nor a `DiamondIsPaused` selector (e.g. RPC error), the script still sent `pauseDiamond()` blindly. In incident response, sending a tx to an unknown-state diamond is dangerous.
4. **Final pause check relied on `cast`'s exit code.** Some RPCs return slightly stale state right after a pause tx is mined (read-after-write lag); `cast` could exit 0 against the stale read and the script would wrongly report a successful pause as a failure.
5. **Bash 64-bit integer overflow on balance check.** Balances above ~9.2 ETH-in-wei overflow bash arithmetic, producing wrong comparisons. Today's small balances mask the bug; it would fail silently as soon as the pauser is funded.
6. **Broken exit-code aggregation.** `main()` called `wait` (no-arg) then `for JOB in $(jobs -p)`. After `wait` reaps background jobs, `jobs -p` returns empty — per-network failures never surfaced in the workflow exit code. The workflow appeared green even when individual networks failed.
7. **Send-loop exhaustion was always terminal.** If all `pauseDiamond()` send attempts failed, the script returned 1 even when another caller (Safe tx, runbook) had paused the diamond concurrently — a false negative under race conditions.
8. **`local` shadowing of `.env` `PRIVATE_KEY`.** `local PRIVATE_KEY=$2` shadowed the `.env` global; subshell helpers (`$(...)` calls) saw an empty key.

**`diamondEmergencyPauseStaging.yml`:**

9. **Slack alert silently skipped on script failure.** The Slack step had no `if:` clause; GitHub Actions' default skip-on-previous-failure hid the alert exactly when on-call most needed it.
10. **Slack message had no status indicator.** Even when the alert fired, the message said "executed" without success/failure context.
11. **Slack message had no deep-link URL.** On-call had to manually navigate to the Actions tab to find the run.
12. **Step name `Send Reminder to Slack …` was inaccurate.** This is a notification of a just-happened event, not a reminder.

## Solution

Numbered to match the problems above:

1. **`rpcCallWithRetry` helper** — retries an RPC-touching command up to `RPC_MAX_ATTEMPTS=5` with `RPC_RETRY_SLEEP_SECONDS=3` between attempts. Captures stdout cleanly via `mktemp` for stderr so callers receive clean values while diagnostics still appear in retry logs / the final error.
2. **Tightened pre-pause regex** — `^0x[0-9a-fA-F]{40}$` requires exactly an address shape. Anything else falls through to fail-closed (#3).
3. **Pre-pause check fails closed** — retries `owner()` until the response is unambiguously address-shaped or matches the `DiamondIsPaused` selector / error name. After retries are exhausted, the script aborts that network — does not send `pauseDiamond()` on unknown state.
4. **Final pause check inspects response content, not `$?`** — retries until it sees the `DiamondIsPaused` selector in the response stream. Decouples the script's decision from RPC read-after-write lag.
5. **Balance check uses `bc`** for arbitrary-precision comparison and validates `^[0-9]+$` before piping into `bc`. Mirrors the fix already shipped in `fundStagingPauserWalletGitHub.sh`.
6. **PID-based exit aggregation** — `main()` captures each `handleNetwork &` PID into an array and aggregates via `wait $PID` per PID. Per-network failures now propagate to the workflow exit code.
7. **Fall-through to on-chain truth on send-loop exhaustion** — if all `MAX_ATTEMPTS=10` `pauseDiamond()` sends fail, the script falls through to the final-pause check. If the chain reports paused (regardless of who paused it), the script returns 0; otherwise 1.
8. **Renamed local parameter to `PAUSER_PRIVATE_KEY`** to avoid shadowing the `.env` global (per `.agents/rules/300-bash.md`). All in-function uses now go through the local parameter; `main()` reads the workflow secret once and threads it through as the function argument.
9. **`if: ${{ always() }}` on the Slack step** so it fires on any exit code.
10. **Status indicator in the message body** — `Success ✅` on `job.status == 'success'`, `Failed ❌` otherwise.
11. **Job-deep-link URL in the message body** — a new "Resolve current job URL" step calls `gh api /repos/.../actions/runs/{run_id}/jobs`, filters by `${{ github.job }}`, and writes the URL to `$GITHUB_OUTPUT`. Falls back to the run-level URL on API failure so the link is always usable. Required adding `actions: read` to the workflow permissions.
12. **Step renamed** to `Notify Slack SC-general channel of pause execution`.

The Slack payload format was also switched from YAML to JSON so the inline ternary for the status emoji interpolates without YAML's single-quote escaping rules.

## Relationship to prod

Staging and prod hit the **same paid RPC tier** — staging is not a parallel system with different network characteristics. Throttling observed during E2E runs (see Cycle B below) is something the script can hit on any RPC under load, which is precisely the load this script runs under (incident response). The pacing/retry logic added here is defensive coding that belongs in prod too; it is not staging-specific.

The follow-up PR will port these changes **line-for-line** into the production mirror scripts (`script/utils/diamondEMERGENCYPauseGitHub.sh` + `.github/workflows/diamondEmergencyPause.yml`), so staging and prod end up running the same script with only environment-specific inputs differing. Staging is hardened first as the lower-stakes preview; once parity ships, "tested in staging" means "tested the exact bytes that will run in prod."

## E2E testing

Three cycles executed via the actual GitHub Actions workflow against the staging diamonds on `bsc/arbitrum/optimism/base`. State reset between cycles via `script/tasks/unpauseAllDiamonds.ts`.

| Cycle | Scenario | Observed outcome | Exit | Validates |
|---|---|---|---|---|
| A | All 4 diamonds already paused going in | 4× `The diamond is already paused.`; **no `pausing diamond` line on any network**; `printStatus` reports 4× paused | 0 | Tightened pre-pause classifier correctly identifies paused state; idempotent re-run |
| B | Failure isolation: bogus RPC on optimism, healthy on the other 3 | optimism: 5 retries of `pre-pause owner() returned unclear response`, then `RPC or network error after 5 attempts: dns error` — **no `pausing diamond` line for optimism**. bsc/base: paused on attempt 1. arbitrum: needed 8/10 send attempts due to `max fee per gas less than block base fee` errors but eventually paused. Runtime: 2m 42s. | 1 | Fail-closed pre-pause check, per-network isolation, PID-based exit aggregation, RPC retry budget |
| C | Mixed-state: 3 paused from B + 1 unpaused (optimism, which had failed in B) | bsc/arbitrum/base hit the already-paused fast path; optimism: full normal flow → balance check → pauserWallet check → pause sent → confirmed | 0 | Parallel subshells stay isolated through divergent branches; aggregation correctly returns 0 in the all-good case |

Run links (visible to lifinance org members):

- Cycle A — [`actions/runs/26278638669/job/77349016074`](https://github.com/lifinance/contracts/actions/runs/26278638669/job/77349016074)
- Cycle B — [`actions/runs/26297165974/job/77412709187`](https://github.com/lifinance/contracts/actions/runs/26297165974/job/77412709187)
- Cycle C — [`actions/runs/26298126706/job/77416027347`](https://github.com/lifinance/contracts/actions/runs/26298126706/job/77416027347)

Cycle B is the headline test for this PR's per-network isolation claim. **It also surfaced problem #9** (Slack silence): `#sc-general` received no notification because the unguarded Slack step was skipped on exit 1. Cycle A and Cycle C both fired Slack (script exited 0). The Slack improvements in this PR resolve that — the next failing run will produce a Slack message containing `Failed ❌` and a clickable deep-link straight to the job log view.

- [x] I have added tests that cover the functionality / test the bug (E2E cycles A/B/C above; no automated unit tests — this is a bash + workflow change validated against real staging diamonds)


script/utils/diamondEMERGENCYPauseStagingGitHub.sh
```diff
@@ -25,7 +73,9 @@ function handleNetwork() {
   echo ""
   echo ">>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>> start network $1 >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>"
   local NETWORK=$1
-  local PRIVATE_KEY=$2
+  # Renamed from `PRIVATE_KEY` to avoid shadowing the `.env` PRIVATE_KEY global
+  # (per .agents/rules/300-bash.md "No local shadowing of .env globals").
+  local PAUSER_PRIVATE_KEY=$2
```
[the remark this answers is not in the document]
mirooon: good point!

script/utils/diamondEMERGENCYPauseStagingGitHub.sh
```diff
@@ -17,6 +17,54 @@ DIAMOND_IS_PAUSED_SELECTOR="0x0149422e"
 # the number of attempts the script will max try to execute the pause transaction
 MAX_ATTEMPTS=10
 
+# RPC pacing for read calls. Send retries use MAX_ATTEMPTS above (intentionally
+# larger). Some public RPCs throttle reads aggressively (~1 call per few seconds),
+# so we pace and retry transient failures.
+#
+# Why staging-only (and not in diamondEMERGENCYPauseGitHub.sh): prod uses paid
+# RPCs that don't throttle under normal load. If prod RPCs ever start throttling
```
0xDEnYO: why are not using exactly the same setup (incl RPC URLs) for staging and PROD?
Also, "if this ever happens in PROD"...then it's too late. Why do we not add it as a standard fallback / security measure if the code is already in our codebase?

mirooon: thank you. the description was wrong - fixed it. staging and prod will be using same logic

0xDEnYO (approved):

maxklenk (approved):

PR #1831, opened 2026-05-20
chore(emergency-pause): replace pauser/pause checks with single checkEmergencyPausePerNetwork

mirooon:

N/A (Phase 2b emergency-pause tooling)

Replaces both `script/tasks/temp/checkPauserWalletPerNetwork.ts` (already on `main`) and the originally-proposed `checkPauseStatusPerNetwork.ts` with a single tool, `checkEmergencyPausePerNetwork.ts`. Resolves the heavy-overlap review feedback by removing the duplicate facet/pauser columns and the drift surface (the two scripts already differed on Tron strategy, rate-limit posture, and selector-check granularity).

Run: `bunx tsx ./script/tasks/temp/checkEmergencyPausePerNetwork.ts`

## What it checks, per `(network, environment)` pair (production + staging)

1. **EmergencyPauseFacet installed** — `DiamondLoupe.facetAddress(selector)` for all four selectors (`pauseDiamond`, `removeFacet`, `unpauseDiamond`, `pauserWallet`). Catches partial diamondCuts that a single-selector probe would miss.
2. **On-chain `pauserWallet()` matches config** — Tron-aware: EVM networks compare against `config/global.json → pauserWallet`, Tron networks against `tronWallets.pauserWallet` (base58 normalised via `normalizeAddressForNetwork`).
3. **Pauser wallet native balance** — viem's `getBalance` over TronGrid's `/jsonrpc` bridge for Tron, formatted with the correct decimals per network (TRX = 6, ETH = 18). An out-of-gas pauser will be visible.
4. **Current pause state** — `owner()` revert probe; a revert that matches the `DiamondIsPaused()` custom-error selector (`0x0149422e`) means the diamond is paused (`OwnershipFacet` is removed during pause, only `EmergencyPauseFacet` remains exposed). Classifies into `PAUSED ⛔` / `NOT PAUSED ✓` / `ERROR`.

## Output

Color-coded table with one row per `(network, environment)` pair, sorted `PAUSED → ERROR → NOT_PAUSED` so anything off-baseline is at the top. PAUSED rows always stay red — that's the headline signal — and in-row markers (`✗ MISMATCH`, `✗` selector flags, `Error` in the balance column) carry the per-cell specifics.

## Use cases

- Before/after an emergency-pause drill — one-shot snapshot of every prod + staging diamond.
- After deploying `EmergencyPauseFacet` to a new network — verifies all four selectors actually registered.
- After rotating `PRIV_KEY_PAUSER_WALLET` — confirms each diamond holds the new address on-chain and that the address has gas.

## Tron support

Diamond address comes from `deployments/tron.json` in base58 (`T…`) and is converted to its 20-byte EVM-equivalent via `normalizeAddressForNetwork`. The viem client talks to TronGrid through its JSON-RPC bridge (`<rpc>/jsonrpc`), and `getTransportConfigFromRpcUrl` injects the `TRONGRID_API_KEY` header when present.


script/tasks/temp/checkPauseStatusPerNetwork.ts
```diff
@@ -0,0 +1,417 @@
...
+    const addr = (r.diamondAddress ?? '').padEnd(addrWidth)
+    const line =
+      `${r.network.padEnd(networkWidth)} | ${String(r.chainId).padStart(8)}` +
+      ` | ${addr} | ${facetStr} | ${pauserStr} | ${statusStr}`
+
+    // Highlight rows that need attention: facet missing or pauser mismatch
+    const needsAttention =
+      r.emergencyPauseFacetInstalled === false ||
+      r.pauserWalletMatchesGlobal === false
+    const lineColor = needsAttention ? 'yellow' : color
```
gvladika: `const lineColor = needsAttention ? 'yellow' : color` lets a pauser-mismatch downgrade a PAUSED row from red to yellow. A paused production diamond should never lose its red — that's the whole headline signal this script exists to surface. Suggest flipping precedence so PAUSED always wins, and showing the mismatch as an extra marker on the row instead of replacing the color.

script/tasks/temp/checkPauseStatusPerNetwork.ts
```diff
@@ -0,0 +1,417 @@
...
+  }
+  results.sort((a, b) => {
+    const diff = statusOrder[a.status] - statusOrder[b.status]
+    return diff !== 0 ? diff : a.network.localeCompare(b.network)
+  })
+
+  printTable(results)
+}
+
+main()
```
gvladika: General comment: 
Heavy overlap with `script/tasks/temp/checkPauserWalletPerNetwork.ts`, which already checks EmergencyPauseFacet installation (all four selectors, not just pauserWallet()) and compares on-chain pauserWallet() to config/global.json. The genuinely new capability here is the owner() revert probe for pause state. Per the repo's "replace, don't deprecate" stance, we could instead: (a) extend `checkPauserWalletPerNetwork.ts` with a Pause-Status column and drop this file, or (b) keep this script but strip the duplicated facet/pauser columns so each tool owns one concern. As-is the two will drift (already differ on Tron strategy, rate-limit posture, and selector-check granularity).

script/tasks/temp/checkPauseStatusPerNetwork.ts
mirooon: Good catch — fixed in 87c83d16. `pauserWallet()` is callable while paused (EmergencyPauseFacet stays exposed during pause), so a PAUSED + mismatch row was indeed reachable and would silently demote red → yellow. Changed the override so PAUSED always wins; the in-row `✗ MISMATCH` / `✗ NOT INSTALLED` markers carry the specifics:

```ts
const lineColor = r.status !== 'PAUSED' && needsAttention ? 'yellow' : color
```

script/tasks/temp/checkPauseStatusPerNetwork.ts
mirooon: You were right — went one step further than either of your options and **merged both files into one**.

## What changed

- ❌ `script/tasks/temp/checkPauserWalletPerNetwork.ts` — deleted (was on `main`)
- ❌ `script/tasks/temp/checkPauseStatusPerNetwork.ts` — deleted (was the original add of this PR)
- ✅ `script/tasks/temp/checkEmergencyPausePerNetwork.ts` — new single tool covering all four checks per `(network, environment)`: all-4-selectors facet probe, on-chain `pauserWallet()` match, pauser native balance, and the `owner()` revert probe for pause state.

Commits: `2447dee7` (merge), `a9b906f5` (`pr-ready:` fix for Tron balance decimals — `formatEther` would have under-reported TRX by 1e12×). Net diff: **+595 / −1188**. PR title + body reframed.

## Response to your (a) / (b) suggestion

Both your options would have resolved the duplication, but neither would have fully resolved the drift surface:

| | Drift surface left | Naming |
|---|---|---|
| (a) Extend `checkPauserWalletPerNetwork.ts` + drop the new file | None (one file) | Old name retained, but its scope now exceeds "pauser wallet" |
| (b) Keep both, strip the new one's facet+pauser columns | **Two files still exist** → Tron transport, rate-limit pacing, and per-tool error handling can still diverge over time | Both names accurate to their narrowed scopes |
| **(c) Merge both into `checkEmergencyPausePerNetwork.ts`** (what we did) | **None** (one file) | New name reflects the broader scope — covers facet + pauser + balance + pause-state |

Picked (c) because:
- The tool's job is materially broader than "check pauser wallet" — keeping the old name would have been misleading.
- The `check…PerNetwork.ts` pattern matches the four neighbours in `temp/`, so it's not an odd-one-out.
- Zero drift surface (your underlying concern) is now mathematically guaranteed: there's only one file, one Tron strategy (viem-via-`/jsonrpc`), one rate-limit posture (shared `sleep(INTER_CALL_DELAY)`), one selector-granularity (all 4 selectors).
- The Tron-aware pauser comparison (`globalConfig.tronWallets.pauserWallet` via `normalizeAddressForNetwork`) is a small improvement over both originals, which compared EVM-hex on Tron.

If you'd prefer the file keep the name `checkPauserWalletPerNetwork.ts` to make the "replace, don't deprecate" framing literal (preserving git history of the older file), happy to `git mv` in a follow-up — but the underlying drift / duplication concern is already gone either way.

gvladika (approved):

PR #1853, opened 2026-05-26
fix(timelock-exec): throttle Tron RPC calls to TronGrid 15 req/s cap

gvladika:

[EXSC-354](https://linear.app/lifi-linear/issue/EXSC-354/throttle-tron-rpc-calls-in-timelock-execution-to-stay-under-trongrid)

TronGrid limits API keys to 15 requests/second and suspends the key for ~25
seconds when exceeded. The Tron timelock execution job was firing the three
readiness reads (`isOperationDone` / `isOperationPending` / `isOperationReady`)
via `Promise.all` per pending op, then running the simulate+broadcast loop
without inter-op pacing. With a queue of 7+ ops, the first parallel read burst
alone exceeds 15 req/s, the key gets suspended, and every subsequent call
fails — both in CI and locally.

Fix is two narrow changes, both gated on `isTronNetworkKey()` so EVM
throughput is unchanged:

- `checkOperationStatus()` runs the three reads sequentially with a 100ms
  gap on Tron (≈10 req/s ceiling for the readiness phase).
- The `for (const operation of readyOperations)` body is wrapped in braces
  and pauses 200ms between iterations on Tron (≈7 req/s ceiling for
  simulate+broadcast).

**Why in the script rather than `tron-caller.ts`**: the rate limit isn't a
property of the chain — TronGrid happily serves 15 req/s. The problem is the
orchestrator firing 20+ in a burst. Other Tron scripts in the repo already
handle their own throttling, so a global throttle in `TronChainCaller` would
risk double-throttling them and impose an always-on tax on one-shot callers
that don't need it. Keeping the fix where the burst is generated is more
honest and surgical; the next workflow run benefits automatically.

Verified locally: a 7-op Tron queue that previously failed on the first
parallel read burst now clears end-to-end with this change.

**Follow-ups (not in this PR)**:

- Move the two sleep constants into `script/deploy/tron/constants.ts`
  alongside other Tron rate-limit knobs.
- Add a retry-on-429 helper in `tron-caller.ts` that reads the
  `suspended for X s` body from TronGrid's 429 and backs off accordingly.
  Would harden the workflow against bursts that do still slip through.

### DIFF
--- a/script/deploy/safe/execute-pending-timelock-tx.ts
+++ b/script/deploy/safe/execute-pending-timelock-tx.ts
@@ -9,6 +9,7 @@
 
 import 'dotenv/config'
 
+import { isTronNetworkKey } from '@lifi/tron-devkit'
 import { defineCommand, runMain } from 'citty'
 import { consola } from 'consola'
 import type { Address, Hex, PublicClient } from 'viem'
@@ -22,6 +23,7 @@ import {
   type SupportedChain,
 } from '../../common/types'
 import { setupEnvironment } from '../../demoScripts/utils/demoScriptHelpers'
+import { sleep } from '../../utils/delay'
 import { getDeployments } from '../../utils/deploymentHelpers'
 import { normalizeAddressForNetwork } from '../../utils/normalizeAddressStringForViem'
 import {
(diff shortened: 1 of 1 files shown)


script/deploy/safe/execute-pending-timelock-tx.ts
```diff
@@ -416,17 +418,46 @@ const cmd = defineCommand({
...
   isReady: boolean
 }> {
+  if (isTronNetworkKey(networkName)) {
+    const isDone = await publicClient.readContract({
+      address: timelockAddress,
+      abi: TIMELOCK_ABI,
+      functionName: 'isOperationDone',
+      args: [operationId],
+    })
+    await sleep(100)
```
mirooon: tbh from my experience, that might be too short for `tron` unfortunately. I’d say 2000 should be fine to be on the save side

script/deploy/safe/execute-pending-timelock-tx.ts
```diff
@@ -654,6 +685,10 @@ async function processNetwork(
         else if (result === 'rejected') operationsRejected++
         else if (result === 'skipped') operationsSkipped++
       }
+      // TronGrid caps the API key at 15 req/s; each op fires ~3-5 RPC calls
+      // (simulate + broadcast + receipt poll), so pace iterations on Tron.
+      if (isTronNetworkKey(network.name)) await sleep(200)
```
mirooon: same here

script/deploy/safe/execute-pending-timelock-tx.ts
gvladika: Done

script/deploy/safe/execute-pending-timelock-tx.ts
gvladika: Done

mirooon (approved):

PR #1847, opened 2026-05-26
EXSC-254 Make PolymerCCTPFacet chainIdToDomainId mapping dynamically updatable [PolymerCCTPFacet v2.1.0]

0xDEnYO:
# Which Linear task belongs to this PR?

https://linear.app/lifi-linear/issue/EXSC-254/make-polymerfacet-chainidtodomainid-mapping-dynamically-updatable

# Why did I implement it this way?

PolymerCCTPFacet previously hardcoded CCTP domain IDs in `_chainIdToDomainId`, requiring a facet redeploy (and audit cycle) for every new network. This replaces that with owner-managed diamond storage, mirroring the DeBridge pattern:

- **`initPolymerCCTP(ChainIdConfig[])`** — batch seed from `config/polymercctp.json` `mappings` (also sets max USDC approval for the TokenMessenger)
- **`setChainIdToDomainId(ChainIdConfig[])`** — add/update mappings via Safe without redeploy (batch)
- **`unsetChainIdToDomainId(uint256)`** — remove a single mapping
- **`getChainIdToDomainId(uint256)`** — onchain verification after init

Mappings store `domainId + 1` so domain `0` (Ethereum) is distinguishable from unset entries.

**Migration note:** existing PolymerCCTP deployments upgrading to v2.1.0 must call `initPolymerCCTP` before bridging works again.

# Checklist before requesting a review

- [x] I have performed a self-review of my code
- [x] This pull request is as small as possible and only tackles one problem
- [ ] I have run `/pr-ready` (local CodeRabbit) on this branch and resolved (or explicitly documented) all findings — see `.agents/commands/pr-ready.md`
- [x] I have added tests that cover the functionality / test the bug
- [ ] For new facets: I have checked all points from this list: https://www.notion.so/lifi/New-Facet-Contract-Checklist-157f0ff14ac78095a2b8f999d655622e
- [x] I have updated any required documentation

# Checklist for reviewer (DO NOT DEPLOY and contracts BEFORE CHECKING THIS!!!)

- [ ] I have checked that any arbitrary calls to external contracts are validated and or restricted
- [ ] I have checked that any privileged calls (i.e. storage modifications) are validated and or restricted
- [ ] I have ensured that any new contracts have had AT A MINIMUM 1 preliminary audit conducted on <date> by <company/auditor>

Made with [Cursor](https://cursor.com)

PR #1857, opened 2026-05-27
fix(husky): unbreak pre-commit hook (silent failure + lint-staged invocation)

0xDEnYO:
## Summary

Fixes two latent bugs in the husky v8 pre-commit hook that combine to let bad commits slip through silently. Discovered while investigating why a TypeScript type-mismatch in PR #1847 wasn't caught locally despite the hook's `bunx tsc-files --noEmit` step.

### Bug 1: `bun lint-staged` → `bunx lint-staged`

`bun lint-staged` (line 524) tries to run a script named `lint-staged` from `package.json` → `scripts`, which doesn't exist — only the lint-staged config + devDep do. Bun reports `Script not found "lint-staged"` and aborts the lint-staged step.

Fix: invoke via `bunx lint-staged`, which resolves the binary from `node_modules`.

### Bug 2: silent syntax-error commits (the bigger one)

`.husky/_/husky.sh` re-execs the hook under `sh -e "$0"`. On macOS, `sh` is bash-in-POSIX-mode, and the hook uses bash-only syntax (process substitution `< <(...)`, array `+=`, `[[ ]]`). The first bash-only token inside a function body triggers a parse-time syntax error — **but the error fires AFTER git has already accepted the commit**, producing the misleading sequence:

```text
.husky/pre-commit: line 381: syntax error near unexpected token '<'
[branch hash] commit subject
 1 file changed, ...
```

Operators see "syntax error" and assume the commit was rejected; it wasn't. **All hook checks (tsc, lint-staged, secret scan) that run inside bash-only function bodies are silently skipped** when this happens.

Fix: replace the `husky.sh` source with an inline re-exec under `bash -e`, replicating the small bits of `husky.sh` init we actually use (`HUSKY=0` skip, `~/.huskyrc` sourcing). Standard `husky_skip_init` guard prevents recursion.

## Verification

End-to-end test: after this change, commits that touch only TS files run the full pre-commit pipeline (tsc-files + lint-staged + secret scan) and abort cleanly on type errors. Demonstrated locally — `bunx tsc-files --noEmit script/tasks/proposePolymerCCTPChainIdMappings.ts` against the buggy state of PR #1847 produces the type error that the hook was supposed to surface:

```text
TS2345: Argument of type '"PRIVATE_KEY_PRODUCTION" | "PRIVATE_KEY_STAGING"' is not
assignable to parameter of type '"PRIVATE_KEY_PRODUCTION" | "SAFE_SIGNER_PRIVATE_KEY" | undefined'.
```

## Test plan

- [ ] Pull this branch, run `bun install` to refresh husky
- [ ] Stage a `.ts` file with a deliberate type error → commit should be **rejected** (not just print "syntax error" and succeed)
- [ ] Stage a clean `.ts` file → commit should succeed with the full pipeline output (tsc-files ✓, lint-staged ✓, secret check ✓)
- [ ] No regression in `HUSKY=0` skip or `~/.huskyrc` sourcing


.husky/pre-commit
```diff
@@ -1,6 +1,20 @@
 #!/usr/bin/env bash
-export HUSKY_SHELL="${HUSKY_SHELL:-bash}"
-. "$(dirname -- "$0")/_/husky.sh"
```
gvladika, line 3: Dead husky.sh now -- pre-commit was the only hook sourcing .husky/_/husky.sh

0xDEnYO: Good catch — fixed in c065b96. Removed the tracked file and dropped the !husky.sh allow-rule in .husky/_/.gitignore so the locally regenerated copy (from husky install) stays ignored. CodeRabbit local review clean.

gvladika (approved):

maxklenk (approved):

PR #1865, opened 2026-05-28
feat(EXP-428): add SupersetFacet v1.0.0 [GenericErrors v1.0.4, SupersetFacet v1.0.0, IOmniTokenAddressBook v1.0.0, ISupersetHubPoolManager v1.0.0, ISupersetPoolManager v1.0.0, ISupersetSpokePoolManager v1.0.0]

gvladika:

[EXP-428
](https://linear.app/lifi-linear/issue/EXSC-329/create-sc-implementation)

Adds `SupersetFacet` v1.0.0 — a hub-and-spoke cross-chain integration with Superset Labs' virtual-pool DEX (hub on Arbitrum, spokes on Base and Unichain).

A few design decisions worth calling out:

- **Single facet, two ABIs**. Superset's `SpokePoolManager` (9-arg `multiHopSwapWithOutputChain`) and `HubPoolManager` (7-arg version) differ in shape because the hub flow doesn't propagate `refundAddress` or `options`. Rather than ship two facets, the constructor reads `block.chainid` and sets an `IS_HUB` immutable; `_startBridge` picks the right call shape from the same facet code. This keeps one selector pair across all chains and avoids deploying two near-identical facets.
- **`refundAddress` from `_supersetData`, not `msg.sender`** — and refunds are routine, not exceptional. The `lzFee` field is an off-chain LayerZero quote that drifts with destination-chain gas price between quote time and execution, so the backend deliberately overpays `msg.value` by ~20% to absorb the drift — meaning a refund of the unused buffer happens on essentially every call, not just on failure. For Permit2 / smart-account flows the diamond's `msg.sender` can be the Permit2Proxy or a session-key wallet, neither of which is a recoverable destination. The facet therefore routes source-side native excess (the LZ buffer + any `_depositAndSwap` swap leftovers) to `_supersetData.refundAddress` — the same address Superset itself uses for its async failure refunds — so refund delivery is consistent end-to-end and the user actually receives the change.
- **Strict `fallbackEoA` validation**. Both Superset (`SwapDelivery.sol:30`) and our facet reject `fallbackEoA.code.length != 0`. We duplicate the check because a clean revert at the facet boundary is cheaper than walking through `safeTransferFrom` + spoke validation. This catches EIP-7702 delegated EOAs at the source — which the demo run on Arbitrum confirmed is a real risk for accounts that have been smart-account-upgraded.
- **Tests on a Base mainnet fork**. Most of `SupersetFacetTest` exercises the live `SpokePoolManager` at `0x57C155a15a9CA0A6C1F759ac6988b4fCa3663Ea4`, so arg-forwarding correctness is implicit (any wrong arg makes the real spoke revert). The hub-branch test still uses a mock — a single Foundry test file can only attach to one fork chain. Replacing that with an Arbitrum fork is a follow-up.

End-to-end validated on mainnet: see the three recorded source/destination tx pairs in `script/demoScripts/demoSuperset.ts:7-13` (Base→Unichain spoke source; Arbitrum→Base hub source; Base→Unichain with WETH→USDC pre-swap).

**Deferred follow-ups**:
- Replace `MockSupersetHubPoolManager` with a real Arbitrum fork in a separate test file.


script/demoScripts/demoSuperset.ts
```diff
@@ -0,0 +1,345 @@
...
+  amount: string // human-readable; ignored when preSwap is present
+  preSwap?: IPreSwap
+  superset: ISupersetParams
+}
+
+const SHARED = {
+  receiver: '0x85A8F3cf6d255BD497Bd1Dd83a338Cce0B14C3B3' as Address,
+  refundAddress: '0x85A8F3cf6d255BD497Bd1Dd83a338Cce0B14C3B3' as Address,
+  fallbackEoA: '0x34E7db45783b50F4e7764258d0Dc0400c3539A57' as Address,
+  deadlineSeconds: 1800, // 30 minutes
```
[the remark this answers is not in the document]
gvladika: Values are intentionally hard-coded. Common dev addresses are used. This is script will not be used outside of the team, it is for single-use demo purposes only.

src/Facets/SupersetFacet.sol
```diff
@@ -0,0 +1,228 @@
...
+        if (msg.value < _supersetData.lzFee) {
+            revert InsufficientNativeValue();
+        }
+    }
+
+    /// @dev Bridge execution: approves the pool manager, then calls the hub or spoke
+    ///      ABI depending on `IS_HUB`.
+    /// @param _bridgeData Core LI.FI bridge data
+    /// @param _supersetData Superset-specific parameters
+    function _startBridge(
```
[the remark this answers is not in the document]
gvladika: Both functions calling _startBridge are non-reentrant

script/deploy/facets/UpdateSupersetFacet.s.sol
```diff
@@ -0,0 +1,39 @@
...
+        SupersetFacet.ChainIdConfig[] memory chainIdConfigs = abi.decode(
+            rawMappings,
+            (SupersetFacet.ChainIdConfig[])
+        );
+
+        return
+            abi.encodeWithSelector(
+                SupersetFacet.initSuperset.selector,
+                chainIdConfigs
+            );
```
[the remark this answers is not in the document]
gvladika: Not a bug — this matches the established repo convention for facets with an `init*` function. Every other update script does the same exclude:

- `UpdatePolymerCCTPFacet.s.sol` excludes `initPolymerCCTP`
- `UpdateOptimismBridgeFacet.s.sol` excludes `initOptimism`
- `UpdateMegaETHBridgeFacet.s.sol` excludes `initMegaETH`
- `UpdateCelerCircleBridgeFacet.s.sol` excludes `initCelerCircleBridge`
- `UpdateHopFacet.s.sol` excludes `initHop`

The init function is invoked atomically by `diamondCut` via its `_init`/`_calldata` payload (one-shot, seeds state during the cut). The selector is intentionally not registered post-cut; ongoing updates go through `setChainIdToEid` (which IS exposed as a callable selector). Tests work because the test diamond cut registers the selector for scaffolding — production doesn't.

`PolymerCCTPFacet` is the closest analogue and ships the same exclude + same "Admin Methods" doc framing. The doc wording was slightly ambiguous, so commit f5b71c38 makes the one-shot semantics explicit. No code change needed.

src/Facets/SupersetFacet.sol
```diff
@@ -0,0 +1,330 @@
...
+            _swapData,
+            payable(_supersetData.refundAddress),
+            _supersetData.lzFee
+        );
+
+        // Adjust destination slippage floor for positive source-side slippage
+        SupersetData memory modifiedSupersetData = _supersetData;
+        modifiedSupersetData.amountOutMin =
+            (_bridgeData.minAmount * _supersetData.amountOutMinPercent) /
+            1e18;
```
mirooon: does amountOutMinPercent (uint64) work correctly for routes where the destination token has more decimals than the bridged token (for example USDC (6 decimals) -> DAI (18 decimals)?

The docs/comments only describe it as a “Fraction (1e18 = 100%)” for positive slippage, and all examples I found are 6 -> 6 decimals (USDC/USDT). But from the implementation it looks more like a price × decimal multiplier.

If that’s correct, then on a 6 -> 18 decimals route the required value could become larger than uint64, which would silently reduce `amountOutMin` close to 0 instead of reverting.

Are omni-tokens internally decimal-normalized so this situation cannot happen? If not, would it make sense to use a uint128 multiplier instead, similar to AcrossFacetV4?

src/Facets/SupersetFacet.sol
```diff
@@ -0,0 +1,330 @@
...
+        // refundAddress also receives source-side excess native and swap leftovers,
+        // so it must be set even on the hub branch where Superset itself ignores it.
+        if (_supersetData.refundAddress == address(0)) {
+            revert InvalidConfig();
+        }
+
+        // Must be a non-zero EOA
+        if (
+            _supersetData.fallbackEoA == address(0) ||
+            _supersetData.fallbackEoA.code.length != 0
```
mirooon, line 340: the `fallbackEoA` check uses raw `code.length != 0`. That rejects EIP-7702-delegated EOAs (23-byte code), which LibAsset.isContract treats as EOAs. Does Superset's spoke contract also reject code.length != 0? If it's 7702-aware, should we switch to LibAsset.isContract to match; if it's strict, can we document that 7702/AA accounts can't be the fallback?

config/superset.json
```diff
@@ -0,0 +1,16 @@
...
+    "arbitrumsepolia": "0x1fDA3aAD6e6B932d992F42C4a808A19fFEd938F0",
+    "base": "0x57C155a15a9CA0A6C1F759ac6988b4fCa3663Ea4",
+    "basesepolia": "0x401DCd3FfD09908e277Ea609A91117C5aDAb4cf6",
+    "unichain": "0x326D0B0357b394EC5bbeFE87554243E5dc76e56E"
+  },
+  "mappings": [
+    { "chainId": 42161, "lzEid": 30110 },
+    { "chainId": 8453, "lzEid": 30184 },
+    { "chainId": 130, "lzEid": 30320 }
+  ]
```
mirooon, line 14: poolManager has 5 chains but mappings only has 3 (the mainnet ones). The two testnets (arbitrumsepolia, basesepolia) have a poolManager address but no mapping entry - is that on purpose? My understanding is poolManager = where the facet is deployed, and mappings = which chains you can bridge to. So if I deploy on a testnet and seed it with this config, it would only know mainnet destinations and couldn't route to the other testnet, right? Should the testnet EIDs be added?

src/Facets/SupersetFacet.sol
```diff
@@ -0,0 +1,330 @@
...
+    /// @dev Overwrites any existing entries for the supplied chain IDs.
+    function initSuperset(ChainIdConfig[] calldata _chainIdConfigs) external {
+        if (_chainIdConfigs.length == 0) revert InvalidConfig();
+        LibDiamond.enforceIsContractOwner();
+
+        Storage storage s = _getStorage();
+
+        for (uint256 i = 0; i < _chainIdConfigs.length; ++i) {
+            s.lzEids[_chainIdConfigs[i].chainId] = _chainIdConfigs[i].lzEid;
+        }
```
mirooon, line 160: initSuperset only emits the batch `SupersetChainMappingsInitialized` event, but `setChainIdToEid` emits `ChainIdToEidSet` per entry. So if someone indexes `ChainIdToEidSet` to track the mappings, they miss everything set at init. Could `initSuperset` also emit `ChainIdToEidSet` per entry, or is that split on purpose?

src/Facets/SupersetFacet.sol
```diff
@@ -0,0 +1,330 @@
...
+    /// @param toEid LayerZero endpoint ID of the destination spoke chain.
+    /// @param options LayerZero executor options for the source → hub request. Ignored
+    ///        on the hub branch (no source → hub LZ leg).
+    /// @param lzFee Native value forwarded to the pool manager (`msg.value`). On a spoke
+    ///        covers all three LZ messages; on the hub covers only the hub → destination
+    ///        delivery message.
+    struct SupersetData {
+        bytes path;
+        uint256 amountOutMin;
+        uint64 amountOutMinPercent;
```
mirooon: the SupersetData struct has both `amountOutMin` and `amountOutMinPercent`, but each path only uses one. `startBridge` uses `amountOutMin` and ignores `amountOutMinPercent`, while `swapAndStartBridge` overwrites `amountOutMin` from `amountOutMinPercent` (so the `amountOutMin` you pass is discarded). Same behaviour as `AcrossFacetV4` with `outputAmount`/`outputAmountMultiplier`, so no logic issue but it's not documented here, and the demo even sets `amountOutMin`: 1341 on the swap scenario where it does nothing. Could we add a one-line NatSpec on each field saying which entrypoint uses it, and set `amountOutMin` to 0 in the swap demo so nobody copies a dead value?

src/Facets/SupersetFacet.sol
mirooon: actually now looking at superset contract our code.length != 0 is the correct mirror of SwapDelivery.sol. then we keep it, just let's document the 7702 limitation

src/Facets/SupersetFacet.sol
```diff
@@ -0,0 +1,330 @@
...
+
+    /// @dev Validates Superset-specific data. Native source asset is rejected
+    ///      by the `noNativeAsset` modifier on each external entry.
+    /// @param _destinationChainId LI.FI chain ID of the destination spoke.
+    /// @param _supersetData Superset-specific parameters
+    function _validateSupersetData(
+        uint256 _destinationChainId,
+        SupersetData calldata _supersetData
+    ) internal view {
+        if (_supersetData.path.length == 0) {
```
mirooon: should we fast-fail on a stale deadline at the source? We don't check deadline at all right now. Confirmed Superset only enforces it on the hub (`HubSwapRouterWithFee.sol` `checkDeadline`), after the source->hub LZ message is already paid so a past/zero deadline burns the user's `lzFee` before it gets rejected. `UnitFacet`/`NEARIntentsFacet`/`EcoFacet` already do this source-side check, so `SupersetFacet` looks inconsistent with them. Could we add `if (block.timestamp > _supersetData.deadline) revert` in `_validateSupersetData` (matches Unit/NEAR and Superset's own <= semantics)?

config/superset.json
gvladika: Good point. I initially deployed to testnets because I could not use staging (Superset's latest contracts were not yet deployed to mainnets). Anyways, since now we can use staging for all testing activities I think it makes no sense to keep testnet config around. I removed poolManager entries:

https://github.com/lifinance/contracts/pull/1865/changes/1793a6adda66658f3ce6473f6769f668ba2f749c

src/Facets/SupersetFacet.sol
gvladika: Good point https://github.com/lifinance/contracts/pull/1865/changes/ec6566a1de5668d669d2f4aa51a6ba15d2db7779

src/Facets/SupersetFacet.sol
gvladika: Fixed by https://github.com/lifinance/contracts/pull/1865/changes/f186e4df08bbdf5dad5c452cdab8419236f69d18 and https://github.com/lifinance/contracts/pull/1865/commits/d597a34a3c54c63e97666ba721b0be3e7b1c3dc5

src/Facets/SupersetFacet.sol
gvladika: https://github.com/lifinance/contracts/pull/1865/changes/04f003d827173a6ff6a851d4635045814ca9dd1a

src/Facets/SupersetFacet.sol
gvladika: Refactored in https://github.com/lifinance/contracts/pull/1865/changes/57a2bee7864ba64a154076dc409f0c2c4929db74

`amountOutMinPercent` is actually not needed. I opted for different approach. The main idea is this one:
- backend provides `bridgeData.minAmount` (src input amount) and `SupersetData.amountOutMin` (dst output amount which includes slippage buffer)
- if source chain pre-swap bump `bridgeData.minAmount` by X%, simply bump `SupersetData.amountOutMin` by X% as well
- that way, the slippage buffer as a ratio between src and dest amount is preserved

src/Facets/SupersetFacet.sol
gvladika: Refactored in https://github.com/lifinance/contracts/commit/57a2bee7864ba64a154076dc409f0c2c4929db74
Check rationale in https://github.com/lifinance/contracts/pull/1865#discussion_r3348058710

src/Facets/SupersetFacet.sol
```diff
@@ -0,0 +1,354 @@
...
+            _supersetData.fallbackEoA.code.length != 0
+        ) {
+            revert InvalidConfig();
+        }
+
+        if (msg.value < _supersetData.lzFee) {
+            revert InsufficientNativeValue();
+        }
+
+        if (block.timestamp > _supersetData.deadline) {
```
mirooon, line 349: the deadline check is good but using `InvalidConfig` for an expired deadline feels a bit off to me. `InvalidConfig` is more for "you sent wrong params", but here the deadline just ran out while the tx was in flight, so it's a different thing.

also we handle this in 3 different ways across facets right now:

| Facet | deadline revert |
|---|---|
| `NEARIntentsFacet` | `QuoteExpired()` |
| `UnitFacet` | `SignatureExpired()` |
| `EcoFacet` | `InvalidConfig()` |
| `SupersetFacet` (this PR) | `InvalidConfig()` |

wdyt about adding a generic `error DeadlineExpired();` in `GenericErrors.sol` (would need a version bump) and using it here? we don't have to touch the other facets now, just leave a `// TODO` so we clean them up next time. not blocking.

config/superset.json
```diff
@@ -0,0 +1,14 @@
+{
+  "IMPORTANT DEV NOTICE": "after adding a new mapping, call initSuperset (first deploy) or setChainIdToEid (incremental update) on every diamond where SupersetFacet is deployed",
+  "Source for LayerZero EIDs: https://docs.layerzero.network/v2/deployments/deployed-contracts": {},
+  "poolManager": {
+    "arbitrum": "0xcC687ed9155F5e49719D5f3147F79FD5F177961c",
+    "base": "0x57C155a15a9CA0A6C1F759ac6988b4fCa3663Ea4",
+    "unichain": "0x326D0B0357b394EC5bbeFE87554243E5dc76e56E"
+  },
```
mirooon, line 9: It might also be worth adding a reference to where the PoolManager addresses can be found after this discussion: https://lifi-protocol.slack.com/archives/C088UJW2DPZ/p1780573238598869

src/Facets/SupersetFacet.sol
```diff
@@ -0,0 +1,354 @@
...
+        // its floor, so the backend's percentage bridge-slippage budget is
+        // preserved against the actual post-swap amount.
+        // `_depositAndSwap` reverts when the swap returns less than the floor,
+        // so by here `_bridgeData.minAmount >= preSwapMinAmount` and the ratio
+        // never tightens the floor. `validateBridgeData` already rejects
+        // `_bridgeData.minAmount == 0`, so `preSwapMinAmount > 0`.
+        SupersetData memory modifiedSupersetData = _supersetData;
+        modifiedSupersetData.amountOutMin =
+            (_supersetData.amountOutMin * _bridgeData.minAmount) /
+            preSwapMinAmount;
```
mirooon: I like this approach. Gj!

config/superset.json
gvladika: Done in 1e7b1358

src/Facets/SupersetFacet.sol
gvladika: Done in 3b2df2bf: added `error DeadlineExpired();` to `GenericErrors.sol` (bumped to 1.0.4), switched the deadline check in `_validateSupersetData` to use it, updated the test that was asserting `InvalidConfig`. Left a `// TODO` next to the new error reminding us to migrate `EcoFacet` / `UnitFacet` / `NEARIntentsFacet` in a follow-up.

mirooon (dismissed):

mirooon (approved):

PR #1866, opened 2026-05-29
fix(erc7730): use actions/checkout for fork clone + emit declaration-form display.formats keys

0xDEnYO:

Follow-up to [EXSC-338](https://linear.app/lifi-linear/issue/EXSC-338/erc-7730-clear-signing-dry-run-first-real-sync-to-ef-registry-post) — unblocks the first real sync. The original sync workflow (PR #1856) fixed the auth-state issue at the clone step but the workflow still failed in production; this PR is the actual, end-to-end-verified fix plus two real-bug fixes in the generator surfaced by the upstream registry's CI.

End-to-end green, verified upstream: [ethereum/clear-signing-erc7730-registry#2597](https://github.com/ethereum/clear-signing-erc7730-registry/pull/2597) is OPEN with `validate descriptors` ✓ and `validate JSON schemas` ✓.

Two distinct fixes, kept in separate commits:

### 1. `syncLedgerClearSigning.yml` — fork clone via `actions/checkout`

The manual `git clone "https://x-access-token:$TOKEN@github.com/..."` pattern in the sync step kept failing with `fatal: could not read Username for 'https://github.com'` — even though the same token, in the same shell, returned 200 from `/user` via `curl` one line earlier. A diagnostic workflow running the *identical* `git clone` command on the same commit/runner succeeded; I could not isolate why the bash-script-context one didn't. Rather than keep digging at a single-step quirk, this PR switches to `actions/checkout` for the fork:

```yaml
- name: Checkout fork (clear-signing-erc7730-registry) with sync token
  uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
  with:
    repository: ${{ secrets.LEDGER_FORK || 'lifinance/clear-signing-erc7730-registry' }}
    token: ${{ secrets.LEDGER_SYNC_TOKEN }}
    path: ledger-registry
    fetch-depth: 0
```

`actions/checkout` is GitHub's well-tested clone path — it also persists `http.https://github.com/.extraheader` inside the cloned repo, so the subsequent `git push origin` works without any manual extraheader setup. Net diff in `syncLedgerClearSigning.yml`: **+13 lines added, ~20 removed**.

### 2. `tasks/buildClearSigningProposal.ts` — emit declaration-form keys

Upstream EF CI on the first attempt surfaced two distinct failures, both stemming from one root cause:

1. `validate JSON schemas` rejected every multi-arg key against the v2 schema's regex, which requires `name(Type _arg, …)` (identifier after every type).
2. `validate descriptors` reported `Descriptor contains 2 formats sections for 0x4630a0d8` — the registry's existing entries are already in declaration form; our canonical-form key reduced to the same selector, producing two entries for one selector after merge.

`signature()` now emits `name(Type _arg, (Type field) _tuple, …)` — both top-level params and tuple components are named. The single change closes both upstream-CI failures simultaneously, because our `swapTokensGeneric(...)` key now matches the upstream entry byte-for-byte (overwrite, not duplicate). Verified by the dry-run merge signature, before and after the fix:

| | Before | After |
|---|---|---|
| added | +101 | +94 |
| replaced | ~0 | **~7** |
| preserved | =7 | **=0** |

The 7 upstream `swapTokens*` entries are now recognized as matches (replaced, not preserved-as-different). Total still 101.

`config/clearSigningProposal.json` is regenerated to match — that's the bulk of the diff in this PR.

### CodeRabbit findings

Two findings from `/pr-ready`, both applied as separate `pr-ready:` commits:

- **Minor** — `declarationType()` was silently emitting `()` when `p.components` was missing from a tuple ABI. Now throws explicitly, matching the surrounding strict-by-default pattern.
- **Major** — `signature()` enforces parameter names, but it was being called from `collectFns()` for *every* ABI item — admin, getter, view — *before* `isKnownNonUserFacing()` could filter them. A single helper with an unnamed param would crash the generator before classification. Split into `signature()` (strict, output-key path) and `collectionSignature()` (canonical, dedup/sort path). Regenerated JSON byte-identical.

### End-to-end verification

| Step | Run | Result |
|---|---|---|
| Token diagnostic on test branch | [26611709476](https://github.com/lifinance/contracts/actions/runs/26611709476) | ✓ |
| First real sync (broken generator) | [26612387286](https://github.com/lifinance/contracts/actions/runs/26612387286) | workflow ✓, upstream CI ✗ (expected — generator bugs) |
| Final dry-run on `fix/...` | [26613218165](https://github.com/lifinance/contracts/actions/runs/26613218165) | ✓ — `+94 added, ~7 replaced, =0 preserved` |
| Final real sync | [26613296049](https://github.com/lifinance/contracts/actions/runs/26613296049) | ✓ — PR #2597 opened |
| Upstream EF CI on PR #2597 | [validate descriptors](https://github.com/ethereum/clear-signing-erc7730-registry/actions/runs/26613311196) + [validate JSON schemas](https://github.com/ethereum/clear-signing-erc7730-registry/actions/runs/26613311196) | ✓ — all green |

- [x] I have run `/pr-ready` (local CodeRabbit) on this branch and resolved (or explicitly documented) all findings — see `.agents/commands/pr-ready.md`. PR created via the documented `PR_READY_OK=1` inline bypass: `/pr-ready` ran cleanly inside the git-worktree this branch lives in (initial pass surfaced 2 findings — both applied as the `pr-ready:` commits at the tip; scoped re-run was clean), but the PreToolUse gate hook resolves its `cwd` from the parent shell, so it looked for the marker in the main checkout's `.git/PR_READY_OK` instead of the worktree's `.git/worktrees/contracts-wt-ledger-sync-debug/PR_READY_OK` where it actually was. Hook bug, not a skipped review.

> **Test plan note**: this is a CI / generator fix without contract-level changes; no Solidity unit tests apply. End-to-end verification is the upstream EF CI passing on PR #2597 (link above) plus the local dry-run on this branch.


tasks/buildClearSigningProposal.ts
```diff
@@ -43,14 +43,77 @@ const ROOT = path.resolve(__dirname, '..')
...
+            `tuple component missing name in field of type "${p.type}" — Foundry ABI artifacts should always include component names; rebuild with \`forge build\` if stale`
+          )
+        return `${declarationType(c)} ${c.name}`
+      })
+      .join(', ')
     return `(${inner})${p.type.slice('tuple'.length)}`
   }
   return p.type
 }
 function signature(fn: IAbiFn): string {
```
gvladika, line 78: This strict, name-required `signature()` is invoked at `main()` (line 713: `const sig = signature(fn)`) on *every* collected function — before `isKnownNonUserFacing(fn)` filters out admin/getter/view helpers two lines down. That's the same crash the `signature()` / `collectionSignature()` split was added to prevent: a non-user-facing function with an unnamed param (legal Solidity) throws there before it gets discarded.

`sig` isn't used before the `continue`, so it's just an ordering issue in that loop — filter first, then compute:

```ts
for (const fn of fns) {
  if (isKnownNonUserFacing(fn)) continue
  const sig = signature(fn)
  ...
```

May not trigger on the current ABIs (depends on whether any non-user-facing function has an unnamed param today), but leaving the order as-is defeats the robustness the split was meant to provide.

.github/workflows/syncLedgerClearSigning.yml
```diff
@@ -65,6 +65,21 @@ jobs:
...
         run: forge build --skip script --skip test --skip Base --skip Test --skip '*.t.sol'
 
+      - name: Checkout fork (clear-signing-erc7730-registry) with sync token
+        # Use `actions/checkout` rather than a manual `git clone` token-in-URL.
+        # The manual pattern failed in this job with `fatal: could not read
+        # Username for 'https://github.com'` even though the same token was
+        # valid (curl /user → 200 in the same shell). `actions/checkout` is
+        # GitHub's own well-tested clone path and handles auth setup robustly
+        # — it also persists `http.https://github.com/.extraheader` inside
+        # the cloned repo so the later push to origin authenticates.
```
gvladika, line 75: Our AI has recently started significantly over-commenting. I added this rule recently:
```
Add a comment ONLY when the *why* is non-obvious: a hidden constraint, a subtle invariant, a workaround for a specific bug, or behavior that would surprise a reader. Never narrate the *what* (the code already does that), and never restate the history, debate, or alternatives behind a line — that belongs in the PR description, commit message, or linked ticket. If a comment would be removed without confusing a future reader, don't write it.
```
It seems rule is not picked up or applied always, will try to figure out why. Anyways, I think this is an example of a comment block which should be way more concise, or removed altogether.

0xDEnYO: I personally feel comments are quite helpful to scan code much faster than reading code. 
We always had a "comment a lot" policy here. 

The only thing that AI does what I dont really like is this sort of historic "now changed to ..." or "because ... didnt work we implemented this now" kinda stuff

0xDEnYO: here we ran into an issue when git cloning that was hard to reproduce locally. So I feel this comment is quite valuable actually as it's a learning from a past mistake that AI can see and not repeat.

tasks/buildClearSigningProposal.ts
0xDEnYO: Good catch — fixed in 74117ecf. Moved the `isKnownNonUserFacing(fn)` filter above `const sig = signature(fn)` in the `main()` loop. `sig` was unused before the `continue`, so it's a pure reordering: re-running the generator produces a byte-identical `config/clearSigningProposal.json` (101 entries, no diff). This restores the robustness the `signature()` / `collectionSignature()` split was meant to provide — a non-user-facing helper with an unnamed param now gets filtered out before the strict signature builder can throw on it.

.github/workflows/syncLedgerClearSigning.yml
gvladika: > historic "now changed to ..." or "because ... didnt work we implemented this now" kinda stuff

Exactly this, I think this kind of description is valuable but should be in commit message or PR description. When it is in source code it clutters and hurts readability IMO. But that's just personal preference, so let's keep it

gvladika (approved):

maxklenk (approved):

PR #1871, opened 2026-05-29
EXSC-370: audit + fund pauser wallet against pauseDiamond() cost

gvladika:

[EXSC-370](https://linear.app/lifi-linear/issue/EXSC-370/cross-chain-audit-pauser-wallet-funds-vs-pausediamond-cost)

EXSC-369 showed Rootstock's emergency pause fail: the pauser wallet held dust — enough to pass the existing `balance > 0` gate in the pause script, but not enough to actually send `pauseDiamond()`. The funding default elsewhere was a flat hardcoded `0.002 ETH`, unrelated to the real per-chain cost of a pause. This PR ties both the audit and the funding default to the actual cost of a pause.

Three changes share one primitive:

- **`estimatePauseCost` (`script/helperFunctions.sh`)** — the single shared helper. Returns the wei cost of one `pauseDiamond()` as `cast estimate × cast gas-price` for the production `LiFiDiamond`. Exit `0` (cost on stdout), `2` (diamond already paused — `pauseDiamond()` reverts `DiamondIsPaused`), `1` (any other failure, reason on stderr). EVM-only; reuses existing resolvers (`getRPCUrl` with a `getRpcUrlFromNetworksJson` fallback, `getContractAddressFromDeploymentLogs`, `getValueFromJSONFile`). `bc` for the multiplication since wei overflows 64-bit shell math.

- **`checkPauserFunds.sh` (`script/utils/`)** — a read-only audit (no transactions). For each production EVM network it computes `NUM OF PAUSES = balance ÷ single-pause cost` and assigns a two-tier status: `CRITICAL` (< 1, can't afford one pause), `WARNING` (1–2.5), `OK` (≥ 2.5), plus `PAUSED`/`ERROR`/`SKIP`. Output is one `column -t`-aligned table sorted worst-first; the `STATUS` column is colored on a TTY only and amounts are shown to 3 significant figures. Progress, the legend and the summary go to stderr so stdout stays a clean, pipeable table; exit code is `1` if any network is `CRITICAL`. Networks with no native currency (`nativeCurrency: "N/A"`, e.g. tempo, which pays gas in a non-native token) are skipped, since a native balance-vs-gas comparison is meaningless there.

- **`deployAllContracts.sh` stage 9** — the PauserWallet funding default is now `2.5 × estimatePauseCost` (≈ two pauses plus buffer) instead of the hardcoded `0.002 ETH`. The existing `balance == 0` trigger and the `gum` accept/override prompt are unchanged, and the DevWallet branch is untouched. If estimation fails the step falls back to the flat default and only `warning`s, so a deployment is never blocked on a transient RPC/estimate error.

Scope decisions (recorded in the design doc): EVM-only — `cast estimate`/`gas-price`/`balance` are EVM, so Tron is skipped and left as a follow-up. The other ticket options are deferred: (b) weekly GitHub Action + Slack alert, (c) an inline guardrail in `diamondEMERGENCYPauseGitHub.sh`, and (d) auto-top-up via the dev-fees flow.

Verification: the repo has no bash unit-test harness, so each script was checked with `bash -n` + `shellcheck` and exercised with live read-only smoke runs (`estimatePauseCost`, `checkPauserFunds.sh`). `/pr-ready` (local CodeRabbit) is clean. The design and step-by-step plan are under `docs/superpowers/`.


script/helperFunctions.sh
```diff
@@ -5418,3 +5418,91 @@ function removeNetworkFromTargetStateJSON() {
...
+  local GAS_PRICE
+  GAS_PRICE=$(cast gas-price --rpc-url "$RPC_URL" 2>/dev/null) || GAS_PRICE=""
+  if ! [[ "$GAS_PRICE" =~ ^[0-9]+$ ]]; then
+    error "estimatePauseCost: could not read gas price for $NETWORK: $GAS_PRICE" >&2
+    return 1
+  fi
+
+  # wei overflows 64-bit bash arithmetic; use bc for the multiplication
+  echo "$GAS_ESTIMATE * $GAS_PRICE" | bc
+  return 0
```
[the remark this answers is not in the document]
gvladika: Addressed in a91b0523 — `estimatePauseCost` now validates the `bc` product and returns 1 on failure instead of unconditionally returning 0.

script/utils/checkPauserFunds.sh
```diff
@@ -0,0 +1,210 @@
...
+#
+# NUM OF PAUSES = balance ÷ cost of ONE pauseDiamond() — how many pauses the wallet can fund.
+# Status:  OK (>= 2.5 pauses)   WARNING (1 to 2.5)   CRITICAL (< 1 pause)
+#          PAUSED (already paused)   ERROR (estimate/RPC failed)   SKIP (filtered/no diamond)
+# Exit code: 1 if any audited network is CRITICAL, else 0.
+#
+# Must be run from the repository root.
+
+# NOTE: intentionally no `set -e` — failures are handled per-network so the sweep continues.
+set -uo pipefail
```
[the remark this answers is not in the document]
gvladika: Addressed in 141eb8d8 — the render pipeline is now guarded (`if ! { … } | column | colorizeStatus`); a render failure exits 1 instead of silently exiting 0.

mirooon (commented):
Had a proper read through — really nice work overall, and it reuses the existing helpers well 👍 I left a few small comments on the files: mostly little consistency/cleanup things, plus one performance idea on the per-network loop. None of them are blocking — feel free to take or leave them.

script/helperFunctions.sh
```diff
@@ -5418,3 +5418,97 @@ function removeNetworkFromTargetStateJSON() {
     return 1
   fi
 }
+
+# estimatePauseCost: Estimate the native-token cost (in wei) of one pauseDiamond() call.
```
mirooon: Small thing 🙂 — this header is pretty long compared to the other functions in this file (most just have a one-line comment). Maybe we could trim it to 2–3 lines: what it does + the exit codes (0/1/2)? A bit easier to scan. Not a blocker.

script/utils/checkPauserFunds.sh
```diff
@@ -0,0 +1,215 @@
...
+
+# colorizeStatus: color the trailing STATUS word of each row. Applied AFTER `column -t` so the
+# non-printing escape codes can't skew column-width alignment, and only on a TTY so piped or
+# redirected output stays plain and parseable.
+function colorizeStatus() {
+  if [[ ! -t 1 ]]; then
+    cat
+    return
+  fi
+  local RED=$'\033[31m' YEL=$'\033[33m' GRN=$'\033[32m' CYN=$'\033[36m' DIM=$'\033[2m' RST=$'\033[0m'
```
mirooon, line 119: Tiny one — `RED` and `GREEN` are already defined up in helperFunctions.sh (~line 21), and we source that file at the top, so we could reuse them here instead of redefining. The other colours (`YEL`/`CYN`/`DIM`) should stay since they're new and need the `$'...'` form for sed — it's really just red/green that are duplicated.

script/utils/checkPauserFunds.sh
```diff
@@ -0,0 +1,215 @@
...
+
+# fmtRow: join the 6 table columns with tabs (column -t aligns them at the end).
+# Usage: fmtRow NETWORK COST REQUIRED BALANCE PAUSES STATUS
+function fmtRow() {
+  local NETWORK="$1" COST="$2" REQUIRED="$3" BALANCE="$4" PAUSES="$5" STATUS="$6"
+  printf '%s\t%s\t%s\t%s\t%s\t%s' "$NETWORK" "$COST" "$REQUIRED" "$BALANCE" "$PAUSES" "$STATUS"
+}
+
+# plainRow: a row with dashes in the numeric columns (for SKIP/PAUSED/ERROR).
+function plainRow() {
```
mirooon: Do we actually need `plainRow`? It only calls `fmtRow` with dashes, so we could just call `fmtRow "$NET" - - - - "$STATUS"` directly at the SKIP/PAUSED/ERROR spots and drop this helper — one less thing to follow. wdyt?

script/utils/checkPauserFunds.sh
```diff
@@ -0,0 +1,215 @@
...
+  exit 1
+fi
+
+# Each entry: "<CAT>|<RATIO>|formatted row". CAT then RATIO order the table:
+#   ERROR first (needs attention), then data rows by ratio ascending, then PAUSED, then SKIP.
+ROWS=()
+HAS_CRITICAL=0
+
+# sort categories (field 1 of each ROWS entry) — keeps ordering independent of ratio magnitude
+readonly CAT_ERROR=0
```
mirooon, line 99: This bit feels a little heavy for what it does — the four `CAT_*` constants plus the `<CAT>|<RATIO>|row` string that gets sorted and cut (line 201), just to print ~40 rows worst-first. It works totally fine, so happy to leave it! If you ever feel like simplifying, a single number prefix you strip at the end would be enough. Just a thought, not blocking.

script/utils/checkPauserFunds.sh
```diff
@@ -0,0 +1,215 @@
...
+
+IDX=0
+for NETWORK in "${NETWORKS[@]}"; do
+  IDX=$((IDX + 1))
+  printf '[%d/%d] %s\n' "$IDX" "$TOTAL" "$NETWORK" >&2
+  if isTestnetNetwork "$NETWORK" >/dev/null 2>&1 || isTronNetwork "$NETWORK" >/dev/null 2>&1; then
+    ROWS+=("$CAT_SKIP|0|$(plainRow "$NETWORK" "SKIP")")
+    continue
+  fi
+  STATUS_FIELD=$(getValueFromJSONFile "./config/networks.json" "${NETWORK}.status")
```
mirooon: There's already an `isNetworkActive` helper in helperFunctions.sh that does exactly this check. Maybe use `if ! isNetworkActive "$NETWORK"; then ... SKIP` instead of reading `.status` by hand? Keeps it consistent with how the rest of the repo checks this.

script/utils/checkPauserFunds.sh
```diff
@@ -0,0 +1,215 @@
...
+    -e "s/(OK)[[:space:]]*\$/${GRN}\\1${RST}/" \
+    -e "s/(SKIP)[[:space:]]*\$/${DIM}\\1${RST}/"
+}
+
+# Progress goes to stderr so the final table on stdout stays clean and pipeable.
+TOTAL=${#NETWORKS[@]}
+echo "Checking pauser-wallet funding on $TOTAL network(s) — reading live gas prices & balances, this can take a minute..." >&2
+
+IDX=0
+for NETWORK in "${NETWORKS[@]}"; do
```
mirooon: This is the main thing I'd suggest looking at (still not blocking 🙂). Right now we go network by network, and each one makes ~3 RPC calls — so for ~40 networks it can take a couple of minutes, and if one RPC is slow everything waits behind it.

Since the networks don't depend on each other, we could run them in parallel and the whole sweep would take about as long as the slowest single network (more like 5–15s). One heads-up if you do: `HAS_CRITICAL` is set inside the loop, and with parallel jobs that runs in a subshell so it won't reach the parent — we'd need to figure out "is anything critical?" after collecting the results instead.

We already do this pattern in a few places if you want something to copy:
- helperFunctions.sh:1691 — `processNetworkLine ... &` with the `while [[ $(jobs | wc -l) -ge $MAX_CONCURRENT_JOBS ]]` throttle (line 1686) and `wait` (1695); worker is at 1718
- helperFunctions.sh:894-927 — runs each item as `( ... > file ) &`, then `wait`, then merges the files
- multiNetworkExecution.sh:2071 — `executeNetworkInGroup ... &`
- `MAX_CONCURRENT_JOBS` (helperFunctions.sh:1557, default 10) is the shared knob for this

mirooon: This per-network loop is something I keep bumping into in our newer bash scripts. I think AI tools almost always start with a normal `for` loop, even when each network is totally independent. And since most of the time is just waiting for RPCs, doing ~40 networks one by one takes minutes when it could take seconds.

The nice part is we already solved this in the repo — there's a ready pattern in `helperFunctions.sh` we can just copy:

- `processNetworkLine` runs per network at `helperFunctions.sh:1691`, with a small throttle (`:1686`) and `wait` (`:1695`)
- the facet resolver at `helperFunctions.sh:894-927` (`( … > file ) &` → `wait` → merge the files)
- `multiNetworkExecution.sh:2071`
- `MAX_CONCURRENT_JOBS` (default 10) is the shared setting

So I was thinking maybe it's worth adding a short note to `.agents/rules/300-bash.md` (and a quick line in `200-typescript.md` for the `Promise.all` version) — basically "if the work is independent per item, run it in parallel, and here's where to copy it from." Just so it's something we remember by default instead of rediscovering each time.

script/helperFunctions.sh
gvladika: Done in `fd6eac33` — condensed to what it does + the exit codes (0/1/2). 👍

script/utils/checkPauserFunds.sh
gvladika: Done in `039e0660` — inlined `fmtRow "$NET" - - - - "$STATUS"` at the SKIP/PAUSED/ERROR sites and dropped `plainRow`.

script/utils/checkPauserFunds.sh
gvladika: Done in `e532a5ad` — switched to `isNetworkActive` (its output is suppressed so an unknown-network arg doesn't leak a line into the table).

script/utils/checkPauserFunds.sh
gvladika: Done in `8656becb` — the sweep now runs networks as throttled background workers (capped by `MAX_CONCURRENT_JOBS`), collects per-network results from files after `wait`, and derives `HAS_CRITICAL` from the collected rows rather than inside the loop (per your subshell heads-up — thanks for that). A 5-network run dropped to ~1.2s.

gvladika: Done in `cc92945f` — added a `[CONV:PARALLEL-WORK]` convention: `300-bash.md` (throttled background jobs via `MAX_CONCURRENT_JOBS`, aggregate after `wait`, the "subshell can't set parent vars" gotcha, and copy `processNetworkLine` / `executeNetworkInGroup`) and `200-typescript.md` (`Promise.all` / `allSettled`, bound concurrency on rate limits). Good call making it a default.

script/utils/checkPauserFunds.sh
gvladika: Looked closely — going to leave this one. The globals are stored as literal-`\033` strings (`RED='\033[0;31m'`) for `printf`/`echo -e`, but `colorizeStatus` injects the code into a `sed` replacement, which does **not** interpret `\033` — reusing them directly prints literal `33[0;31m` (verified on GNU sed 4.9). Converting them with `printf '%b'` would add a line per colour and hit a name-collision (`local RED=$(printf '%b' "$RED")` reads the now-empty local), and `YEL`/`CYN`/`DIM` are new anyway. So the local `$'\033…'` definitions are the minimal correct form here. Happy to revisit if we ever move the shared colour globals to real-ESC form repo-wide.

script/utils/checkPauserFunds.sh
gvladika: Leaving this one too  — the single-number-prefix version is actually what it was originally, and it had a bug: extreme ratios on cheap-gas chains (berachain/gnosis ~1e9, tempo ~1e58) collided with the `PAUSED`/`SKIP` sentinels and sorted into the wrong group. The two-part `<CAT>|<RATIO>` key decouples grouping from ratio magnitude so that can't happen — keeping it for that reason.

.agents/rules/200-typescript.md
```diff
@@ -73,6 +73,15 @@ paths:
...
 - When a module is only needed for one chain (e.g., TronWeb, Tron executor), use `await import('./path')` instead of top-level imports. This prevents loading chain-specific dependencies when running on other chains. See `create-chain-caller.ts` for the pattern.
 
+### Parallelize independent async work [CONV:PARALLEL-WORK]
+
+When the same async operation runs over many independent items (per-network balance/price
+fetches, per-chain queries), use `Promise.all` — or `Promise.allSettled` when one failure must
+not reject the whole batch — instead of `await`-ing in a sequential loop; most of the cost is
+network latency, so this turns minutes into seconds. If you hit provider rate limits, bound
+concurrency (chunk the array or use a small pool) rather than reverting to sequential. See
+`script/balances.ts` for the pattern.
```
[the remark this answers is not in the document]
gvladika: Done in `c78de4d0` — the rule now points at `script/tasks/moveNativeFundsToNewWallet.ts` for the batched, `MAX_CONCURRENT_JOBS`-bounded case (alongside `balances.ts` for the simple `Promise.all`).

script/utils/checkPauserFunds.sh
```diff
@@ -0,0 +1,240 @@
...
+
+  if isTestnetNetwork "$NETWORK" >/dev/null 2>&1 || isTronNetwork "$NETWORK" >/dev/null 2>&1; then
+    echo "$CAT_SKIP|0|$(fmtRow "$NETWORK" "-" "-" "-" "-" "SKIP")"
+    return
+  fi
+  # output suppressed: isNetworkActive logs "not found" for unknown args, which would
+  # otherwise land on stdout and pollute the table.
+  if ! isNetworkActive "$NETWORK" >/dev/null 2>&1; then
+    echo "$CAT_SKIP|0|$(fmtRow "$NETWORK" "-" "-" "-" "-" "SKIP")"
+    return
```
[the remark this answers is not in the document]
gvladika: Done in `469952f4` — explicit network args are now validated against `networks.json` up front; an unknown one errors and exits 1 instead of becoming a silent SKIP. The no-arg sweep is unaffected (networks come from `getAllNetworksArray`). Good catch.

script/utils/checkPauserFunds.sh
```diff
@@ -0,0 +1,240 @@
...
+RESULT_DIR=$(mktemp -d)
+trap 'rm -rf "$RESULT_DIR"' EXIT
+echo "Checking pauser-wallet funding on $TOTAL network(s), up to $MAX_JOBS in parallel — reading live gas prices & balances..." >&2
+
+IDX=0
+for NETWORK in "${NETWORKS[@]}"; do
+  IDX=$((IDX + 1))
+  printf '[%d/%d] %s\n' "$IDX" "$TOTAL" "$NETWORK" >&2
+  # throttle to MAX_JOBS concurrent workers
+  while (($(jobs -rp | wc -l) >= MAX_JOBS)); do wait -n; done
```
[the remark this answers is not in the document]
gvladika: Done in `76370365` — `MAX_JOBS` is now clamped to a positive integer (`^[1-9][0-9]*$`, default 10), so `MAX_CONCURRENT_JOBS=0` or a non-numeric value can't spin the throttle loop.

mirooon (approved):

PR #1896, opened 2026-06-10
feat: batch MongoDB deployment-log queries (EXSC-133)

0xDEnYO:

[EXSC-133](https://linear.app/lifi-linear/issue/EXSC-133/batch-mongodb-deployment-log-queries-reduce-repeated-calls-in-scripts) — Batch MongoDB deployment-log queries, reduce repeated calls in scripts and confirm-safe-tx

Each deployment-log lookup previously spawned its own `bun script/deploy/query-deployment-logs.ts` process with its own MongoDB round-trip, producing a wall of "MongoDB query" log lines (e.g. one process per facet address in `saveDiamondFacets`, one per network in `getContractDeploymentStatusSummary`).

**1. Batch API (`query-deployment-logs.ts batch`)**
- New `batch` subcommand accepts a JSON array of queries (`get` / `latest` / `find` / `exists` / `history`, each optionally with a per-query `env`) inline or via stdin (`--queries=-`; the equals form is required because the pinned citty version parses a bare `-` after a flag as empty).
- Execution logic lives in `script/deploy/shared/batch-deployment-queries.ts` (100% test coverage). One querier is created per distinct environment in the batch, so the whole batch costs one cache load (or one Mongo connection in `--no-use-cache` mode) per environment instead of one per lookup. Per-query failures are reported in the result array without failing the batch.

**2. helperFunctions.sh uses the batch API where multiple lookups happen in one step**
- New `batchQueryMongoDeployments` helper (stdin transport, retry + JSON validation, mirrors the existing per-query retry helpers).
- `saveDiamondFacets`: all facet addresses (× both environments, mirroring `findContractInMasterLogByAddress`'s env fallback) are resolved in ONE invocation; the per-address parallel Mongo loop remains only as a fallback when the batch call fails entirely, and the local diamond-file/deployment-file fallbacks for misses are unchanged.
- `getConstructorArgsFromMasterLog` (no explicit version): one `history` lookup replaces the previous two invocations (highest-version filter + get); explicit-version calls keep the existing single `get` path.
- `getContractDeploymentStatusSummary`: one batch invocation for all networks (was one process per network, 76 on current config); also reads `address`/`verified` with both key casings so Mongo-record results are parsed correctly.
- jq gotcha encoded as comments at both call sites: `$ENV` is a jq *builtin* (the process environment) and silently shadows `--arg ENV`, so the env arg is passed as `$ENVIRON`.

**3. confirm-safe-tx deployment reuse**
- `getDeployments` in `script/utils/deploymentHelpers.ts` now memoizes per `(chain, environment)` for the lifetime of the process, so `confirm-safe-tx` (via `safe-decode-utils`' per-address labelling and periphery checks) loads each network's deployments once per run. Failed loads are not cached, so transient errors remain retryable. Covered by a new colocated test.

## Verification

- `bun test script/deploy/shared/batch-deployment-queries.test.ts script/utils/deploymentHelpers.test.ts` — 16 pass, 100% line/function coverage on the new shared module and `deploymentHelpers.ts`
- `bunx eslint` + `bunx tsc-files --noEmit` clean on all touched TS files (incl. all `getDeployments` callers)
- `bash -n script/helperFunctions.sh` + strict-mode (`set -euo pipefail`) source check pass
- Live smoke tests against the real deployment log:
  - batch CLI via stdin across all 5 ops + mixed staging/production envs
  - `saveDiamondFacets` (facets-only mode, mainnet): batch-resolved facets match the existing diamond log entries
  - `getConstructorArgsFromMasterLog "Executor" "mainnet" "production"` (batch path) returns byte-identical args to the legacy two-step path
  - `getContractDeploymentStatusSummary production Permit2Proxy`: 76 networks in one invocation
- `/pr-ready` (local CodeRabbit, full org mode): **no findings** on first pass; scoped re-run skipped per skill rules

- [ ] For new facets: I have checked all points from this list: https://www.notion.so/lifi/New-Facet-Contract-Checklist-157f0ff14ac78095a2b8f999d655622e (N/A — no facet changes)


script/utils/deploymentHelpers.ts
```diff
@@ -7,26 +7,46 @@ import { fileURLToPath } from 'url'
...
+ * Shape of an imported `deployments/<network>[.staging].json` module: contract name →
+ * address. The module namespace also carries a `default` export object at runtime.
+ */
+type DeploymentsFileModule = Record<string, string> & {
+  default?: Record<string, string>
+}
+
+// In-run memoization so repeated lookups for the same network/environment (e.g.
+// per-address labelling in confirm-safe-tx) load each deployments file only once.
+const deploymentsCache = new Map<string, Promise<DeploymentsFileModule>>()
```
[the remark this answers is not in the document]
0xDEnYO: @AikidoSec ignore: intentional in-run memoization in a CLI script (not a server) — the cache holds static deployments/*.json file contents keyed by (chain, environment), no request- or user-specific data; process lifetime is the desired scope.

script/helperFunctions.sh
```diff
@@ -889,22 +982,85 @@ function saveDiamondFacets() {
...
+  # shared across staging/production on the same chain, so query both environments
+  # (requested env wins), mirroring findContractInMasterLogByAddress.
+  local OTHER_ENV
+  if [[ "$ENVIRONMENT" == "staging" ]]; then
+    OTHER_ENV="production"
+  else
+    OTHER_ENV="staging"
+  fi
+
+  local BATCH_QUERIES="[]"
```
mirooon, line 1042: **🔴 The batch payload is built two different ways in this PR, and this one is O(N²)**

This block builds the query payload by spawning `jq` **once per facet**, and each call re-serializes the whole growing array:

```bash
BATCH_QUERIES="[]"
for RAW_ADDR in "${FACET_ADDRESSES[@]}"; do
  FACET_ADDRESS=$(echo "$RAW_ADDR" | tr -d '"' | tr -d ' ')
  BATCH_QUERIES=$(echo "$BATCH_QUERIES" | jq --arg ADDR ... '. + [{...},{...}]')
done
```

For ~30–40 facets that's 30–40 `jq` process spawns plus quadratic re-parsing of the accumulator — to build the payload for a call whose entire purpose is *reducing* process spawns. It still beats N `bun`+Mongo spawns, but `getContractDeploymentStatusSummary` (line 5465, same PR) already does the equivalent in a **single** pass:

```bash
BATCH_QUERIES=$(printf '%s\n' "${NETWORKS[@]}" | jq -R . | jq -s --arg CONTRACT ... '[.[] | select(. != "") | {...}]')
```

Please build the facets payload the same single-pass way (`printf '%s\n' "${FACET_ADDRESSES[@]}" | jq -R . | jq -s '...'`) so both functions are consistent and the per-facet `jq` spawns go away.

script/helperFunctions.sh
```diff
@@ -889,22 +982,85 @@ function saveDiamondFacets() {
...
+  local BATCH_RESULT=""
+  local BATCH_OK="false"
+  for RAW_ADDR in "${FACET_ADDRESSES[@]}"; do
```
mirooon: **🟠 `FACET_ADDRESSES` is now iterated three times with the same sanitization repeated each pass**

This PR adds a build loop (line 999) and a parse loop (line 1017) on top of the existing fallback loop (line 1041) — three passes over the same array, each repeating:

```bash
FACET_ADDRESS=$(echo "$RAW_ADDR" | tr -d '"' | tr -d ' ')
```

Suggest sanitizing once into a normalized array and reusing it; the build+parse passes can likely collapse into one. The resolve-all-then-fallback structure is defensible for clarity, but the triplicated `tr -d` is pure duplication and avoidable.

script/deploy/shared/batch-deployment-queries.ts
```diff
@@ -0,0 +1,278 @@
...
+  'find',
+  'exists',
+  'history',
+]
+
+function isValidOp(value: unknown): value is BatchQueryOp {
+  return typeof value === 'string' && VALID_OPS.includes(value as BatchQueryOp)
+}
+
+function isValidEnv(value: unknown): value is keyof typeof EnvironmentEnum {
```
mirooon, line 76: **🟡 `isValidEnv` hardcodes the enum members instead of deriving from `EnvironmentEnum`**

```ts
function isValidEnv(value: unknown): value is keyof typeof EnvironmentEnum {
  return value === 'staging' || value === 'production'
}
```

`EnvironmentEnum` is the source of truth (`script/common/types.ts:38`). Deriving the check — e.g. `typeof value === 'string' && value in EnvironmentEnum` — means a future third environment won't silently fail validation here.

Counterpoint: every existing CLI command in `query-deployment-logs.ts` hardcodes the same pair, so this matches current convention — but this new shared module is the right place to do it properly. Non-blocking.

script/helperFunctions.sh
```diff
@@ -484,6 +484,67 @@ function getLatestMongoDeployment() {
...
+
+  local ATTEMPT=1
+  local RETRY_DELAY=1
+  while [[ $ATTEMPT -le $MONGO_MAX_RETRIES ]]; do
+    echoDebug "Querying MongoDB (batch) with $(echo "$QUERIES_JSON" | jq 'length' 2>/dev/null || echo "?") queries (attempt $ATTEMPT/$MONGO_MAX_RETRIES)"
+    local BATCH_RESULT
+    local EXIT_CODE
+    # Queries are passed via stdin to avoid argv length limits for large batches.
+    # NOTE: must use --queries=- (equals form); citty parses a bare "-" after a flag as empty
+    BATCH_RESULT=$(echo "$QUERIES_JSON" | bun script/deploy/query-deployment-logs.ts batch \
```
mirooon: **🟢 Note (not a regression): `bun` vs `bunx tsx`**

This uses `bun script/deploy/query-deployment-logs.ts`, whereas rule `200-typescript` says invoke TS via `bunx tsx`. This is **correct to match surroundings** — all 9 existing `query-deployment-logs.ts` calls in this file already use bare `bun`. Flagging only so the team is aware the whole file is pre-existing debt against that rule; not something to fix in this PR's scope.

script/helperFunctions.sh
0xDEnYO: Done in 97bef1ae — the facets payload is now built in a single `printf | jq -R | jq -s` pass like `getContractDeploymentStatusSummary`, with each address expanding into its staging+production query pair inside the one jq program. No more per-facet jq spawns.

script/helperFunctions.sh
0xDEnYO: Done in 97bef1ae — addresses are now sanitized once (single `tr` over the whole list) into `CLEAN_FACET_ADDRESSES`, which the build, parse, fallback, and (bonus) the pre-existing merge loop all reuse. Kept parse and fallback as separate passes: the fallback loop spawns throttled background subshells and must skip facets the batch already resolved, so collapsing them would tangle the concurrency control with the batch parsing.

script/deploy/shared/batch-deployment-queries.ts
0xDEnYO: Done in 97bef1ae — now derived via `Object.values<string>(EnvironmentEnum).includes(value)`. Went with `Object.values` rather than `value in EnvironmentEnum` because `in` walks the prototype chain (`isValidEnv('toString')` would pass), and values are the semantically right check for raw CLI input. Required flipping the import from `import type` to a value import.

script/helperFunctions.sh
0xDEnYO: Ack — left as bare `bun` to match the 9 existing call sites; happy to do a file-wide `bunx tsx` sweep in a separate chore PR if the team wants the debt cleared.

0xDEnYO: Update: the debt is now being cleared — EXSC-502 / #1921 sweeps all 10 pre-existing bare-`bun` call sites in this file to `bunx tsx`. The `batch` call site added by this PR will be swept by whichever of the two PRs merges second (noted on the ticket).

mirooon (approved):

PR #1715, opened 2026-04-15
feat: migrate GenericSwapFacetV3 to granular allowlist + LibAsset routing [GenericSwapFacetV3 v2.0.0]

mirooon:
# Which Linear task belongs to this PR?
https://linear.app/lifi-linear/issue/EXSC-241/fix-tron-usdt-transfers-failing-on-safetransferlib-return-data-check

# Why did I implement it this way?

This PR migrates `GenericSwapFacetV3` to the granular `LibAllowList` v2 interface and routes its ERC20 transfers and approvals through `LibAsset`. It is the only change `main` needs for the Tron USDT rollout; the actual fix and the WithdrawablePeriphery port land in [lifinance/contracts-tron#9](https://github.com/lifinance/contracts-tron/pull/9), and the rollout itself requires both PRs to merge plus the operational redeploys / diamond cut on Tron.

```
┌─────────────────────────────────────────────────────────────┐
│  PR #1715 (this PR — main contracts repo)                   │
│  GenericSwapFacetV3 v2.0.0:                                 │
│    • LibAllowList v2 granular (contract,selector) allowlist │
│    • ERC20 transfers/approvals through LibAsset             │
│  No changes to LibAsset itself; no other periphery touched. │
└────────────────────────┬────────────────────────────────────┘
                         │  rebase
                         ▼
┌─────────────────────────────────────────────────────────────┐
│  PR #9 (contracts-tron fork)                                │
│  • LibAsset v2.1.3-tron:  chain+address gated Tron USDT     │
│    bypass in transferERC20                                  │
│  • WithdrawablePeriphery v1.0.0-tron: route withdrawToken   │
│    through LibAsset.transferAsset                           │
└─────────────────────────────────────────────────────────────┘
```

## Background — Tron USDT

Tron's canonical USDT (`TR7NHqjeKQxGTCi8q8ZY4pL8otSzgjLj6t` / `0xa614f803B6FD780986A42c78Ec9c7f77e6DeD13C`) was compiled with old Solidity (~0.4.x). Its `transfer()` is declared `returns (bool)` but does not `return true;`. The transfer succeeds and balances update, but callers receive 32 zero bytes (`false`). `SafeTransferLib.safeTransfer` rejects that → `TransferFailed`. `transferFrom()` is unaffected.

[...]

PR #1900, opened 2026-06-10
ci: enforce LibAsset routing for ERC20 transfers

0xDEnYO:

[EXSC-300](https://linear.app/lifi-linear/issue/EXSC-300/add-solhint-rule-or-ci-grep-enforcing-libasset-routing-for-erc20) — Add solhint rule (or CI grep) enforcing LibAsset routing for ERC20 transfers

Audit follow-up to PR #1715 (Sujith, informational finding #4).

The ticket offered two options: a custom solhint rule or a CI grep step. I went with the **CI grep step** (a new workflow, `enforceLibAssetRouting.yml`):

- A custom solhint rule would require shipping a local JS solhint plugin package (`solhint-plugin-*`), which conflicts with the repo's TypeScript/Bash-only scripting rule and adds dependency surface for a single rule.
- The grep backstop is explicitly sanctioned by the ticket as "adequate", is trivially auditable, and follows the established house pattern of inline-bash checker workflows (`jsonChecker.yml`, `protectSecurityRelevantCode.yml`).

Behavior:

- Fails CI if any `.sol` file in `src/` outside the allowlist contains `.safeTransfer(` or `.safeTransferFrom(`.
- `safeTransferETH(` is not matched and remains allowed everywhere (the regex requires `(` directly after `safeTransfer`/`safeTransferFrom`).
- The four packed facets (`HopFacetPacked`, `AcrossFacetPacked`, `CBridgeFacetPacked`, `AcrossFacetPackedV3`) are explicitly allowlisted with a comment, per the AC.
- Beyond the four facets named in the AC, six more pre-existing offenders exist on `main` (`TokenWrapper`, `LiFiDEXAggregator`, `ReceiverStargateV2`, `ReceiverAcrossV3`, `ReceiverChainflip`, `WithdrawablePeriphery` — consistent with the audit finding "Out-of-scope facets and LidoWrapper bypass LibAsset"). They are grandfathered in a separately-commented allowlist section so the check is green on `main` from day one while still rejecting any NEW file or any new call site in a currently-clean file.

Verification evidence (run locally against this branch):

- `bash -n` on the extracted check script: pass
- Clean tree (current `src/`): exit 0
- Injected `src/EvilFacet.sol` with `IERC20(t).safeTransfer(...)`: exit 1, offending file + line printed
- Injected `src/EthOnlyFacet.sol` with only `safeTransferETH` calls (both method-style and library-style): exit 0
- `bunx prettier --check` on the workflow file: pass

## DRAFT status / pr-ready note

`/pr-ready` (local CodeRabbit) could not be run to a clean state: the CLI returned "Review limit reached" (org usage credits exhausted), twice. Per the pre-PR workflow this PR is therefore opened as a **draft**; cloud CodeRabbit will review on this PR as the safety net. A `self-review-pass` sweep (mechanical/semantic/executable) was completed with no findings.

- [ ] I have run `/pr-ready` (local CodeRabbit) on this branch and resolved (or explicitly documented) all findings — see `.agents/commands/pr-ready.md` (attempted; rate-limited — documented above)
- [x] I have added tests that cover the functionality / test the bug (manual positive/negative/exemption runs of the check script, evidence above; no Solidity changes)
- [ ] For new facets: I have checked all points from this list: https://www.notion.so/lifi/New-Facet-Contract-Checklist-157f0ff14ac78095a2b8f999d655622e (N/A — no facets touched)
- [x] I have updated any required documentation (workflow header documents purpose, triggers, behavior, limitations)

### DIFF
--- a/.github/workflows/enforceLibAssetRouting.yml
+++ b/.github/workflows/enforceLibAssetRouting.yml
@@ -1,0 +1,88 @@
+# Enforce LibAsset Routing
+# - enforces the architectural invariant that all ERC20 value-flow in src/ goes through LibAsset
+#   (audit follow-up: Sujith finding #4 on PR #1715 / Tron USDT routing migration, ticket EXSC-300)
+# - fails the build if any .sol file in src/ outside the allowlist calls .safeTransfer( or
+#   .safeTransferFrom( directly (Solmate / Solady / OZ SafeTransferLib-style calls)
+# - native-ETH transfers via safeTransferETH( remain allowed everywhere
+# - pre-existing offenders are grandfathered via an explicit allowlist in the check step;
+#   new files (or new call sites in previously clean files) are rejected
+# - known limitations: text-based grep, so a match inside a comment or string literal would
+#   also be flagged (acceptable as a backstop; remove the comment or route through LibAsset)
+
+name: Enforce LibAsset Routing
+on:
+  pull_request:
+    types: [opened, synchronize, reopened, ready_for_review]
+
+permissions:
+  contents: read # required to fetch repository contents
+
+jobs:
+  enforce-libasset-routing:
+    runs-on: ubuntu-latest
+
+    steps:
+      # Step 1: Checkout code
+      - name: Checkout code
+        uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
+
+      # Step 2: Scan src/ for direct ERC20 safeTransfer / safeTransferFrom calls
+      - name: Check for direct ERC20 transfers bypassing LibAsset
+        run: |
+          set -euo pipefail
+
+          ##### Files allowed to call .safeTransfer( / .safeTransferFrom( directly.
+          ##### Everything else in src/ must route ERC20 transfers through LibAsset.
+          ALLOWLIST=(
+            # canonical implementation - this is where ERC20 transfers are supposed to live
+            "src/Libraries/LibAsset.sol"
+            # gas-optimized packed facets bypass LibAsset by design (audit: Sujith #4, EXSC-300);
+            # migrating them is explicitly out of scope - see EXSC-300
+            "src/Facets/AcrossFacetPacked.sol"
+            "src/Facets/AcrossFacetPackedV3.sol"
+            "src/Facets/CBridgeFacetPacked.sol"
+            "src/Facets/HopFacetPacked.sol"
+            # pre-existing direct call sites grandfathered when this check was introduced (EXSC-300);
+            # do NOT add new files here without an explicit architectural decision
+            "src/Helpers/WithdrawablePeriphery.sol"
+            "src/Periphery/LiFiDEXAggregator.sol"
+            "src/Periphery/ReceiverAcrossV3.sol"
+            "src/Periphery/ReceiverChainflip.sol"
+            "src/Periphery/ReceiverStargateV2.sol"
+            "src/Periphery/TokenWrapper.sol"
+          )
+
+          ##### Matches .safeTransfer( and .safeTransferFrom( but NOT .safeTransferETH(
+          PATTERN='\.safeTransfer(From)?\('
+
+          ##### Initialize empty variables
+          VIOLATIONS=""
+
+          ##### collect every src/ Solidity file containing a direct ERC20 transfer call
+          ##### (grep exits 1 when nothing matches, which is fine - the loop body just never runs)
+          while IFS= read -r FILE; do
+            ##### skip allowlisted files
+            for ALLOWED in "${ALLOWLIST[@]}"; do
+              if [[ "$FILE" == "$ALLOWED" ]]; then
+                continue 2
+              fi
+            done
+            VIOLATIONS="${VIOLATIONS}${FILE}"$'\n'
+          done < <(grep -rlE "$PATTERN" src --include='*.sol' || true)
+
+          ##### fail with the offending lines if any non-allowlisted file matched
+          if [[ -n "$VIOLATIONS" ]]; then
+            echo -e "\033[31mError: direct ERC20 .safeTransfer/.safeTransferFrom calls found outside LibAsset:\033[0m"
+            while IFS= read -r FILE; do
+              if [[ -n "$FILE" ]]; then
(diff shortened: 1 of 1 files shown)


.github/workflows/enforceLibAssetRouting.yml
```diff
@@ -0,0 +1,88 @@
...
+        uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
+
+      # Step 2: Scan src/ for direct ERC20 safeTransfer / safeTransferFrom calls
+      - name: Check for direct ERC20 transfers bypassing LibAsset
+        run: |
+          set -euo pipefail
+
+          ##### Files allowed to call .safeTransfer( / .safeTransferFrom( directly.
+          ##### Everything else in src/ must route ERC20 transfers through LibAsset.
+          ALLOWLIST=(
```
mirooon, line 40: **The hardcoded allowlist is in the wrong place — this is a smart-contract decision, not a CI-infra one**

Because this list lives in the workflow under `.github/`, any edit to it triggers `protectSecurityRelevantCode.yml`, which requires an **InformationSecurityManager** approval. That gate is there to protect our CI integrity (so checks can't be quietly disabled) — it is not the team that reviews contract logic.

But "may this contract bypass LibAsset?" is a **smart-contract security** decision. We review those in normal SC PR review, not through the InfoSec-infra gate. So with the list here, every time someone grandfathers or removes a contract, they pull in the wrong approvers for a decision that should sit with SC reviewers. The list is also disconnected from the contracts it describes, so it goes stale on a rename or deprecation.

I'd suggest moving the list into `config/global.json`, next to the lists we already keep there (`coreFacets`, `corePeriphery`, `blacklistedFunctionSelectors`). For example:

```json
"libAssetRoutingExempt": [
  "AcrossFacetPacked",
  "AcrossFacetPackedV3",
  "CBridgeFacetPacked",
  "HopFacetPacked",
  "WithdrawablePeriphery",
  "LiFiDEXAggregator",
  "ReceiverAcrossV3",
  "ReceiverChainflip",
  "ReceiverStargateV2",
  "TokenWrapper"
]
```

The workflow then reads it with `jq` instead of carrying the array inline. `config/` is not under `.github/`, so edits go through normal SC review (the right people) — while the workflow *logic* stays protected by the InfoSec gate, which is what that gate is actually for. What do you think?

.github/workflows/enforceLibAssetRouting.yml
```diff
@@ -0,0 +1,88 @@
...
+            "src/Helpers/WithdrawablePeriphery.sol"
+            "src/Periphery/LiFiDEXAggregator.sol"
+            "src/Periphery/ReceiverAcrossV3.sol"
+            "src/Periphery/ReceiverChainflip.sol"
+            "src/Periphery/ReceiverStargateV2.sol"
+            "src/Periphery/TokenWrapper.sol"
+          )
+
+          ##### Matches .safeTransfer( and .safeTransferFrom( but NOT .safeTransferETH(
+          PATTERN='\.safeTransfer(From)?\('
```
mirooon: **Coverage gap: raw `.transfer` / `.transferFrom` are not caught**

The regex only matches `.safeTransfer(` and `.safeTransferFrom(`. It does **not** match raw `.transfer(` / `.transferFrom(`. I checked `src/` and `LidoWrapper.sol` uses the raw form:

```
src/Periphery/LidoWrapper.sol:71   IERC20(address(ST_ETH)).transferFrom(...)
src/Periphery/LidoWrapper.sol:85   IERC20(WST_ETH_ADDRESS).transfer(msg.sender, wrappedAmount);
```

This stood out to me because audit finding #4 is literally titled *"Out-of-scope facets **and LidoWrapper** bypass LibAsset"*. So the one contract the auditor named is not caught by the check **and** is not in the allowlist — it just passes silently.

The bigger risk is the bug class itself: the original Tron USDT problem was a raw `transfer()` returning bad data. With the current regex, a new facet that calls `IERC20(token).transfer(...)` directly would pass CI and could bring that exact bug back.

Two ways to handle it:
1. Widen the pattern to also catch raw transfers, e.g. `\.(safe)?[Tt]ransfer(From)?\(`, and add `LidoWrapper.sol` to the allowlist with a short comment.
2. Keep the narrow scope, but then document in the header that raw `.transfer`/`.transferFrom` is intentionally out of scope and LidoWrapper is a known gap.

I'd lean toward option 1, since it closes the exact bug class this check exists to prevent. What do you think?

.github/workflows/enforceLibAssetRouting.yml
0xDEnYO: Thanks for your comment. I see your point but I have a different opinion. 
Using a bypass should be avoided at any cost. The exemptions are there only for legacy so we dont need to change a number of already-deployed contracts. 

It should not be easy for a developer to not use LibAsset. If we just allow setting a flag in global.json, we make it rather easy. Having it where it is right now requires the additional security approval which is an additional check.

 I would suggest to leave it where it is, since we wont plan to use this regularly as a feature, it is a legacy exemption only.

.github/workflows/enforceLibAssetRouting.yml
0xDEnYO: Went with option 1 — implemented in 4ea9719d:

- Pattern widened to `\.(safeT|t)ransfer(From)?\(` so raw `.transfer(`/`.transferFrom(` are now caught too (still excludes `.safeTransferETH(`).
- `LidoWrapper.sol` added to the allowlist as grandfathered (it's the contract named in audit finding #4), with a comment marking it as not endorsed.
- The widened pattern also flagged `Executor.sol:168` — that one is a false positive: it calls `transferFrom` on our own `ERC20Proxy`, which itself routes through `LibAsset.transferFromERC20`. Allowlisted with a comment explaining that.

Verified locally: the check passes on the current tree, fails on a synthetic raw-`transfer` violation, and all four call forms match while `safeTransferETH(` doesn't.

0xDEnYO:
@coderabbitai The `persist-credentials: false` nitpick is addressed in 88ad8319.

mirooon (approved):

maxklenk (approved):

PR #1901, opened 2026-06-10
feat(EXSC-331): static validation of touched TS scripts (pre-push hook + CI workflow)

0xDEnYO:

[EXSC-331 — CI: dedicated workflow to validate scripts when they are touched](https://linear.app/lifi-linear/issue/EXSC-331/ci-dedicated-workflow-to-validate-scripts-when-they-are-touched)

Per the approach decided on the ticket: scripts are **not executed** (flaky/slow). Instead a deterministic static check, delivered as both a local pre-push hook (fast feedback) and a path-gated CI workflow (enforcement backstop, since hooks are bypassable with `--no-verify`):

- **`script/utils/validateScripts.ts`** (also exposed as `bun validate-scripts`) does two things:
  1. **Type check** — `tsc-files --noEmit` over the TS files under `script/` changed between a base ref and head. Removing a used import fails with TS2304; removing a dependency a script imports fails with TS2307. A full-repo `tsc` gate was not feasible: `tsc --noEmit -p tsconfig.json` is currently red (missing generated `typechain/` plus pre-existing errors in a handful of legacy scripts), so the check is scoped to changed files — same fix-on-touch semantics the pre-commit hook already applies via lint-staged.
  2. **Import resolution sweep** — when dependency manifests (`package.json` / `bun.lock` / `tsconfig.json`) change, every bare import specifier across ALL `script/**` TS files must be declared in `package.json` (**strict declared-only**, per review decision — relying on transitively installed packages breaks silently when the parent dependency drops them; `bs58`, previously relied on via `@layerzerolabs/lz-v2-utilities`, is now declared explicitly and its eslint suppression removed). Type-only imports are exempt (erased at runtime by `tsx`). This catches the exact scenario from the PR #1791 discussion — dependency removed while `script/mongoDb/fetch-rpcs.ts` still imports it — even when no `.ts` file was touched.
- **`.husky/pre-push`** — runs the validator per pushed ref (base = remote sha when it exists locally, otherwise merge-base with `origin/main`). Uses the same husky-v8 `bash -e` re-exec workaround as `.husky/pre-commit`.
- **`.github/workflows/validateScripts.yml`** — follows the `forge-unit-tests.yml` house pattern: cheap `detect-changes` (dorny/paths-filter) + path-gated heavy job + an always-reporting `validate-scripts-required` aggregator so it can be registered as a required check on `main`. TypeChain types are generated in the job because several scripts (demoScripts, tron deploy scripts via `demoScriptHelpers`) import generated `typechain/` types. `forge-unit-tests.yml` stays solidity-only (untouched).

Coordination with EXSC-368: no overlap — EXSC-368 guards the *bash* `universalCast`/`sendOrPropose` routing behaviourally (Anvil smoke tests / testnet dry-runs); this PR is a *TypeScript static* check only and does not touch `.sh` scripts.

**Verification evidence (local dry-runs):**

- removed `MongoClient` import from `fetch-rpcs.ts` → type check fails (TS2304), exit 1 ✅
- removed `mongodb` from `package.json` (+ simulated fresh install) → sweep flags 10 importing files, exit 1 ✅
- manifest touched with deps intact → sweep passes over 138 script files, exit 0 ✅
- no relevant changes → "nothing to validate", exit 0 ✅
- strict-mode negative test: with `bs58` undeclared, the sweep flags `demoScriptHelpers.ts`, exit 1 ✅; declared → sweep passes over 138 files, exit 0 ✅
- pre-push hook stdin simulation (new-branch ref line) → exit 0 ✅ and the hook ran live on the actual push of this branch
- `bash -n .husky/pre-push`, `bunx eslint` + `bunx tsc-files --noEmit` + prettier on `validateScripts.ts`: clean ✅

**For the reviewer / follow-up:** after merge, register `validate-scripts-required` as a required status check on `main` branch protection (repo settings — needs admin).

- [x] I have run `/pr-ready` (local CodeRabbit) on this branch and resolved (or explicitly documented) all findings — see `.agents/commands/pr-ready.md` (zero findings)
- [x] I have added tests that cover the functionality / test the bug (deterministic dry-run evidence above; the validator is itself a static check harness)
- [ ] For new facets: I have checked all points from this list: https://www.notion.so/lifi/New-Facet-Contract-Checklist-157f0ff14ac78095a2b8f999d655622e (N/A — no contracts touched)

### DIFF
--- a/.github/workflows/validateScripts.yml
+++ b/.github/workflows/validateScripts.yml
@@ -1,0 +1,130 @@
+# Validate Scripts (EXSC-331)
+# - Static validation (no script execution) of TS scripts under script/
+# - Type check: `tsc-files --noEmit` over changed TS files under script/ —
+#   removing a used import (TS2304) or a dependency a script imports (TS2307)
+#   fails fast instead of breaking at runtime
+# - Import resolution: when package.json / bun.lock / tsconfig.json change,
+#   every bare import in script TS files must resolve to a declared or
+#   installed package (catches "dependency removed while still imported"
+#   even when no .ts file was touched)
+# - Path-gated: the heavy job runs only when scripts or dependency manifests
+#   change. A `validate-scripts-required` aggregator job always reports a
+#   final status so this workflow can be registered as a required check on
+#   `main` branch protection (see .agents/rules/500-github-actions.md)
+# - TypeChain types are generated before the type check because several
+#   scripts import generated typechain/ types (requires foundry + forge build)
+# - Local counterpart: the .husky/pre-push hook runs the same validator for
+#   fast feedback; this workflow is the enforcement backstop since git hooks
+#   can be bypassed with --no-verify
+
+name: Validate Scripts
+on:
+  pull_request:
+    types: [opened, synchronize, reopened, ready_for_review]
+
+  # Allows to run this workflow manually from the Actions tab
+  workflow_dispatch:
+
+permissions:
+  contents: read # required to fetch repository contents
+  pull-requests: read # required by dorny/paths-filter to read changed files on PRs
+
+jobs:
+  detect-changes:
+    # Determines whether files that can influence script validity were touched.
+    # Always runs (cheap) so the downstream validate-scripts-required can report status.
+    runs-on: ubuntu-latest
+    outputs:
+      scripts: ${{ steps.filter.outputs.scripts }}
+    steps:
+      - uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
+      - name: Detect relevant changes
+        id: filter
+        uses: dorny/paths-filter@de90cc6fb38fc0963ad72b210f1f284cd68cea36 # v3.0.2
+        with:
+          filters: |
+            scripts:
+              - 'script/**/*.ts'
+              - 'package.json'
+              - 'bun.lock'
+              - 'tsconfig.json'
+              - '.github/workflows/validateScripts.yml'
+
+  validate-scripts:
+    # Runs on PRs (once out of draft) and on workflow_dispatch.
+    # For non-PR events `github.event.pull_request` is null, so the draft check
+    # is gated behind an event-name guard.
+    needs: detect-changes
+    if: >-
+      ${{
+        needs.detect-changes.outputs.scripts == 'true' &&
+        (github.event_name != 'pull_request' ||
+         github.event.pull_request.draft == false)
+      }}
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
+        with:
+          # Full history so the validator can compute the merge-base with the
+          # PR base branch; submodules for `forge build src` (TypeChain input)
+          fetch-depth: 0
+          submodules: recursive
+
+      - name: Set up Bun
+        uses: oven-sh/setup-bun@0c5077e51419868618aeaa5fe8019c62421857d6 # v2
+
+      - name: Install dev dependencies
+        run: bun install
+
+      - name: Install Foundry
+        uses: ./.github/actions/setup-foundry
+
+      - name: Install forge Dependencies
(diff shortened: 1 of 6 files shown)


.github/workflows/validateScripts.yml
```diff
@@ -0,0 +1,130 @@
...
+
+jobs:
+  detect-changes:
+    # Determines whether files that can influence script validity were touched.
+    # Always runs (cheap) so the downstream validate-scripts-required can report status.
+    runs-on: ubuntu-latest
+    outputs:
+      scripts: ${{ steps.filter.outputs.scripts }}
+    steps:
+      - uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
```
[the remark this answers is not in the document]
0xDEnYO: Applied in dad48ef6 for the `detect-changes` checkout (`persist-credentials: false` — read-only job, no git operations after checkout).

Deliberately NOT applied to the `validate-scripts` checkout: it uses `submodules: recursive`, and `persist-credentials: false` has known interactions with the post-job submodule credential cleanup (can fail the job). The job is read-only either way (`permissions: contents: read`, no authenticated git writes), so the hardening value there is marginal versus the breakage risk — consistent with the rest of this repo's workflows, none of which set it on submodule checkouts.

script/utils/validateScripts.ts
```diff
@@ -0,0 +1,280 @@
...
+      consola.success(
+        `No script/**/*.ts or dependency manifest changes between ${base} and ${head} — nothing to validate`
+      )
+      return
+    }
+
+    let passed = true
+    if (changedScriptFiles.length > 0)
+      passed = runTypeCheck(repoRoot, changedScriptFiles) && passed
+    if (manifestsChanged) passed = runImportResolutionCheck(repoRoot) && passed
```
gvladika: **Import-resolution sweep only runs on manifest changes — undeclared imports can be introduced without detection.**

This gate is the only thing that triggers `runImportResolutionCheck`, but the transitive-dependency footgun this PR targets can be *introduced* in a PR that adds an import yet touches no manifest. In that case:

- the type check (`tsc-files`) resolves the import fine — the package is physically in `node_modules` via another dependency, so no TS2307; and
- the sweep is skipped, since no manifest changed.

So the bad import lands and only surfaces later, when some unrelated `package.json`/`bun.lock`/`tsconfig.json` edit finally trips the sweep — exactly the "breaks silently" scenario the strict declared-only policy exists to prevent, just deferred.

Since the sweep is pure AST parsing + a `Set` lookup over ~138 files (no I/O, sub-second), gating it on manifest changes is a micro-optimization that costs real coverage. Suggest also running it whenever script files changed, so undeclared imports are caught at introduction:

```ts
if (manifestsChanged || changedScriptFiles.length > 0)
  passed = runImportResolutionCheck(repoRoot) && passed
```

0xDEnYO: Good catch — agreed, and fixed in 6106e88e.

You're right that the manifest-only gate left a hole: an undeclared import added in a script (no manifest touched) resolves fine for `tsc-files` (the package is physically present via a transitive dep, so no TS2307) and skipped the sweep, only surfacing when a later unrelated manifest edit tripped it — the deferred-"breaks silently" case the strict policy is meant to prevent.

The sweep now runs whenever **script files OR manifests** changed:

```ts
if (manifestsChanged || changedScriptFiles.length > 0)
  passed = runImportResolutionCheck(repoRoot) && passed
```

As you noted it's pure AST + `Set` lookup over ~138 files (sub-second, no I/O), so the added coverage is effectively free. The CI workflow already path-gates the heavy job on scripts-or-manifests, so no workflow change was needed. Updated the header docstring and dropped the now-inaccurate "Dependency manifest changed" log prefix.

Verified locally:
- undeclared import in a script, no manifest change → caught, exit 1 ✅
- declared import in a script → passes, no false positive ✅
- no relevant change → early return, exit 0 ✅

gvladika (approved):

maxklenk (approved):

PR #1902, opened 2026-06-10
feat: combine whitelist removal and addition into a single timelock proposal

0xDEnYO:

[EXSC-247](https://linear.app/lifi-linear/issue/EXSC-247/combine-whitelist-removal-and-addition-into-a-single-timelock-proposal) — Combine whitelist removal and addition into a single timelock proposal

`batchSetContractSelectorWhitelist(address[],bytes4[],bool)` takes a single bool for the whole batch, so removals and additions cannot be mixed in one diamond call. The timelock controller's `scheduleBatch(address[],uint256[],bytes[],...)` however already accepts arrays of inner calls — so the combination happens at the timelock-wrapping layer instead:

- **`propose-to-safe.ts`** now accepts repeated `--to`/`--calldata` pairs. Multiple pairs require `--timelock` (without the timelock there is no batch mechanism short of a Safe MultiSend, which is out of scope); inputs are validated up front (count match, no `--calldataFile` with multiple pairs) before any network/Safe interaction.
- **`wrapWithTimelockSchedule()`** (`safe-utils.ts`) accepts arrays of targets/calldatas instead of always encoding batch-of-one. The pure encoding was extracted to `encodeTimelockScheduleBatch()` in `timelock-abi.ts` (kept dependency-light) so it is unit-testable without RPC mocks.
- **`sendOrPropose()`** (`helperFunctions.sh`) accepts comma-separated calldatas (calldata is hex, so commas are unambiguous). Routing stays in this one helper per `[CONV:ROUTING-PREDICATE]`: EVM propose route with timelock → ONE combined `scheduleBatch` proposal; all other routes (direct send, Tron propose, propose without timelock) → multiple calldatas are **rejected** (the only multi-calldata caller is gated on the timelock route; sequential fan-out elsewhere would be untested dead generality). Inputs are validated strictly (no whitespace, well-formed hex per element).
- **`diamondSyncWhitelist.sh`**: when a sync needs BOTH removals and additions on the EVM timelock route, Stage 4a defers the removal calldata, Stage 4b defers each addition-batch calldata, and a new Stage 4c submits them as inner calls of a single proposal — removal first, so execution ordering (remove before add) is guaranteed by `scheduleBatch` array order. Add-only, remove-only, staging/direct-send, testnet, and Tron flows are unchanged. If the *removal* calldata generation fails, the script falls back to the previous two-proposal flow; if an *addition*-batch calldata generation fails, the sync aborts before Stage 4c with an explicit warning that the deferred removal was not proposed either — nothing has been submitted at that point, so a re-run recovers cleanly.

Downstream tooling (`confirm-safe-tx.ts`, `execute-pending-timelock-tx.ts` via `executeBatch`) is already array-aware, so combined proposals decode and execute without changes.

## Hardening round (2026-06-11, multi-agent design + code review)

A 34-agent adversarially-verified review confirmed the timelock-wrapping-layer design as the right approach (vs. contract-level change, Safe MultiSend, signing-UX batching — all evaluated and rejected with evidence) and surfaced six findings, all fixed:

1. **Combined-proposal size cap (≤300 pairs)** — `executeBatch` runs all inner calls in ONE tx, so unbounded combination undid Stage 4b's 150-pair gas batching and could yield a scheduled-but-unexecutable proposal that atomically blocks the removal. Above the cap the sync falls back to sequential proposals.
2. **Strict `cast calldata` capture** — stdout-only (stderr to a temp file for diagnostics) + strict single-line hex match, so foundry warnings/update notices can't poison the comma-joined calldata list.
3. **citty repeated-flag guard** — citty 0.2.x silently keeps only the LAST `--to`/`--calldata` value (would drop inner calls). Runtime argv cross-check fails loudly + a regression test pins the 0.1.x array contract.
4. **Hex/address validation** in `normalizeProposeCalls` + `encodeTimelockScheduleBatch` — viem silently zero-pads non-hex calldata into bytes that only fail at execution time, days after signing. Empty (`0x`) payloads rejected in multi-call batches (allowed for single calls, e.g. nonce-gap fillers).
5. **Multi-calldata rejected on non-timelock routes** in `sendOrPropose` — the sequential fan-out loops on direct-send/Tron/non-timelock routes were unreachable dead generality; now an explicit error.
6. **CLI normalization extracted to `propose-calls.ts` and unit-tested** (15 new tests; `propose-to-safe.ts` can't be imported in tests since top-level `runMain` fires the CLI).

## Verification

- `bun test script/deploy/safe/` — 78 pass, 0 fail (includes new `timelock-abi.test.ts`: multi-call encoding, remove-before-add ordering, batch-of-one, error paths; and `propose-calls.test.ts`: multi-pair CLI normalization, validation paths, citty repeated-flag contract)
- `bunx eslint` + `bunx tsc-files --noEmit` clean on all touched TS files; `bunx prettier --check` clean
- `bash -n` clean on `helperFunctions.sh` and `diamondSyncWhitelist.sh`
- `sendOrPropose` routing exercised against stubs for all four routes (staging multi → sequential sendRaw; production multi+timelock → single combined proposal invocation; production single; production multi w/o timelock → sequential proposals; invalid calldata → rejected)
- `cast calldata` invocation format for the combined route verified against the real `cast` binary

## /pr-ready result (2026-06-10, after merge with main)

Local CodeRabbit ran after the org credit reset: **1 finding (minor / potential_issue)** — in combined mode, a failed addition-batch calldata generation aborts before Stage 4c without saying that the deferred removal was also not proposed. Fixed in commit `a775f5dc` (explicit warning; no behavioral change — nothing has been submitted at that point, so abort + re-run was already the safe recovery). Scoped re-verify of that commit could not run (hourly review rate limit); cloud CodeRabbit on this PR is the safety net for it.

- [ ] For new facets: I have checked all points from this list: https://www.notion.so/lifi/New-Facet-Contract-Checklist-157f0ff14ac78095a2b8f999d655622e (N/A — no new facets)


script/tasks/diamondSyncWhitelist.sh
```diff
@@ -240,6 +240,14 @@ function diamondSyncWhitelist {
     # sendOrPropose: propose vs send is decided by SEND_PROPOSALS_DIRECTLY_TO_DIAMOND and environment. TIMELOCK_FLAG only used when proposing (wrap in timelock).
     local TIMELOCK_FLAG=$(getTimelockFlag "$NETWORK" "$ENVIRONMENT")
 
+    # Combined remove+add proposal route [CONV:ROUTING-PREDICATE]:
+    # When a sync needs BOTH removals and additions on the EVM timelock-proposal route,
+    # all calldatas are wrapped into ONE timelock scheduleBatch proposal (single signing
+    # round, removals execute before additions). Tron is excluded (propose-to-safe-tron.ts
+    # has no batch support). The flag is finalized once REMOVED_PAIRS/NEW_PAIRS are known.
```
mirooon: Minor nit on the rationale here: "no batch support" slightly overstates the barrier — I think `propose-to-safe-tron.ts` could be adapted to batch fairly easily, and the rest of the pipeline is already array-aware:

- The Tron timelock uses the same `scheduleBatch(address[],uint256[],bytes[],...)` — `propose-to-safe-tron.ts` already calls it, just with hardcoded single-element arrays (e.g. `[[targetEvm], [0n], [innerCalldata], ...]`).
- `execute-pending-timelock-tx.ts` already handles **both** Tron (`isTronNetworkKey` branches) **and** array-based `executeBatch`, so a combined Tron proposal would decode/execute through existing tooling unchanged. (The "when Tron execution is supported" note in `propose-to-safe-tron.ts` looks stale.)

So the propose-side change looks small: make `--to`/`--calldata` repeatable (+ the same citty repeated-flag guard), map `tronBase58ToEvm20Hex` over the targets, and **reuse `encodeTimelockScheduleBatch()`** instead of the inline single-element encode. Then drop `isTronNetwork` from the `sendOrPropose` multi-calldata guard and `IS_TRON == false` from this gate.

Totally fine to keep Tron out of scope for this PR (it's a parallel CLI/normalization refactor that deserves its own tests + Tron-specific verification). But could we reword the comment so it reflects the real reason — "Tron batching deferred; the propose script takes a single call, while the timelock and executor already support arrays" — rather than implying a capability gap? Happy to open a follow-up ticket for the Tron parity refactor.

script/deploy/safe/propose-calls.ts
```diff
@@ -0,0 +1,75 @@
...
+
+  // Multiple calls are only combinable via the timelock's scheduleBatch
+  if (targets.length > 1 && !options.timelock)
+    throw new Error(
+      'Multiple --to/--calldata pairs require --timelock (combined into a single scheduleBatch proposal)'
+    )
+
+  // Reject malformed inputs here, before signing: viem would silently zero-pad
+  // non-hex calldata into valid-looking bytes that only fail at execution time
+  for (const [i, target] of targets.entries())
```
mirooon: Minor optimization I noticed (non-blocking): this address/calldata validation loop — `isAddress(..., { strict: false })` + `isHex(..., { strict: true })` — is duplicated almost verbatim in `encodeTimelockScheduleBatch()` (`timelock-abi.ts:57-65`), down to the "viem silently zero-pads non-hex" rationale comment.

In the actual call path the two run back-to-back on the same arrays (`normalizeProposeCalls` → `wrapWithTimelockSchedule` → `encodeTimelockScheduleBatch`), and since `encodeTimelockScheduleBatch` has only that one caller, the second pass re-validates input the first already guaranteed. Looks like there's room to consolidate the per-element validation into a single place. Not blocking — just flagging it for a possible follow-up.

script/tasks/diamondSyncWhitelist.sh
0xDEnYO: Reworded in a9bbfbff — the comment now states the actual barrier (propose-to-safe-tron.ts takes a single call today, so Tron stays on the sequential route) and notes that the Tron timelock and execute-pending-timelock-tx.ts already handle arrays, instead of implying a capability gap. Tron batch proposals stay out of scope for this PR; a follow-up ticket for the parity refactor sounds right.

script/deploy/safe/propose-calls.ts
0xDEnYO: Consolidated in a9bbfbff: the per-element loops now live once as validateCallPairs() in timelock-abi.ts, parameterized by error-message label. Kept both call sites rather than dropping one — normalizeProposeCalls also guards the single-call non-timelock route (which never reaches the encoder), and the encoder-side check protects future direct callers (e.g. the Tron batch follow-up) — but the validation logic itself is now single-sourced.

script/tasks/diamondSyncWhitelist.sh
mirooon: Sorry for the confusion here — following up on my own comment to correct it: I was too quick to wave Tron out of scope. I think Tron parity actually belongs **in this PR**, not a follow-up ticket.

My reasoning: the whole point of EXSC-247 is the single-proposal, remove-before-add ordering guarantee. Leaving Tron on the old two-proposal sequential route means the ordering invariant this PR establishes silently doesn't hold on Tron — so the fix is incomplete exactly where it's least visible. And the remaining gap is genuinely small and isolated to the propose side:

- The Tron timelock already exposes the same `scheduleBatch(address[],uint256[],bytes[],...)` — `propose-to-safe-tron.ts` just hardcodes single-element arrays (`[targetEvm]`, `[0n]`, `[innerCalldata]`).
- `execute-pending-timelock-tx.ts` already executes `executeBatch(...)` through its chain-agnostic `chainCaller` (Tron-aware via `isTronNetworkKey`), so a combined Tron proposal decodes and executes through existing tooling unchanged. The `"...when Tron execution is supported"` note in `propose-to-safe-tron.ts` is stale and can go.

I know this grows the PR, but it keeps the invariant honest across all timelock chains and avoids shipping a known asymmetry.

0xDEnYO: Agreed — Tron parity now lands in this PR (pushed in `6e2931d3`). You were right that the gap is propose-side only; the Tron timelock and `execute-pending-timelock-tx.ts` already handle arrays.

What changed:
- **`propose-to-safe-tron.ts`**: `--to`/`--calldata` are now repeatable (+ the same citty repeated-flag argv guard as the EVM script), targets are mapped through `tronBase58ToEvm20Hex`, and it reuses **`encodeTimelockScheduleBatch()`** instead of the inline single-element encode. Stale "when Tron execution is supported" note removed.
- **`propose-calls-tron.ts`** (new): pure, unit-tested Tron call-pair normalizer (mirrors `propose-calls.ts`, which exists because `runMain` fires on import).
- **`sendOrPropose`**: multiple calldatas now allowed on the Tron production timelock route (dropped `isTronNetwork` from the reject guard); the Tron branch loops repeated pairs into one `scheduleBatch`.
- **`diamondSyncWhitelist.sh`**: the Stage-4 combine gate now allows Tron **production**.

One Tron-specific detail that wasn't obvious from the propose side: the deferred combined `cast calldata` blocks (Stage 4a/4b) were encoding the **base58** address arrays. That's fine for EVM but breaks `address[]` ABI-encoding on Tron, so they now encode `$CONTRACTS_FOR_SEND` (hex-converted on Tron production; identical to the base58 array on EVM). Tron **staging** therefore stays on the sequential route — only the production branch produces hex-converted addresses.

Verified: `bun test script/deploy/{safe,tron}/` green (incl. new `propose-calls-tron.test.ts`), `sendOrPropose` routing exercised against stubs for every route (Tron prod+timelock multi → one combined proposal; Tron-direct-multi / Tron-staging-multi / non-timelock → rejected; EVM unchanged), `bash -n` / tsc / eslint clean. The one piece I can't run locally is a live Tron Shasta combined-proposal dry-run (needs keys) — flagging that as the residual manual check.

script/deploy/safe/propose-to-safe.ts
```diff
@@ -50,6 +56,8 @@ import {
  * @param options - Options including network, rpcUrl, privateKey, to address, and calldata
  */
 export async function runPropose(options: IProposeToSafeOptions) {
+  const { targets, calldatas } = normalizeProposeCalls(options)
+
```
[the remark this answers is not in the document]
0xDEnYO: The conflict guard is implemented one frame down, as the first statement of `normalizeProposeCalls` (`propose-calls.ts`), which `runPropose` calls on its very first line (`propose-to-safe.ts:59`) — so it fires before any signing and covers every entry point that builds proposal options (including the CLI `run()` path at 356-361):

```ts
if (options.calldataFile) {
  // CLI defaults --calldata to '' when omitted, so "provided" means non-empty
  if (options.calldata && options.calldata.length > 0)
    throw new Error('Provide either --calldata or --calldataFile, not both')
```

I kept it inside `normalizeProposeCalls` (its only caller is `runPropose`) rather than duplicating it at the call site, to match the single-source-of-validation pattern this PR already consolidated for the address/calldata checks (`validateCallPairs`). Note the empty-string nuance: the citty CLI passes `calldata: (args.calldata ?? '')`, so a bare `!== undefined` check would false-positive on legitimate file-only usage — hence the non-empty test. Covered by a new unit test (`propose-calls.test.ts`: "rejects calldata and calldataFile provided together" + "treats an empty --calldata default as absent alongside calldataFile"). Resolving.

mirooon (approved):

PR #1909, opened 2026-06-10
feat(safe): use openchain.xyz for selector lookup, drop anyabi.xyz

gvladika:

Fixes EXSC-491

The Safe transaction decoding flow resolves function selectors to signatures via an external signature database, used in two places:

- `safe-decode-utils.ts` → `decodeTransactionData` (drives `confirm-safe-tx.ts` calldata display)
- `safe-utils.ts` → `lookupSelectorFromFourByte` / `decodeDiamondCut` (drives diamondCut selector names)

**1. Prefer openchain.xyz, with Sourcify as fallback.** Both sites previously queried only the Sourcify mirror (`api.4byte.sourcify.dev`). That mirror was adopted (canceled ticket EXSC-175) on the premise that openchain.xyz would migrate away — but openchain.xyz is the signature database **Foundry's `cast 4byte` uses**, and since this repo is Foundry-native, aligning on it gives the most complete and consistent decoding. That is the main reason it's more complete, and the reason for migrating back to it. Now both query **openchain.xyz first** and fall back to the Sourcify mirror only when openchain is unreachable or has no match. The two endpoints share the same response shape (parsing unchanged) and both keep `filter=true` to suppress spam/collision signatures. The local `diamond.json` ABI remains the primary resolver before any remote lookup. Lookups stay HTTP via `fetchWithTimeout` (per `[CONV:FETCH-TIMEOUT]`) rather than shelling out to `cast`.

**2. Drop the anyabi.xyz facet-ABI fetch from `decodeDiamondCut`.** When the local `diamond.json` map was unavailable (it's generated/gitignored, so it's absent in fresh checkouts), the old code fetched the facet's full ABI from `anyabi.xyz` keyed by facet address. For Remove actions the facet address is the zero address, so that lookup always failed and spammed `Could not fetch ABI for facet 0x0…` once per cut entry. It also used a bare `fetch()` that bypassed `fetchWithTimeout`. Since selectors are already resolved individually (local diamond ABI → openchain → Sourcify), the facet-by-address fetch is redundant. Removed it; selector resolution is now a single map-optional loop that works for Add/Replace/Remove without depending on a live facet contract.

### DIFF
--- a/script/deploy/safe/safe-decode-utils.ts
+++ b/script/deploy/safe/safe-decode-utils.ts
@@ -34,8 +34,10 @@ import { decodeDiamondCut } from './safe-utils'
 
 /**
  * Signature-database lookup endpoints, tried in order. openchain.xyz is the
- * canonical source `cast 4byte` queries; the Sourcify mirror is a fallback for
- * when openchain is unreachable or has no match. Both share the same API shape.
+ * database Foundry's `cast 4byte` uses; since this repo is Foundry-native,
+ * aligning on it gives the most complete, consistent decoding. The Sourcify
+ * mirror is a fallback for when openchain is unreachable or has no match. Both
+ * share the same API shape.
  */
 const SIGNATURE_LOOKUP_ENDPOINTS = [
   {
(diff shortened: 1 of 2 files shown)


gvladika:
Closing — the premise of this PR is invalid.

`api.openchain.xyz` and `api.4byte.sourcify.dev` are the **same database**: they return byte-identical responses for every selector tested (incl. spam like `0xdeadbeef` → `CodeIsLawZ95677371()` on both). Sourcify absorbed the OpenChain signature DB; `openchain.xyz` (the UI) now redirects to `4byte.sourcify.dev`, and `api.openchain.xyz` only stays up because Foundry's `cast` is hardwired to it. So:

- "openchain is more complete" — false (identical data).
- "`filter=true` suppresses spam" — false (spam returned with `"filtered":false` on both).
- Sourcify is the canonical home; openchain is the legacy host being wound down (as EXSC-175 predicted). The existing Sourcify-only behavior was already correct.

The genuinely useful parts here were endpoint-independent and will be redone in a separate focused PR if wanted:
- removing the dead `anyabi.xyz` facet-ABI fetch in `decodeDiamondCut` (it spammed `Could not fetch ABI for facet 0x0…` on Remove cuts);
- robustness guards (`response.ok` check before `json()`, payload `ok` validation).

PR #1919, opened 2026-06-11
feat(EXSC-499): add non-interactive multi-network deploy script

gvladika:

[EXSC-499](https://linear.app/lifi-linear/issue/EXSC-499/add-non-interactive-multi-network-deploy-script)

`scriptMaster.sh` use case 1 is the standard way to deploy a facet and register it in the diamond, but its gum menus make it impossible to drive from automation, and deploying one contract to N networks means re-running the interactive flow once per network. This PR adds a thin non-interactive entry point:

```
./script/deploy/deployContractToNetworks.sh MayanFacet arbitrum base optimism [--production]
```

Design decisions:

- **Zero duplicated deploy logic.** The wrapper sources the exact same framework files as scriptMaster and calls the existing `deployAndAddContractToDiamond` / `deploySingleContract` functions. CREATE3 salts, retries, verification, log updates, and Safe-propose-vs-direct routing all stay where they are.
- **Production double opt-in** replaces the interactive "last chance" prompt: `--production` on the CLI **and** `PRODUCTION=true` in `.env` must both hold; a mismatch in either direction is a hard error. Default is staging.
- **Sequential network loop** (not parallel): deployment-log writes are per-network files plus atomic MongoDB upserts, so parallelism would be safe data-wise — sequential was chosen for v1 simplicity and readable output (parallel output interleaves, and verification shares one Etherscan V2 API rate limit). Parallelizing via the existing `multiNetworkExecution.sh` pattern is a possible follow-up. One failing network doesn't abort the rest; failures are collected and reported in a summary, exit code 1 if any failed.
- **No `set -euo pipefail`** (deliberate, same as scriptMaster): the sourced framework relies on `$?` checks and retry loops that strict mode would abort.
- **`TARGET_`-prefixed locals**: the sourced framework assigns generic names (`CONTRACT`, `NETWORK`, `VERSION`, ...) without `local`, which clobbers same-named caller variables via dynamic scoping — observed live (the summary printed a wrong contract name before the rename).
- **Prerequisite fix** in `deployAndAddContractToDiamond` (helperFunctions.sh): the facet and diamond branches hardcoded `return 0`, swallowing failures. Now `return $?`. **Behavior-change callout:** scriptMaster's `checkFailure` after this call (use case 1) previously could never fire for facets; after this fix a failed facet deployment correctly aborts scriptMaster instead of reporting success.

Testing performed (no bash test harness exists in the repo):

- `bash -n`, `shellcheck` (zero findings), `shfmt -i 2` (no diff).
- 9 validation-path smoke tests (help, missing args, unknown option, `--diamond` without value, invalid diamond, unknown network, unknown contract, production mismatch) — all exit with the specific expected error.
- Live end-to-end staging run: MayanFacet v1.3.0 deployed to polygon staging at `0xEB6803A40B8350286e1E8684068E28c5c57a94AE`, polygonscan-verified, diamond cut executed and confirmed on-chain (`facetAddress(0xce90a721)` on the staging diamond returns the new facet), deployment logs updated (committed here as artifacts). Idempotent re-run produced a correct summary and no artifact churn.
- Exit-code fix verified with stubbed success/failure callees before and after.

Known limitations / follow-ups (documented, intentionally out of scope):

- The periphery branch of `deployAndAddContractToDiamond` returns success if **either** deploy **or** registry update succeeded (`||` at helperFunctions.sh:3742-3746). Should this be `&&`? Left unchanged pending reviewer input; the wrapper's per-network OK/FAILED classification inherits this for periphery contracts (the facet path is correct).
- scriptMaster use case 2 (deploy to all networks) still ignores per-network exit codes — pre-existing, now more visible since the codes are meaningful.
- Rare framework paths hard-`exit` instead of returning (e.g. missing `.zksync.s.sol` variant on a zkEVM network), which bypasses the wrapper's failure collection for the remaining networks.
- On zkEVM networks, redeploying the same version with the same salt hits the framework's interactive `gum confirm`; in a non-TTY run this aborts (the right failure mode).
- No localanvil auto-start (scriptMaster feature; not needed for real chains).


gvladika:
@coderabbitai Both findings addressed in commit 1a383601: forge build exit status is now checked under COMPILE_ON_STARTUP (aborts before any deployment), and the task-script sourcing loop uses [[ ]] per the bash guidelines. Note: both fixes share one commit; the commit subject names only the forge-build fix.

mirooon (approved):

PR #1931, opened 2026-06-16
feat: add verify-contracts skill

0xDEnYO:

[EXSC-510](https://linear.app/lifi-linear/issue/EXSC-510/add-verify-contracts-skillcommand) — grew out of the Outlaw verification work in [EXSC-249](https://linear.app/lifi-linear/issue/EXSC-249/sc-jun-30-public-launch-outlaw), but the skill itself is generic tooling.

Adds a `/verify-contracts` command that captures the verified, working flow for verifying a network's deployed contracts on its block explorer — learned the hard way and previously undocumented.

- **Canonical-file + symlinks**, per the repo's agent-agnostic convention: the real file lives at `.agents/commands/verify-contracts.md`, with `.cursor/commands/verify-contracts.md` and `.claude/skills/verify-contracts/SKILL.md` as symlinks into it (scaffolded via the `/add-rule-or-skill` structure). README command table updated.
- **Direct loop over `deployments/<network>.json`** rather than `scriptMaster.sh` option 8 (`verifyAllUnverifiedContractsInLogFile`): that menu path reads the local `_deployments_log_file.json` cache, which is empty/stale for a freshly-deployed network. Looping the deployment JSON through `verifyContract` is the reliable path.
- **Two gotchas called out explicitly**, because each one cost real time:
  1. Submodules must be `git submodule update --init --recursive` in a fresh worktree, or `forge` can't compile and verification fails with the misleading `Details: Fail - Unable to verify`.
  2. On-chain verification does **not** write the MongoDB `verified` flag — only `logContractDeploymentInfo()` does. The skill flips the flag explicitly via `update-deployment-logs.ts`, matching records by address (case-insensitive) so superseded versions aren't falsely marked verified.
- **Constructor args pass `""`** — blockscout/sourcify match runtime bytecode, so they're not required.

All CLI flags referenced (`query-deployment-logs.ts list`, `update-deployment-logs.ts update`, `verifyContract`) were verified against the scripts at write time.

> **Note on `/pr-ready`**: ran locally against this branch in a git worktree and returned **No findings** (full mode, org-attributed). PR creation used the documented `PR_READY_OK=1` path only because the pre-PR gate evaluates the main-checkout `cwd` (where `HEAD` differs from the worktree), not the worktree where the marker is genuinely newer than `HEAD` — a known worktree/cwd gate misfire, not a skipped review.


.agents/commands/verify-contracts.md
```diff
@@ -0,0 +1,116 @@
...
+**Match records by ADDRESS (case-insensitive)** to the deployment map — a contract name can have multiple versions in Mongo (e.g. a superseded facet). Only flip the versions you actually verified on-chain; never blindly mark a superseded version `true` — its older bytecode won't match current `src/` and the claim would be false.
+
+### 6. Verify the outcome
+
+Re-query with `--no-use-cache` and confirm the verified count matches the deployment map:
+
+```bash
+bunx tsx script/deploy/query-deployment-logs.ts list \
+  --env "$ENVIRONMENT" --network "$NETWORK" --limit 200 --no-use-cache --format json \
+  | jq '[.[] | {contract, version, verified}]'
```
gvladika: This `jq` filter is wrong on two counts, and Step 6 is the outcome-confirmation step the skill tells the reader to trust — so as written it silently emits garbage instead of a per-contract list.

1. `list --format json` does **not** print a bare array. It wraps the records in an object — `JSON.stringify({ data: [...], pagination: {...} })` (`query-deployment-logs.ts` `outputJSON({ data, pagination })`). So `.[]` iterates the object's *values* (the `data` array and the `pagination` object), not the records. The accessor needs to be `.data[]`.
2. The record field is `contractName`, not `contract` (the Mongo schema in `update-deployment-logs.ts` uses `contractName` everywhere). `{contract, ...}` would emit `"contract": null` for every row.

Correct form:

```bash
... --format json | jq '[.data[] | {contract: .contractName, version, verified}]'
```

(Same `.[]`-vs-`.data[]` assumption is worth double-checking anywhere else the skill pipes `list --format json` into `jq`.)

.agents/commands/verify-contracts.md
```diff
@@ -0,0 +1,116 @@
...
+
+## Workflow
+
+### 1. Get the deployment address map onto disk
+
+- **By network** (already deployed, files on `main`): use `deployments/<network>.json` in the current checkout.
+- **By PR** (deploy PR — the deployment files exist only on the PR branch): check out that branch in a worktree:
+
+  ```bash
+  ~/.claude/scripts/contracts-wt-add.sh <pr-branch>   # symlinks .env; never copy it
```
gvladika: `~/.claude/scripts/contracts-wt-add.sh` is a personal, home-directory script — it isn't checked into the repo, so any other engineer (or an agent on a clean machine) running this skill won't have it and this step just fails for them. A repo-committed skill should be self-contained: either reference a repo-native worktree helper or give the plain `git worktree add` equivalent, and note that `contracts-wt-add.sh` is an optional personal convenience (it symlinks `.env`).

.agents/commands/verify-contracts.md
0xDEnYO: Good catch on both — fixed in 6bc5999b. `list --format json` emits `{data,pagination}` (verified in `query-deployment-logs.ts` `outputJSON({ data, pagination })`) and the record field is `contractName` (per the Mongo schema in `update-deployment-logs.ts`). Step 6 now reads `jq '[.data[] | {contract: .contractName, version, verified}]'`. The only other `jq` in the skill (Step 4) is `to_entries[]` over the deployment JSON, not `list` output, so it's unaffected.

.agents/commands/verify-contracts.md
0xDEnYO: Agreed — made it self-contained in 6bc5999b. Step 1 now uses repo-native `git worktree add` plus an explicit `.env` symlink (`ln -s "$(git rev-parse --show-toplevel)/.env" …`), and notes that a personal `contracts-wt-add.sh` helper is optional, not required.

gvladika (approved):

gvladika (approved):

PR #1936, opened 2026-06-16
feat(EXSC-417): vault wrapper factory (S8) [LiFiVaultWrapperFactory v1.0.0, LiFiVaultWrapperTypes v1.0.0, ILiFiVaultWrapper v1.0.0, MockVaultWrapper v1.0.0]

gvladika:

- [EXSC-417 — S8 Factory](https://linear.app/lifi-linear/issue/EXSC-417/s8-factory)
- [EXSC-514 — Yield adapter abstraction in vault wrapper factory](https://linear.app/lifi-linear/issue/EXSC-514/yield-adapter-abstraction-in-vault-wrapper-factory) (consolidated in from PR #1938)

Implements the **Programmable Vault Wrapper factory** (S8) plus the **yield-adapter abstraction** (EXSC-514, previously the stacked PR #1938, now merged into this branch). The factory is the single on-chain entry point that deploys per-integrator wrapper instances as deterministic beacon proxies, gated by a curated underlying allowlist, an approved-adapter registry, per-fee-type bounds/caps, deploy authorization, and a factory-level global circuit breaker.

Key decisions:

- **Standalone subsystem, not Diamond periphery.** The factory is not called by the Diamond, so it lives in a dedicated top-level `src/VaultWrapper/` folder rather than under `src/Periphery/`.
- **Beacon proxy + storage-init for clone identity.** Clones are deployed via Solady `LibClone.deployDeterministicERC1967BeaconProxy(beacon, salt)`. Solady's beacon proxies don't carry CWIA immutable args, so per-clone identity `(asset, underlying, adapter, vaultWrapperAdmin, integratorShareBps)` is set write-once in the clone's `initialize()`. Bytecode-immutable CWIA identity is **S1's** concern; the factory interface is unaffected either way.
- **Pluggable yield adapter (EXSC-514).** Asset derivation routes through a governance-approved `IYieldAdapter` (`resolveAsset(underlying) → asset`, reverts if it can't be resolved) rather than a hardcoded ERC-4626 asset lookup, so future yield sources (Aave v3, Compound, Morpho) need no factory change and no second beacon. `ERC4626Adapter` is the first adapter; the factory contains zero ERC-4626 references. The chosen adapter is folded into the CREATE2 salt, the `WrapperDeployed` event, `predictAddress`, and `initialize()` so each clone records which adapter it routes through at runtime. Wrapper runtime methods + a concrete `AaveV3Adapter` remain **S1 / wrapper-team** scope.
- **Mock implementation behind the beacon.** S8 is unblocked ahead of S1 (core ERC-4626 + CWIA) and S9 (beacon upgrade wiring) by pointing the beacon at `MockVaultWrapper`, built on Solady's ERC-4626 (constructor-free, so it initializes correctly through the beacon proxy) and recording its init args so tests can assert clone wiring. It is removed when S1's real implementation lands.
- **Namespace identity + deploy authorization (B+C).** Integrator identity is a `bytes32 namespace` (e.g. `"Coinbase"`) — chain-independent, so a wrapper keeps the same address across chains regardless of the integrator's per-chain addresses. Salt = `keccak256(namespace, adapter, underlying, nonce)`; duplicates revert. `deploy()` is callable by the onboarding manager (LI.FI) **or** the deployer assigned to that namespace via `approvedIntegratorDeployer[namespace]` (set by the onboarding manager) — binding the namespace to its deployer prevents squatting. The per-vault controller is a separate `vaultWrapperAdmin` (passed to `initialize`, kept out of the salt so each vault has its own rotatable admin), and the integrator fee split (`integratorShareBps`) is resolved per deploy (sentinel inherits the factory default; explicit values validated `<= 100%`) and snapshotted into the instance.
- **Underlying allowlist + adapter asset resolution.** Cooldown/lock rejection (A6) is curation-side (off-chain vetting per Q6.6); on-chain `deploy()` requires both the adapter approved and the underlying allowlisted, then the adapter resolves and returns the underlying's asset.
- **Fees.** Per-fee-type bounds are owner/timelock-set within immutable bytecode caps (perf 50% / mgmt 10% / deposit 20% / withdrawal 20%); disabled fee types must carry a zero rate; unset bounds fail closed. The integrator's share of the split is snapshotted per instance (`integratorShareBps`), while LI.FI's share routes to a factory-governed `lifiFeeRecipient` (owner-set, read live by wrappers at distribution) so an integrator cannot redirect LI.FI's cut via `initData`. Integrator-side receiver wallets remain **S1** `initData`.
- **Roles & breaker.** Three roles — `owner` (TransferrableOwnership, the 48h-timelock placeholder), `emergencyPauser` (global circuit breaker; clones read the auto-getter `globalPaused()`), `onboardingManager` (namespace→deployer assignment); owner rotates the latter two. The factory is constructed with `(beacon, owner, emergencyPauser, onboardingManager, lifiFeeRecipient)`; `lifiFeeRecipient` is owner-set/rotatable. Contract-existence checks use the repo's `LibAsset.isContract` (rejects EIP-7702 delegated EOAs).
- **Events/errors in an interface; specific errors.** All factory events and errors live in `ILiFiVaultWrapperFactory` (which the factory inherits), and the generic `GenericErrors.UnAuthorized` / `InvalidConfig` reverts are replaced with specific errors — `NotEmergencyPauser`, `NotOnboardingManager`, `NotApprovedDeployer`, `ZeroAddress`, `ZeroNamespace`, `InvalidFeeBounds`, `InvalidSplit`. The inherited `onlyOwner` (TransferrableOwnership) still reverts `UnAuthorized` and is unchanged.

Built task-by-task with TDD; 58 factory unit tests cover deploy wiring, namespace/deployer authorization (incl. squatting and zero-namespace), per-vault admin, the integrator-share override + validity (incl. the self-serve floor that bounds a self-serve deployer to the factory default so it cannot cut LI.FI's share), the `lifiFeeRecipient` config, allowlist + adapter approval/asset-resolution, fee bounds/caps, CREATE2 determinism (incl. per-adapter), global pause, role rotation, and instance tracking. Factory coverage is 100% lines / 100% functions / ~96% branches.

**Scope boundaries / follow-ups:**
- The deploy script is a functional `Script` bootstrap (now also deploys `ERC4626Adapter`); full `DeployScriptBase` / CREATE3 / deployment-log integration is **S14 (EXSC-420)**.
- The script transfers beacon ownership to the governance `owner`; repointing beacon ownership to the real 48h timelock is **S9/S10**. Governance must `setAdapterApproved` + `setUnderlyingAllowed` post-deploy before any wrapper can be deployed.
- Depends on **S1 (EXSC-409)** and **S9 (EXSC-416)** — the `MockVaultWrapper` is a temporary stand-in until S1's implementation is wired behind the beacon.

**S8 review — spec deviations (recorded in Linear):** review findings are addressed in commit `2b12e4ab` (self-serve fee-split bound + NatSpec/custody/version cleanups). Where the code deliberately diverges from ticket scope / product stories, the deltas are logged rather than silently dropped:
- **[EXSC-417 — deviations from ticket scope](https://linear.app/lifi-linear/issue/EXSC-417/s8-factory#comment-8121a0fe-c1b9-4281-b90a-bf6306d8fb4c):** on-chain cooldown rejection is off-chain vetting by design (Q6.6); the single-chain-lock / `chainLockId` stories (A17/I2/A4) add nothing on-chain and need PO refinement; the self-serve fee-split floor (`default = floor`) needs a one-line PO confirm.
- **[EXSC-420 — S14 carry-over](https://linear.app/lifi-linear/issue/EXSC-420/s14-deployment-and-configuration-scripts#comment-089ed250-57d8-4273-8761-9a9bfc4136af):** the real deploy must route factory + beacon through CREATE3 for cross-chain address parity, and add the factory's `deployRequirements.json` entry when it joins the deploy framework.

- [x] I have run `/pr-ready` (local CodeRabbit) on this branch and resolved (or explicitly documented) all findings — see `.agents/commands/pr-ready.md`. A high-effort multi-angle code review was also run; actionable findings (LibAsset.isContract, deduped errors, `_cap` array, `lifiFeeRecipient`) are fixed. Reviewer (mirooon) findings are addressed in `2b12e4ab` with replies on each thread. Local `forge test` (78/78 VaultWrapper) + `solhint` (0 errors) pass; cloud CodeRabbit runs in CI as the safety net.


gvladika:
## Addressed review findings (pushed in `f7ee0f2d`, `45c2fe6a`, `82debc07`)

**CodeRabbit — all resolved:**
- 🔴 Critical — `_owner` now validated `!= address(0)` in the constructor.
- 🟠 Major — deploy script validates `OWNER`/`EMERGENCY_PAUSER`/`ONBOARDING_MANAGER` are non-zero (custom errors, per repo `gas-custom-errors`).
- 🟠 Major — removed redundant `isGlobalPaused()`; clones read the auto-generated `globalPaused()` getter.
- Nitpicks — `WrapperDeployed` now emitted before `initialize()` (CEI); added paginated `getInstances(offset, limit)` + test; NatSpec on mock getters; test state vars moved to top.

Factory suite: 30/30 passing. Local CodeRabbit re-run: no findings.

**Olympix — triaged:**
- `owner-single-point-of-failure` (×5 setters): dismissed (won't-fix) — `owner` is the governance 48h Timelock + multisig, not an EOA; intended trust boundary with timelock-gated rotation (A19).
- `uninitialized-state-variable` (`allInstances`, mock fields ×4): dismissed (false positive) — array is intentionally empty-then-appended; mock fields set once in `initialize()`.
- `reentrancy-events` + `no-parameter-validation-in-constructor`: not dismissed — fixed in code (CEI reorder + `_owner` check), will clear on rescan.

src/VaultWrapper/LiFiVaultWrapperFactory.sol
```diff
@@ -0,0 +1,346 @@
...
+        address _underlying,
+        bool _allowed
+    ) external onlyOwner {
+        if (_underlying == address(0)) revert ZeroAddress();
+        allowedUnderlying[_underlying] = _allowed;
+        emit UnderlyingAllowedSet(_underlying, _allowed);
+    }
+
+    /// @notice Approve or revoke a yield adapter usable in deployments.
+    function setAdapterApproved(
```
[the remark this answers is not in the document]
gvladika: By design: this setter is governance-gated (`onlyOwner`, where owner is the 48h `TimelockController` per `[CONV:ARCH-GOVERNANCE]`). Owner-gating of privileged config is a required security control in this subsystem, not a vulnerability — removing it would violate the architecture rule against admin shortcuts. Resolving as won't-fix.

src/VaultWrapper/LiFiVaultWrapperFactory.sol
```diff
@@ -0,0 +1,346 @@
...
+        FeeType _feeType,
+        uint16 _lifiBps
+    ) external onlyOwner {
+        if (_lifiBps > BPS_DENOMINATOR) revert InvalidSplit();
+        defaultLifiShareBps[_feeType] = _lifiBps;
+        emit DefaultSplitSet(_feeType, _lifiBps);
+    }
+
+    /// @notice Rotate the emergency pauser role.
+    function setEmergencyPauser(address _newPauser) external onlyOwner {
```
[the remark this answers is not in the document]
gvladika: By design: this setter is governance-gated (`onlyOwner`, where owner is the 48h `TimelockController` per `[CONV:ARCH-GOVERNANCE]`). Owner-gating of privileged config is a required security control in this subsystem, not a vulnerability — removing it would violate the architecture rule against admin shortcuts. Resolving as won't-fix.

src/VaultWrapper/LiFiVaultWrapperFactory.sol
```diff
@@ -0,0 +1,353 @@
...
+        if (_minBps > _maxBps || _maxBps > _cap(_feeType))
+            revert InvalidFeeBounds();
+        feeBounds[_feeType] = FeeBounds(_minBps, _maxBps);
+        emit FeeBoundsSet(_feeType, _minBps, _maxBps);
+    }
+
+    /// @notice Set the default integrator share (bps) of the underlying-generated fees;
+    ///         LI.FI implicitly receives the remaining (100% - this value). Read by vault
+    ///         wrappers when applying the split (see S1).
+    function setDefaultSplit(uint16 _integratorBps) external onlyOwner {
```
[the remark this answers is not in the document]
gvladika: By design: this setter is governance-gated (`onlyOwner`, where owner is the 48h `TimelockController` per `[CONV:ARCH-GOVERNANCE]`). Owner-gating of privileged config is a required security control in this subsystem, not a vulnerability — removing it would violate the architecture rule against admin shortcuts. Resolving as won't-fix.

src/VaultWrapper/mocks/MockVaultWrapper.sol
```diff
@@ -0,0 +1,56 @@
...
+/// @title MockVaultWrapper
+/// @author LI.FI (https://li.fi)
+/// @notice Temporary beacon implementation used until S1 (CWIA core) lands.
+///         Records init args so factory tests can assert vault wrapper wiring.
+/// @custom:version 1.0.0
+contract MockVaultWrapper is ILiFiVaultWrapper {
+    error AlreadyInitialized();
+
+    bool public initialized;
+    address public asset;
```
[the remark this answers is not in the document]
gvladika: `MockVaultWrapper` is a temporary beacon implementation used only until S1 (CWIA core) lands; it records its init args, which are set write-once in `initialize()`. It is not production code and is removed when the real wrapper is wired behind the beacon. Resolving as won't-fix.

src/VaultWrapper/mocks/MockVaultWrapper.sol
```diff
@@ -0,0 +1,56 @@
...
+    address public underlying;
```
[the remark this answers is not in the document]
gvladika: `MockVaultWrapper` is a temporary beacon implementation used only until S1 (CWIA core) lands; it records its init args, which are set write-once in `initialize()`. It is not production code and is removed when the real wrapper is wired behind the beacon. Resolving as won't-fix.

src/VaultWrapper/mocks/MockVaultWrapper.sol
```diff
@@ -0,0 +1,56 @@
...
+    address public adapter;
```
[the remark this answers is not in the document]
gvladika: `MockVaultWrapper` is a temporary beacon implementation used only until S1 (CWIA core) lands; it records its init args, which are set write-once in `initialize()`. It is not production code and is removed when the real wrapper is wired behind the beacon. Resolving as won't-fix.

src/VaultWrapper/mocks/MockVaultWrapper.sol
```diff
@@ -0,0 +1,56 @@
...
+    address public integrator;
```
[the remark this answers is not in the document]
gvladika: `MockVaultWrapper` is a temporary beacon implementation used only until S1 (CWIA core) lands; it records its init args, which are set write-once in `initialize()`. It is not production code and is removed when the real wrapper is wired behind the beacon. Resolving as won't-fix.

script/deploy/vaultWrapper/DeployLiFiVaultWrapperFactory.s.sol
```diff
@@ -0,0 +1,60 @@
...
+    function run()
+        public
+        returns (
+            LiFiVaultWrapperFactory factory,
+            UpgradeableBeacon beacon,
+            MockVaultWrapper impl,
+            ERC4626Adapter erc4626Adapter
+        )
+    {
+        uint256 deployerPrivateKey = uint256(vm.envBytes32("PRIVATE_KEY"));
```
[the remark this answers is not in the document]
gvladika: Addressed in 98fcbe72 — the deploy script now reverts `ZeroPrivateKey` before `vm.startBroadcast`. Resolving.

gvladika:
@coderabbitai review

src/VaultWrapper/LiFiVaultWrapperFactory.sol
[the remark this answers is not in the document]
gvladika: False positive / by design: this is governance-gated config (`onlyOwner`, where owner is the 48h `TimelockController` per `[CONV:ARCH-GOVERNANCE]`). Owner control of privileged config is a required security control in this subsystem, not a vulnerability — the architecture rule forbids ungoverned admin shortcuts. Resolving as won't-fix.

src/VaultWrapper/mocks/MockVaultWrapper.sol
```diff
@@ -0,0 +1,59 @@
...
+/// @author LI.FI (https://li.fi)
+/// @notice Temporary beacon implementation used until S1 (CWIA core) lands.
+///         Records init args so factory tests can assert vault wrapper wiring.
+/// @custom:version 1.0.0
+contract MockVaultWrapper is ILiFiVaultWrapper {
+    error AlreadyInitialized();
+
+    bool public initialized;
+    address public asset;
+    address public underlying;
```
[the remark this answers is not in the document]
gvladika: False positive: `MockVaultWrapper` is a temporary beacon stand-in (removed when S1 lands). These vars are set write-once in `initialize()`, which the factory calls atomically in the same tx as the proxy deploy — they are never read uninitialized. Resolving as won't-fix.

src/VaultWrapper/mocks/MockVaultWrapper.sol
```diff
@@ -0,0 +1,59 @@
...
+    address public adapter;
```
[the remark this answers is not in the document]
gvladika: False positive: `MockVaultWrapper` is a temporary beacon stand-in (removed when S1 lands). These vars are set write-once in `initialize()`, which the factory calls atomically in the same tx as the proxy deploy — they are never read uninitialized. Resolving as won't-fix.

src/VaultWrapper/mocks/MockVaultWrapper.sol
```diff
@@ -0,0 +1,59 @@
...
+    address public vaultWrapperAdmin;
```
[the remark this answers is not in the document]
gvladika: False positive: `MockVaultWrapper` is a temporary beacon stand-in (removed when S1 lands). These vars are set write-once in `initialize()`, which the factory calls atomically in the same tx as the proxy deploy — they are never read uninitialized. Resolving as won't-fix.

src/VaultWrapper/LiFiVaultWrapperFactory.sol
```diff
@@ -0,0 +1,362 @@
...
+    }
+
+    /// Deploy ///
+
+    /// @notice Deploy a new wrapper instance under an integrator namespace.
+    /// @dev Caller must be the onboarding manager, or the deployer assigned to
+    ///      `_params.namespace`.
+    /// @param _params The deployment parameters.
+    /// @return instance The address of the newly deployed vault wrapper.
+    function deploy(
```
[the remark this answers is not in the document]
gvladika: False positive: `deploy()` has no reentrancy vector — its only external calls are a `view` adapter staticcall, the CREATE2 proxy creation (no user callback), and `initialize()` on the freshly-created clone, with `isInstance` written before init. Emitting `WrapperDeployed` ahead of the trusted `initialize()` causes no out-of-order-event or reentrancy risk. Confirmed by security review. Resolving as won't-fix.

src/VaultWrapper/mocks/MockVaultWrapper.sol
```diff
@@ -0,0 +1,77 @@
...
+/// @notice Temporary beacon implementation used until S1 (CWIA core) lands. Built
+///         on Solady's ERC-4626 so the stand-in behind the beacon is a real vault
+///         share token (constructor-free, so it works through the beacon proxy),
+///         and records its init args so factory tests can assert clone wiring.
+/// @custom:version 1.0.0
+contract MockVaultWrapper is ERC4626, ILiFiVaultWrapper {
+    error AlreadyInitialized();
+
+    bool public initialized;
+    address internal _vaultAsset;
```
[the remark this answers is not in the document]
gvladika: False positive: `MockVaultWrapper` is a temporary beacon stand-in (removed when S1 lands). These vars are set write-once in `initialize()`, which the factory calls atomically in the same tx as the proxy deploy — they are never read uninitialized. Resolving as won't-fix.

src/VaultWrapper/LiFiVaultWrapperFactory.sol
```diff
@@ -0,0 +1,379 @@
...
+    /// @notice Set the default integrator fee share (bps) applied to deploys that don't
+    ///         override it; LI.FI implicitly receives the remaining (100% - this value).
+    function setDefaultSplit(uint16 _integratorBps) external onlyOwner {
+        if (_integratorBps > BPS_DENOMINATOR) revert InvalidSplit();
+        defaultIntegratorShareBps = _integratorBps;
+        emit DefaultSplitSet(_integratorBps);
+    }
+
+    /// @notice Set the recipient of LI.FI's fee share, read live by vault wrappers.
+    function setLifiFeeRecipient(address _recipient) external onlyOwner {
```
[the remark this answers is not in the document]
gvladika: False positive / by design: this is governance-gated config (`onlyOwner`, where owner is the 48h `TimelockController` per `[CONV:ARCH-GOVERNANCE]`). Owner control of privileged config is a required security control in this subsystem, not a vulnerability — the architecture rule forbids ungoverned admin shortcuts. Resolving as won't-fix.

test/solidity/VaultWrapper/LiFiVaultWrapperFactory.t.sol
```diff
@@ -0,0 +1,615 @@
...
+    function test_ConstructorRevertsOnZeroBeacon() public {
+        vm.expectRevert();
+        new LiFiVaultWrapperFactory(
+            address(0),
+            owner,
+            pauser,
+            onboarder,
+            lifiRecipient
+        );
+    }
```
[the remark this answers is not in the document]
gvladika: Addressed in e616493d — the zero-beacon test now asserts `ILiFiVaultWrapperFactory.ZeroAddress.selector` instead of a bare `vm.expectRevert()`.

test/solidity/VaultWrapper/mocks/MockZeroAdapter.sol
```diff
@@ -0,0 +1,12 @@
+// SPDX-License-Identifier: LGPL-3.0-only
+pragma solidity ^0.8.17;
+
+import { IYieldAdapter } from "lifi/VaultWrapper/interfaces/IYieldAdapter.sol";
+
+/// @notice Adapter that resolves to the zero address without reverting, used to
+///         exercise the factory's zero-asset guard in _resolveAssetViaAdapter.
+contract MockZeroAdapter is IYieldAdapter {
+    function resolveAsset(address) external pure returns (address) {
```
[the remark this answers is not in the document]
gvladika: False positive. Solidity has not required `override` for single-interface implementations since 0.8.8 (this repo compiles with solc 0.8.29); `MockZeroAdapter` builds cleanly and all 59 VaultWrapper tests pass. The production `ERC4626Adapter.resolveAsset` intentionally uses the same no-`override` style, so adding it here would diverge from the established pattern. No compile error exists. Resolving.

src/VaultWrapper/LiFiVaultWrapperTypes.sol
```diff
@@ -0,0 +1,43 @@
...
+/// @notice Inputs for a single `deploy` call.
+struct DeployParams {
+    bytes32 namespace; // Integrator identity seeding the salt (e.g. "Coinbase"); must be assigned to the caller.
+    address vaultWrapperAdmin; // Per-vault controller granted the instance admin role.
+    address adapter; // Approved yield adapter; resolves the underlying's ERC20 asset.
+    address underlying; // Protocol-specific yield source (e.g. an ERC-4626 vault).
+    uint256 nonce; // Disambiguates instances sharing the same (namespace, adapter, underlying).
+    FeeConfig fees; // Per-fee-type rates and enabled flags; validated against bounds/caps.
+    uint16 integratorShareBps; // Integrator's fee share (bps); type(uint16).max = factory default, else must be <= 100%.
+    bytes initData; // Opaque wrapper-side config forwarded to the instance's initialize.
```
mirooon, line 44: single-chain lock / `chainLockId` (DeployParams has no chainLockId field)

Quick note on the vault wrapper factory and the "single-chain lock" idea.

What the stories ask for:
- A17 + I2: a built-in lock so a vault can only work on ONE chain, and you can prove it just by looking at the code. This is meant for regulators.
- A4: put the underlying and a "chainLockId" into the contract code at deploy time, so nobody can change them later.

The worry in the review: the factory has no way to set this chainLockId, so we can't give that single-chain proof.

Why I don't think it's a real problem for this PR:
- A deployed contract already lives only on the chain you deploy it to. So it is already "locked" to that chain, just by existing there. And nothing moves between chains here — money stays on the chain you put it on. So there's no extra lock to add.
- There are no signatures anywhere in the code. The usual reason to tie something to a chain is to stop a signature being reused on another chain — but we have no signatures, so there's nothing to reuse.
- We already decide where it gets deployed. Only the right owner can deploy, and only approved adapters/underlyings are allowed.

Suggestion: this isn't really a code fix — it's the user stories that don't match what we actually need to build. So we should refine or delete A17/I2/A4 and let the product owner know. The "chainLockId" mechanism they describe doesn't add anything, because deploying to a chain already keeps the contract on that chain.

Proposed story changes (BEFORE / AFTER):

A17
As a vault owner I want an opt-in, bytecode-enforced single-chain lock, so regulators get structural proof the instance only runs on one chain.
The thing is that vault instance only exists on the chain it's deployed to — that's already true by default, no special lock is needed. 
imo we should remove this open question. Or am I getting this wrong?

I2
BEFORE: As an integrator I want to declare at deploy that my instance is single-chain.
No declaration needed — a deployed instance is single-chain by nature. Delete unless tied to the open question below.
imo we should remove this open question. Or am I getting this wrong?

A4
BEFORE: Bake the underlying and a chainLockId into the clone bytecode so they can't be changed after deploy.
AFTER: Bake the underlying into the instance at deploy so it can't be changed afterwards. (Drop the chainLockId part — it has no purpose here.) 
wdyt?

src/VaultWrapper/LiFiVaultWrapperFactory.sol
```diff
@@ -0,0 +1,379 @@
...
+            _params.underlying
+        );
+
+        _validateFees(_params.fees);
+
+        uint16 integratorShareBps = _params.integratorShareBps;
+        if (integratorShareBps == USE_DEFAULT_SPLIT) {
+            integratorShareBps = defaultIntegratorShareBps;
+        } else if (integratorShareBps > BPS_DENOMINATOR) {
+            revert InvalidSplit();
```
mirooon, line 272: fee split should be bounded for self-serve deployers

Quick note on the vault wrapper factory fee split (integratorShareBps decides how the collected fee is shared between integrator and LI.FI - default 20% to LI.FI).

The spec says this split is LI.FI's to set, only via the 48h timelock (A9, A22, Q1.13), but in deploy() the integrator can freely pick any value up to 100% - so an approved deployer can set 9000 and quietly cut LI.FI to 10%, or 10000 and cut us to 0%.

Question to confirm: a self-serve integrator should NOT be able to set LI.FI's share below the default (going above, i.e. giving LI.FI more, is fine) - correct?

If yes, the fix is one line: when the caller isn't the onboarding manager, revert if integratorShareBps > defaultIntegratorShareBps.

Worth doing now, because self-serve is already live (just a setApprovedIntegratorDeployer call away) and the split is immutable after deploy, so any bad split is permanent.

src/VaultWrapper/LiFiVaultWrapperFactory.sol
```diff
@@ -0,0 +1,379 @@
...
+
+    /// @notice Deploy a new wrapper instance under an integrator namespace.
+    /// @dev Caller must be the onboarding manager, or the deployer assigned to
+    ///      `_params.namespace`.
+    /// @param _params The deployment parameters.
+    /// @return instance The address of the newly deployed vault wrapper.
+    function deploy(
+        DeployParams calldata _params
+    ) external returns (address instance) {
+        if (_params.namespace == bytes32(0)) revert ZeroNamespace();
```
mirooon, line 251: deploy() does not check globalPaused

deploy() doesn't check globalPaused, so new proxies can be created during an emergency. They are not a deposit risk — being beacon proxies, they read the same live globalPaused flag and are frozen from birth. The only cost is operational noise (events, salt consumption, onboarding during an incident). Either add if (globalPaused) revert GloballyPaused(); as cheap hygiene, or document that deploys are intentionally allowed because new instances inherit the pause anyway.

src/VaultWrapper/LiFiVaultWrapperFactory.sol
```diff
@@ -0,0 +1,379 @@
...
+
+    /// Config (owner / timelock) ///
+
+    /// @notice Add or remove a yield source from the deploy allowlist.
+    function setUnderlyingAllowed(
+        address _underlying,
+        bool _allowed
+    ) external onlyOwner {
+        if (_underlying == address(0)) revert ZeroAddress();
+        allowedUnderlying[_underlying] = _allowed;
```
mirooon, line 136: The ticket promises "reject cooldown vaults," but the code doesn't do that (on purpose)

The problem

The original ticket's acceptance list says the allowlist should "reject vaults with withdrawal cooldowns or lock periods." That sounds like the contract is supposed to look at a vault and automatically refuse it if it has a cooldown.

But the code doesn't work that way — and that's intentional. A later product decision (Q6.6) changed the plan: a human checks each vault before it's added. There is no on-chain detection of cooldowns. The allowlist is just a simple yes/no list that the owner edits by hand:

function setUnderlyingAllowed(address _underlying, bool _allowed) external onlyOwner {
    if (_underlying == address(0)) revert ZeroAddress();
    allowedUnderlying[_underlying] = _allowed;   // owner flips a flag — no cooldown inspection
    emit UnderlyingAllowedSet(_underlying, _allowed);
}

The contract trusts whatever the owner puts on the list. It never reads the vault to see if it has a cooldown. So the code matches the real decision (manual vetting), but not the old wording of the ticket.

Example

An auditor reads "cooldown rejection" in the ticket, then searches the contract for any logic that detects a cooldown or lock period. They find nothing. They conclude a required safety feature is missing — and possibly flag it as a bug. In reality nothing is missing: the check was deliberately moved off-chain to human review. The auditor just wasted time chasing a ghost, because the ticket and the code tell different stories.

How to fix

No code change. Just align the words. Update the ticket / PR description to say something like:

  "Vaults with withdrawal cooldowns or lock periods are excluded by off-chain vetting (Q6.6) before being added to the allowlist — there is no on-chain cooldown check by design."

That one sentence stops the confusion during audit.

Impact

Documentation clarity only. Not blocking.

src/VaultWrapper/LiFiVaultWrapperTypes.sol
```diff
@@ -0,0 +1,43 @@
+// SPDX-License-Identifier: LGPL-3.0-only
+pragma solidity ^0.8.17;
+
+/// @custom:version 1.0.0
+/// @notice Shared value types for the LI.FI vault wrapper subsystem.
```
mirooon: Small convention and documentation gaps

The problem
A few repo conventions are not fully met:
- LiFiVaultWrapperTypes.sol is missing the @title and @author NatSpec tags.
- NatSpec tag order is not always consistent. The repo convention is @title -> @author -> @notice -> @custom:version, with @custom:version as the last tag sitting directly above the contract/interface declaration. ILiFiVaultWrapper.sol places @custom:version before @notice, so it should be reordered to match the factory and the rest of the repo.
- The factory and adapter don't state whether they hold user funds (rule 101 asks every contract to say this — here the answer is "no funds held," it just needs to be written).
- No entry was added to deployRequirements.json for the factory's constructor arguments.

script/deploy/vaultWrapper/DeployLiFiVaultWrapperFactory.s.sol
```diff
@@ -0,0 +1,91 @@
...
+            MIN_DELAY,
+            proposers,
+            executors,
+            address(0)
+        );
+
+        impl = new MockVaultWrapper();
+        beacon = new UpgradeableBeacon(address(impl));
+        beacon.transferOwnership(address(timelock));
+        factory = new LiFiVaultWrapperFactory(
```
mirooon, line 86: "Same address on every chain" only works if the factory and beacon are deployed deterministically

The problem
Where the promise is stated: this is not from a user story or product document — the contract's own NatSpec asserts it. LiFiVaultWrapperFactory.sol:313-315 (the _salt helper) says: "Identical inputs yield the same address on every chain; the namespace is chain-independent (e.g. 'Coinbase'), so address parity holds even when the integrator uses different deployers/admins per chain." It is reinforced at line 15 ("deterministic beacon proxies") and line 295 (predictAddress — "The deterministic address a vault wrapper will have for the given key").

That NatSpec makes the cross-chain guarantee unconditional, but it omits the real precondition: address parity only holds if the factory (the CREATE2 deployer, address(this) in predictAddress) and the beacon (baked into the creation code) are themselves at identical addresses on every chain. So this is a doc-vs-reality mismatch inside the same contract — the code documents a cross-chain guarantee the current deployment method does not back up.

Vault instances are already deployed deterministically with CREATE2 (LibClone.deployDeterministicERC1967BeaconProxy at LiFiVaultWrapperFactory.sol:263), and the beacon address is baked into the proxy's creation code. A CREATE2 address depends on (deployer, salt, creationCode), so an instance only lands on the same address across chains if BOTH the factory (the CREATE2 deployer) and the beacon (part of the creation code) sit at identical addresses on every chain.

The current bootstrap script does not guarantee that. DeployLiFiVaultWrapperFactory.s.sol uses plain `new` (CREATE) for the implementation, the beacon, and the factory (lines 77-80). CREATE addresses depend on the deployer's nonce on each chain, so the factory and beacon can land on different addresses on Ethereum vs. Polygon. That breaks the "same vault address everywhere" promise, which also breaks the single-chain-lock logic (Finding 1).

This is known and intentional for the bootstrap script — its own NatSpec (line 28) states: "Full DeployScriptBase / CREATE3 / deployment-log integration is S14 (EXSC-420)." Nothing production-facing is broken yet, because this script is a stand-in that also deploys MockVaultWrapper rather than the real vault.

The requirement
The real deployment (ticket S14 / EXSC-420) must deploy the factory and the beacon with a deterministic method so they land on identical addresses on all chains. The repo already has the prior art for this:
- CREATE3 is the repo-wide standard: every deployable contract goes through CREATE3Factory via DeployScriptBase (script/deploy/facets/utils/DeployScriptBase.sol). ERC20Proxy even pre-authorizes the Executor at its predicted CREATE3 address, which only works because CREATE3 gives identical, nonce-independent addresses across chains.
- CREATE2 is already used for the vault instances themselves.

The solution
Route the factory + beacon deploy through the existing CREATE3Factory path (the same one DeployScriptBase / every other deployable contract uses) in S14, instead of plain `new`. That makes the factory and beacon addresses converge across chains, which in turn makes the CREATE2 instance addresses converge — restoring the "same vault address everywhere" guarantee and the chain-lock logic in Finding 1. No new mechanism is needed; reuse the repo standard. For now, just record this CREATE3 requirement clearly for S14.

Either way, also fix the wording in the _salt NatSpec (LiFiVaultWrapperFactory.sol:313) so it stops promising more than it delivers — e.g. add "address parity holds only when the factory and beacon are deployed at identical addresses on each chain (deterministic deploy, S14)." That removes the doc-vs-reality mismatch regardless of when S14 lands.

Impact
Informational. The current script is an intentional bootstrap stand-in; the deterministic-deployment work is correctly deferred to S14. The action is to ensure S14 uses CREATE3 for the factory and beacon.

src/VaultWrapper/LiFiVaultWrapperFactory.sol
```diff
@@ -0,0 +1,379 @@
...
+        emergencyPauser = _emergencyPauser;
+        onboardingManager = _onboardingManager;
+        lifiFeeRecipient = _lifiFeeRecipient;
+        defaultIntegratorShareBps = DEFAULT_INTEGRATOR_SHARE_BPS;
+    }
+
+    /// Config (owner / timelock) ///
+
+    /// @notice Add or remove a yield source from the deploy allowlist.
+    function setUnderlyingAllowed(
```
mirooon, line 131: Eight external functions are missing `@param`/`@return`

[CONV:NATSPEC] requires public/external functions to document params/returns. These have only a one-line `@notice`:

| Function | line | Missing |
|---|---|---|
| `setUnderlyingAllowed` | 127 | both `@param` |
| `setAdapterApproved` | 137 | both `@param` |
| `setFeeBounds` | 149 | 3× `@param` |
| `setDefaultSplit` | 162 | `@param` |
| `setLifiFeeRecipient` | 169 | `@param` |
| `setEmergencyPauser` | 176 | `@param` |
| `setOnboardingManager` | 184 | `@param` |
| `predictAddress` | 296 | 4× `@param` **and** `@return` |

Odd inconsistency: the **internal** helpers (`_salt`, `_cap`, `_resolveAssetViaAdapter`, `_validateFees`) are fully documented, while the **external, ABI-facing** setters are not — it's backwards. It slipped through because `.solhint.json` enables no NatSpec rule, so CI never flagged it.

Example fix:
```solidity
/// @notice Add or remove a yield source from the deploy allowlist.
/// @param _underlying The yield source (e.g. an ERC-4626 vault) to toggle.
/// @param _allowed    True to allow as a wrapper underlying, false to remove.
function setUnderlyingAllowed(address _underlying, bool _allowed) external onlyOwner {
```

Impact: convention violation across the most security-relevant functions; harder to audit the privileged surface. Cheap to fix — should fix before audit.

src/VaultWrapper/LiFiVaultWrapperFactory.sol
```diff
@@ -0,0 +1,379 @@
...
+import { InvalidContract } from "../Errors/GenericErrors.sol";
+import { FeeType, FeeBounds, FeeConfig, DeployParams } from "./LiFiVaultWrapperTypes.sol";
+import { IYieldAdapter } from "./interfaces/IYieldAdapter.sol";
+import { ILiFiVaultWrapper } from "./interfaces/ILiFiVaultWrapper.sol";
+import { ILiFiVaultWrapperFactory } from "./interfaces/ILiFiVaultWrapperFactory.sol";
+import { LibClone } from "solady/utils/LibClone.sol";
+
+/// @title LiFiVaultWrapperFactory
+/// @author LI.FI (https://li.fi)
+/// @notice Deploys per-integrator vault wrapper instances as deterministic beacon
```
mirooon, line 15: Missing "does/doesn't custody funds" statement 

Neither `LiFiVaultWrapperFactory` (this `@notice`) nor `ERC4626Adapter` (`adapters/ERC4626Adapter.sol:10`) states whether it holds funds. Neither does. Rule 101 asks every `src/` contract to say so for bug-bounty clarity.

Fix — add one line to each `@notice`, e.g.:
> "This contract does not custody user funds; it only deploys and configures wrapper instances."

Impact: low; documentation / bounty hygiene.

src/VaultWrapper/LiFiVaultWrapperFactory.sol
```diff
@@ -0,0 +1,379 @@
...
+        _;
+    }
+
+    /// Constructor ///
+
+    /// @notice Initializes the factory with a beacon and role addresses.
+    /// @param _beacon        Address of the UpgradeableBeacon holding the wrapper implementation.
+    /// @param _owner         Address that will own the factory.
+    /// @param _emergencyPauser Address authorized to trigger global pause.
+    /// @param _onboardingManager Address authorized to manage integrator allowlist.
```
mirooon: Constructor `@param _onboardingManager` is misleading

> `/// @param _onboardingManager Address authorized to manage integrator allowlist.`

The onboarding manager does **not** manage any allowlist. It assigns/revokes the **deployer per namespace** (`setApprovedIntegratorDeployer`). The underlying allowlist is managed by the owner (timelock) via `setUnderlyingAllowed`. Calling it "integrator allowlist" blurs two different things.

Fix:
> "Address authorized to assign/revoke the deployer for each integrator namespace."

Impact: a reader could wrongly think the onboarding manager controls which vaults can be wrapped.

src/VaultWrapper/LiFiVaultWrapperTypes.sol
```diff
@@ -0,0 +1,43 @@
+// SPDX-License-Identifier: LGPL-3.0-only
+pragma solidity ^0.8.17;
+
+/// @custom:version 1.0.0
```
mirooon: `LiFiVaultWrapperTypes.sol` has no `@title`/`@author`

Has `@notice`/`@dev`/`@custom:version` but not `@title`/`@author`. Gray area (file-level structs/enums, no contract/interface to title), so solhint won't catch it, but it deviates from the "every file" spirit of [CONV:NATSPEC].

Fix — add to the file-level block:
```solidity
/// @title LiFiVaultWrapperTypes
```
Impact: cosmetic.

src/VaultWrapper/mocks/MockVaultWrapper.sol
```diff
@@ -0,0 +1,77 @@
...
+
+/// @title MockVaultWrapper
+/// @author LI.FI (https://li.fi)
+/// @notice Temporary beacon implementation used until S1 (CWIA core) lands. Built
+///         on Solady's ERC-4626 so the stand-in behind the beacon is a real vault
+///         share token (constructor-free, so it works through the beacon proxy),
+///         and records its init args so factory tests can assert clone wiring.
+/// @custom:version 1.0.0
+contract MockVaultWrapper is ERC4626, ILiFiVaultWrapper {
+    error AlreadyInitialized();
```
mirooon, line 17: `AlreadyInitialized` error and script `run()` lack NatSpec

Acceptable under the "test/script scaffolding = minimal headers" rule, but worth a `@notice` on this error and a short `@notice`/`@return` on the deploy script's `run()` entry point (`DeployLiFiVaultWrapperFactory.s.sol:40`).

Impact: none / nice-to-have.

src/VaultWrapper/LiFiVaultWrapperFactory.sol
```diff
@@ -0,0 +1,379 @@
...
+
+    /// @notice Address authorized to toggle the global circuit breaker.
+    address public emergencyPauser;
+    /// @notice Address authorized to approve and revoke integrators.
+    address public onboardingManager;
+    /// @notice Recipient of LI.FI's fee share; read live by every vault wrapper at
+    ///         distribution, so an integrator cannot redirect LI.FI's cut.
+    address public lifiFeeRecipient;
+    /// @notice Whether deposits are globally halted; read by every vault wrapper.
+    bool public globalPaused;
```
mirooon, line 64: Forward-looking descriptions (flagging, not an error)

Several factory comments describe **clone-side** behavior this PR doesn't exercise yet — e.g. `globalPaused` "read by every vault wrapper" (here) and `lifiFeeRecipient` "read live by every vault wrapper at distribution" (line 60). The clones (S1) don't exist yet and `MockVaultWrapper` reads neither value. The text is accurate about the intended system role, so it's fine — just be aware these document future S1/S5 behavior, not behavior provable in this PR.

script/deploy/vaultWrapper/DeployLiFiVaultWrapperFactory.s.sol
```diff
@@ -0,0 +1,91 @@
...
+///      beacon ownership transferred to the timelock.
+///      Timelock roles: the LI.FI multisig is proposer AND canceller (OZ grants both to
+///      each proposer); the executor role is open (address(0)); the optional admin is
+///      renounced (address(0)), so the timelock is self-administered.
+///      Reads MULTISIG, EMERGENCY_PAUSER, ONBOARDING_MANAGER, and LIFI_FEE_RECIPIENT from
+///      environment. Because the factory owner is the 48h timelock, post-deploy governance
+///      actions — setAdapterApproved(adapter) and setUnderlyingAllowed(underlying), required
+///      before any wrapper can be deployed — must be scheduled through the timelock.
+///      Full DeployScriptBase / CREATE3 / deployment-log integration is S14 (EXSC-420).
+/// @custom:version 1.1.0
```
mirooon: Script DeployLiFiVaultWrapperFactory.s.sol is `@custom:version 1.1.0`; all contracts are `1.0.0`.

mirooon (commented):
left comments below. Great job one this!

src/VaultWrapper/LiFiVaultWrapperTypes.sol
gvladika: Agreed — and this is a product-story question, not a code change. A deployed instance already exists only on its chain, there are no signatures anywhere to replay across chains, and deploy authorization + the adapter/underlying allowlists already gate where instances land, so a bytecode `chainLockId` adds nothing here. I've recorded the deviation and the recommendation to refine/drop A17/I2 and drop the `chainLockId` half of A4 (keeping "bake the underlying into the instance") in Linear for the PO: https://linear.app/lifi-linear/issue/EXSC-417/s8-factory#comment-8121a0fe-c1b9-4281-b90a-bf6306d8fb4c (item 2). No code change in this PR.

src/VaultWrapper/LiFiVaultWrapperFactory.sol
gvladika: Fixed in `2b12e4ab`. `deploy()` now reverts `IntegratorShareAboveDefault` when the caller is **not** the onboarding manager and `integratorShareBps > defaultIntegratorShareBps` — so a self-serve deployer can give LI.FI more than the default cut but never less; the onboarding manager keeps the full 0–100% range; the default itself stays owner/timelock-set via `setDefaultSplit`. Added tests for both the self-serve revert and the onboarding-manager-above-default path. One open confirm logged for the PO (is `default = floor` the intended rule, or a separately-governed minimum?): https://linear.app/lifi-linear/issue/EXSC-417/s8-factory#comment-8121a0fe-c1b9-4281-b90a-bf6306d8fb4c (item 3).

src/VaultWrapper/LiFiVaultWrapperFactory.sol
gvladika: Went with documentation rather than a redundant check. Added a `@dev` note on `deploy()` (`2b12e4ab`): deploys are intentionally allowed while `globalPaused` because a new instance is a beacon proxy that reads the same live flag and is frozen from birth — no deposit risk. Adding a `globalPaused` read + new error here would be extra surface for zero security gain.

src/VaultWrapper/LiFiVaultWrapperFactory.sol
gvladika: Confirmed — no code change, by design. There is no on-chain cooldown/lock detection; per Q6.6 the vetting is manual off-chain before a vault is allowlisted. Logged the AC-wording fix in Linear (https://linear.app/lifi-linear/issue/EXSC-417/s8-factory#comment-8121a0fe-c1b9-4281-b90a-bf6306d8fb4c, item 1) and added it to the PR body so the ticket and code stop telling different stories.

src/VaultWrapper/LiFiVaultWrapperTypes.sol
gvladika: Mostly fixed in `2b12e4ab`: added `@title`/`@author` to `LiFiVaultWrapperTypes`, fixed the `@custom:version` tag order in `ILiFiVaultWrapper`, and added the "does not custody user funds" line to the factory and `ERC4626Adapter`. The `deployRequirements.json` entry is **deliberately deferred**: the deploy script is standalone and intentionally doesn't use `DeployScriptBase`/the Diamond deploy framework (agent rule `108-vault-wrapper`), so the entry would be unused config until S14 folds it into the framework. Tracked in https://linear.app/lifi-linear/issue/EXSC-420/s14-deployment-and-configuration-scripts#comment-089ed250-57d8-4273-8761-9a9bfc4136af (item 2).

script/deploy/vaultWrapper/DeployLiFiVaultWrapperFactory.s.sol
gvladika: Agreed it's an intentional bootstrap stand-in. Clarified the `_salt` NatSpec in `2b12e4ab` to state the precondition: cross-chain address parity requires the factory (CREATE2 deployer) and the beacon (in the proxy creation code) to be at identical addresses on each chain. Recorded the CREATE3 deterministic-deploy requirement as a hard S14 item: https://linear.app/lifi-linear/issue/EXSC-420/s14-deployment-and-configuration-scripts#comment-089ed250-57d8-4273-8761-9a9bfc4136af (item 1).

src/VaultWrapper/LiFiVaultWrapperFactory.sol
gvladika: Fixed in `2b12e4ab` — added `@param`/`@return` to all eight: the seven external setters and `predictAddress` (4× `@param` + `@return`).

src/VaultWrapper/LiFiVaultWrapperFactory.sol
gvladika: Fixed in `2b12e4ab` — added a no-fund-custody line to both `LiFiVaultWrapperFactory` ("does not custody user funds; it only deploys and configures wrapper instances") and `ERC4626Adapter` ("does not custody user funds; it is a stateless view helper").

src/VaultWrapper/LiFiVaultWrapperFactory.sol
gvladika: Fixed in `2b12e4ab` — the constructor `@param _onboardingManager` now reads "Address authorized to assign/revoke the deployer for each integrator namespace."

src/VaultWrapper/LiFiVaultWrapperTypes.sol
gvladika: Fixed in `2b12e4ab` — added `@title LiFiVaultWrapperTypes` / `@author LI.FI` to the file-level block.

src/VaultWrapper/mocks/MockVaultWrapper.sol
gvladika: Fixed in `2b12e4ab` — added `@notice` to `MockVaultWrapper.AlreadyInitialized` and `@notice`/`@return` to the deploy script's `run()`.

src/VaultWrapper/LiFiVaultWrapperFactory.sol
gvladika: Acknowledged, no change — these describe intended S1/S5 clone-side behavior and are accurate about the system role. (I did add a `deploy()`-side `@dev` note clarifying that new instances inherit the live `globalPaused` flag.)

script/deploy/vaultWrapper/DeployLiFiVaultWrapperFactory.s.sol
gvladika: Set to `1.0.0` in `2b12e4ab`. Agreed with the principle that `@custom:version` should track shipped releases, not in-dev iterations — so all S8 artifacts (script included) are `1.0.0` until go-live.

PR #1943, opened 2026-06-18
EXSC-520: Cache Foundry RPC state in CI

0xDEnYO:

https://linear.app/lifi-linear/issue/EXSC-520

Pinned fork blocks make RPC cache deterministic. Restoring ~/.foundry/cache/rpc in setup-foundry keyed on block-pin hash avoids cold archive fetches on every CI run.

CodeRabbit: clean (local /pr-ready).

---

**pr-ready note (2026-06-25):** local CodeRabbit review timed out after ~59 min on the latest commit (`a3c260274`) — CR backend was degraded today, not a real review (timeout ID `3fe20168`). Diff is a single bash script and was verified manually (syntax, hash recompute, all six pins captured, default single-sourced). Pushed via the documented `PR_READY_OK=1` bypass; CI CodeRabbit remains the safety net.


script/utils/hashForkBlockPins.sh
```diff
@@ -0,0 +1,15 @@
...
+
+ROOT="$(git rev-parse --show-toplevel 2>/dev/null || pwd)"
+cd "$ROOT"
+
+DEFAULT_BLOCK="15588208"
+
+{
+  echo "$DEFAULT_BLOCK"
+  grep -rhoE 'customBlockNumberForForking = [0-9]+' test/solidity archive/test 2>/dev/null \
+    | sed 's/customBlockNumberForForking = //' || true
```
gvladika: The header comment says this hashes *"all pinned fork block numbers in test files,"* but this grep only matches the `customBlockNumberForForking = N` style. It misses pins set via a local var + `vm.createSelectFork(rpcUrl, blockNumber)` — and two are **active** tests:
- `test/solidity/Facets/CBridge/CBridgeRefund.t.sol:78` → `25085298` (Polygon)
- `test/solidity/Facets/HopFacet.t.sol:276` → `36004499`

Consequence: if one of those pins changes, the hash doesn't change, so `actions/cache` restores the stale blob and never re-saves → that block is fetched cold on every run until some other pin bumps the key. Broaden the pattern to also catch `blockNumber = [0-9]+` / `createSelectFork(..., <num>)`, or drive the hash off the manifest (see cross-PR note).

Separately: grepping `archive/test` busts the cache when archived tests change yet contributes nothing (the archived test uses the unmatched local-var style anyway) — drop it.

script/utils/hashForkBlockPins.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
+# Prints a stable SHA-256 of all pinned fork block numbers in test files.
+# Used as a GitHub Actions cache key segment for ~/.foundry/cache/rpc.
+set -euo pipefail
+
+ROOT="$(git rev-parse --show-toplevel 2>/dev/null || pwd)"
+cd "$ROOT"
+
+DEFAULT_BLOCK="15588208"
```
gvladika: Hardcoding `15588208` here makes this the **third** copy of the default block (alongside `TestBase.DEFAULT_BLOCK_NUMBER_MAINNET` and #1944's `fork-blocks.manifest`). If the default ever changes, all three must move in lockstep or the cache key is silently wrong. Derive it from a single source of truth instead — read it out of `TestBase.sol` or the manifest rather than duplicating the literal.

gvladika (commented):
Premise is sound (pinned blocks → immutable RPC responses → safe to cache) and the wiring is correct — I confirmed the hash step runs and `actions/cache` keys as expected in `deploy-smoke-test`. No correctness risk: Foundry's RPC cache is content-addressed by chain+block, so a stale key only degrades cache refresh, never returns wrong data. Two things worth addressing, plus a cross-PR decision. Details inline.

⚠️ **Cross-PR coupling with #1944:** the hash greps `customBlockNumberForForking = <number>` literals. #1944 establishes the manifest + the `DEFAULT_BLOCK_NUMBER_MAINNET` constant, and a natural cleanup there is to replace those literals with the constant — which would make them invisible to this grep. So the literal-grep here and the constant-consolidation in #1944 are at odds. I'd settle it across both PRs rather than as a fast-follow: either keep the literals in #1944 (and just fix the grep below), or make `fork-blocks.manifest` the single source of truth that both `TestBase` and this script read. Avoid clean-constants-in-1944 + literal-grep-here.

Minor: this hash + RPC-cache restore runs in *every* job that uses `setup-foundry`, including non-fork jobs (deploy-smoke-test, generate-tag), which restore a blob they don't use. Consider gating the cache step to fork-test jobs.

script/utils/hashForkBlockPins.sh
0xDEnYO: Fixed in a3c260274. You were right — and it was actually **5** active pins missed, not 2. Broadened the grep to capture all three styles:

- `customBlockNumberForForking = N`
- local var: `blockNumber = N` (+ `vm.createSelectFork(rpcUrl, blockNumber)`)
- inline literal: `vm.createSelectFork(rpc, N)`

That brings in the two you flagged (`CBridgeRefund.t.sol:78` → 25085298, `HopFacet.t.sol:276` → 36004499) plus three the old pattern also missed: `Executor.t.sol:89`, `AcrossV4SwapFacet.Sponsored.t.sol:81` → 445622003, and `LiFiDEXAggregator.t.sol:1904` → 29831758.

Also dropped the `archive/test` scan per your note — archived tests use the unmatched local-var style and only churn the key for nothing.

Note: these pins live in `CBridgeRefund` / `HopFacet`, which #1944 does **not** touch — #1944 only deletes the `customBlockNumberForForking` pins in `CBridgeAndFeeCollection` / `Gas/Hop`. So the grep genuinely needed broadening regardless of #1944's merge order.

script/utils/hashForkBlockPins.sh
0xDEnYO: Fixed in a3c260274. Now read from `TestBase.DEFAULT_BLOCK_NUMBER_MAINNET` (the manifest is gone after #1944, so `TestBase.sol` is the single source), with a fail-loud guard if the constant can't be parsed:

```bash
DEFAULT_BLOCK="$(grep -oE 'DEFAULT_BLOCK_NUMBER_MAINNET = [0-9]+' "$TESTBASE" | grep -oE '[0-9]+$' || true)"
if [ -z "$DEFAULT_BLOCK" ]; then
  echo "::error::could not read DEFAULT_BLOCK_NUMBER_MAINNET from $TESTBASE" >&2
  exit 1
fi
```

No more duplicated literal — bump the constant and the cache key follows.

script/utils/hashForkBlockPins.sh
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env bash
```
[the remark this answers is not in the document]
0xDEnYO: Fixed in 69512dfd7 — valid finding per `.agents/rules/300-bash.md` (this script was the lone `#!/usr/bin/env bash` outlier among 31 `.sh` files):

- shebang → `#!/bin/bash`
- `[ -z "$DEFAULT_BLOCK" ]` → `[[ -z "$DEFAULT_BLOCK" ]]`

gvladika (approved):

gvladika (approved):

maxklenk (approved):

PR #1944, opened 2026-06-18
EXSC-522: Consolidate pinned fork block numbers

0xDEnYO:

https://linear.app/lifi-linear/issue/EXSC-522

Align default fork block with DEFAULT_BLOCK_NUMBER_MAINNET (15588208), migrate explicit legacy pins, and add fork-blocks.manifest as consolidation documentation.

CodeRabbit: clean (local /pr-ready).


test/solidity/Facets/CBridge/CBridgeAndFeeCollection.t.sol
```diff
@@ -22,7 +22,7 @@ contract CBridgeAndFeeCollectionTest is TestBase {
     TestCBridgeFacet internal cBridge;
 
     function setUp() public {
-        customBlockNumberForForking = 14847528;
+        customBlockNumberForForking = 15588208;
```
gvladika: This (and the identical line in `Gas/Hop.t.sol`) re-pins the literal `15588208`, which is now exactly the default. Since this test extends `TestBase`, the line is a no-op — it sets `customBlockNumberForForking` to the value the fallback would already use. The actual consolidation is to **delete the line** (fall through to the default) or, if an explicit pin is wanted, `customBlockNumberForForking = DEFAULT_BLOCK_NUMBER_MAINNET;`. As written we now have six loose copies of `15588208` across the suite that must be hand-synced with the constant — the drift this PR set out to remove.

⚠️ Note the #1943 coupling in the top-level comment before changing this: #1943 greps the numeric literal, so switching to the constant or deleting the line affects its cache hash.

test/solidity/Periphery/Executor.t.sol
```diff
@@ -86,7 +86,7 @@ contract ExecutorTest is Test {
 
     function fork() internal {
         string memory rpcUrl = vm.envString("ETH_NODE_URI_MAINNET");
-        uint256 blockNumber = 14847528;
+        uint256 blockNumber = 15588208;
```
gvladika, line 91: `ExecutorTest` extends `Test`, not `TestBase`, so `DEFAULT_BLOCK_NUMBER_MAINNET` isn't in scope here — hence the bare literal. For real consolidation, import the constant (or move this test onto `TestBase`); at minimum add a one-line comment on why the block is pinned. As-is this is another hand-synced copy of the default.

test/solidity/fork-blocks.manifest
```diff
@@ -0,0 +1,8 @@
+# Canonical mainnet fork block for tests that do not set customBlockNumberForForking.
+15588208
+
+# Frequently reused pins (consolidation targets — migrate tests to these when possible):
+# 15588208 — default mainnet (TestBase.DEFAULT_BLOCK_NUMBER_MAINNET)
+# 58467500 — Arbitrum-heavy integrations (Hop/CBridge packed)
+# 22989702 — Across V4 family
+# 19979843 — Emergency pause / calldata verification
```
gvladika: The groupings here are accurate (verified each pin exists in the suite), but the file's design is drift-prone:

- **Nothing reads it.** Line 2's bare `15588208` duplicates `TestBase.DEFAULT_BLOCK_NUMBER_MAINNET` (and #1943 hardcodes it a third time). If the default changes, this silently goes stale and no check catches it.
- **It documents not-yet-done work** ("migrate tests to these when possible") — a TODO dressed as documentation, which brushes against the repo's "don't document features that aren't implemented" standard.

Stronger alternative that actually consolidates: promote the three reused pins to **named constants in `TestBase`** (e.g. `BLOCK_ARBITRUM_PACKED = 58467500`, `BLOCK_ACROSS_V4 = 22989702`, `BLOCK_EMERGENCY_PAUSE = 19979843`) and reference them from the tests. Then the code *is* the manifest, drift is impossible, and #1943 can hash the constants' source of truth. If you keep the file, make it machine-checked (a unit test asserting every pin literal in the suite is one of the manifest values) so it can't rot.

gvladika (commented):
Direction is right and CI is green — the `TestBase` fallback going from a magic literal to `DEFAULT_BLOCK_NUMBER_MAINNET` is a genuine improvement. Two things to consider before merge: the consolidation is only half-done (touched tests still hardcode the literal), and the manifest is a drift-prone doc. Details inline.

⚠️ **Cross-PR coupling with #1943:** #1943's cache-key script hashes `customBlockNumberForForking = <number>` *literals* by grep. So the "use the named constant instead of the literal" suggestions below would make those lines invisible to #1943's hash. Don't action these in isolation — decide the consolidation strategy across both PRs (see my review on #1943). The one combination to avoid is clean-constants-here + literal-grep-there.

test/solidity/Facets/CBridge/CBridgeAndFeeCollection.t.sol
0xDEnYO: Done in 3356172d25bbff03b1f3740e81bee5978e3322db — deleted the line entirely so the test falls through to the `TestBase` default. Same for `Gas/Hop.t.sol`. Chose deletion over `= DEFAULT_BLOCK_NUMBER_MAINNET` deliberately: it removes the loose literal *and* stays compatible with #1943's grep (the default `15588208` is still hashed via that script's hardcoded `DEFAULT_BLOCK`). Switching to the constant here would have hidden the line from #1943's `customBlockNumberForForking = [0-9]+` cache hash — the combination you flagged to avoid.

test/solidity/Periphery/Executor.t.sol
0xDEnYO: Added a comment in 3356172d25bbff03b1f3740e81bee5978e3322db explaining why the literal stays: `ExecutorTest extends Test` (not `TestBase`), so `DEFAULT_BLOCK_NUMBER_MAINNET` isn't in scope, and this `blockNumber =` line isn't matched by #1943's grep anyway. Kept it a literal rather than importing/inheriting to avoid scope creep on a test that's otherwise standalone.

test/solidity/fork-blocks.manifest
0xDEnYO: Agreed — deleted the manifest in 3356172d25bbff03b1f3740e81bee5978e3322db. It was TODO-as-docs that nothing reads, exactly the drift you called out. Skipped the 'promote reused pins to named constants' alternative on purpose: converting the `58467500`/`22989702`/`19979843` literals to constants would make them invisible to #1943's literal grep, so that's a coordinated change to make *across both PRs* (update `hashForkBlockPins.sh` at the same time) rather than unilaterally here. Left as a follow-up; this PR now just keeps the `TestBase` literal→constant improvement + removes the redundant pins.

gvladika (approved):

gvladika (approved):

PR #1946, opened 2026-06-18
EXSC-524: Shard forge test CI across runner matrix

0xDEnYO:

https://linear.app/lifi-linear/issue/EXSC-524

Parallel matrix over Libraries/Facets/Periphery/Gas/misc reduces wall-clock CI time. Rebased on EXSC-523 — merge EXSC-523 first (includes FOUNDRY_PROFILE=ci + matrix).

Depends on EXSC-523. CodeRabbit: not run (CLI rate limit after 523).

### DIFF
--- a/.github/workflows/enforceTestCoverage.yml
+++ b/.github/workflows/enforceTestCoverage.yml
@@ -96,7 +96,24 @@ jobs:
         env:
           FOUNDRY_PROFILE: ci
         run: |
-          forge coverage --report lcov --force --ir-minimum
+          set -euo pipefail
+          # forge coverage hits live RPCs for fork tests and has no built-in
+          # retry, so a single transient RPC blip sinks the whole report. Retry
+          # the full command (--rerun can't be used — coverage needs the complete
+          # run). Kept low: --ir-minimum coverage is slow (~8 min/run).
+          for i in 1 2 3; do
+            echo "::group::forge coverage (attempt $i/3)"
+            if forge coverage --report lcov --force --ir-minimum; then
+              echo "::endgroup::"
+              break
+            fi
+            echo "::endgroup::"
+            if [[ "$i" -eq 3 ]]; then
+              echo "forge coverage failed after 3 attempts" >&2
+              exit 1
+            fi
+            sleep 15
+          done
 
           echo "Filtering coverage report to only contain coverage info for 'src/'' folder now"
 
--- a/.github/workflows/enforceTestCoverage.yml
+++ b/.github/workflows/enforceTestCoverage.yml
@@ -93,6 +93,8 @@ jobs:
         run: forge install
 
       - name: Generate Coverage Report
+        env:
+          FOUNDRY_PROFILE: ci
         run: |
           forge coverage --report lcov --force --ir-minimum
 
--- a/.github/workflows/forge-unit-tests.yml
+++ b/.github/workflows/forge-unit-tests.yml
@@ -99,6 +99,8 @@ jobs:
       - name: Run forge tests
         # retry failed tests up to 10 times (RPC flakiness)
         shell: bash
+        env:
+          FOUNDRY_PROFILE: ci
         run: |
           forge test && exit 0
           for i in $(seq 1 10); do
--- a/.github/workflows/forge-unit-tests.yml
+++ b/.github/workflows/forge-unit-tests.yml
@@ -108,29 +108,24 @@ jobs:
           FOUNDRY_PROFILE: ci
         run: |
           set -euo pipefail
-          MATCH_PATHS=()
+          # forge accepts only a single --match-path/--no-match-path argument, so
+          # each shard maps to exactly one glob. `misc` is a catch-all by
+          # exclusion: any new top-level dir under test/solidity/ auto-runs there
+          # instead of silently dropping out of CI with all checks green.
           case "${{ matrix.shard }}" in
-            libraries) MATCH_PATHS+=("test/solidity/Libraries/**") ;;
-            facets) MATCH_PATHS+=("test/solidity/Facets/**") ;;
-            periphery) MATCH_PATHS+=("test/solidity/Periphery/**") ;;
-            gas) MATCH_PATHS+=("test/solidity/Gas/**") ;;
-            misc)
-              MATCH_PATHS+=("test/solidity/Helpers/**")
-              MATCH_PATHS+=("test/solidity/Security/**")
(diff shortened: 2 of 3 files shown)


0xDEnYO:
⚠️ **Depends on #1945 (EXSC-523)** — that PR adds `[profile.ci]` (reduced fuzz runs) and the matrix groundwork this PR builds on. **Merge #1945 first.** This branch is rebased on it, so the `foundry.toml` profile.ci diff overlaps until 1945 lands.

.github/workflows/forge-unit-tests.yml
```diff
@@ -96,15 +101,36 @@ jobs:
...
+              MATCH_PATHS+=("test/solidity/Helpers/**")
+              MATCH_PATHS+=("test/solidity/Security/**")
+              MATCH_PATHS+=("test/solidity/LiFiDiamond.t.sol")
+              ;;
+          esac
+
+          FORGE_ARGS=(test)
+          for path in "${MATCH_PATHS[@]}"; do
+            FORGE_ARGS+=(--match-path "$path")
+          done
```
gvladika: 🔴 **Blocker — this breaks every shard with more than one path, which is exactly why `misc` is the only red shard.**

`forge test` rejects a repeated `--match-path`:
```
error: the argument '--match-path <GLOB>' cannot be used multiple times
```
Reproduced on forge 1.7.1, and it's in the `misc` job log failing all 10 retries. The four single-path shards pass only because the loop never appends a second `--match-path`.

Use a single glob with brace alternation instead of N args, and drop the loop:
```bash
FORGE_ARGS=(test --match-path "$MATCH_PATH")
```
with `MATCH_PATH` set per shard (see the comment on the `case` block for the `misc` value).

.github/workflows/forge-unit-tests.yml
```diff
@@ -96,15 +101,36 @@ jobs:
...
+            libraries) MATCH_PATHS+=("test/solidity/Libraries/**") ;;
+            facets) MATCH_PATHS+=("test/solidity/Facets/**") ;;
+            periphery) MATCH_PATHS+=("test/solidity/Periphery/**") ;;
+            gas) MATCH_PATHS+=("test/solidity/Gas/**") ;;
+            misc)
+              MATCH_PATHS+=("test/solidity/Helpers/**")
+              MATCH_PATHS+=("test/solidity/Security/**")
+              MATCH_PATHS+=("test/solidity/LiFiDiamond.t.sol")
+              ;;
+          esac
```
gvladika, line 122: 🟠 **Should fix — two things:**

**(a)** Collapse `misc`'s three paths into one brace glob (required by the `--match-path` single-arg limit above). Verified locally this matches exactly `Helpers/**`, `Security/**`, and `LiFiDiamond.t.sol`:
```bash
misc) MATCH_PATH="test/solidity/{Helpers/**,Security/**,LiFiDiamond.t.sol}" ;;
```

**(b) Silent-coverage risk:** this is a hardcoded allowlist. Today the union covers every real test file (only `test/solidity/utils/**` is excluded, and those have no runnable `test*` functions — confirmed). But the day someone adds a new top-level dir (e.g. `test/solidity/Integration/`), it lands in **no shard** and silently stops running in CI, with all checks green. Safer to make `misc` a catch-all by exclusion so new dirs auto-fall into it:
```bash
misc) NO_MATCH="test/solidity/{Libraries,Facets,Periphery,Gas}/**" ;;
```
then pass `--no-match-path "$NO_MATCH"` for that shard (`--no-match-path` has the same single-arg limit, hence the brace glob again). At minimum (a) is needed to make `misc` run; (b) is the more durable fix.

.github/workflows/enforceTestCoverage.yml
```diff
@@ -93,6 +93,8 @@ jobs:
         run: forge install
 
       - name: Generate Coverage Report
+        env:
+          FOUNDRY_PROFILE: ci
         run: |
           forge coverage --report lcov --force --ir-minimum
```
gvladika: 🟡 **Non-blocking — the coverage failure on this PR is RPC flakiness, not logic.** 3 fork tests failed with `operation timed out` / `error sending request for url` (`EcoFacet.t.sol`, `LiFiDEXAggregator.t.sol`). The unit-test job survives this via its 10× retry loop; this single `forge coverage` call has none, so one transient blip sinks coverage.

Since this PR already edits this step, consider a retry. Note you can't use `--rerun` (coverage needs the full report), so retry the whole command, and keep the count low since `--ir-minimum` coverage is slow (~8 min/run). Reasonable to defer to a follow-up and just re-run the job to get green — flagging since you're touching the file.

foundry.toml
```diff
@@ -16,6 +16,16 @@ ffi = true
 libs = ["node_modules", "lib"]
 cache = true
 
+# CI profile: lower fuzz runs for faster PR feedback. Local default keeps 256 runs.
+[profile.default.fuzz]
+runs = 256
```
gvladika, line 21: ⚪ **Nit (optional):** `[profile.default.fuzz] runs = 256` is redundant — 256 is already Foundry's default. Harmless and it documents intent, so fine to keep. The `[profile.ci]` inheritance is correct: the selected profile merges over `default`, so overriding only `fuzz.runs = 32` is exactly right.

gvladika (commented):
Automated review notes for the CI sharding change. One blocker (the `misc` shard can't run as written — verified locally on forge 1.7.1), one design risk worth fixing, plus two non-blocking notes. Details inline.

.github/workflows/forge-unit-tests.yml
0xDEnYO: Fixed in 53ee3761 — verified locally on forge 1.7.1 that repeated `--match-path` is rejected. Each shard now maps to a single arg, and `misc` is a catch-all by exclusion (`--no-match-path "test/solidity/{Libraries,Facets,Periphery,Gas}/**"`) so new top-level test dirs auto-run instead of silently dropping. Added an unknown-shard guard. All 5 sharded jobs (incl. misc) are green in CI now.

foundry.toml
0xDEnYO: Keeping as-is per your note — agreed it documents intent. The `[profile.ci]` `fuzz.runs = 32` override is intentional. (Also wrapped `forge coverage` in a 3× retry to match the unit-test job's RPC-flakiness handling, since the PR already touches that workflow.)

gvladika (approved):

maxklenk (approved):

PR #1959, opened 2026-06-18
fix(safe): don't advance expectedNonce on reverted Safe tx (EXSC-533)

gvladika:

[EXSC-533](https://linear.app/lifi-linear/issue/EXSC-533/fix-confirm-safe-tx-nonceduplicate-row-bug-when-transactions-revert)

Completes the **nonce-advancement** half of EXSC-533. The duplicate-row half (signatures landing on the wrong MongoDB row) was already fixed and merged in #1956; this builds on top of it.

`confirm-safe-tx`'s `executeTransaction` returned `true` for a `reverted` outcome, and every caller advances `expectedNonce` on a `true` return. That is wrong: a top-level revert rolls back all state, including the Safe's own `nonce++`, so the nonce is **not** consumed. (In this repo the executor submits with `safeTxGas=0`, which is why an inner-call failure surfaces as a top-level `execTransaction` revert (GS013) rather than an `ExecutionFailure` with a consumed nonce — but the `reverted ⇒ nonce-not-consumed` rule holds regardless of gas params.)

Treating `reverted` as nonce-consuming desynced `expectedNonce` within a run: after a revert the loop would treat the next-nonce proposal as current/executable, then it would revert on-chain with **GS026** (wrong nonce). This was observed during the 2026-06-18 emergency-pause test on the Rootstock unpause and required manual DB surgery.

Changes:
- Added `NONCE_CONSUMING_STATUSES` in `safe-utils.ts` as the single, documented source of truth (`['executed']`), referenced by both the runtime predicate `safeTxStatusConsumedNonce` and reconcile's highest-consumed-nonce query. A MongoDB filter can't call the predicate, so a shared constant — not the function alone — is what genuinely unifies the two sites. Extracting a pure helper makes the rule unit-testable (`executeTransaction` is a private closure) and pins it so a future edit can't silently re-add `|| 'reverted'`.
- `executeTransaction` now returns `safeTxStatusConsumedNonce(nextStatus)` instead of `=== 'executed' || === 'reverted'`.
- **`reconcile.ts`**: `computeExpectedNonce` previously queried the highest-consumed nonce with `status: { $in: ['executed', 'reverted'] }`, encoding the old (wrong) rule by hand. A `reverted` row's nonce inflated `maxConsumed`, which masked a real gap in the revert → re-execute → lost-hash case and silently skipped Sweep B's idempotent back-fill (the EXSC-248 "Case 1" safety net). Switched the query to `{ $in: [...NONCE_CONSUMING_STATUSES] }` and added a test pinning a `reverted` row as the highest consumed-status row.
- Corrected the misleading `On-chain nonce has advanced` log line, the `SafeTxStatus` lifecycle doc (lines 106-108, which still claimed `reverted` consumed the nonce "because signatures verified"), and the `executeTransaction` comment (now accurate for the no-receipt Tron path).
- Added tests: `executed`→true, `reverted`→false (the bug), `submitted`/`pending`→false; plus the reconcile back-fill test above. Mutation-checked the reconcile test (fails against the old `['executed','reverted']` query, passes with the fix).

Only the `reverted` path changes behavior. `executed` (including the Tron no-receipt path) and `submitted` are unchanged.

## Tron status resolution (same bug class, fixed here)

The Tron path previously marked **every** execution `executed` as soon as a broadcast hash returned, even if the tx reverted on-chain — and `reconcile.ts` skips Tron entirely, so nothing corrected a wrong `executed`. A failed Tron Safe tx advanced `expectedNonce` with no safety net: the bug this PR closes for EVM, left open on Tron. Tron's TVM rolls back `nonce++` on a top-level revert exactly like EVM, so the rule is identical.

- Replaced `IChainExecutionResult.receipt` (a viem `TransactionReceipt`) with a normalized `status?: 'success' | 'reverted'`. Its only consumer was `confirm-safe-tx` (the timelock path uses the separate `IChainCallResult`), so the contract change is contained.
- `evm-executor` derives `status` from the receipt (unchanged behavior, incl. the poll-timeout → `undefined` → `submitted` path).
- `tron-executor` now resolves the result **synchronously** via `getTransactionInfo` (`waitForConfirmation`) and maps `receipt.result` through an extracted, unit-tested `resolveTronExecutionStatus` helper (anything set and ≠ `'SUCCESS'` → `reverted`). Synchronous because Tron has no reconcile back-fill — a deferred `submitted` row would never be resolved.
- `confirm-safe-tx` drops the Tron special-case and uses the same `status` check for both chains. An `undefined` status (EVM poll timeout) still leaves the row `submitted` for reconciliation.

Verification: `bunx tsc-files --noEmit` and `bunx eslint` clean on all changed files; `bun test script/deploy/safe/` → 147 pass / 0 fail (incl. new `tron-executor.test.ts`); `/pr-ready` (local CodeRabbit) → no findings.


script/deploy/safe/safe-utils.ts
```diff
@@ -837,6 +837,25 @@ export class SafeClient {
...
+ * unknown outcome (no receipt yet) and must not be treated as consuming the
+ * nonce until reconciliation resolves it. Callers gate `expectedNonce++` on
+ * this — returning true for `reverted` would desync the expected nonce and make
+ * the next-nonce proposal revert with GS026.
+ *
+ * @param status - Resolved Safe-tx status after an execution attempt.
+ * @returns true only for `executed`; false for `reverted`, `submitted`, and `pending`.
+ */
+export function safeTxStatusConsumedNonce(status: SafeTxStatus): boolean {
+  return status === 'executed'
```
mirooon: **"Single source of truth" is not yet true — reconcile still treats `reverted` as nonce-consuming.**

This helper is described in the PR as the "single, documented source of truth" that pins the rule so a future edit can't silently re-add `|| 'reverted'`. But `reconcile.ts` still encodes the literal equivalent by hand, un-pinned and not sharing this rule:

- `reconcile.ts:515` — `status: { $in: ['executed', 'reverted'] }`
- `reconcile.ts:499-500` — doc: "highest consumed (executed or reverted) nonce + 1"
- `reconcile.ts:533` — `const consumedHead = maxConsumed >= 0n ? maxConsumed + 1n : 0n`

By this PR's own logic a `reverted` row did NOT consume its nonce, so it must not bump `consumedHead`. This isn't cosmetic — `computeExpectedNonce` gates Sweep B (the lost-hash back-fill safety net, EXSC-248 "Case 1"):

```
executed @0–3            -> on-chain nonce 4
proposal @4 reverts      -> row reverted@4, on-chain nonce STILL 4
re-proposal @4 executes  -> on-chain nonce 5, but hash lost before reaching Mongo

computeExpectedNonce = maxConsumed(4)+1 = 5
onChainNonce(5) <= expected(5)  ->  gap NOT detected  ->  Sweep B skipped
->  the missed execution is never back-filled.
```

If `reverted` were excluded (count only executed@3 -> expected 4), then 5 > 4 would trigger Sweep B and back-fill correctly. So reconcile silently defeats the safety net in exactly the revert-then-re-execute case this PR is about. Sweep B is read-only and idempotent, so erring toward running it more is strictly safer; the current code errs the harmful way.

**Suggested:** change `reconcile.ts:515` to `status: { $eq: 'executed' }` and update the `:499-500` doc. Note line 515 is a MongoDB query filter, not a runtime check, so it can't call `safeTxStatusConsumedNonce(...)` directly — to make the "single source of truth" real, hoist the rule into a shared constant both sites reference:

```ts
export const NONCE_CONSUMING_STATUSES: readonly SafeTxStatus[] = ['executed']
// helper:    status === NONCE_CONSUMING_STATUSES[0]   (or .includes(status))
// reconcile: status: { $in: [...NONCE_CONSUMING_STATUSES] }
```

No test pins a `reverted` row as highest-consumed (`reconcile.test.ts:465,501` only exercise `executed` for the head), so flipping to `executed`-only breaks no test. If you'd rather keep this PR narrowly scoped, please add a follow-up ticket and a note in the PR body rather than leaving the divergence silent.

script/deploy/safe/safe-utils.ts
```diff
@@ -837,6 +837,25 @@ export class SafeClient {
...
+ * always submits with `safeTxGas=0`, so an inner-call failure reverts the
+ * entire `execTransaction` (GS013); the reverted tx rolls back the nonce
+ * increment, leaving the nonce open for re-proposal. A `submitted` row is an
+ * unknown outcome (no receipt yet) and must not be treated as consuming the
+ * nonce until reconciliation resolves it. Callers gate `expectedNonce++` on
+ * this — returning true for `reverted` would desync the expected nonce and make
+ * the next-nonce proposal revert with GS026.
+ *
+ * @param status - Resolved Safe-tx status after an execution attempt.
+ * @returns true only for `executed`; false for `reverted`, `submitted`, and `pending`.
```
mirooon, line 866: **Contradictory doc in the same file this PR edits.**

The canonical `SafeTxStatus` lifecycle doc at lines 106-108 still says the opposite of this new helper:

```
* - `reverted`  : on-chain receipt confirmed revert (Safe nonce still
*                 consumed because signatures verified; flagged for manual
*                 review).
```

"Safe nonce still consumed because signatures verified" is exactly the wrong mental model this PR exists to correct — signature verification is irrelevant; the top-level revert rolls back the `nonce++`. A reader hitting that doc first will re-introduce the bug. Please update lines 106-108 to match this helper: a `reverted` tx does NOT consume the nonce; the slot can be re-proposed.

script/deploy/safe/confirm-safe-tx.ts
```diff
@@ -199,9 +200,13 @@ const processTxs = async (
    * @param txDoc - The pendingTransactions row being processed
    * @param safeClient - The Safe client to use for execution (defaults to main safe client)
    */
-  // Returns true if the Safe nonce was consumed on-chain (executed or
-  // reverted with valid signatures). Returns false when the outcome is
-  // unknown ('submitted') — the caller should not advance expectedNonce.
+  // Returns true only when the execution tx itself succeeded
+  // (receipt.status === 'success') — the only outcome that consumes the Safe
```
mirooon: **Function comment is inaccurate for the Tron path — and Tron could mirror EVM here.**

"Returns true only when the execution tx itself succeeded (`receipt.status === 'success'`)" isn't quite accurate: the Tron branch (lines 235-239) sets `nextStatus='executed'` with NO receipt and this function then returns true.

**Minimal fix (this PR):** reword to "Returns true only for the `executed` status (receipt success, or the Tron no-receipt path) — the only outcome that consumes the Safe nonce."

**Worth a follow-up — give Tron the same status resolution as EVM.** This isn't only a comment nit. Today the Tron path eagerly marks **every** execution `executed` as soon as a hash comes back, even if the tx later failed on-chain — and `reconcile.ts:108` skips Tron entirely, so there's no safety net to correct a wrong `executed`. A failed Tron Safe tx would advance the nonce and never be fixed: the exact bug class this PR closes for EVM, left open on Tron. (Tron's TVM rolls back the nonce on a top-level revert just like EVM, so the rule is identical.)

The info is already available — the repo polls `tronWeb.trx.getTransactionInfo(txId)` via `waitForConfirmation()` (`script/troncast/utils/tronweb.ts:92`) and reads `receipt.result` (`troncast/.../send.ts:222`, `=== 'FAILED'`). So `tron-executor.ts` could resolve the result and return a normalized `receipt.status`, letting `confirm-safe-tx.ts` drop the special Tron branch and use the same `exec.receipt.status === 'success' ? 'executed' : 'reverted'` line as EVM:

```ts
// tron-executor.ts — after broadcast
const info = await waitForConfirmation(tronWeb, hash)
const result = info?.receipt?.result          // undefined / 'SUCCESS' = ok
return {
  hash, displayHash, explorerUrl,
  receipt: { status: result && result !== 'SUCCESS' ? 'reverted' : 'success' },
}
```

Caveat: Tron has no reconcile back-fill, so the executor must resolve **synchronously** (wait for the result and decide `executed`/`reverted` there) rather than leaving a `submitted` row. Suggest a follow-up ticket for this; the reword above is fine for this PR.

script/deploy/safe/confirm-safe-tx.ts
```diff
@@ -299,7 +304,7 @@ const processTxs = async (
       )
       consola.log(' ')
 
-      return nextStatus === 'executed' || nextStatus === 'reverted'
+      return safeTxStatusConsumedNonce(nextStatus)
```
mirooon, line 305: **Fix is more robust than the description implies (non-blocking praise).**

Correctness here doesn't actually depend on `safeTxGas=0`, which the PR description leans on. With `safeTxGas>0` or `gasPrice>0` a failed inner call does NOT revert — the Safe emits `ExecutionFailure`, the tx succeeds, `receipt.status==='success'` -> 'executed' -> nonce correctly consumed. The `reverted` status only ever arises from a genuine top-level revert, where the EVM rolls back the nonce increment unconditionally. So `reverted => nonce-not-consumed` is universal, not conditional on the executor's gas params — the reasoning is stronger than "because safeTxGas=0".

One latent caveat: if a relayer ever wraps `execTransaction` in try/catch so the outer tx succeeds while the inner Safe reverts, `receipt.status` would be 'success' and we'd mis-mark 'executed'. Not a concern today — `evm-executor.ts:64-72` calls `execTransaction` directly via `writeContract` and returns that call's own receipt (no wrapping tx), and the executor submits `safeTxGas=0/baseGas=0/gasPrice=0` at `evm-executor.ts:40-42`.

script/deploy/safe/safe-utils.ts
```diff
@@ -837,6 +837,25 @@ export class SafeClient {
...
+ */
+export function safeTxStatusConsumedNonce(status: SafeTxStatus): boolean {
```
mirooon, line 868: **A one-line predicate with a single call site — make the rule genuinely shared.**

The body is literally `return status === 'executed'`, and right now it has exactly one caller (`confirm-safe-tx.ts:307`). The PR's justification ("single source of truth" so a future edit can't re-add `|| 'reverted'`) only holds if it's actually the single source. It isn't: `reconcile.ts:515` (`status: { $in: ['executed', 'reverted'] }`) encodes the same rule by hand and contradicts it (see the other comment on the function body).

**Preferred — make the rule genuinely shared:** hoist a `NONCE_CONSUMING_STATUSES` constant that both this helper and reconcile's consumed-nonce query reference. reconcile uses a MongoDB filter, so it can't call the predicate directly — a shared constant, not the function, is what unifies them. With both sites enforcing one rule the claim becomes true.

I lean toward keeping the function regardless of call count — its value isn't the one-liner body, it's (a) the documented EVM-rollback rationale and (b) unit-testability (`executeTransaction` is a private closure and can't be tested directly). Pair it with the shared constant and you get the testable predicate **and** a genuinely single source of truth.

mirooon (commented):
Line-anchored review of the reverted-nonce fix.

script/deploy/safe/safe-utils.ts
gvladika: Fixed in 808ebdb0. Hoisted `NONCE_CONSUMING_STATUSES` (`['executed']`) in `safe-utils.ts` as the shared rule; `reconcile.ts:515` now queries `status: { $in: [...NONCE_CONSUMING_STATUSES] }` instead of `['executed', 'reverted']`, and the `computeExpectedNonce` doc explains why counting `reverted` would mask the gap and skip Sweep B.

Added a `reconcile.test.ts` case pinning your exact scenario (executed@3, reverted@4, re-proposal@4 with lost hash): it now triggers Sweep B and back-fills. Mutation-checked — it fails against the old `['executed','reverted']` query and passes with the fix.

script/deploy/safe/safe-utils.ts
gvladika: Fixed in 808ebdb0 — lifecycle doc (lines 106-108) now states a `reverted` tx did **not** consume the nonce (a top-level revert rolls back `nonce++`, so the slot can be re-proposed), dropping the "because signatures verified" rationale.

script/deploy/safe/confirm-safe-tx.ts
gvladika: Both done — and I implemented the Tron follow-up rather than deferring it.

- Comment reworded in 808ebdb0: "Returns true only for the `executed` status (receipt success, or the Tron no-receipt path)".
- Tron status resolution implemented in 118e84c4: `tron-executor` now resolves the result **synchronously** via `getTransactionInfo` (`waitForConfirmation`) and maps `receipt.result` through an extracted, unit-tested `resolveTronExecutionStatus` (anything set and ≠ `'SUCCESS'` → `reverted`). `confirm-safe-tx` drops the special Tron branch and uses the same status check for both chains. `IChainExecutionResult.receipt` → normalized `status?: 'success' | 'reverted'`. Synchronous as you noted, since reconcile skips Tron — a deferred `submitted` row would never resolve.

script/deploy/safe/confirm-safe-tx.ts
gvladika: Agreed. Updated the helper doc and PR description so the rule reads as universal (a top-level revert always rolls back `nonce++`), with `safeTxGas=0` framed as *why* an inner-call failure surfaces as a top-level revert here — not as the reason the rule holds. The try/catch-wrapping caveat doesn't apply today: `evm-executor` calls `execTransaction` directly via `writeContract` and returns that call's own receipt.

script/deploy/safe/safe-utils.ts
gvladika: Done in 808ebdb0 — added `NONCE_CONSUMING_STATUSES`, referenced by both the predicate and reconcile's Mongo filter, so it's genuinely the single source of truth. Kept `safeTxStatusConsumedNonce` for the documented EVM-rollback rationale and unit-testability, as you suggested.

mirooon (approved):

mirooon (approved):

PR #1983, opened 2026-06-25
feat(EXSC-411): S3 — fee receivers & sweep distribution

gvladika:

[EXSC-411 — S3: Fee receivers & sweep distribution](https://linear.app/lifi-linear/issue/EXSC-411/s3-fee-receivers-and-sweep-distribution)

**Stacked on S2.** The base of this PR is the S2 branch (#1993, the fee accrual engine — now including S2b's performance fee and per-fee-type split, and the S5 pause module via the merged lineage): `dev-vault-wrapper ← S1 ← S2 ← S3`. The diff is only the S3 distribution layer; retarget to `dev-vault-wrapper` once S2 lands.

S2 splits every fee between LI.FI and the integrator **the moment it accrues**, into four per-recipient counters: `lifiFeeAssets`/`integratorFeeAssets` (idle asset, held in the wrapper, excluded from `totalAssets()`) and `lifiFeeShares`/`integratorFeeShares` (dilution shares minted to `address(this)`). S3 is the payout layer: it folds receiver configuration and a permissionless `distributeFees()` directly into the wrapper (S2 keeps all fee logic there, so no separate mixin), paying out exactly those tracked entitlements — **nothing is re-split at distribution**, so a later default-split change can never retroactively re-split already-earned fees.

- **Payout in native denomination.** `distributeFees()` first calls `_accrueFees()` (so pending management/performance fees crystallize even while deposits are paused — tested), then pays out each fee pool in its own token: the idle asset as the asset, the dilution shares as wrapper shares. **No adapter touch**, so the distribution can never be blocked by underlying illiquidity. LI.FI's booked parts go to the factory's **live** `lifiFeeRecipient()` (an integrator can never redirect it); the integrator's booked parts are fanned across its 1..N wallets (sanity-capped at 50) by their bps, with the **last wallet absorbing the integer-division remainder** so the pool zeroes exactly.
- **Single `distributeFees()`, never blockable.** Each integrator transfer uses OZ's non-reverting `trySafeTransfer`; a reverting wallet (e.g. a USDC-blacklisted address) has its share **redirected to LI.FI** in the same distribution instead of reverting. There is no separate LI.FI-only escape hatch — redirect makes one unnecessary. LI.FI's own transfer uses `safeTransfer` and is not caught (a blacklisted LI.FI recipient is LI.FI's own, factory-governed, concern).
- **Receivers are mandatory.** Set and validated at `initialize` via a typed `FeeReceiver[]` deploy param (1..N non-zero wallets, sanity-capped at 50, bps summing to exactly 100%); the `setIntegratorFeeReceivers` mutator (gated on `vaultWrapperAdmin`) re-validates, so an empty/unconfigured receiver set is structurally impossible — no griefing window where a permissionless distribution could divert unconfigured integrator fees.
- **Storage & CEI.** Receiver arrays are appended after S2's storage before a reduced `__gap` (`uint256[46]`); the subsystem is pre-release and the layout freezes before first deployment. `distributeFees()` zeroes all four counters before any transfer (CEI) and is `nonReentrant`. If a counter ever saturated (S2's uint128 clamp), the distribution pays the tracked amount — the excess stays in the wrapper rather than reverting anything.

Tests drive **real accrual** (deposit/withdraw + time warp through S2's engine), not seeded pools: payout equals the at-accrual tracked parts across both fee pools, 1..N-wallet fan-out with last-receiver remainder, the blacklist→redirect-to-LI.FI path, permissionless distribution, live-recipient reads, distribution-crystallizes-pending-management, **distribution accrues and pays out while deposits are paused** (the ticket behaviour previously deferred to S5, now in the base), zero-split books everything to LI.FI, and the init/mutator validation reverts.

## /pr-ready (local CodeRabbit) findings

- Re-run against the S2 base after merging S2 (incl. S2b's per-fee-type split/performance fee and the S2 review fixes) into this branch and porting the distribution to the per-recipient counters (see the merge commit for the port details): **no findings**. One pre-existing lint gate crossing: the merged wrapper reaches 16 state declarations, over solhint's `max-states-count` — scoped disable with justification (single beacon-proxy storage host; mixins deliberately rejected).


gvladika:
@coderabbitai review

gvladika:
@coderabbitai full review

src/VaultWrapper/interfaces/ILiFiVaultWrapper.sol
```diff
@@ -62,6 +57,32 @@ interface ILiFiVaultWrapper {
     /// @param by The owner (integrator) that toggled it.
     event PauseSet(bool paused, address indexed by);
 
+    /// @notice Emitted when the integrator's receiver set is configured.
+    /// @param receivers The integrator payout wallets.
+    /// @param bps The per-receiver basis points (sum to 100%).
+    event ReceiversSet(address[] receivers, uint16[] bps);
```
0xDEnYO: could consider to introduce a struct that ties a receiver address to a bps value. That would make the relationship between them clearer and also we dont need to check if arrays have same length and can remove `ReceiversLengthMismatch`

src/VaultWrapper/interfaces/ILiFiVaultWrapper.sol
```diff
@@ -62,6 +57,32 @@ interface ILiFiVaultWrapper {
...
+
+    /// @notice Emitted once per non-empty reservoir distributed by `sweep`.
+    /// @param token The reservoir token (the vault asset, or this wrapper's shares).
```
0xDEnYO: that's not clear to me....the vault asset (e.g. USDC), or this wrapper's shares (e.g. vaultUSDC)....arent those two different things? How can it be treated as the same here?
And is "reservoir" a common term for it?
So far we used things like underlying asset, vault asset, etc.

src/VaultWrapper/interfaces/ILiFiVaultWrapper.sol
```diff
@@ -62,6 +57,32 @@ interface ILiFiVaultWrapper {
...
+    /// @param lifiAmount Amount delivered to the LI.FI recipient (LI.FI's split + any redirected).
+    /// @param integratorAmount Amount delivered across the integrator wallets.
+    event ReservoirSwept(
+        address indexed token,
+        uint256 lifiAmount,
+        uint256 integratorAmount
+    );
+
+    /// @notice Emitted when an integrator payout fails (e.g. a blacklisted wallet) and the
+    ///         amount is redirected to the LI.FI recipient instead of reverting the sweep.
```
0xDEnYO: Sounds like a legal red-flag for me. 
All of a sudden we become the custodian for sanctioned (and frozen) funds. I am not sure if this is something legal would want or agree to. This must be signed off by legal team explicitly

src/VaultWrapper/interfaces/ILiFiVaultWrapper.sol
0xDEnYO: Image

you already have it, just not yet using it everywhere

src/VaultWrapper/LiFiVaultWrapper.sol
```diff
@@ -57,6 +58,15 @@ contract LiFiVaultWrapper is
...
     using MetadataReaderLib for address;
+    using Math for uint256;
+
+    /// Constants ///
+
+    /// @notice Maximum number of integrator receiver wallets. This is only a sanity cap to
+    ///         bound the permissionless `sweep` fan-out loop and prevent griefing (an oversized
+    ///         receiver set could otherwise make `sweep` run out of gas and lock distribution),
+    ///         not a product limit.
+    uint256 internal constant MAX_FEE_RECEIVERS = 50;
```
0xDEnYO, line 69: shouldnt this be 5 according to our requirements?

src/VaultWrapper/LiFiVaultWrapper.sol
```diff
@@ -107,6 +117,11 @@ contract LiFiVaultWrapper is
     ///         mints shares.
     uint192 public perfHighWaterMarkPps;
 
+    /// @notice Integrator payout wallets (1..50) with their bps split, each packed into one
+    ///         slot (address + uint16). Set at `initialize` and mutable by the integrator;
+    ///         always non-empty after deploy. Bps sum to 100%.
+    Receiver[] public integratorReceivers;
```
0xDEnYO: would recommend to add the word "fee" in the variable name as "receiver" is a similar yet differently used term in this contract

src/VaultWrapper/LiFiVaultWrapper.sol
```diff
@@ -141,8 +157,9 @@ contract LiFiVaultWrapper is
             _vaultWrapperAdmin == address(0)
         ) revert ZeroAddress();
         for (uint256 i; i < 4; ++i) {
```
0xDEnYO: recommend to use our gas-efficient pattern for increasing loop counters

also, instead of hardcoding a 4 I would suggest to use a constant (e.g. `MaxIntegratorFeeReceivers`) with value 5 and then use this value - 1 here

src/VaultWrapper/LiFiVaultWrapper.sol
```diff
@@ -548,6 +559,69 @@ contract LiFiVaultWrapper is
...
+    ///      while deposits are paused, when the deposit/mint accrual cannot run), then pays
+    ///      out the four per-recipient counters booked at accrual — the LI.FI/integrator
+    ///      split already happened when each fee accrued, so nothing is re-split here.
+    ///      LI.FI's parts go to the factory's live `lifiFeeRecipient`; the integrator's
+    ///      parts are fanned across its wallets by bps (last absorbs the rounding
+    ///      remainder). CEI: all four counters are zeroed before any transfer, and the call
+    ///      is `nonReentrant`. A failing integrator transfer (e.g. a blacklisted wallet) is
+    ///      redirected to LI.FI rather than reverting the sweep, so the integrator can
+    ///      never block it. No-op when every counter is empty.
+    function sweep() external nonReentrant {
```
0xDEnYO: Would personally prefer to call such an external function more explicitly according to what it does, e.g. "withdrawIntegratorFees" or so

src/VaultWrapper/LiFiVaultWrapperTypes.sol
```diff
@@ -49,4 +49,19 @@ struct DeployParams {
     FeeConfig fees; // Per-fee-type rates (0 = disabled); validated against bounds/caps.
     uint16[4] integratorShareBps; // Integrator's fee share (bps) per FeeType (index = ordinal); type(uint16).max = factory default, else must be < 100%.
     bytes initData; // Opaque wrapper-side config forwarded to the instance's initialize.
+    IntegratorReceivers receivers; // Integrator payout wallets + their bps split; validated on the instance at initialize.
+}
+
+/// @notice Integrator payout wallets and their basis-point split, validated together on the
+///         wrapper instance (1..50 non-zero wallets, bps summing to exactly 10_000 = 100%).
+struct IntegratorReceivers {
```
0xDEnYO: I would recommend to remove this struct entirely and use Receiver[] instead

src/VaultWrapper/LiFiVaultWrapperTypes.sol
```diff
@@ -49,4 +49,19 @@ struct DeployParams {
...
+    address[] wallets; // Payout wallets (1..50, non-zero).
+    uint16[] bps; // Per-wallet bps summing to 10_000 (100%); parallel to wallets.
+}
+
+/// @notice Storage form of a single integrator payout wallet: the wallet and its bps share
+///         packed into one slot (address + uint16). The set sums to 10_000 (100%).
+struct Receiver {
```
0xDEnYO: I would recommend to name this FeeReceiver

A receiver is just wallet. We pack it as a struct cause we want to tie it to a fee bps value. So it makes sense (at least to me) to call it a FeeReceiver (or so)

src/VaultWrapper/interfaces/ILiFiVaultWrapper.sol
gvladika: Agree that terminology is confusing - instead of `reservoir` let's use `feePool`. The concept is as following:
- every time deposit/withdrawal fee is triggered  we calculate fees in asset (USDC) and increase `lifiFeeAssets` and `integratorFeeAssets`
- every time management/performance fee is triggered we calculate fees in shares (vaultUSDC) and increase `lifiFeeShares` and `integratorFeeShares`
- periodically we need to pay out / distribute those fees to end fee receivers (function `sweep` does it, to be renamed `distributeFees`)
- 'reservoir' is just short name for fee pool in assets (tracked by `lifiFeeAssets` and `integratorFeeAssets`) and fee pool in shares (tracked by `lifiFeeShares` and `integratorFeeShares`)
- so what `sweep` does is distributing the 2 "reservoirs" (transfering assets and shares to fee recipients)

Hopefully after renaming the terminology and flow will be clearer.

src/VaultWrapper/interfaces/ILiFiVaultWrapper.sol
gvladika: Good point, let me raise this again with Alessandro

src/VaultWrapper/LiFiVaultWrapper.sol
gvladika: Requirements said 5 in some places, and in some N wallets. Alessandro confirmed we can use N wallets (https://lifi-protocol.slack.com/archives/C0B9Q24JQM9/p1783603846308439?thread_ts=1783421350.972529&cid=C0B9Q24JQM9). '50' as upper bound is just a sanity check, to prevent malicious integrator adding hundreds of wallets and thus preventing distribution of fees to both himself and lifi (every distribution attempt would run out of gas)

src/VaultWrapper/LiFiVaultWrapper.sol
gvladika: Re loop counters, check here: https://github.com/lifinance/contracts/pull/1993#discussion_r3558817828

src/VaultWrapper/interfaces/ILiFiVaultWrapper.sol
gvladika: Done — collapsed the parallel `address[] wallets` / `uint16[] bps` into a single `FeeReceiver[]` at the boundary (initialize, setter, DeployParams, the event), which removed `ReceiversLengthMismatch` entirely since the arrays can no longer diverge.

https://github.com/lifinance/contracts/commit/83a07d103214a4ab8c61da2908ea26a740dcfef7

src/VaultWrapper/LiFiVaultWrapperTypes.sol
gvladika: Done — removed the `IntegratorReceivers` struct; `DeployParams.receivers` and the whole boundary are now `FeeReceiver[]`.

https://github.com/lifinance/contracts/commit/83a07d103214a4ab8c61da2908ea26a740dcfef7

src/VaultWrapper/LiFiVaultWrapperTypes.sol
gvladika: Done — renamed `Receiver` → `FeeReceiver`.

https://github.com/lifinance/contracts/commit/adec4c9dcd10c955cae6c286cc3decc996172524

src/VaultWrapper/LiFiVaultWrapper.sol
gvladika: Done — `integratorReceivers` → `integratorFeeReceivers` (and the mutator `setIntegratorReceivers` → `setIntegratorFeeReceivers` for consistency with the getter).

https://github.com/lifinance/contracts/commit/ef3557aeeebf3ccc2eb224fb7207ecf7f0ed3602

src/VaultWrapper/interfaces/ILiFiVaultWrapper.sol
gvladika: Renamed `reservoir` → `feePool` throughout, and `sweep` → `distributeFees`. 

https://github.com/lifinance/contracts/commit/37eb994e8d96050c1ad2ae47f93e098c1982331f

src/VaultWrapper/LiFiVaultWrapper.sol
gvladika: Renamed `sweep()` → `distributeFees()`. Kept it fee-scoped rather than `withdrawIntegratorFees` because it pays out **both** LI.FI's and the integrator's accrued entitlements (across the asset and share pools), so an integrator-only name would misdescribe it.

https://github.com/lifinance/contracts/commit/37eb994e8d96050c1ad2ae47f93e098c1982331f

src/VaultWrapper/interfaces/ILiFiVaultWrapper.sol
gvladika: No code change — leaving the redirect-to-LI.FI behaviour as-is pending explicit legal sign-off (raising with Alessandro). If legal isn't comfortable with LI.FI receiving a blacklisted wallet's share, the alternative is to re-book the failed portion in the wrapper instead of routing it to LI.FI (distribution still never reverts); happy to implement that if we go that route.

src/VaultWrapper/LiFiVaultWrapper.sol
gvladika: Now using `FEE_TYPE_COUNT` instead of hardcoded 4

gvladika:
Thanks — responses to the cycle-3 review, item by item.

**1. [High] `MAX_FEE_RECEIVERS = 50` — intentional.**
50 is a gas-safety sanity cap on the permissionless `distributeFees` fan-out loop, **not** a product limit — this is already documented inline at `LiFiVaultWrapper.sol:65-68`. Alessandro approved N wallets (Slack, 2026-07-10). No code change; the product-limit decision is tracked on the Linear ticket, not enforced on-chain.

**2. [Medium] Redirect of failed transfers to `lifiFeeRecipient` — intentional, pending legal.**
The non-reverting `trySafeTransfer` + redirect-to-LI.FI design is deliberate: it guarantees one hostile/blacklisted integrator wallet can never block the whole distribution. We're keeping it as-is pending explicit legal/compliance sign-off; if legal isn't comfortable, the fallback is a dedicated escrow target or a silent skip. This stays open until legal responds — no code change now.

**3. [Medium] Receiver-rotation semantics — done.**
NatSpec added to `setIntegratorFeeReceivers` documenting that accrued-but-undistributed fees follow the *new* receiver set (single-total accounting, no per-wallet bookkeeping), plus `test_SetReceiversThenDistributeUsesNewReceivers`.
→ 00d234d96

**4. [Low] 50-wallet acceptance test — done.**
`test_SetIntegratorFeeReceiversAccepts50Wallets` — 50 receivers summing to 10_000 bps, asserts success and exactly-50 storage.
→ 0ad36e5a6

**5. [Low] Zero-share skip test — done.**
`test_DistributeFeesSkipsZeroShareReceiver` — a 1-bps receiver on a tiny fee pool whose share floors to 0, asserting it's skipped and the last receiver absorbs the whole integrator total (no stranded funds).
→ 6888fce10

FIND-04 (requires-types) and FIND-06 (dedup `_defaultReceivers`) — FIND-06 is already resolved (no local `_defaultReceivers` remain; all suites use the shared `defaultReceivers()` helper). FIND-05 (Olympix) confirmed resolved in your last pass.

Full VaultWrapper suite: 232 passing.

src/VaultWrapper/LiFiVaultWrapper.sol
0xDEnYO: ok

0xDEnYO (approved):

PR #1992, opened 2026-06-26
feat(EXSC-409): core ERC-4626 vault wrapper (S1)

gvladika:
> **Re-opened for review.** The original S1 PR (#1971) was merged into `dev-vault-wrapper` prematurely via auto-merge before SC-team review. That merge has been reverted (#1991), and this PR re-opens the identical S1 changes (same branch, tip `1d2392653`) so it can be reviewed properly before landing. No code changes vs. #1971.

[EXSC-409 — S1: Core ERC-4626 + immutable args](https://linear.app/lifi-linear/issue/EXSC-409/s1-core-erc-4626-immutable-args-cwia)

**Stacked PR.** Base is `feature/exsc-544-…` (#1969), which is itself stacked on the S8 factory branch. Review the incremental diff against the base branch, and do not merge before its parents land.

S1 is the base vault wrapper implementation that sits behind the beacon the factory (#1969's parent) already deploys.

- **OpenZeppelin `ERC4626Upgradeable` base (v5.6.1), beacon implementation, initialize-based identity — not CWIA.** Despite the ticket title, the merged factory deploys `LibClone` ERC-1967 beacon proxies and calls `initialize` write-once (see `[CONV:VW-BEACON]`). `LiFiVaultWrapper` is built on OZ's upgradeable ERC-4626 so the asset, decimals, and share metadata live in proxy storage (a beacon proxy cannot carry constructor-set immutables per instance), configured via the existing `ILiFiVaultWrapper.initialize` signature under OZ's `initializer` guard. CWIA bytecode-immutable identity stays deferred. OZ over Solady because a source-verified survey of major vault protocols found no production use of Solady's ERC-4626 and no documented standalone audit of it, while OZ's base has broad precedent (MetaMorpho, Gearbox, Ethena) — and Kiln DeFi, the closest comparable (integrator-facing beacon-proxy vaults with delegatecall connectors + per-vault fees), is built on `ERC4626Upgradeable`. The VaultWrapper subsystem runs on **OZ v5.6.1** while the Diamond stays on the vendored v4.9.2 (see the dependency note below).
- **Delegatecall yield adapter.** `IYieldAdapter` is extended with `deposit`/`withdraw` (delegatecall-only, so they run in the wrapper's context and the wrapper holds the yield-source position) and `totalAssets(underlying, holder)` (plain view call). The call-type split is the security contract — delegatecalled methods must be stateless — and is documented in the interface NatSpec. Adapter routing lives in the `_deposit`/`_withdraw` overrides (OZ has no after/before hooks). This keeps "new yield source = new adapter, no wrapper/beacon change" (`[CONV:VW-ADAPTERS]`).
- **Safety:** core `ReentrancyGuard` on `deposit/mint/withdraw/redeem` (v5's guard is proxy-safe — namespaced storage, no initializer); OZ virtual-share offset for inflation-attack mitigation; adapter revert data bubbled verbatim.
- **Factory reference captured** write-once in `initialize` (`factory = msg.sender`) to unblock the S5 global circuit breaker read.
- **Per-instance share metadata.** `name`/`symbol` derive from the underlying asset symbol (`LI.FI Earn USDC` / `lfUSDC`) via Solady `MetadataReaderLib` (a utility — handles `string`/`bytes32`/missing), falling back to `VW` when the asset exposes no symbol, so instances are distinguishable in wallets and explorers.
- **Mock removed.** The temporary `MockVaultWrapper` is deleted; the factory, beacon, and timelock tests now exercise the real `LiFiVaultWrapper`, and only the underlying yield source is mocked (Solmate `MockERC4626`).
- **Fees/access/pause are scaffolded as no-op seams (S2/S4/S5).** `_beforeOperation()` (pause → access → fee accrual) is wired into the four entrypoints, the `preview*` functions route through `_entryFee`/`_exitFee`, and the deposit/withdraw hooks skim/route fees via `_routeFee`. Every stub is a no-op today (fees return 0, guards pass), so behaviour is unchanged; the bodies land in the fee/access/pause tickets. `initData` and the fee config remain stored opaquely.
- **Hardened against direct/invalid initialization.** The implementation is locked with `_disableInitializers()` (only beacon proxies initialize); `initialize` rejects zero asset/underlying/adapter/admin, `integratorShareBps > 10000`, and an adapter that resolves the underlying to a different asset than the one passed. These are defense-in-depth — redundant for the trusted factory flow, but cheap insurance on a custody contract, with `testRevert_` coverage for each path.

**Deferred CodeRabbit finding (final local re-run is fully clean):** earlier `/pr-ready` runs flagged that the wrapper stores/exposes fee config and `initData` while fee, pause, and access enforcement are still no-op seams — so a vault could appear fee-enabled/permissioned without runtime enforcement. Deferred intentionally: storing config ahead of enforcement is the **merged factory's** deliberate staged design (the factory validates fees against bounds/caps and passes them through), so rejecting it in the wrapper would break the factory's plumbing and its fee-enabled deploy tests. The subsystem is unreleased and will not be production-deployed before the enforcement modules (S2 fees / S4 access / S5 pause) land — which is when these guards come off the seams.

**New dependencies (OZ v5 for the VaultWrapper only):** the subsystem builds on **OpenZeppelin v5.6.1** while the Diamond stays on the vendored **v4.9.2** core. Because v5's `-upgradeable` package imports the non-upgradeable core as a peer, this vendors a second core submodule `lib/openzeppelin-contracts-v5` (v5.6.1) and bumps `lib/openzeppelin-contracts-upgradeable` to v5.6.1. Version routing is by **path-scoped remappings**: `@openzeppelin/contracts/` resolves to v5 for `src/VaultWrapper/`, `test/solidity/VaultWrapper/`, `script/deploy/vaultWrapper/`, and the upgradeable lib itself; the global stays v4.9.2, so the Diamond is unaffected (`[CONV:VW-OZ-VERSION]`). v5 specifics absorbed: core `ReentrancyGuard`/`IERC20`/`SafeERC20`/`Initializable`, two-arg `UpgradeableBeacon(impl, owner)`, custom-error reverts, and `_initErc4626Metadata` extracted to avoid a v5 stack-too-deep (no repo-wide `via_ir`).

**Pending security triage (gates Ready/merge):** Olympix flags 4 design-inherent alerts on `LiFiVaultWrapper.sol` — `reentrancy-events` ×2 (events after the guarded adapter interaction) and `arbitrary-delegatecall` + `call-without-gas-budget` on the delegatecall adapter route. These are the wrapper's intended design (governance-approved adapter, `nonReentrant` + CEI) and predate the v5 change; they must be dismissed-with-justification (or addressed) in code scanning before the `securityAlertsReview` gate will let this PR stay Ready for Review.

**Divergences from the `wrapper-research` flow doc** (which predates the merged factory) that are intentional: delegatecall adapter vs the doc's call-based/no-adapter; `namespace`+`adapter` salt vs `integrator`+`chainLockId` (chainLockId dropped); single `integratorShareBps` vs per-fee-type splits; opaque `initData` vs explicit init fields.


src/VaultWrapper/LiFiVaultWrapper.sol
```diff
@@ -0,0 +1,422 @@
...
+        uint256 owed = _assets + fee;
+        uint256 withdrawn = _routeThroughAdapter(
+            abi.encodeCall(
+                IYieldAdapter.withdraw,
+                (assetToken, underlying, owed)
+            )
+        );
+        if (withdrawn < owed) revert AdapterWithdrawShortfall(owed, withdrawn);
+        _routeFee(fee);
+        SafeERC20.safeTransfer(IERC20(assetToken), _to, _assets);
```
mirooon, line 387: The `enforce-libasset-routing` check is currently failing on this line. That check scans `src/` for direct `.safeTransfer(` calls and only skips files on its allowlist. Here the subsystem uses OZ `SafeERC20` on purpose (it's standalone OZ v5, not a Diamond contract), so this looks like an intended exception rather than a real issue — but the build stays red until it's handled. Might be worth giving the VaultWrapper folder a carve-out in that workflow, similar to how `remappings.txt` already scopes OZ v5. (Side note: the adapter's `forceApprove` is the same choice but isn't matched by the pattern, so a folder-level exception would keep it consistent.)

src/VaultWrapper/LiFiVaultWrapper.sol
```diff
@@ -0,0 +1,422 @@
...
+///      per instance). This contract DOES custody funds: it holds the yield-source position on
+///      behalf of depositors and transiently holds the asset while routing a deposit or
+///      withdrawal. Identity (`underlying`/`adapter`/`owner`/`factory`) and the fee
+///      configuration are set write-once in `initialize`. The per-vault admin role is OZ's
+///      two-step `owner` (`transferOwnership`/`acceptOwnership`); renouncing it is disabled.
+///      Fee, access, and pause logic are
+///      scaffolded as no-op seams (`_accrueFees`, `_entryFee`/`_exitFee`, `_routeFee`,
+///      `_requireNotPaused`, `_checkAccess`) wired into the entrypoints and the deposit/withdraw
+///      flow; their bodies land in follow-up tickets. Inflation-attack protection relies on the
+///      ERC-4626 virtual-share offset.
```
mirooon, line 33: Small thought here: `_decimalsOffset()` is not overridden, so the vault runs at the default offset `0`. That already makes the classic attack unprofitable, so this is **not a bug**. But because `totalAssets()` reads a live balance from the underlying vault (line 193), a first depositor could still be harmed in a fresh/empty vault — even if the attacker loses money doing it. For a fund-holding vault it might be worth a little extra protection. Not blocking; details and what comparable protocols do are below.

<details>
<summary>Full analysis + what other protocols actually do (research)</summary>

Two things worth keeping separate:

**1. "Unprofitable for the attacker" ≠ "the victim can't lose."** OZ's docs, the `ERC4626.sol` NatSpec, and their security blog all say virtual shares + assets make the attack unprofitable at offset 0 ("the attacker's loss is at least equal to the user's deposit"), and the one public counterexample (OZ issue #5223) was retracted by its own author. But OZ also concedes offset 0 "does not fully prevent the attack" — a griefer willing to *lose* money can still round a victim's deposit down. Raising the offset only turns "unprofitable" into "orders-of-magnitude catastrophic" — it's hardening, not a fix for a broken default. *(oz docs 5.x/erc4626, ERC4626.sol NatSpec, PR #3979)*

**2. The scary audited findings were against wrappers with NO virtual shares — which tempers this for us.** The claims that Sherlock Burve #219 (HIGH) and Napier #125 prove OZ's offset-0 defense is insufficient don't hold up: those wrappers used raw `mulDiv` with neither virtual shares nor internal accounting — that's why victims lost funds. They show the *vector*, not that OZ virtual shares (which this wrapper does have) are exploitable for profit. Matches the math: at offset 0 the attacker nets ≈ −V/2 even for this donatable wrapper.

So the override is **defense-in-depth, not a fix for an exploitable bug.** The relevant point for this design: because `totalAssets()` is a live external read (`underlyingVault.convertToAssets(balanceOf(wrapper))`, line 193), internal accounting is off the table (it would break yield reflection, U19), and a higher offset hardens the share math but does **not** close the empty-vault first-depositor window. The thing that does is a **non-empty-vault invariant** (a seed / dead-share at deploy).

**What real protocols do (verified):**

| Protocol | Approach | Detail |
|---|---|---|
| OZ ERC4626 default | virtual shares, offset 0 | Baseline; "unprofitable" but "not fully prevented". |
| **MetaMorpho** (closest comparable) | dynamic offset + documented seed deposit, no dead shares | `DECIMALS_OFFSET = max(0, 18 - assetDecimals)` → 12 for USDC (6-dec), 0 for 18-dec. `deposit` NatSpec: *"For tokens with 18 decimals, the protection … is low. Vault deployers should make an initial deposit of a non-trivial amount … or depositors should check that the share price does not exceed a certain limit."* Vault V2 later added a dead-deposit mechanism. |
| Uniswap V2 | dead shares | Burns `MINIMUM_LIQUIDITY` (1000) to `address(0)` on first mint. |
| Napier / Burve auditors (Sherlock) | raise offset **or** seed/dead-shares **or** internal accounting | Napier #125: *"A MIN_LIQUIDITY amount of shares needs to exist within the vault … Or, just use a `_decimalsOffset > 0`."* #94: *"Track the total assets internally"* / *"Mint … shares to zero address (similar to Uniswap V2)."* |

*Coverage gap: MetaMorpho verified in depth; exact values for Yearn V3, Gearbox, Euler v2, Aave's static-aToken wrapper, Spark, Sommelier, Superform did not survive verification, so not asserted here. (Lido wstETH isn't ERC-4626.)*

**One option for reference (mirrors MetaMorpho) — two cheap measures, skip internal accounting:**

1. Override `_decimalsOffset()` with MetaMorpho's normalization:
```solidity
// stored once in initialize (beacon proxy → cannot use immutable); consumes one slot from __gap
uint8 private _sharesOffset;

function _initErc4626Metadata(address _asset) private {
    // ...
    __ERC4626_init(IERC20(_asset));
    // MetaMorpho parity: 18 - assetDecimals, floored at 0 (offset 12 for USDC, 0 for 18-dec)
    uint8 dec = IERC20Metadata(_asset).decimals();
    _sharesOffset = dec < 18 ? 18 - dec : 0;
}

function _decimalsOffset() internal view override returns (uint8) {
    return _sharesOffset;
}
```
2. Enforce a non-trivial seed / dead-share at deploy so the vault is never empty — this is the part that actually addresses the live-read donation vector, and the factory already deploys + initializes atomically, so it's the natural place. This is a **product decision**, not something to hardcode blindly.

Don't switch to internal-asset accounting — correctly tracking a yield-bearing external position defeats the purpose of a wrapper.

*Trade-offs: share decimals = asset decimals + offset; `max(0, 18-dec)` keeps everything at 18-decimal-equivalent (the aggregator norm), and the max-deposit headroom loss is negligible against `uint256`.*
</details>

mirooon (commented):
left some comments

src/VaultWrapper/LiFiVaultWrapper.sol
```diff
@@ -0,0 +1,422 @@
...
+        return super.previewDeposit(assets - _entryFee(assets));
+    }
+
+    /// @inheritdoc ERC4626Upgradeable
+    function previewMint(
+        uint256 shares
+    ) public view override returns (uint256) {
+        uint256 assets = super.previewMint(shares);
+
+        return assets + _entryFee(assets);
```
mirooon, line 320: A note for when the fee logic is filled in later (S2): `previewMint` takes the fee on the **net** amount here, while `_deposit` (line 309) takes `_entryFee` on the **gross** amount. With a non-zero rate those bases don't match, so on the `mint` path the quoted fee and the invested amount can drift apart (the `deposit` path is fine; `withdraw`/`redeem` have the mirror of this). Harmless today because the fee returns `0`.

Might be worth: (1) keeping deposit/mint (and withdraw/redeem) consistent when the fee is real — OZ's `ERC4626Fees` splits this into `_feeOnRaw`/`_feeOnTotal` for exactly this reason; (2) sharpening the existing `TODO(fees)` on `_entryFee`/`_exitFee` to name that split so the next implementer sees it; and (3) a round-trip test with a non-zero rate. Worth capturing in the S2 fee ticket too, so it isn't lost when this thread resolves.

src/VaultWrapper/LiFiVaultWrapper.sol
```diff
@@ -0,0 +1,422 @@
...
+    /// Events ///
+
+    /// @notice Emitted once when the instance is configured.
+    /// @param asset The ERC20 asset the vault is denominated in.
+    /// @param underlying The yield source the wrapper deposits into.
+    /// @param adapter The yield adapter the wrapper routes through.
+    /// @param vaultWrapperAdmin The per-vault controller granted the instance admin role.
+    /// @param factory The factory that deployed and initialized the instance.
+    /// @param integratorShareBps The integrator's fee share (bps) snapshotted at deploy.
+    event Initialized(
```
mirooon: Tiny thing: this event is named `Initialized`, which is also the name of the event OpenZeppelin's `Initializable` emits (with a different signature). Both end up in the ABI and both fire during `initialize`, which can be a bit confusing for indexers and log readers. A more specific name (for example `VaultWrapperConfigured`) might read clearer. Just a small suggestion.

src/VaultWrapper/LiFiVaultWrapper.sol
```diff
@@ -0,0 +1,422 @@
...
+    /// @return True if the fee type is enabled.
+    function feeEnabled(uint8 _feeType) external view returns (bool) {
+        if (_feeType > 3) revert InvalidFeeType(_feeType);
+        return _feeConfig.enabled[_feeType];
+    }
+
+    /// ERC-4626 entrypoints (reentrancy-guarded) ///
+
+    /// @inheritdoc ERC4626Upgradeable
+    function deposit(
```
mirooon, line 235: Minor style note: the repo usually prefixes parameters with an underscore (`_amount`, etc.), but these overrides keep OpenZeppelin's names (`assets`, `receiver`, `shares`, `owner`). That's reasonable since they match the `@inheritdoc` signatures, and the custom internals (like `_transferOut(_to, _assets)`) do follow the convention — so this is fine to keep. Flagging it only for consistency.

src/VaultWrapper/interfaces/IYieldAdapter.sol
```diff
@@ -4,21 +4,72 @@ pragma solidity ^0.8.17;
...
+///      - `resolveAsset` and `totalAssets` are invoked as ordinary (static) calls and
+///        run in the adapter's own context; they take an explicit `_holder`/`_underlying`
+///        and MUST be free of side effects.
+///      - `deposit` and `withdraw` are invoked via `delegatecall` and therefore run in
+///        the wrapper's context: `address(this)`, token balances, and yield-source
+///        positions are the wrapper's. They MUST be stateless with respect to adapter
+///        storage (no reads or writes of adapter state) so a shared adapter cannot
+///        corrupt or be corrupted by the wrapper's storage layout; they may only act on
+///        their arguments and external calls.
+/// @custom:version 2.0.0
```
mirooon: Small version note: this interface is at `2.0.0`, but the VaultWrapper subsystem hasn't shipped yet, so there's no released `1.0.0` to bump from — a MAJOR bump normally signals a breaking change against something already deployed. It's also the only file in the subsystem not at `1.0.0` (the wrapper, factory, adapter, types and `ILiFiVaultWrapper` are all `1.0.0`). Might be cleaner to reset it to `1.0.0` while everything is still pre-release. Small consistency point.

src/VaultWrapper/LiFiVaultWrapper.sol
gvladika: Fixed in f61489ab9 — added a folder-level exemption (`ALLOWLIST_PREFIXES=("src/VaultWrapper/")`) to `enforceLibAssetRouting.yml` rather than listing files one by one, so the adapter's `forceApprove` is covered too and new subsystem files stay exempt without allowlist churn. The comment mirrors the OZ-v5 remapping rationale. Check is green now.

src/VaultWrapper/LiFiVaultWrapper.sol
gvladika: Agreed — this is defense-in-depth, not an exploitable-for-profit bug, and a higher offset alone does not close the empty-vault window (the live `totalAssets()` read is the vector). I captured your full analysis (MetaMorpho-parity dynamic `_decimalsOffset` + a mandatory seed/dead-share at deploy in the factory, plus the no-internal-accounting caveat) on the S6 ticket so it gets decided deliberately there rather than bolted onto S1: https://linear.app/lifi-linear/issue/EXSC-414 — thanks for the thorough writeup.

src/VaultWrapper/LiFiVaultWrapper.sol
gvladika: Good catch — and this is already resolved in S2, which routes both bases through `LibVaultWrapperMath.feeOnRaw`/`feeOnTotal` (the OZ `ERC4626Fees` raw/total split you mention), so deposit/mint and withdraw/redeem stay consistent under a non-zero rate. I will sharpen the S1 `TODO(fees)` to name the split and make sure S2 carries the non-zero-rate round-trip test.

src/VaultWrapper/LiFiVaultWrapper.sol
gvladika: Agreed, the clash with `Initializable.Initialized` is confusing for indexers/log readers. I will rename it to `VaultWrapperConfigured`. Since the event lives in the interface and is shared across S1–S5, I will do it as one cross-branch propagation pass (same as the Ownable2Step change) rather than only here.

src/VaultWrapper/LiFiVaultWrapper.sol
gvladika: Agreed — keeping OZ's `assets`/`receiver`/`shares`/`owner` names on the overrides for `@inheritdoc` parity while the custom internals keep the `_` convention. Leaving as-is.

src/VaultWrapper/interfaces/IYieldAdapter.sol
gvladika: Done in 0ea78eddb — reset to `1.0.0`. Pre-release, so there was no shipped `1.0.0` to MAJOR-bump from, and it now matches the rest of the subsystem.

PR #1993, opened 2026-06-29
feat(EXSC-410, EXSC-556): vault wrapper fee engine — 4 fee types, per-fee split, HWM performance fee (S2)

gvladika:

[EXSC-410](https://linear.app/lifi-linear/issue/EXSC-410/s2-fee-engine-4-types-accrual-caps-per-fee-split) — S2, the vault-wrapper fee engine — and [EXSC-556](https://linear.app/lifi-linear/issue/EXSC-556/s2b-performance-fee-hwm-dilution-holding-period-accrual) — S2b, the performance fee, which was developed as stacked PR #2015 and merged into this branch.

Stacked on top of S1 (#1992, merged); base is `dev-vault-wrapper`, so the diff is the full fee engine: all four fee types, per-fee-type split, accrual, caps.

Fills in the no-op fee seams S1 left behind.

- **Management fee (dilution).** Linear pro-rata on gross AUM over a fixed 365-day year, crystallized as fee-shares minted to the wrapper on every deposit/withdraw. No external `harvest()`. The baseline (`lastMgmtAccrual`) always advances on accrual: elapsed time whose fee floors to zero shares is dropped (holder-favouring) rather than carried, so dormant or dust-scale time is never re-priced against later deposits or a changed rate.
- **Performance fee (HWM dilution).** Gated by a global price-per-share high-water mark: PPS scaled by 1e18, defined as `convertToAssets(1e18)` under the same OZ virtual-offset convention as all share pricing (`LibVaultWrapperMath.pricePerShare`), stored as `uint192 perfHighWaterMarkPps` packed into one slot with `lastMgmtAccrual`. In `_accrueFees`, management crystallizes first, then performance is charged on the **post-management** share price — performance is net of management. Gain = `supply * (pps - hwm) / 1e18` on real holder shares, fee = `feeOnRaw(gain, rateBps)` (ceil), converted through the same `dilutionShares` floor as the management fee. The watermark ratchets to the post-crystallization PPS **only when fee-shares were actually minted** — a gain whose fee floors to zero shares stays chargeable (unlike management's elapsed time, a price level can be re-measured later without overcharging anyone). Enabling performance from disabled via `setFeeRate` re-anchors the watermark **up-only** at the current PPS — gains made while disabled are never charged retroactively, and toggling the fee off/on at a trough can never lower the watermark to re-charge a recovery back to the old peak; disabling crystallizes pending gains first.
- **Deposit/withdrawal fees (asset-side).** Skimmed at transaction time; the fee is a percentage of the net amount (gross = net + fee), rounded up so it favours holders. Fee assets are held idle (excluded from `totalAssets()`), so the share price is unaffected by a fee event. Zero-amount forwards short-circuit on both sides (`_deposit` and `_transferOut`), so operations that preview as 0 execute without reverting on sources that reject zero-amount calls. Any overage a round-up source pays beyond the owed withdrawal amount is booked with the fee, so every idle asset stays attributed for payout instead of stranding as untracked dust.
- **Lazy preview == execution.** `_convertToShares`/`_convertToAssets` value against an effective supply that includes the fee-shares (management + performance, in accrual order) pending since the last accrual. `_accrueFees` and `_pendingFeeShares` share one computation, so the value a `preview*` returns immediately before a call equals what the caller gets, to the wei. This is the headline invariant, asserted across deposit/mint/withdraw/redeem after time warps and yield events.
- **Per-fee-type split, applied at accrual.** `DeployParams.integratorShareBps` and `initialize` take a `uint16[4]` indexed by `FeeType` ordinal. Each element resolves independently in the factory: `type(uint16).max` inherits the single factory default, an explicit value must be `< 100%`, and a self-serve deployer cannot set any element above the default. The split is applied **the moment a fee accrues**, into per-recipient counters — `lifiFeeShares`/`integratorFeeShares` and `lifiFeeAssets`/`integratorFeeAssets`, each pair two packed `uint128`s so one accrual updates a single slot. The split lives in one shared helper (`_splitFee`, 512-bit mulDiv) and the counters **saturate** at the uint128 max instead of reverting: the accrual runs on every operation including exits and the fee-disable path, so it must never be able to brick withdrawals. No split-at-distribution step remains: S3's sweep only pays out tracked entitlements, and a later default-split change can never retroactively re-split already-earned fees. Until that S3 payout module lands, accrued fees are intentionally unclaimable — the wrapper only books entitlements. `DilutionFeeAccrued`/`AssetFeeCharged` carry the integrator's part next to the total.
- **`setFeeRate`** is `vaultWrapperAdmin`-gated for all four types, reads the factory's bounds live (the immutable cap is enforced transitively), and accrues at the old rate before applying a change. The split setter and timelock-gating are S10.
- **Pure math in one library.** All fee math *and* the fee-adjusted share/asset conversions live in `LibVaultWrapperMath`, so the novel arithmetic has a single side-effect-free unit/fuzz surface; the contract holds state, minting, and routing only.
- **Events and errors** for the wrapper live in `ILiFiVaultWrapper` (mirroring how `ILiFiVaultWrapperFactory` hosts the factory's), not inline in the contract.

## Default decisions taken (documented for review)

1. **Split rounding**: the integrator's part rounds down; the ≤1 wei remainder per accrual goes to LI.FI.
2. **Factory default split stays a single value** applied to any sentinel element. Per-fee-type *defaults* were not added — no requirement for them yet; explicit per-type values cover the need at deploy time.
3. **Watermark anchor at `initialize` is computed pure** at the empty-vault price (supply and position are provably 0 on a fresh single-shot proxy) instead of reading `totalAssets()` through the adapter — an underlying whose empty-position query reverts must not brick deployment. Consequence: assets donated to the predicted address pre-deployment count as gain at the first accrual; charging a donation is harmless and disarms watermark-seeding.
4. **Management fee-shares participate in the performance gain measurement** (they exist at crystallization time). The effect is fee-on-fee dust that only shifts value between the two fee recipients, never from depositors.
5. **Global HWM as holding-period approximation** is carried from the ticket unchanged (a post-drawdown entrant rides free until the prior peak is reclaimed; product sign-off on the U15 rewording is tracked in the ticket, not here).
6. **Counters are `uint128` and saturate**: a fee balance ≥ 2^128 wei of any real asset is unreachable. If a pathological asset manufactures one anyway, the counters clamp at the max (under-attributing payout for the excess) instead of reverting — the accrual path runs on every exit, so a reverting counter would permanently lock depositor funds with no escape hatch (`setFeeRate` accrues first, so even disabling the fee could not recover).

## Tests

- Shared scaffolding lives in `VaultWrapperFeeTestBase` (beacon-proxy setup, factory stack, deposit/yield/loss helpers, crystallize dust), inherited by both fee suites.
- `LiFiVaultWrapperFees.t.sol`: management/deposit/withdrawal accrual, caps, preview==execution invariants (including an executed `mint()` with the deposit fee enabled), full exit via `withdraw(maxWithdraw(owner))` with the withdrawal fee, per-fee-type split applied at accrual, rounding remainder to LI.FI, counter saturation never bricks exits, adapter withdrawal overage booked, events carry the integrator part.
- `LiFiVaultWrapperPerformanceFee.t.sol` (15 tests): charge on gain above HWM with exact dilution math and split, never on net loss, deposits alone never charge, no double-charge across drawdown→recovery, HWM ratchet to post-crystallization PPS + idempotence, sub-threshold gain stays chargeable, preview==execution with pending perf, combined management+performance sequence (exact), management dilution alone never triggers perf, setFeeRate enable re-anchor / toggle-at-trough cannot lower the watermark / disable crystallize-first / bounds.
- Factory tests: per-element sentinel resolution, per-element `InvalidSplit` and above-default rejection.
- `LibVaultWrapperMath.t.sol`: unit + fuzz over all fee math; `pricePerShare == convertToAssets(1e18)` (fuzz), performance fee zero-inputs / at-or-below-watermark / known value / bounded-by-rate×gain (fuzz) / monotonic in yield (fuzz, comparing two above-watermark levels).
- Key behaviors mutation-verified (split indexing, HWM ratchet, mgmt→perf ordering, enable re-anchor, up-only re-anchor, counter saturation, overage booking, mint-path fee recomputation: each mutation breaks the matching tests; code restored).
- Full VaultWrapper suite: **213 tests, 0 failures**.

## /pr-ready (local CodeRabbit) findings

- Applied: deploy NatSpec wording ("below 100%", `LiFiVaultWrapperFactory.sol`) — own `pr-ready:` commit (`6907625ad`).
- Rejected (documented): "Preserve the existing beacon-proxy storage layout" — the flagged reordering would matter only for already-initialized proxies. **No wrapper instance or beacon is deployed anywhere**; the subsystem is pre-release and the storage layout is finalized before first deployment. Same rationale as the S1 review; `__gap` intentionally stays `uint256[50]` until the pre-deployment layout freeze.
- Review round 2 (2026-07-06): first commit re-reviewed clean by local CodeRabbit; the last two commits (`FeeConfig.enabled` removal, saturating-add hardening) were pushed with the local review explicitly skipped on the author's instruction (CLI rate limits). Cloud CodeRabbit on this PR remains the net.

## PR review round 2 (2026-07-06)

- Per-member NatSpec on the `FeeType` enum (charging semantics of each type).
- `_beforeOperation` indirection removed; entrypoints call `_checkAccess` + `_accrueFees` directly.
- Generic `fee` locals renamed (`depositFee`/`withdrawalFee`/`mgmtFeeAssets`/`perfFeeAssets`); `maxDeposit`/`maxMint` grouped with the preview/limit views.
- **`FeeConfig.enabled` dropped** — a fee type's rate is the single source of enablement (0 = disabled). `feeEnabled()` derives from the rate, `FeeConfigUpdated` lost its derived bool arg, factory validation skips zero rates, and `DisabledFeeMustBeZero` (plus its now-unrepresentable test) is gone. Enabling the performance fee from a zero rate now always re-anchors the watermark.
- `_saturatingAddUint128` hardened to compare-before-add: `dilutionShares` is only bounded by the `_mint` supply check, so the previous checked add could itself overflow uint256 on the accrual path the function must keep alive.
- Olympix code-scanning alerts triaged: 7 dismissed as false positives with justification comments (4× reentrancy-events on guarded/trusted external-call patterns, 2× unsafe-downcast on the range-guarded saturating cast, 1× uninitialized-state-variable on write-once receiver arrays).

## Code-review findings addressed (multi-agent review, 2026-07-03)

- **Fixed**: up-only HWM re-anchor in `setFeeRate` (the fee-receiving owner could previously reset the watermark to a trough by toggling the fee off/on); fee counters saturate instead of SafeCast-reverting (a reverting counter would brick exits with `setFeeRate` unable to recover); adapter withdrawal overage booked with the fee (was silently stranded); split logic deduped into `_splitFee`; test gaps closed (executed `mint()` with deposit fee, `withdraw(maxWithdraw)` full exit, meaningful monotonicity fuzz, shared test base).
- **Deferred to S3** (by design, documented above): the fee payout/claim path — accrued fees are unclaimable until the sweep module lands.
- **Rejected**: storage-layout preservation vs the base branch — pre-release, no proxy deployed anywhere; layout freezes before first deployment (same rationale as the S1 review).

## PR review round 3 (2026-07-10)

Addressed @0xDEnYO's review comments (commit `22c0527`):

- **`FEE_TYPE_COUNT` constant** replaces the hardcoded `uint16[4]` array sizes and fee-type loop bounds across the subsystem (types, wrapper, factory, both interfaces). A `type(FeeType).max`-derived array length is rejected by the compiler (*"Invalid array length, expected integer literal or constant expression"*), so the constant is a named literal documented as the single source for the member count.
- **`USE_DEFAULT_SPLIT` renamed to `DEFAULT_SPLIT_SENTINEL`** — describes the value, not its usage.
- **Over-length withdrawal NatSpec line reflowed** to the 100-char cap.
- Loop-counter `unchecked` suggestion **not applied**: solc ≥ 0.8.22 elides the redundant loop-counter overflow check automatically (not via-IR-gated); measured 0 gas difference on our 0.8.29 / no-`via_ir` config.
- Access-seam signature question **deferred to #2032** (EXSC-412), which replaces the `_checkAccess` seam entirely.
- Pause-modifier suggestion **not applied**: the pause guard has two behaviors (revert on `deposit`/`mint`, `return 0` on `maxDeposit`/`maxMint`), which a single revert-style modifier can't express.

Local CodeRabbit (round 3) came back clean on the five changed files. It also raised one **major** finding on pre-existing PR code — a theoretical overflow in `LibVaultWrapperMath.managementFeeAssets` (`uint256(_rateBps) * _elapsed` before `mulDiv`). **Rejected as a false positive**: the only caller derives `_elapsed = block.timestamp - lastMgmtAccrual` where `lastMgmtAccrual` is `uint64`, so `_elapsed < 2^64` and `_rateBps ≤ 65535 < 2^17`, giving a product `< 2^81` — nowhere near a uint256 overflow; the existing `>= _totalAssets` clamp handles the rest.


gvladika:
@coderabbitai review

src/VaultWrapper/LiFiVaultWrapperTypes.sol
```diff
@@ -39,7 +46,7 @@ struct DeployParams {
     address adapter; // Approved yield adapter; resolves the underlying's ERC20 asset.
     address underlying; // Protocol-specific yield source (e.g. an ERC-4626 vault).
     uint256 nonce; // Disambiguates instances sharing the same (namespace, adapter, underlying).
-    FeeConfig fees; // Per-fee-type rates and enabled flags; validated against bounds/caps.
-    uint16 integratorShareBps; // Integrator's fee share (bps); type(uint16).max = factory default, else must be <= 100%.
+    FeeConfig fees; // Per-fee-type rates (0 = disabled); validated against bounds/caps.
+    uint16[4] integratorShareBps; // Integrator's fee share (bps) per FeeType (index = ordinal); type(uint16).max = factory default, else must be < 100%.
```
0xDEnYO: would recommend to use a a constant (-1) instead of hardcoded 4

src/VaultWrapper/LiFiVaultWrapperFactory.sol
```diff
@@ -35,7 +35,7 @@ contract LiFiVaultWrapperFactory is
     uint16 internal constant CAP_WITHDRAWAL_BPS = 2000;
     /// @notice Integrator fee share (bps) applied when a deploy does not override it; 80% (LI.FI receives the remaining 20%).
     uint16 internal constant DEFAULT_INTEGRATOR_SHARE_BPS = 8000;
-    /// @notice Sentinel for DeployParams.integratorShareBps meaning "inherit the factory default".
+    /// @notice Sentinel for a DeployParams.integratorShareBps element meaning "inherit the factory default".
     uint16 internal constant USE_DEFAULT_SPLIT = type(uint16).max;
```
0xDEnYO: nitpick:

Are description and naming of this variable precise and correct?

USE.... suggests to me like I am expecting a boolean
SPLIT suggests to me I should expect two (or more) values (an actual split)

The description explains how it's used, not what it represents imho

src/VaultWrapper/LiFiVaultWrapper.sol
```diff
@@ -146,8 +140,10 @@ contract LiFiVaultWrapper is
             _adapter == address(0) ||
             _vaultWrapperAdmin == address(0)
         ) revert ZeroAddress();
-        if (_integratorShareBps >= 10_000)
-            revert InvalidIntegratorShareBps(_integratorShareBps);
+        for (uint256 i; i < 4; ++i) {
```
0xDEnYO: use gas-efficient loop counter increase

src/VaultWrapper/LiFiVaultWrapper.sol
```diff
@@ -237,7 +284,8 @@ contract LiFiVaultWrapper is
         address receiver
     ) public override nonReentrant returns (uint256) {
         if (depositsPaused()) revert DepositsPaused();
-        _beforeOperation();
+        _checkAccess(msg.sender);
```
0xDEnYO, line 287: The seam pins access to msg.sender, but is caller-gating actually the invariant we want? If the access mode means "only permissioned addresses hold shares", we need to check receiver on deposit/mint (and arguably the transfer hook), not the caller — and caller-gating breaks for anything routed through the diamond/zaps, where msg.sender is the router. Also: do we really want exits (withdraw/redeem) gated at all? A delisted user shouldn't be trapped. Suggest deciding the invariant before the access PR and adjusting the seam signature/NatSpec accordingly.

src/VaultWrapper/LiFiVaultWrapper.sol
```diff
@@ -333,7 +379,32 @@ contract LiFiVaultWrapper is
...
+            });
+    }
+
+    /// @inheritdoc ERC4626Upgradeable
+    /// @dev Reports 0 while any pause source is engaged so EIP-4626 consumers see the vault
+    ///      as closed to deposits and do not build deposits that would revert.
+    function maxDeposit(
+        address receiver
+    ) public view override returns (uint256) {
+        if (depositsPaused()) return 0;
```
0xDEnYO, line 396: could consider to create one or several modifiers, DepositsNotPaused etc.

src/VaultWrapper/LiFiVaultWrapper.sol
```diff
@@ -369,21 +449,32 @@ contract LiFiVaultWrapper is
...
+        uint256 owed = _assets + withdrawalFee;
         uint256 withdrawn = _routeThroughAdapter(
             abi.encodeCall(
                 IYieldAdapter.withdraw,
                 (assetToken, underlying, owed)
             )
         );
         if (withdrawn < owed) revert AdapterWithdrawShortfall(owed, withdrawn);
-        _routeFee(fee);
+        // A round-up source may pay more than owed; book the overage with the fee so
```
0xDEnYO: typo

src/VaultWrapper/LiFiVaultWrapperTypes.sol
gvladika: Done in 22c0527 — extracted a `FEE_TYPE_COUNT` constant in `LiFiVaultWrapperTypes.sol` and threaded it through every `uint16[4]` and fee-type loop across the subsystem (wrapper, factory, both interfaces).

One caveat worth recording: a `type(FeeType).max`-derived expression is **not** accepted as an array length — the compiler rejects it with *"Invalid array length, expected integer literal or constant expression"*. So `FEE_TYPE_COUNT` is a named literal with a comment tying it to the FeeType member count. It's now the single place to change if a fifth fee type is added.

src/VaultWrapper/LiFiVaultWrapperFactory.sol
gvladika: Agreed — renamed to `DEFAULT_SPLIT_SENTINEL` in 22c0527. It's a per-element sentinel meaning "this element inherits the factory default", so the new name describes the value rather than how it's consumed.

src/VaultWrapper/LiFiVaultWrapper.sol
gvladika: Checked this against our actual compiler config rather than assume. Since solc **0.8.22** the optimizer removes the redundant overflow check on loop counters automatically — Changelog: *"Remove redundant overflow checks of certain `for` loops when the counter variable cannot overflow."* — and it's not via-IR-gated.

We ship 0.8.29 (`optimizer = true`, no `via_ir`). I measured checked vs `unchecked { ++i }` over 100k iterations under exactly that config: **123 gas/iteration both ways, 0 difference**. So `unchecked` buys nothing on our compiler; I left the increments as-is to keep the loop bodies clean. Happy to apply it purely for house-style consistency if you'd prefer.

src/VaultWrapper/LiFiVaultWrapper.sol
gvladika: Agreed the `msg.sender` shape is wrong for a permissioned vault — receiver is what matters for inflows, exits shouldn't be trappable, and routed/zap flows make the caller the router.

This is being resolved in the access PR #2032 (EXSC-412), which replaces this seam entirely with a single pluggable gate — `_checkAccess(msg.sender)` doesn't survive there. Rather than churn the throwaway placeholder in this PR, let's settle the receiver-vs-caller and exit-gating decisions on #2032.

src/VaultWrapper/LiFiVaultWrapper.sol
gvladika: Considered it, but the pause check has two behaviors: `deposit`/`mint` revert (`DepositsPaused`), while `maxDeposit`/`maxMint` return 0 so EIP-4626 consumers see the vault as closed. A revert-style modifier only covers the first two and can't express the `return 0` path, so it would split pause handling across two idioms for a net saving of two one-line guards. Leaning toward keeping the explicit inline checks — open to it if you feel strongly.

src/VaultWrapper/LiFiVaultWrapper.sol
gvladika: Tracked it down — the word is `overage`, on this line. It's a real accounting term (an amount *in excess*, the opposite of a shortage), so not strictly a misspelling, but it's uncommon enough to read as one. Reworded it to `excess` at all three spots on the withdrawal path (this comment plus the surrounding NatSpec) in 9c650ae — the plainer word is clearer for the next reader/auditor. Thanks for the nudge.

0xDEnYO (approved):

PR #1996, opened 2026-06-30
feat(PaxosTransitFacet): Paxos Transit bridge facet (EXSC-547) [PaxosTransitFacet v1.0.0, IPaxosTransit v1.0.0]

0xDEnYO:

[EXSC-547](https://linear.app/lifi-linear/issue/EXSC-547/assessmentsc-paxos-transit-sc-feasibility)

Implements `PaxosTransitFacet` — a bridge facet for Paxos Transit (ETH ↔ Robinhood Chain via a Paxos-signed quote), plus 100%-covered unit tests, deploy scaffolding, docs, and a runnable local demo.

**Design (per Paxos' answers to our integration questionnaire):**

- **The Diamond is the payer.** Paxos confirmed `submitOrder`'s `msg.sender` always provides the offer asset, and the want asset goes to `quote.receiver` regardless of submitter. So the facet keeps our normal funds flow: deposit/swap into the Diamond → `maxApproveERC20` the station → `submitOrder`, which pulls the offer asset **from the Diamond**. `quote.receiver` (validated `== bridgeData.receiver`) directs the output to the end user. No `submitOrderWithPermit`/on-behalf call is needed — the `Quote` struct has no payer field, only `receiver`. Modeled on `GlacisFacet`.
- **ERC20-only offer asset** (`noNativeAsset`); `msg.value` carries the LayerZero messaging fee (`nativeFee`), forwarded as `submitOrder{value: nativeFee}`.
- **Locked rate, no slippage param** (Paxos rate is fixed by the signed quote). The swap variant requires swap output ≥ `quote.offerAmount`, refunds positive slippage, and bridges exactly `offerAmount`.
- **All user-owned value routes to an explicit `refundRecipient`** (review feedback from @gvladika): swap leftovers, positive slippage, **and excess source-side native** — including LZ fee overage, since the TransitStation forwards the whole `msg.value` to the LZ endpoint with `refundAddress = msg.sender`, so the endpoint refunds overage to the Diamond mid-call on essentially every tx (the backend must buffer against fee drift). `msg.sender` may be a relayer or the Permit2Proxy (where a native refund would strand as owner-recoverable dust), so nothing refunds to `msg.sender` anymore. Both entrypoints zero-check `refundRecipient` (`InvalidCallData`) — without the guard a zero recipient would revert only when fee drift actually left an excess (`LibAsset.transferNativeAsset` reverts on `address(0)`, it does not burn), i.e. data-dependently and late. Mirrors `SupersetFacet`'s `refundAddress`; codified as `[CONV:FACET-REFUNDS]` in `.agents/rules/102-facets.md`. Known accepted footgun: a `refundRecipient` that rejects native reverts the bridge (self-inflicted; same as Superset).
- **Quote consistency check** in `_startBridge`: `sendingAssetId`/`minAmount`/`receiver` must match the signed quote, and `distributorCode == "LIFI"` (`0x4c49464900…`) for attribution; else `InformationMismatch`.
- **`destEID`/`wantAsset` are trusted from the signed quote** (not cross-checked against `destinationChainId`) — funds always follow the signed quote, same trust model as `AcrossFacetV4`'s `outputAmount`. Documented in code + docs.

**Verified against the live deployment.** Paxos deployed + verified the `TransitStation` on 2026-06-29 at `0x49AAA987b1a7e9E4AE091dcD8332c39F322D7d28` on Ethereum (1) and Robinhood Chain (4663). I checked the verified source + on-chain state and confirmed: `submitOrder` pulls the offer asset from `msg.sender` (so the Diamond is the payer — no on-behalf/permit needed); its capability is **public** (`Authority.canCall` → true, the Diamond can call it); the on-chain ABI matches `IPaxosTransit` exactly (exercised directly — the test suite calls the deployed station on a mainnet fork); and no failure path refunds the caller. Real addresses are now in `config/paxosTransit.json`.

**Verification:** 39 tests, **100% facet coverage** (lines/statements/branches/functions), now run as **mainnet-fork tests against the real `TransitStation`** per `[CONV:FORK-FIRST]` — no mock. (The rule itself ships in this PR: `.agents/rules/400-solidity-tests.md` gains the fork-first section — mocks are a last resort behind gate-lifting on the fork.) The station's Paxos-signer gate is lifted on the fork (owner-pranked `setQuoteSigner` to a test key; quotes EIP-712-signed in-test via the new `TestPaxosTransitBackendSig` helper against the station's live domain), the LZ fee comes from the station's own `quoteSend`, and the fork is pinned to a block where Paxos has the USDC→USDG route to EID 30416 configured (guarded in `setUp`). This adds real-behavior coverage a mock couldn't provide: protocolFee/integratorFee/net splits to the real Paxos recipients, replay guard (`SignatureAlreadyUsed` + `usedDigests`), `QuoteExpired`, `InvalidSigner`, `LZ_InsufficientFee` on underpaid fee, and LZ fee-overage refunds flowing back to `refundRecipient`. `MockTransitStation` is deleted; the demo (`demoPaxosTransit.ts`) now runs the same real-station flow end-to-end on an anvil mainnet fork (verified locally).

**Operational dependencies (Paxos / backend, not this facet):** the route `(destEID, offerAsset, wantAsset)` must be approved by Paxos (`setRouteApprovals`, routes go live ~2026-07-01); quotes must be backend-signed. **Out of scope:** on-chain deployment; backend integration.

Reviewed locally with an adversarial code-review pass (funds-safety / correctness) and a self-review pass. Local `/pr-ready` (CodeRabbit CLI) ran on the latest push: two findings were against an untracked local dev script that is not part of this PR; the one in-scope suggestion is deferred to reviewers below.

**Deferred local-CodeRabbit suggestion (explicit reviewer decision wanted):** harden `swapAndStartBridgeTokensViaPaxosTransit` with a native-balance-delta check (require the post-swap native balance gain to cover `nativeFee` before `_startBridge`), so the LZ fee could never be paid from pre-existing stray diamond native balance. Not applied pre-push: it is a logic change to the already-reviewed fee-funding design (`nativeFee` may be funded by `msg.value` or by the ERC20→native pre-swap's reserved output; the Diamond holds no native between txs by invariant, and `refundExcessNative` only refunds balance gained within the call), so it should be a deliberate reviewer call rather than a silent pre-push edit.

- [x] For new facets: I have checked all points from this list: https://www.notion.so/lifi/New-Facet-Contract-Checklist-157f0ff14ac78095a2b8f999d655622e


config/paxosTransit.json
```diff
@@ -0,0 +1,7 @@
+{
+  "_comment": "Paxos Transit station (source-chain) addresses, deployed + verified by Paxos 2026-06-29 (EXSC-547). offerAsset is ERC20-only; submitOrder carries msg.value for the LayerZero messaging fee. submitOrder is a PUBLIC capability (verified on-chain), so the Diamond can call it; routes must be approved by Paxos and quotes signed by their quoteSigner.",
+  "transitStation": {
+    "mainnet": "0x49AAA987b1a7e9E4AE091dcD8332c39F322D7d28",
+    "outlaw": "0x49AAA987b1a7e9E4AE091dcD8332c39F322D7d28"
```
gvladika: Change outlaw -> robinhood

src/Facets/PaxosTransitFacet.sol
```diff
@@ -0,0 +1,184 @@
...
+    /// @param _bridgeData The core information needed for bridging
+    /// @param _paxosData Data specific to Paxos Transit
+    function startBridgeTokensViaPaxosTransit(
+        ILiFi.BridgeData memory _bridgeData,
+        PaxosTransitData calldata _paxosData
+    )
+        external
+        payable
+        nonReentrant
+        refundExcessNative(payable(msg.sender))
```
gvladika: Issue: LZ fee refund is misrouted for the Permit2Proxy path — on essentially every tx

TransitStation charges the actual LZ fee at inclusion and reverts if underpaid, while the on-chain fee drifts up with gas between quote-time and inclusion. So the backend must always send a native buffer, or txs revert. That means a nonzero refund is the normal, every-transaction outcome — not something that only happens on a backend mistake or a rare fee spike. The refund path is exercised on virtually every bridge.

Why that matters. Both facet entrypoints refund excess native via `refundExcessNative(payable(msg.sender))`:
- Direct EOA / relayer call → fine, goes back to the payer.
- Via Permit2Proxy → msg.sender is the proxy. The refund lands in Permit2Proxy (caught by its receive()) but is never forwarded — no sweep after the diamond call. It strands as owner-recoverable dust, not returned to whoever funded the buffer.

Combined with the point above: for the Permit2Proxy path this stranding happens every time, not occasionally. It's a steady leak, not a corner case.

Fix. Refund native to the backend-supplied `refundRecipient` instead of msg.sender, on both paths — as `SupersetFacet` (the closest analog: LZ-messaged, ERC20 offer, native = LZ fee) already does with a single refundAddress for excess native + swap leftovers + positive slippage. Strictly more expressive than msg.sender (backend picks the correct payee), and closes the leak without touching shared periphery.

Required with the change:
1. Add a refundRecipient != address(0) guard to the non-swap path (only the swap path validates it today) — else excess native burns to address(0).
2. Accept the minor footgun: a refundRecipient that rejects native reverts the bridge (self-inflicted; Superset accepts the same).
3. Remove the now-inverted "Excess native stays with msg.sender — the caller funds the fee" comment in the swap path and the matching line in the PR description.

0xDEnYO: Addressed in 005857da9:

- Both entrypoints now use `refundExcessNative(payable(_paxosData.refundRecipient))` — nothing refunds to `msg.sender` anymore.
- Added the `refundRecipient != address(0)` guard to the non-swap path (`InvalidCallData`, consistent with the swap path). One correction on the rationale there: a zero recipient wouldn't burn — `LibAsset.transferNativeAsset` reverts on `address(0)`. The actual failure mode was a *data-dependent late revert* (tx success depending on whether LZ fee drift left an excess to refund); the guard makes it deterministic and early.
- Fixed the inverted swap-path comment; struct NatSpec + `docs/PaxosTransitFacet.md` now document the routing and the accepted native-rejecting-recipient footgun (mirrors `SupersetFacet.refundAddress`).
- Codified in `[CONV:FACET-REFUNDS]` (`.agents/rules/102-facets.md`); tests added: excess native → `refundRecipient` on both paths, zero `refundRecipient` reverts `InvalidCallData` on the non-swap path.

One factual nit on the mechanism: the TransitStation doesn't charge-and-refund itself — it forwards the whole `msg.value` to the LZ endpoint with `refundAddress = msg.sender`, so it's the endpoint that reverts on underpay / refunds the overage to the diamond mid-call. Same conclusion (a refund is the every-tx norm given the backend's buffer), different plumbing.

gvladika (dismissed):

gvladika:
Checklist checkbox (item #1) is now ticked. For items **#2** and **#3**, no code change will be applied — justification below.

### #2 — No zkSync deploy script (accepted, by design)

Paxos Transit is a source-chain-only integration: the facet calls `TransitStation.submitOrder` on the chain where the station is deployed, and Paxos has deployed the station only on Ethereum and Robinhood Chain. `config/paxosTransit.json` reflects exactly that — `transitStation` is configured for `mainnet` and `robinhood`, both EVM (`type: mainnet` in `networks.json`); no zkSync-family chain is present or planned.

A `script/deploy/zksync/DeployPaxosTransitFacet.zksync.s.sol` would be dead code: it could never resolve a `transitStation` address (the config has no zkSync key, and `allowToDeployWithZeroAddress` is `false`), so any zkSync deploy would revert `InvalidConfig()` in the constructor by construction. The zkSync counterpart convention applies to facets that are actually deployed to zkSync chains; adding an unusable script for a facet that never will be is exactly the kind of misleading dead code our conventions tell us to avoid.

This comment is the explicit record of that decision.

### #3 — `PaxosTransitData` field ordering (no `receiverAddress`; refund field stays last)

This facet has no user-supplied receiver field, so the `receiverAddress`-first mandate in `[CONV:FACET-REQS]` does not map to it. The destination receiver is carried inside the Paxos-signed `Quote` (`quote.receiver`) and is validated on-chain in `_startBridge` (`_bridgeData.receiver != quote.receiver → InformationMismatch`), on top of the backend EIP-712 signature over the whole quote. Receiver safety is therefore enforced by both mechanisms the rule asks for — the ordering convention exists to surface a receiver field that this struct legitimately does not have.

`refundRecipient` is a refund destination, not the bridge receiver. Renaming it to `receiverAddress` would be semantically wrong (it never receives the bridged funds — the end user does, on the destination chain) and would mislead integrators.

The convention file already treats this exact struct as canonical: `[CONV:FACET-REFUNDS]` in `.agents/rules/102-facets.md` cites `PaxosTransitFacet.sol` by name as the reference implementation for the `refundRecipient` pattern, and the native-fee-guard bullet cites it as well. Reordering the struct now would contradict the rule that points at it.

No change; keeping the struct as-is. Happy to fold an explicit opaque-calldata / quote-bound-receiver carve-out into `[CONV:FACET-REQS]` in a separate docs PR if the SC team wants the exemption written down rather than inferred.

gvladika (approved):

PR #2001, opened 2026-07-01
feat(safe-decode): resolve role names for revokeRole/renounceRole

0xDEnYO:

No dedicated Linear ticket — small self-contained improvement to the Safe/timelock
proposal decoder surfaced while reviewing a live `revokeRole` proposal. Happy to
attach a ticket or apply the `trivial` label per the Ticket Linkage policy if
preferred.

When reviewing multisig/timelock proposals, `grantRole` already rendered the
human-readable OpenZeppelin role name next to the raw `bytes32` hash (e.g.
`(CANCELLER_ROLE)`), but `revokeRole` fell through to the generic argument
printer and showed only the raw hash — making revocations harder to verify at a
glance. `renounceRole` had the same gap.

`bytes32` roles are `keccak256("ROLE_NAME")` (a one-way hash), so there is no way
to reverse a hash back to its name — the resolver has to match against a known
set. The repo already precomputes that set (`KNOWN_ROLE_NAMES` + `getRoleName`);
it just wasn't wired into the revoke/renounce paths.

Changes:

- Generalized `formatGrantRole` → `formatRoleChange(functionName, …)` and route
  `grantRole` / `revokeRole` / `renounceRole` through it via a single shared ABI
  and a function-name set. All three now label the role.
- Added `DEFAULT_ADMIN_ROLE` (`bytes32(0)`) to the known-role map — it's the
  literal zero value, not a keccak hash of its name, so it needed an explicit
  entry. It's the role most likely to appear unlabeled in real proposals.
- Added a colocated `safe-decode-utils.test.ts` covering `getRoleName` (against
  the real keccak hashes, incl. the `CANCELLER_ROLE` hash from the proposal that
  prompted this) and `formatRoleChange` (grant/revoke/renounce all resolve the
  label; unknown roles stay unlabeled; incomplete args no-op).

Kept the change scoped to the display layer per `[CONV:SAFE-DECODE]` /
`[CONV:SAFE-DECODE-UTILS]`; no on-chain behavior, storage, or selector layout is
touched. A dynamic "read the contract's `*_ROLE()` getters over RPC" fallback for
project-specific roles was deliberately left out to avoid an RPC round-trip per
proposal — the static map covers every role currently defined in `src/`.

### DIFF
--- a/script/deploy/safe/safe-decode-utils.test.ts
+++ b/script/deploy/safe/safe-decode-utils.test.ts
@@ -1,0 +1,112 @@
+/**
+ * Tests for role-name resolution and role-change display in safe-decode-utils.
+ * Covers getRoleName (hash -> OZ AccessControl role name) and formatRoleChange
+ * (the grantRole / revokeRole / renounceRole display path).
+ */
+import {
+  describe,
+  expect,
+  it,
+  afterEach,
+  spyOn,
+  // eslint-disable-next-line import/no-unresolved
+} from 'bun:test'
+import { consola } from 'consola'
+
+import { getRoleName, formatRoleChange } from './safe-decode-utils'
+
+const DEFAULT_ADMIN_ROLE = `0x${'00'.repeat(32)}`
+const CANCELLER_ROLE =
+  '0xfd643c72710c63c0180259aba6b2d05451e3591a24e58b62239378085726f783'
+const PROPOSER_ROLE =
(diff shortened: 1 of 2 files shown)


0xDEnYO:
Superseded by #2003, which combines this with #2002 (both are display-layer improvements to the Safe-signing review flow under `script/deploy/safe/`) into a single review for @gvladika. Content is included verbatim in #2003.

PR #2018, opened 2026-07-03
fix(tron): route deployment MongoDB logging through the working update-deployment-logs.ts path

0xDEnYO:

Fixes EXSC-589 — [EXSC-589](https://linear.app/lifi-linear/issue/EXSC-589/fix-hardcoded-optimizer-runs200-in-tron-mongodb-deployment-log) was originally filed as a follow-up for the hardcoded `optimizer-runs=200` value (see below), now also fixed in this same PR.

While investigating why the Tron deploy runbook has a manual "update `deployments/tron.json`/`tron.diamond.json`" step, I found that step is actually stale — those files are already written automatically by `saveContractAddress()` / `updateDiamondJsonBatch()` during every Tron deploy.

But that investigation surfaced a real, separate bug: `logDeployment()` in `script/utils/utils.ts` (used by every Tron deploy script) has been shelling out to `script/helperFunctions.sh logDeployment ...` to write the MongoDB deployment-log record — except `helperFunctions.sh` has never had a `logDeployment` function (only `logContractDeploymentInfo`) and has no argument dispatcher at the bottom of the file. So the shell-out has silently exited 0 and done nothing since it was introduced (`c93177d10`). Every Tron deploy since then has been missing from the MongoDB deployment log that `query-deployment-logs.ts` reads.

This is Tron-specific — EVM deploys never go through this function; `deploySingleContract.sh` calls the working `logContractDeploymentInfo` bash function directly, which in turn invokes `bunx tsx script/deploy/update-deployment-logs.ts add ...`.

The fix: point `logDeployment()` at that same, already-working `update-deployment-logs.ts add` CLI instead of the nonexistent bash function, with the same field mapping `logContractDeploymentInfo` uses (contract/network/version/address/optimizer-runs/timestamp/constructor-args/verified/solc-version/evm-version). No behavior change for EVM; Tron deploys now actually land in MongoDB.

I also left a comment on the Notion runbook explaining both findings and suggesting step 6 be simplified to point at the "end-to-end deploy flow" section instead of restating it.

**Update**: the hardcoded `--optimizer-runs 200` finding below was originally deferred to EXSC-589 as a separate follow-up, but ended up landing on this same branch instead of a new PR — see the second `/pr-ready` entry.

## `/pr-ready` (local CodeRabbit) — 2 findings, both resolved

- `script/utils/utils.ts:304` — `--optimizer-runs` was hardcoded to `'200'`, but `foundry.toml`'s `[profile.default].optimizer_runs` is actually `1000000`, so the MongoDB deployment-log record stored an inaccurate value. Pre-existing (not introduced by this PR — the old broken command hardcoded the same `"200"`). Fixed by adding `getFoundryDefaultOptimizerRuns()` (mirroring `getFoundryDefaultSolcVersion()`/`getFoundryDefaultEvmVersion()`), tracked as [EXSC-589](https://linear.app/lifi-linear/issue/EXSC-589/fix-hardcoded-optimizer-runs200-in-tron-mongodb-deployment-log).
- `script/utils/utils.ts:102` — the follow-up commit's regex fallback for `optimizer_runs` only captured leading digits (`1_000` → `1`) and didn't validate the result as a real integer. Fixed: regex now accepts underscore-separated digits, and the final value is validated with `Number.isSafeInteger`.


gvladika (dismissed):

0xDEnYO:
Addressed the QA review in 6bfa3eb28:

- **[Medium] error branches untested** — added cases for missing `optimizer_runs`, negative value, non-integer value, and a missing `[profile.default]` section, plus the underscore-separated `1_000_000` value through the regex fallback path (forced by a digit-leading `rpc_endpoints` key that makes `Bun.TOML.parse` throw).
- **[Low] coupled to live `foundry.toml`** — tests now mock `fs.readFileSync` via `mock.module` from `bun:test` (transparent passthrough unless a test sets content), so they no longer assert a hardcoded repo value. A single integration sanity check remains but only asserts the real value is a non-negative safe integer.

Verified locally: `bun test script/` — 312 pass / 0 fail (all script test files in one process, so the fs mock provably doesn't leak).

gvladika (approved):

PR #1981, opened 2026-06-25
feat(EXSC-408): S11 — access-control interfaces + reference adapter

gvladika:
# Which Linear task belongs to this PR?

[EXSC-408 — S11: Access interfaces + reference adapter](https://linear.app/lifi-linear/issue/EXSC-408/s11-access-interfaces-reference-adapter)

**Stacked PR.** Base is `feature/exsc-409-s1-core-erc-4626-immutable-args-cwia` (#1971, S1). Review the incremental diff against the base; do not merge before S1 lands. First link in the stack S1 → **S11** → S5 → S3.

# Why did I implement it this way?

S11 is foundational and dependency-free: it ships only the external access-control surface (interfaces + a reference adapter) that S4's in-wrapper backends will import. It does **not** touch the wrapper — `_checkAccess` stays a no-op seam until S4.

- **`IVaultAccessControl` (`isAllowed`/`isDenied`), not `IAccessControl`.** The ticket names it `IAccessControl`, but this subsystem now vendors **OZ v5**, whose `@openzeppelin/contracts/access/IAccessControl.sol` (role-based: `hasRole`/`grantRole`) would hard-clash on the identifier the moment any S4 file imports both. Renamed to avoid that; the methods (`isAllowed`/`isDenied`) are unrelated to OZ's role model anyway. Two independent predicates so one adapter serves allowlist, denylist, or both modes.
- **`ISanctionsOracle` matches Chainalysis verbatim.** `isSanctioned(address) → bool` is the exact `SanctionsList` signature, so a live Chainalysis oracle drops in as the screening source with no shim.
- **`ReferenceAccessControl` is a template, not a LI.FI-operated contract.** Each integrator deploys/forks its own. `TransferrableOwnership`-gated allowlist + denylist + optional sanctions oracle. Policy: denied = denylisted OR sanctioned; allowed = `!denied && (!allowlistEnabled || allowlisted)`. Single + batch setters; the oracle is called directly (no try/catch) because `ISanctionsOracle` contractually must not revert — a flaky oracle is the integrator's adapter choice, documented in NatSpec. Holds no funds.

# Checklist before requesting a review

- [x] I have performed a self-review of my code
- [x] This pull request is as small as possible and only tackles one problem
- [x] I have run `/pr-ready` (local CodeRabbit) on this branch and resolved (or explicitly documented) all findings — see `.agents/commands/pr-ready.md`
- [x] I have added tests that cover the functionality / test the bug
[...]

PR #2026, opened 2026-07-05
feat(EXSC-412): vault wrapper access control — whitelist/blacklist/both, sanctions oracle (S4)

gvladika:

[EXSC-412](https://linear.app/lifi-linear/issue/EXSC-412/s4-access-control-whitelistblacklistboth-oracle-hook) — S4, vault-wrapper access control (whitelist/blacklist/both, sanctions-oracle hook).

**Stacked PR.** Base is `feature/exsc-410-s2-fee-engine-4-types-accrual-caps-per-fee-split` (#1993, S2) — review the incremental diff; do not merge before S2 lands.

Note: the ticket lists S11 (EXSC-408 / #1981) as a dependency; per agreement the interfaces are defined here from scratch instead, and #1981 is superseded for the interface part (EXSC-408 remains for the integrator-facing reference adapter).

Fills the `_checkAccess` no-op seam S1 left behind, as an abstract module (`VaultWrapperAccessControl`) the wrapper inherits.

- **Two independent gates, per-gate backend.** Allow gate and deny gate each select `{Disabled, Mapping, Merkle, External}` — "whitelist", "blacklist", and "both" fall out of the combination, including mixed backends (e.g. Merkle allowlist + mapping denylist). One `IVaultAccessControl` adapter serves both External gates (`isAllowed`/`isDenied` are independent predicates). The deny gate can never be Merkle: that would require non-inclusion proofs, which plain Merkle trees cannot provide — the natural blacklist shapes are the mapping (ops adds addresses as discovered) or an external adapter.
- **Interface names.** The ticket says `IAccessControl`, but OZ v5 (which this subsystem vendors) exports a role-based `IAccessControl` that would clash on import — hence `IVaultAccessControl`. `ISanctionsOracle.isSanctioned(address)` matches Chainalysis' `SanctionsList` verbatim so a live oracle drops in without a shim; `address(0)` disables the hook.
- **Receiver-validated, dual-path safe.** All deposit-path checks validate the **share receiver**, never `msg.sender`, so direct ERC-4626 calls and Composer-proxied deposits are gated identically (I29/U5). Withdraw/redeem run no access check at all — withdrawals-always-open is structural (U20), including for holders who fall off the list or get sanctioned after entry.
- **Merkle via permissionless prove-and-cache.** `deposit(assets, receiver)` has no proof slot and U25 locks the standard ERC-4626 surface, so proofs are delivered out-of-band: `proveAllowed(account, proof)` verifies against the current root and caches `provenAllowed[root][account]`. Anyone can prove for any account (an integrator backend can pre-prove a Composer receiver). The cache is keyed by root, so a rotation (I7) invalidates every cached entry with zero iteration. OZ standard-tree double-hashed leaves, compatible with `@openzeppelin/merkle-tree` tooling.
- **Transfer ↔ access-mode coupling.** Any active gate makes shares non-transferable (`_update` hook, I9/U10); mints, burns, and transfers **from the wrapper itself** are exempt (fee-share accrual and S3's sweep keep working). With only the oracle active, shares stay transferable but the **recipient** is screened — otherwise a sanctioned address could acquire shares by secondary transfer and exit via the ungated withdrawal path. The sender is deliberately not screened (freezing a holder's outbound movement is a custodial posture the wrapper avoids, A24). This is a small extension over the story wording; stories are being updated.
- **Fail-closed, loudly.** Adapter/oracle calls have no try/catch on the execution path: a reverting backend blocks deposits (never exits) and surfaces its own error verbatim. The EIP-4626 limit views (`maxDeposit`/`maxMint`) must not revert, so they report 0 for a gated receiver via a non-reverting mirror (`isDepositAllowed`), consistent with how pause already reports 0.
- **Typed init config.** `initialize`'s `_initData` is now a decoded-once `AccessConfig` (empty bytes = fully open instance); the raw `bytes initData` storage var is deleted — persisting an already-decoded blob was pure waste, and the subsystem is pre-deployment so the layout change is free. Config invariants (External ⇒ adapter set, Merkle ⇒ root set, deny ≠ Merkle) are enforced at `initialize` and on every setter, and the initial config emits the same events the setters would (A20 indexers need no calldata decoding). No seed arrays in init — a mapping-backed instance is closed until the owner seeds it post-deploy.
- **Module storage & admin gating.** State lives in an ERC-7201 namespaced slot (same idiom as the OZ v5 modules already in the inheritance list), so the module adds no sequential slot. It deliberately does not inherit `OwnableUpgradeable` — the wrapper's `Ownable2Step` branch would diamond-clash on `transferOwnership` — so the wrapper supplies the owner check through the abstract `_requireAccessAdmin` seam (mutators revert `OwnableUnauthorizedAccount` as usual). Mutation authority is the per-vault `owner` only: the integrator brings their own authority model (EOA/multisig/timelock, I7) by holding that role.
- **Underlying-vault perimeter (ticket note).** There is no standard interface to enforce a third-party vault's own access list from our side. What this PR guarantees: if the underlying blocks the wrapper, the revert bubbles verbatim (test with a gated-4626 mock). The per-user question (wrapper aggregates users the underlying's KYB never sees) is an onboarding criterion for the factory's underlying allowlist (A5/A6, Q6.6 → S8 docs), with the External backend as the mitigation — an integrator wrapping a gated vault points their allow gate at the underlying's own list.

Tests: 54 new cases in `LiFiVaultWrapperAccess.t.sol` (mode matrix incl. none, receiver-vs-sender validation, transfer coupling incl. sweep exemption, oracle enable/disable, fail-closed backends, Merkle prove/rotate/invalidate, EIP-4626 limit views, gated-underlying bubbling) + factory forwarding test reworked to the typed config. Full suite: 1944 passed.

### DIFF
--- a/src/VaultWrapper/LiFiVaultWrapper.sol
+++ b/src/VaultWrapper/LiFiVaultWrapper.sol
@@ -14,6 +14,7 @@ import { ILiFiVaultWrapperFactory } from "./interfaces/ILiFiVaultWrapperFactory.
 import { IYieldAdapter } from "./interfaces/IYieldAdapter.sol";
 import { FeeConfig, FeeType } from "./LiFiVaultWrapperTypes.sol";
 import { LibVaultWrapperMath } from "./libraries/LibVaultWrapperMath.sol";
+import { VaultWrapperAccessControl } from "./VaultWrapperAccessControl.sol";
 
 /// @title LiFiVaultWrapper
 /// @author LI.FI (https://li.fi)
@@ -35,13 +36,17 @@ import { LibVaultWrapperMath } from "./libraries/LibVaultWrapperMath.sol";
 ///      Every fee is split between LI.FI and the integrator at accrual time using the
 ///      fee type's own share, into per-recipient counters. This contract does not
 ///      distribute the accrued fees. Pause is enforced on the deposit/mint path only
-///      (withdrawals stay open);
-///      access logic remains a no-op seam (`_checkAccess`) with its body landing in a
-///      follow-up ticket. Inflation-attack protection relies on the ERC-4626 virtual-share
-///      offset.
+///      (withdrawals stay open). Access control (VaultWrapperAccessControl) gates the
+///      deposit path on the share receiver and freezes transfers while a gate is
+///      active; withdrawals are never gated. Inflation-attack protection relies on the
+///      ERC-4626 virtual-share offset.
 /// @custom:version 1.0.0
 contract LiFiVaultWrapper is
     ERC4626Upgradeable,
+    // Access control keeps its state in an ERC-7201 namespaced slot (see the module),
+    // so inheriting it adds no slot to this layout and is collision-free behind a
+    // beacon proxy.
+    VaultWrapperAccessControl,
     // OZ v5's Ownable/Ownable2Step keep `_owner`/`_pendingOwner` in fixed ERC-7201
     // namespaced slots, not sequential ones, so they add no slot to this layout and are
     // collision-free behind a beacon proxy. They provide the per-vault admin role: `owner()`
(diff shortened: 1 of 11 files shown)


gvladika:
Replaced by https://github.com/lifinance/contracts/pull/2032

PR #2031, opened 2026-07-08
fix(diamondSyncWhitelist): treat decimals()==0 as non-token (EXSC-365)

gvladika:

[EXSC-365](https://linear.app/lifi-linear/issue/EXSC-365/improve-istokencontract-detection-for-fallback-contracts)

`isTokenContract` in `diamondSyncWhitelist.sh` classified any address as a token when `decimals()` did **not** revert. Contracts with a catch-all fallback that returns 32 zero bytes for unknown selectors decode `decimals()` as `0`, so they were misclassified as tokens and the whitelist sync aborted.

This surfaced while rolling out the Nexroute post-hook wrapper on BSC (`0xdc7A4C55E9573eAA9399124b92019fF68874044C`, whitelisting selector `0xf8989325` = `backrun(bytes)`). The address has no real `decimals()` — its fallback answers every selector (`0xdeadbeef`, `0x12345678`, …) with `0` — and `name()`/`symbol()`/`totalSupply()` are all empty/zero, so it is not a token.

The fix keeps the `decimals()` heuristic for 1–255 (normal tokens classified exactly as before) and treats a returned `0` as ambiguous: it then probes `symbol() returns (string)`. A genuine 0-decimal token returns a non-empty symbol and is still classified as a token; a catch-all fallback decodes `symbol()` to an empty string and is rejected. This addresses the review feedback that the initial reject-all-zeros version introduced a false negative for genuine 0-decimal tokens.

Implementation notes:

- The `symbol()` probe runs **only** when `decimals() == 0`, not for every address — requiring a `string`-decodable `symbol()` unconditionally would misclassify tokens whose `symbol()` returns `bytes32` (MKR-style) even though their `decimals()` is normal.
- `cast` prints decoded strings wrapped in literal quotes (`"USDT"`, and `""` when empty) while `troncast` strips quotes entirely, so surrounding quotes are removed before the emptiness check.
- Verified with a 10-case mocked matrix (cast/troncast formats, reverts, whitespace-only symbols, out-of-range decimals) and live on BSC: USDT → token, the Nexroute wrapper → not a token, PancakeRouter → not a token.

Note: `/pr-ready` (local CodeRabbit CLI) was **not** run to completion. The first attempt wedged on "Preparing review" for 30+ minutes (service-side issue); the re-run after the symbol() probe commit hit the plan's review rate limit and the local gate was then skipped on explicit dev instruction. Cloud CodeRabbit runs on this PR as the safety net.

### DIFF
--- a/script/tasks/diamondSyncWhitelist.sh
+++ b/script/tasks/diamondSyncWhitelist.sh
@@ -91,16 +91,17 @@ function diamondSyncWhitelist {
   fi
 
   # Function to check if an address is a token contract
-  # tries to call decimals() function and returns true if a number value is returned
+  # tries to call decimals() and returns true only if a non-zero uint8 is returned
   function isTokenContract {
     local ADDRESS=$1
     local RPC_URL=$2
     local NETWORK=$3  # Add network parameter
     local RESULT
-    
+
     if RESULT=$(universalCast "call" "$NETWORK" "$ADDRESS" "decimals() returns (uint8)" 2>/dev/null); then
-      # Validate 0-255 strictly
-      if [[ "$RESULT" =~ ^([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])$ ]]; then
+      # Accept 1-255 only. A returned 0 means the address has no real decimals()
+      # and is answering via a catch-all fallback that returns zeros, not a token.
+      if [[ "$RESULT" =~ ^([1-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])$ ]]; then
         return 0
       fi
     fi
--- a/script/tasks/diamondSyncWhitelist.sh
+++ b/script/tasks/diamondSyncWhitelist.sh
@@ -91,18 +91,33 @@ function diamondSyncWhitelist {
   fi
 
   # Function to check if an address is a token contract
-  # tries to call decimals() and returns true only if a non-zero uint8 is returned
+  # tries to call decimals() and returns true if a uint8 value is returned.
+  # A returned 0 is ambiguous: genuine 0-decimal tokens return it, but so do
+  # contracts whose catch-all fallback answers every selector with 32 zero bytes.
+  # Disambiguate via symbol(): a real token returns a non-empty string, while a
+  # zero-returning fallback decodes to an empty one.
   function isTokenContract {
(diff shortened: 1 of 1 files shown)


script/tasks/diamondSyncWhitelist.sh
```diff
@@ -91,16 +91,17 @@ function diamondSyncWhitelist {
...
     local NETWORK=$3  # Add network parameter
     local RESULT
-    
+
     if RESULT=$(universalCast "call" "$NETWORK" "$ADDRESS" "decimals() returns (uint8)" 2>/dev/null); then
-      # Validate 0-255 strictly
-      if [[ "$RESULT" =~ ^([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])$ ]]; then
+      # Accept 1-255 only. A returned 0 means the address has no real decimals()
+      # and is answering via a catch-all fallback that returns zeros, not a token.
+      if [[ "$RESULT" =~ ^([1-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])$ ]]; then
```
0xDEnYO, line 108: we fix one issue but introduce another one (tokens with decimals == 0) are not detected. 
While 0 decimals tokens might be rare (or even entirely non-existent, I dont know), I feel we could also approach this differently, e.g. by calling another function a token contract always has (e.g. symbol())  and checking if there is any meaningful response.

gvladika: Implemented the symbol() probe in e752c034a. The 1-255 fast path is unchanged; only the ambiguous `decimals() == 0` case now probes `symbol() returns (string)` — a genuine 0-decimal token returns a non-empty symbol and is still classified as a token, while a zero-returning fallback decodes to an empty string and is rejected.

The probe is deliberately scoped to the `decimals() == 0` case rather than required for every address: tokens whose `symbol()` returns `bytes32` (MKR-style) fail string decoding, so an unconditional symbol requirement would misclassify them despite normal decimals.

Verified live on BSC: USDT (decimals=18) → token; the Nexroute wrapper `0xdc7A…044C` (decimals=0, empty symbol) → not a token; PancakeRouter (decimals reverts) → not a token. Note `cast` prints an empty decoded string as literal `""`, so the check strips surrounding quotes first (troncast strips them itself).

0xDEnYO (approved):

PR #2034, opened 2026-07-09
feat(PolymerCCTPFacet): pin HyperCore forwarder and validate hook receiver (EXSC-169) [PolymerCCTPFacet v3.0.0]

gvladika:

https://linear.app/lifi-linear/issue/EXSC-169/polymer-hypercore-integration-as-destination-smart-contract

**Stacked on #1652** (base: `polymer-hypercore-ex-286`). Strengthens the HyperCore hook flow so its security properties are enforced on-chain instead of relying on API-generated calldata.

Per [Circle's HyperCore transfer docs](https://developers.circle.com/cctp/transfer-usdc-from-ethereum-to-hypercore), a HyperCore deposit must set **both** `mintRecipient` and `destinationCaller` to the `CctpForwarder` on HyperEVM, with the real recipient encoded in `hookData[32:52]`. In #1652 those values were caller-supplied and unvalidated, which meant: (a) `LiFiTransferStarted.receiver` would carry the forwarder, not the user, (b) compromised calldata could redirect funds via `hookData` with no on-chain check, and (c) a wrong `destinationCaller` would burn USDC unrecoverably.

This PR pins and validates the flow (same pattern as MayanFacet's `MAYAN_HYPERCORE_DEPOSITOR`, #1881):

- `mintRecipient == destinationCaller == HYPERCORE_CCTP_FORWARDER` (hardcoded constant; rotation requires a facet upgrade).
- `hookData[32:52]` must equal `BridgeData.receiver` (revert `InvalidReceiver`), so events always carry the real end user and calldata cannot silently redirect funds. No further hook-content validation (magic/version/dex flag) by design — the recipient offset is the security-critical field; a malformed tail cannot redirect funds.
- Hook ⇔ HyperCore: hooks toward any other destination revert `InvalidCallData` (fail closed — corridors without a pinned forwarder have no validated hook consumer), and hookless transfers to HyperCore revert as well (they would mint on HyperEVM and never reach HyperCore). The corridor dispatch is structured so future corridors (e.g. Stellar, CCTP domain 27, same CctpForwarder pattern) land as one additional arm plus constants.
- `destinationCaller` is removed from `PolymerCCTPData` — the facet supplies it from the pinned constant, so the field could only ever be redundant or wrong.

Version stays `3.0.0`: v3 was never production-deployed, so this folds into the existing major bump (struct/ABI change vs released v2.1.0).

**Open items before merge:**

- [ ] ~~Confirm the production `CctpForwarder` address~~ Resolved (`366a40496`): the pinned address is now Circle's published mainnet HyperEVM `CctpForwarder` `0xb21D281DEdb17AE5B501F6AA8256fe38C4e45757` ([Circle reference](https://developers.circle.com/cctp/references/hypercore-contract-addresses)); the EXSC-169 address `0x9765...2815` turned out to be Polymer's test diamond (on-chain bytecode is a LibDiamond dispatch proxy), not a forwarder. Remaining ask to Polymer: confirm their relayer executes messages through Circle's forwarder.
- [ ] API must populate `BridgeData.receiver` with the end user (not the forwarder) for HyperCore quotes — the facet supplies the forwarder itself.
- [ ] Follow-up (not in this PR, per review agreement): `script/demoScripts/demoPolymerCCTP.ts` builds `PolymerCCTPData` without `hookData` and will need `hookData: '0x'` once `lifi-contract-types` regenerates.

Tests: `forge test --match-path test/solidity/Facets/PolymerCCTPFacet.t.sol` — 44/44 pass on a mainnet fork against the real TokenMessengerV2, including a HyperCore e2e with `vm.expectCall` asserting the forwarder as both `mintRecipient` and `destinationCaller`, and revert coverage for hook-to-wrong-destination, hookless HyperCore, receiver mismatch, non-EVM receiver, and truncated hook data.

- [ ] I have run `/pr-ready` (local CodeRabbit) on this branch and resolved (or explicitly documented) all findings — see `.agents/commands/pr-ready.md` — **partially**: first pass ran and its single finding was fixed (`747dcffe4`); the second pass reported 1 Major finding whose full text was lost to the CLI rate limit. Opening as draft with cloud CodeRabbit triggered instead; the cloud review must be clean/addressed before this leaves draft.

### DIFF
--- a/src/Facets/PolymerCCTPFacet.sol
+++ b/src/Facets/PolymerCCTPFacet.sol
@@ -289,7 +289,8 @@ contract PolymerCCTPFacet is
             // pinned forwarder. Add new corridors (e.g. Stellar) as additional arms.
             if (
                 _bridgeData.destinationChainId == LIFI_CHAIN_ID_HYPERCORE &&
-                _bridgeData.receiver != NON_EVM_ADDRESS
+                _bridgeData.receiver != NON_EVM_ADDRESS &&
+                _polymerData.hookData.length >= HOOK_RECIPIENT_OFFSET_END
             ) {
                 // The forwarder credits the account encoded in the hook data, so it
                 // must equal the declared receiver or events would misreport the
--- a/src/Facets/PolymerCCTPFacet.sol
+++ b/src/Facets/PolymerCCTPFacet.sol
@@ -37,8 +37,9 @@ contract PolymerCCTPFacet is
     /// @notice Circle's CctpForwarder on HyperEVM. HyperCore hook flows mint to this contract,
     ///         which alone may execute the message and deposits the USDC into HyperCore for the
     ///         receiver encoded in the hook data (see _startBridge).
+    ///         https://developers.circle.com/cctp/references/hypercore-contract-addresses
     address internal constant HYPERCORE_CCTP_FORWARDER =
-        0x9765885053Da4835e98546139017EEee41712815;
+        0xb21D281DEdb17AE5B501F6AA8256fe38C4e45757;
 
     /// @dev Byte offsets of the recipient address within CctpForwarder hook data
     ///      (24 bytes magic + 4 bytes version + 4 bytes payload length, then the recipient)
--- a/src/Facets/PolymerCCTPFacet.sol
+++ b/src/Facets/PolymerCCTPFacet.sol
@@ -34,12 +34,13 @@ contract PolymerCCTPFacet is
     /// @notice bytes32(0) allows any address to complete the CCTP transfer on destination chain
     bytes32 private constant UNRESTRICTED_DESTINATION_CALLER = bytes32(0);
 
-    /// @notice Circle's CctpForwarder on HyperEVM. HyperCore hook flows mint to this contract,
-    ///         which alone may execute the message and deposits the USDC into HyperCore for the
-    ///         receiver encoded in the hook data (see _startBridge).
+    /// @notice Circle's CctpForwarder on HyperEVM (0xb21D281DEdb17AE5B501F6AA8256fe38C4e45757),
+    ///         pre-encoded as the bytes32 mintRecipient/destinationCaller. HyperCore hook flows
+    ///         mint to this contract, which alone may execute the message and deposits the USDC
+    ///         into HyperCore for the receiver encoded in the hook data (see _startBridge).
     ///         https://developers.circle.com/cctp/references/hypercore-contract-addresses
-    address internal constant HYPERCORE_CCTP_FORWARDER =
-        0xb21D281DEdb17AE5B501F6AA8256fe38C4e45757;
+    bytes32 internal constant HYPERCORE_CCTP_FORWARDER =
+        bytes32(uint256(uint160(0xb21D281DEdb17AE5B501F6AA8256fe38C4e45757)));
 
     bytes32 internal constant NAMESPACE =
         keccak256("com.lifi.facets.polymercctp");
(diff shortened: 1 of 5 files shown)


gvladika:
@coderabbitai review

src/Facets/PolymerCCTPFacet.sol
```diff
@@ -27,6 +34,17 @@ contract PolymerCCTPFacet is
...
+    ///         which alone may execute the message and deposits the USDC into HyperCore for the
+    ///         receiver encoded in the hook data (see _startBridge).
+    address internal constant HYPERCORE_CCTP_FORWARDER =
+        0x9765885053Da4835e98546139017EEee41712815;
+
+    /// @dev Byte offsets of the recipient address within CctpForwarder hook data
+    ///      (24 bytes magic + 4 bytes version + 4 bytes payload length, then the recipient)
+    uint256 private constant HOOK_RECIPIENT_OFFSET_START = 32;
+    uint256 private constant HOOK_RECIPIENT_OFFSET_END = 52;
+
```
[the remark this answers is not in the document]
gvladika: Fixed in 366a40496 — pinned Circle's published mainnet CctpForwarder `0xb21D281DEdb17AE5B501F6AA8256fe38C4e45757`. Verified on-chain: the previous `0x9765...2815` is a LibDiamond dispatch proxy (Polymer's test diamond from the Linear ticket), not a forwarder.

src/Facets/PolymerCCTPFacet.sol
```diff
@@ -265,14 +278,33 @@ contract PolymerCCTPFacet is
...
+                if (
+                    address(bytes20(_polymerData.hookData[32:52])) !=
+                    _bridgeData.receiver
+                ) {
+                    revert InvalidReceiver();
+                }
+            } else {
+                revert InvalidCallData();
+            }
+        } else if (_bridgeData.destinationChainId == LIFI_CHAIN_ID_HYPERCORE) {
```
0xDEnYO: you are doing this destinationChainId check twice. 

In the same context...first you check if hookdata > 0 (you check this for all cases) and set isHookData, then you check again if hookdata is >= 52 (so on that branch you check hookdata length twice.

I think you can reshuffle this and optimize so that checks are only done once.

src/Facets/PolymerCCTPFacet.sol
```diff
@@ -299,17 +331,20 @@ contract PolymerCCTPFacet is
...
             }
 
-            bytes32 mintRecipient = bytes32(
-                uint256(uint160(_bridgeData.receiver))
-            );
+            if (isHookFlow) {
+                // Mint to the pinned forwarder (never to the receiver directly) and
+                // restrict execution to it; the forwarder then deposits into
+                // HyperCore for the receiver validated above
+                bytes32 forwarder = bytes32(
```
0xDEnYO: why dont we store this as bytes32 value so we dont have to convert this every single time?

src/Facets/PolymerCCTPFacet.sol
```diff
@@ -49,9 +63,8 @@ contract PolymerCCTPFacet is
         // the minimum finality at which a burn message will be attested to, will be passed directly to tokenMessenger.depositForBurn method.
         // 1000 = fast path, 2000 = standard path
         uint32 minFinalityThreshold;
-        // Restricted caller on the destination domain for hook flows (e.g. HyperCore forwarding); only the forwarder contract may execute.
-        bytes32 destinationCaller;
```
0xDEnYO, line 53: So we will not need this for other chains? Is it safe to remove?

gvladika: It is safe. This param is only introduced by base PR 1652 (not merged or deployed anywhere). But it is not needed with this re-design, because we use the constant address `HYPERCORE_CCTP_FORWARDER` (will introduce similair for stellar). This is safer because address is pinned, and not provided from outside.

src/Facets/PolymerCCTPFacet.sol
gvladika: Good catch — restructured in 129dcf423: the dispatch now keys on `destinationChainId` once, the hookless-HyperCore case folds into the single `length < 52` bound, and the downstream deposit branch reuses the chainId check, so `isHookFlow` is gone entirely. Each condition is checked exactly once.

src/Facets/PolymerCCTPFacet.sol
gvladika: Done in 5a9bd84a8 — constant is now stored pre-encoded as `bytes32` and used directly at both call sites (plain address kept in the NatSpec for reference). Small note: since it's a `constant`, solc folds the cast at compile time, so there was no runtime conversion cost either way — this is purely a readability win.

src/Facets/PolymerCCTPFacet.sol
0xDEnYO: So destinationCaller is only used when bridging to hypercore, not when bridging to other chains (important cause we only have one hardcoded forwarder address in this contract), correct?

0xDEnYO (approved):

PR #2020, opened 2026-07-03
chore(pr-ready): remove local CodeRabbit pre-flight gate repo-wide

0xDEnYO:
# Which Linear task belongs to this PR?

# Why did I implement it this way?

The local CodeRabbit CLI pre-flight (`/pr-ready`) has been hitting rate limits and
wasting review cycles across the whole SC team, so it's being removed entirely rather
than patched — there's no separable non-CodeRabbit functionality in the skill worth
keeping. This PR:

- Deletes `.agents/commands/pr-ready.md` and its Cursor/Claude Code symlinks.
- Deletes `.claude/scripts/pr-ready-gate.ts` and the `PreToolUse` hook in
  `.claude/settings.json` that enforced it.
- Removes the CodeRabbit CLI auto-install block from `preinstall.sh` (it existed only
  to support `/pr-ready`).
- Updates `.agents/rules/099-finish.md`, `.agents/commands/create-pr.md`,
  `deploy-contract.md`, `multisig-rollout.md`, `check-open-prs.md`, the
  `aikido-false-positive-catalog.md`, and `.github/pull_request_template.md` to drop
  all references to the removed gate/skill.
- Updates `.agents/README.md`'s command table.

Cloud CodeRabbit still runs in GitHub CI on every PR, unaffected — only the local
pre-flight step is removed. `self-review-pass` (a separate mechanical/semantic/
executable sweep, unrelated to CodeRabbit) is untouched and remains the mandatory
step before pinging for review.

**On `/autofix-pr` (raised in the `#dev-sc-review` thread on this PR)**: evaluated as a
replacement mechanism and declined for now. `/autofix-pr` is a built-in Claude Code CLI
command (not a plugin/skill) that hands a PR off to a Claude Code on the web session and
turns on a persistent per-PR auto-fix toggle — Claude then reacts to CI failures and
review comments via GitHub webhooks until turned off. It (a) requires installing the
Claude GitHub App on the `lifinance` org, which we haven't done, and (b) its cloud
sessions draw from the same shared Claude/Claude Code rate-limit pool as everything else
on the account — no separate dollar charge, but continuous webhook-triggered fixing
across multiple open PRs is unpredictable quota consumption we'd rather not take on.
Individuals who want an async safety net for catching CodeRabbit's cloud comments can
build a personal scheduled task instead (no new org-level dependency); this isn't made a
team requirement since it needs per-user schedule setup.

**Bootstrap note**: this PR removes the very gate hook that would otherwise block
[...]

PR #2042, opened 2026-07-13
feat(MayanFacet): route native ETH via swapAndForwardEth (EXSC-364) [MayanFacet v2.0.0, IMayan v1.1.0]

0xDEnYO:

https://linear.app/lifi-linear/issue/EXSC-364/sc-mayanfacet-route-between-forwardeth-and-swapandforwardeth-by

Also closes **EXSC-606** (route refunds to an explicit `refundRecipient` instead of `msg.sender`) — folded in here to avoid a second `MayanData` ABI break, per the refund-routing convention call (agreed by Daniela, Goran, Max).

Unblocks BE **EXBE-355** — this PR defines the `MayanData` calldata shape the backend consumes.

`MayanFacet` previously routed all native-ETH sends through the Mayan forwarder's `forwardEth`. A Mayan Swift v2 order takes a specific ERC20 input token, not native ETH, so to offer **native ETH** as a sending token the forwarder exposes `swapAndForwardEth`, which performs a source-side swap (via `swapProtocol`/`swapData`) converting the native input into the order's input token (`middleToken`) before creating the order. The `middleToken` is chosen by Mayan per route/quote — it may be WETH **or another token such as a stablecoin** — and the conversion is a real DEX swap, not a fixed 1:1 wrap. (The verified staging run below used **USDT** as `middleToken` for an ETH→Polygon-USDT route.)

**Design — extend `MayanData`, discriminate at the native branch.** Four swap-layer fields are appended to `MayanData` for the native path, plus one refund field (see below). The full struct:

```solidity
struct MayanData {
    bytes32 nonEVMReceiver;
    address mayanProtocol;
    bytes    protocolData;      // inner Mayan order → _parseReceiver reads this — UNCHANGED
    address swapProtocol;       // NEW (EXSC-364)
    bytes    swapData;          // NEW (EXSC-364)
    address middleToken;        // NEW (EXSC-364)
    uint256  minMiddleAmount;   // NEW (EXSC-364)
    address refundRecipient;    // NEW (EXSC-606)
}
```

`_startBridge` native branch:
- ERC20 → `forwardERC20` (unchanged)
- native + `swapProtocol != address(0)` → `swapAndForwardEth(minAmount, swapProtocol, swapData, middleToken, minMiddleAmount, mayanProtocol, protocolData)`
- native + `swapProtocol == address(0)` → `forwardEth` (retained)

The backend selects the path per-quote by populating (or not) the swap-layer fields — the facet needs no knowledge of Mayan's ETH/WETH order semantics. `_parseReceiver`/`_parseHypercoreReceiver` are **byte-for-byte unchanged**, so on-chain receiver validation (derived from `protocolData`, run *before* the branch) still gates every path. The new opaque `swapData`/`swapProtocol` only govern the source-side conversion executed inside the immutable, trusted `MAYAN` forwarder (which itself allowlists `swapProtocol`). No EIP-712 gate is required — unlike rule 102's opaque-receiver case, the final receiver stays on-chain-validatable.

**Refund routing (EXSC-606, folded in).** Both entrypoints previously refunded excess native and swap leftovers to `msg.sender`. Per `.agents/rules/102-facets.md` `[CONV:FACET-REFUNDS]`, `msg.sender` is the relayer (or the Permit2Proxy on gasless flows), never assumed to be the user, so user-owned refunds get stranded. This PR adds an explicit `refundRecipient` to `MayanData`, zero-address-checked on **both** entrypoints (revert `InvalidCallData`), and routes both `refundExcessNative(payable(refundRecipient))` and the `_depositAndSwap` leftover receiver there. The backend sets `refundRecipient` to the owner of the source-side input per order (relayer when the relayer fronts the native, user when the user funds it). Since we were already breaking the `MayanData` ABI for EXSC-364, folding this in adds one struct field with **no additional selector churn**.

**Versioning.** Appending struct fields changes both entrypoint selectors, a breaking ABI change for existing callers → MAJOR bump `1.3.0 → 2.0.0` (one MAJOR bump covers the whole PR):
- `startBridgeTokensViaMayan` `0xb621b032 → 0xa2eff2da`
- `swapAndStartBridgeTokensViaMayan` `→ 0x2abf2290`

`IMayan` `1.0.0 → 1.1.0` (genuinely new `swapAndForwardEth` function).

**Testing.** `forge test --match-contract MayanFacet` → **39 pass**, incl. new tests for the native `swapAndForwardEth` path (arg pass-through against a recorder etched onto the real `MAYAN` immutable), the `swapAndStartBridge` double-swap edge (proves `_replaceInputAmount` stays skipped for native), a receiver-mismatch revert on the new path, and the EXSC-606 refund tests (zero-`refundRecipient` reverts on both entrypoints; excess native lands on `refundRecipient`, not `msg.sender`).

**Staging E2E (real Mayan data).** `script/demoScripts/demoMayan.ts` was rewritten to exercise the native `swapAndForwardEth` path end-to-end with a live Mayan quote (previously it only covered the ERC20 path). This required bumping `@mayanfinance/swap-sdk` `8.5.0 → 15.0.0` (the pinned SDK is rejected by Mayan's quote API: `SDK_VERSION_TOO_OLD`). Verified run on Arbitrum staging: **0.001 ETH → ~1.8388 USDT** delivered to the receiver on Polygon, Mayan status **`ORDER_SETTLED`**. Source tx: [`0xa67bb19b…dd1f6c`](https://arbiscan.io/tx/0xa67bb19b7fbf19d4e866b3dd743114ef1b7918533b08576a249a2f58ffdd1f6c).

**Review.** Local CodeRabbit (`/pr-ready`) bypassed per the sanctioned interim (rate limits — see #2020); an independent adversarial `code-reviewer` pass was run instead (correctness + rule-102 opaque-`swapData` lens; `IMayan` verified byte-exact against the live forwarder ABI) with no findings. Cloud CodeRabbit + Aikido run on this PR.

**Deploy.** MayanFacet **v2.0.0** deployed + verified on the Arbitrum **staging** diamond at `0x9923249c1f14E4BB68F71C689C3fCC323E5410aC` and cut in (selectors `startBridgeTokensViaMayan 0xa2eff2da`, `swapAndStartBridgeTokensViaMayan 0x2abf2290`, `MAYAN() 0xce90a721`). Note: an earlier staging deploy in this PR predated the `refundRecipient` field, so it carried the pre-`refundRecipient` selectors; the diamond has now been re-cut to the current build (confirmed via DiamondLoupe).

**Follow-up.** EXSC-604 (signed arbitrary destination-call payloads, EIP-712) has been **descoped / cancelled** (PR #2043 closed) — it is no longer stacked on this branch. The remaining Mayan hardening item is **EXSC-605** (reject a non-empty `customPayload` on the unsigned bridge path), tracked separately; it is a pre-existing surface not widened by this PR.

- [ ] I have run `/pr-ready` (local CodeRabbit) on this branch and resolved (or explicitly documented) all findings — **bypassed per the sanctioned interim (rate limits); independent adversarial review done instead, documented above**
- [ ] For new facets: I have checked all points from this list — **N/A (existing facet, not a new one)**


script/demoScripts/demoMayan.ts
```diff
@@ -1,96 +1,322 @@
...
+        null,
+        address,
+        ARBITRUM_CHAIN_ID,
+        null,
+        null,
+        { apiKey: MAYAN_API_KEY }
+      )
+    } catch {
+      continue
+    }
```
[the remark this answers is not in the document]
0xDEnYO: Fixed in fcc41380 — the catch now logs the route + error before continuing, so auth/network/payload failures are surfaced rather than folded into the generic no-quote path.

script/demoScripts/demoMayan.ts
```diff
@@ -1,96 +1,322 @@
...
     null,
-    null
+    null,
+    { apiKey: MAYAN_API_KEY }
   )
 
-  const iface = IMayan__factory.createInterface()
   const parsed = iface.parseTransaction({ data: payload.data as string })
+  // Derive the exact amount Mayan expects instead of hard-coding it.
+  const amountIn = BigNumber.from(parsed.args.amountIn)
```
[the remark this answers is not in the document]
0xDEnYO: Fixed in fcc41380 — the ERC20 path now rejects any decoded method other than forwardERC20 before reading amountIn/protocolData, mirroring the native-path guard.

script/demoScripts/demoMayan.ts
```diff
@@ -1,96 +1,322 @@
...
+  const mode = (process.env.MAYAN_DEMO_MODE ?? 'native').toLowerCase()
+
+  const provider = new ethers.providers.JsonRpcProvider(RPC_URL)
+  const signer = new ethers.Wallet(PRIVATE_KEY, provider)
+
+  if (mode === 'erc20') {
+    await runErc20ForwardErc20(signer, provider)
+  } else {
+    await runNativeSwapAndForwardEth(signer, provider)
+  }
```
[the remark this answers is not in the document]
0xDEnYO: Fixed in fcc41380 — unsupported MAYAN_DEMO_MODE values now throw instead of defaulting to native (the script broadcasts a real tx).

src/Facets/MayanFacet.sol
```diff
@@ -203,6 +224,16 @@ contract MayanFacet is
                 _mayanData.mayanProtocol,
                 _mayanData.protocolData
             );
+        } else if (_mayanData.swapProtocol != address(0)) {
+            MAYAN.swapAndForwardEth{ value: _bridgeData.minAmount }(
+                _bridgeData.minAmount,
+                _mayanData.swapProtocol,
+                _mayanData.swapData,
+                _mayanData.middleToken,
+                _mayanData.minMiddleAmount,
```
gvladika, line 306: `minMiddleAmount` is forwarded straight through as the **only** slippage guard on the native ETH→`middleToken` source swap, but the facet never requires it to be `> 0`. At zero, the forwarder's `middleAmount >= minMiddleAmount` check is a no-op and the intermediate swap can be fully sandwiched — the order gets created for a dust amount and the user receives almost nothing on the destination chain.

This is a new value-at-risk surface introduced by the `swapAndForwardEth` path. It's caller/backend-supplied per the aggregator pattern, so this may be an accepted "backend responsibility" — but if so, please state that explicitly in the audit notes, and consider whether a `require(minMiddleAmount > 0)` (or a receiver-derived floor) is warranted here given the source swap executes with the user's native value.

deployments/arbitrum.diamond.staging.json
```diff
@@ -217,9 +217,25 @@
         "Name": "SupersetFacet",
         "Version": "1.0.0"
       },
-      "0xEB6803A40B8350286e1E8684068E28c5c57a94AE": {
+      "0x6b380169D4C02446703b97E6c3a5dA514039A897": {
```
gvladika: This diamond staging map now lists **MayanFacet 2.0.0 at two addresses** — `0x6b380169…` here and `0x9923249c…` a few lines below — while the flat `arbitrum.staging.json` registers only `0x9923249c…`. The same hunk also adds CalldataVerificationFacet / PolymerCCTPFacet / SymbiosisFacet addresses that don't match the flat file's entries for those facets.

The PR body explains a re-cut happened after the `refundRecipient` field was added, so `0x6b380169…` is most likely a stale pre-`refundRecipient` artifact. Please reconcile the two deployment logs before merge — deploy/verify tooling that iterates the diamond facets map (verify-contracts, multisig-rollout) could otherwise act on a facet address that was never cut into the live diamond, or report a spurious mismatch that blocks the staging rollout.

src/Facets/MayanFacet.sol
0xDEnYO: Fixed in 5769e96 — added an explicit guard in `_startBridge` before the `swapAndForwardEth` call:

```solidity
} else if (_mayanData.swapProtocol != address(0)) {
    // minMiddleAmount is the only slippage guard on the native input -> middleToken
    // source swap; at zero the swap can be fully sandwiched, so reject it explicitly.
    if (_mayanData.minMiddleAmount == 0) {
        revert InvalidAmount();
    }
    ...
```

This rejects the fully-sandwichable zero-floor case at the contract level (consistent with `validateBridgeData`'s `minAmount > 0`). Covered by `testRevert_SwapAndForwardEthWithZeroMinMiddleAmount`. The residual slippage *tightness* above zero stays backend-supplied per the aggregator model. Facet redeployed to arb staging with the guard at `0xc8D1bb44ED16b05518D7E32BD7fb2746594C9045`.

deployments/arbitrum.diamond.staging.json
0xDEnYO: Fixed in 7f7fd2f — reconciled the diamond staging log to a single `MayanFacet` entry at `0xc8D1bb44ED16b05518D7E32BD7fb2746594C9045`, matching the flat `arbitrum.staging.json` and the live on-chain diamond (verified via `facetAddresses()` on `0xD3b2b0aC0AFdd0d166a495f5E9fca4eCc715a782` — only `0xc8D1bb44` is cut; `0x6b380169`, `0x9923249c` and the old `0xEB6803A4` v1.3.0 are not).

Note: the guard change altered the bytecode, so the facet was redeployed (new CREATE3 address `0xc8D1bb44`) and `diamondCut`-replaced the prior `0x9923249c`. The `CalldataVerificationFacet` / `PolymerCCTPFacet` / `SymbiosisFacet` entries you flagged were ambient noise from the deploy tool's full-log re-sync; after basing the log on `main` + the surgical Mayan edit, the PR's net diamond-log diff is now MayanFacet-only. This also cleared the merge conflict.

gvladika (approved):

PR #2047, opened 2026-07-13
feat(cleanUpProdDiamond): auto-detect & remove deprecated facets via target-state diff (EXSC-193)

0xDEnYO:

https://linear.app/lifi-linear/issue/EXSC-193/deprecate-contracts-in-smart-contracts-after-backend (EXSC-193)

Follow-up #2 of **EXSC-193** (on-chain removal of the facets deprecated in
#2046), parent epic **EXP-350** "Deprecation of GenericSwapFacet and old Across
facets". EXSC-193 is already tracked (In Review, assigned to Daniel) and linked
to this PR.

**Problem.** `/deprecate-contract` only removes a facet from the codebase and
`_targetState.json`; the on-chain facet stays registered and callable on every
production diamond forever. The only remedy today is running
`cleanUpProdDiamond.ts` by hand, per network — and that script can't even remove
an *already-deprecated* facet, because it derives selectors from the deleted
`out/` artifact.

**Design.** Full spec: [`docs/FacetRemovalReconciliation.md`](docs/FacetRemovalReconciliation.md).
Two selection signals, one shared plumbing:

- **Primary — deprecation-driven & explicit, now PARKED (spec §10).**
  `/deprecate-contract` *names* the facet, so removal doesn't need to be
  inferred. For each named facet we find every PROD diamond whose **deployment
  log** lists it (the authoritative facet→address map) and **park** one removal
  task per (facet, network) into the deferred diamond-cleanup queue via
  `enqueue-parked-task.ts` (the store + CLI merged in #2051). **No Safe proposal
  is created at deprecation time** — the parked task is drained into the removal
  proposal on a later rollout, reading the facet's selectors from the on-chain
  loupe at drain time. `/deprecate-contract` step 6 drives the park.
- **Backstop — target-state diff.** For orphans nobody explicitly named
  (historical, wrong-chain), diff the loupe against `_targetState.json`. Here a
  source-presence gate is essential to distinguish *deprecated* (source gone)
  from *drift* (target state merely lags).

Both feed `cleanUpProdDiamond.ts`'s existing timelock-wrap + Safe-propose
plumbing — no new governance path, Solidity deploy scripts untouched.

> **§10 fold.** An earlier revision of this PR made `/deprecate-contract` step 6
> *create* the removal proposals eagerly (one Safe proposal per network at
> deprecation time). Per the agreed deferred-cleanup-queue design
> ([`docs/DeferredDiamondCleanupQueue.md`](docs/DeferredDiamondCleanupQueue.md),
> §10), step 6 now **parks** instead, so signers aren't asked to sign a
> standalone per-chain removal at deprecation time — the removal rides the next
> rollout's signing session. The removal **engine** and the drift `--auto`
> backstop are unchanged; only *when* the named removal becomes a proposal moved
> (park now → drain later). The deferred-cleanup-queue **spec is now included in
> this PR** (`docs/DeferredDiamondCleanupQueue.md`, folded in from the former
> #2049, now closed); the queue **store + enqueue/list CLIs** are already merged
> in #2051, and the **drain** consumer is a deliberate follow-up (spec §13).

> This primary/backstop framing came from earlier review feedback: an earlier
> revision made the target-state diff the primary trigger. But the deployment
> logs *are* the live deployment map and exist at deprecation time, and
> deprecation names the facet — so the explicit, log-driven path is both simpler
> and safer, and the diff is just the backstop.

**What landed**

- `script/deploy/safe/diamondRemovalDiff.ts` — pure, injectable engine
  (+ colocated tests): `computeNamedFacetRemovals` (primary) and
  `computeFacetRemovalDiff` (backstop). Selectors always from the loupe, so both
  work after the facet's source/artifact is gone.
- `script/tasks/cleanUpProdDiamond.ts` — `--facets '[...]'` is now loupe-driven
  (fixes the can't-remove-a-deprecated-facet bug) and supports `--all-networks`;
  `--auto`/`--auto --all-networks` is the diff backstop; dry-run unless `--yes`.
  Consumed by the parked-task **drain** and by on-demand manual removal.
- `.agents/commands/deprecate-contract.md` — **step 6** now **parks** the named
  facets' removals into the deferred cleanup queue (one task per (facet, network)
  via `enqueue-parked-task.ts`), run as the **last** step after `gh pr create`
  returns the deprecation PR URL so `--prUrl` carries the real URL. Sequencing
  rule updated: don't delete the facet→address log entries until the parked task
  **retires** (executed/cancelled/superseded).
- `docs/FacetRemovalReconciliation.md` — named-path narrative realigned from
  propose-at-deprecation to **park-then-drain** (engine, CLI `--facets`, and the
  drift `--auto` backstop unchanged); notes the named-park vs drift-`--auto`
  overlap and that the drain hook is a follow-up.
- `.agents/commands/multisig-rollout.md` — opt-in **Phase 3.5** so a backstop
  (drift) removal proposal can ride the next rollout's signing session.
- `docs/DeferredDiamondCleanupQueue.md` — the scheduling-layer **spec** (design of
  record; folded in from the former #2049). Specifies the park-then-drain model,
  the resolved store/chokepoint decisions, the parked-task lifecycle, and the
  drain/reconcile/TTL follow-ups. The store + enqueue/list CLIs it references are
  merged (#2051); the drain consumer is the next PR (spec §13).

**Guardrails (governance-preserving).**

- Hardcoded never-remove allowlist — `DiamondCutFacet` / `DiamondLoupeFacet` /
  `OwnershipFacet` / `EmergencyPauseFacet` — plus `getCoreFacets()` /
  `getCorePeriphery()` and the diamonds; protected even if dropped from target
  state.
- Skips `LiFiDiamondImmutable` (can't be cut).
- **Source-presence gate (backstop path)** — in the diff path a facet is removed
  only if its `src/` source is gone (true deprecation), *not* merely absent from
  target state. A live smoke run against the real mainnet diamond found **8**
  facets on-chain and absent from target state that are still live
  (`PaxosTransitFacet`, `MegaETHBridgeFacet`, `PolymerCCTPFacet`,
  `AcrossV4SwapFacet`, `NEARIntentsFacet`, `EcoFacet`, `GardenFacet`, `UnitFacet`)
  — a naïve diff would have proposed removing all 8; they're now classified as
  *drift* and never removed. The named path doesn't need this gate (intent is
  explicit) but still enforces the never-remove allowlist.
- **Deploy-log sequencing (named path)** — the parked task snapshots
  facet→address from the deploy log at enqueue time; `/deprecate-contract` warns
  not to delete the entries until the parked task retires.
- Held-back selectors an active facet is expected to own; unresolved on-chain
  addresses never auto-removed; conspicuous `⚠️ IRREVERSIBLE FACET REMOVAL`
  banner per network before any propose (at drain / backstop time).
- Governance path unchanged: production always wraps in the timelock schedule
  and proposes to the Safe; staging/testnet keep the existing sanctioned
  direct-send.

**Open decisions for review** (spec §10): rollout step defaults to **opt-in**;
first-PR scope is engine + CLI + opt-in doc + the §10 park wiring; periphery
reconciliation deferred; the drain hook (parked-task consumer + named-park vs
drift reconciliation) is a follow-up.

**Draft** because the design (esp. the three open decisions and the §10 park
fold) should be reviewed before the first live sweep, and because the parked
tasks have no drain consumer until the follow-up drain-hook PR.

**Testing.** `bun test script/deploy/safe/diamondRemovalDiff.test.ts` → 29 pass
(named + diff paths); the queue store `parked-tasks.test.ts` (from #2051) → 32
pass, so the enqueue seam `/deprecate-contract` step 6 drives is covered end to
end; `bunx eslint` + `bunx tsc-files --noEmit` clean on all changed TS;
markdownlint clean on all changed docs. Live mainnet smoke runs
(dry-run, nothing proposed): diff path → 0 removals + 8 drift held back; named
path → refused protected `GenericSwapFacet`, flagged `PaxosTransitFacet` with its
4 on-chain selectors, reported a made-up name as not registered. No Solidity
changed → no `forge` impact. The §10 fold is docs-only (no TS/Solidity change).


.agents/commands/multisig-rollout.md
```diff
@@ -101,13 +101,33 @@ deploy mode already executed via `deploy-contract` in Phase 2. For whitelist mod
...
 
-Expect one `pending` proposal per succeeded network with `signatureCount: 1` (the signature added at creation) — **two** per network when a diamond-called periphery's allowlist synced (registration + whitelist). Targets are the chain's `LiFiTimelockController` (proposals wrap in a timelock `scheduleBatch`). Keep `nonce` per network — the PR table needs it. Missing networks here mean the propose step failed even though the deploy succeeded — investigate before continuing; a periphery network showing only one proposal means its allowlist sync didn't land.
+Expect one `pending` proposal per succeeded network with `signatureCount: 1` (the signature added at creation) — **two** per network when a diamond-called periphery's allowlist synced (registration + whitelist), or when Phase 3.5's stale-facet removal ran (deploy/register + removal). Targets are the chain's `LiFiTimelockController` (proposals wrap in a timelock `scheduleBatch`). Keep `nonce` per network — the PR table needs it. Missing networks here mean the propose step failed even though the deploy succeeded — investigate before continuing; a periphery network showing only one proposal means its allowlist sync didn't land.
```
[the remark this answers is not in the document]
0xDEnYO: Fixed in 6dfe56af0. `multisig-rollout.md` Phase 4 now documents the additive case: a network that ran a diamond-called periphery allowlist sync **and** a Phase 3.5 stale-facet removal shows **three** proposals — the doc now says to expect **two or three** per network when either or both apply.

script/deploy/safe/diamondRemovalDiff.ts
```diff
@@ -0,0 +1,427 @@
...
+  names: string[],
+  selectorsOf: (name: string) => `0x${string}`[] = getFunctionSelectors
+): Set<string> {
+  const selectors = new Set<string>()
+  for (const name of names)
+    try {
+      for (const sel of selectorsOf(name)) selectors.add(lower(sel))
+    } catch {
+      // No artifact for this name (deprecated/never-built) — nothing to protect.
+    }
```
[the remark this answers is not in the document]
0xDEnYO: Fixed in 6dfe56af0. `collectActiveSelectors` now **fails closed**: a missing/unreadable artifact for an active (target-state ∩ on-chain) facet throws with a "run forge build" message instead of silently returning an incomplete protected-selector set, so a removal can never sweep out selectors an active facet still owns.

script/tasks/cleanUpProdDiamond.ts
```diff
@@ -429,6 +465,156 @@ const command = defineCommand({
...
+  displayEnvironmentConfiguration(environment, network)
+
+  const removalCalldata = buildDiamondCutRemoveCalldata(diff.removals)
+  const { targetAddress, calldata: finalCalldata } =
+    await prepareTimelockCalldata(
+      removalCalldata,
+      diff.diamondAddress,
+      network,
+      environment
+    )
```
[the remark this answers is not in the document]
0xDEnYO: Fixed in 6dfe56af0. New `assertGovernedProductionRemoval` gate (invoked at the top of `proposeRemovals`) hard-refuses `SEND_PROPOSALS_DIRECTLY_TO_DIAMOND=true` on a production, non-testnet removal and exits 1 — irreversible facet removals must go through Safe + timelock. Staging/testnet keep their sanctioned direct-send.

script/tasks/cleanUpProdDiamond.ts
```diff
@@ -429,6 +465,156 @@ const command = defineCommand({
...
+  for (const network of networkIds)
+    try {
+      await runAutoRemoval(network, environment, yes ? 'yes' : 'dry-run')
+    } catch (err) {
+      consola.error(
+        `[${network}] failed: ${
+          err instanceof Error ? err.message : String(err)
+        }`
+      )
+    }
```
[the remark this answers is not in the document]
0xDEnYO: Fixed in 6dfe56af0. Both fleet sweeps now collect per-network failures and call `exitOnFleetFailures`, which exits non-zero listing the failed networks — a partial sweep surfaces as a red CI check instead of a green pass.

script/tasks/cleanUpProdDiamond.ts
```diff
@@ -166,14 +173,50 @@ const command = defineCommand({
...
+      if (facets)
+        // Named-facet fleet removal (the deprecation-driven path).
+        await runNamedFleetRemoval(env, parseFacetNames(facets), Boolean(yes))
+      else if (auto)
+        // Target-state-diff sweep (backstop for orphans nobody named).
+        await runFleetRemoval(env, Boolean(yes))
+      else {
+        consola.error('--all-networks requires --facets or --auto')
+        process.exit(1)
+      }
```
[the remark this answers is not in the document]
0xDEnYO: Fixed in 34907f5d4. `--auto` + `--facets` is now rejected as mutually exclusive; `--all-networks` with no `--facets` runs the diff sweep (honouring the documented "implies --auto"); and single-network `--auto` is dry-run without `--yes`, matching the fleet sweep and the `--yes` help text (auto modes never submit unconfirmed). Verified the mutual-exclusion guard exits 1 before any RPC.

.agents/commands/deprecate-contract.md
```diff
@@ -18,6 +18,7 @@ This command completely removes one or more contracts (facets or periphery) from
 - Removing source files, deployment scripts, docs, and config entries
 - Updating whitelist for periphery contracts
 - Running test suite to verify remaining tests pass
+- **Parking the on-chain removal** (facets) into the deferred diamond-cleanup queue so the facet is actually removed from the production diamonds that still register it — not just the codebase — when the parked task is later drained into a proposal
```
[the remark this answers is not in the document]
0xDEnYO: Fixed in 34907f5d4. Reworded to: the facet "can eventually be removed … once the parked task is later drained into a proposal **and that proposal executes**" — draining only creates the proposal; on-chain removal happens on execution.

.agents/commands/deprecate-contract.md
```diff
@@ -93,18 +94,60 @@ The command performs these steps in order:
...
+   - **What to park**: for each deprecated facet, one parked task per PRODUCTION
+     network whose **deploy log** (`deployments/<network>.json`) lists it — that
+     log is the authoritative facet → address map. Read the diamond address and
+     the facet address from that log; do **not** delete those entries yet
+     (see step 7).
+   - **Park last, after the deprecation PR exists.** Each parked task **requires**
+     the originating PR URL (`--prUrl`) so the reviewer sees it at signing, and
+     that URL only exists once `gh pr create` has returned it. So the enqueue is
+     the **final action** of this workflow — run it after the deprecation PR is
+     open, passing the real PR URL. One invocation per (facet, network):
```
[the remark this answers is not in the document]
0xDEnYO: Fixed in 34907f5d4. Now reads: the enqueue is the "**last removal-related action** — run it once the deprecation PR is open … and only then continue with steps 7–8 (remaining-occurrences review and final reminders)", so it is no longer called the final action of the whole workflow.

.agents/commands/deprecate-contract.md
```diff
@@ -243,6 +286,11 @@ Please review the above list and indicate which files/occurrences should be remo
 2. ⚠️  Review codebase search results above:
    Action: Clean up any remaining occurrences as needed (deployments, typechain will regenerate, etc.)
 
+3. ⚠️  On-chain removal (step 6): parked into the deferred diamond-cleanup queue
+   for RelayFacet across the PROD diamonds that still register it — drained into a
+   Safe proposal on a later rollout. Do NOT clean deployments/*.json RelayFacet
+   entries until those parked tasks retire (executed/cancelled/superseded).
+
```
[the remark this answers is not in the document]
0xDEnYO: Fixed in 34907f5d4. The "deployments only" option now carries the prerequisite: **never** drop the `deployments/*.json` facet→address entry of a facet whose parked removal task (step 6) is still pending — those stay until the task retires (executed/cancelled/superseded), consistent with steps 6–7.

script/deploy/safe/diamondRemovalDiff.ts
```diff
@@ -0,0 +1,551 @@
...
+    driftDetected: [],
+  }
+  if (!diamondAddress) return empty
+
+  const [onChainFacets, addressToName] = await Promise.all([
+    resolved.getOnChainFacets(diamondAddress, network),
+    resolved.getAddressToName(network, environment),
+  ])
+
+  const expectedNames = resolved.getExpectedNames(network, environment)
```
gvladika, line 454: **[1 · correctness · high] Missing-network target state collapses to "expects zero facets" → can remove live facets.**

`getExpectedFacetNames` returns an empty set both when a network legitimately has zero facets and when it is simply **absent from `_targetState.json`** — the two cases are indistinguishable. `computeFacetRemovalDiff` only bails on a missing diamond *address*, not an empty expected set, so for a diamond-deployed-but-not-yet-in-target-state network, every non-protected on-chain facet becomes a removal candidate.

The only remaining guard is then `getSourceContractNames('src')` (line 423), a **CWD-relative** path: run this from anywhere but the repo root (or with `src` unreadable) and the drift guard silently returns empty → proposes an irreversible removal of in-use facets.

_Fix:_ distinguish "network missing from target state" from "empty facet map" and skip/throw in the former; and resolve `src` from an absolute path rather than `process.cwd()`.

script/deploy/safe/diamondRemovalDiff.ts
```diff
@@ -0,0 +1,551 @@
...
+  const protectedNames = getProtectedNames()
+
+  // Selectors that must never be swept out from under an active facet: the union
+  // owned by EVERY facet target state expects to keep — not only those already
+  // routed on-chain. A replacement facet listed in target state but not yet
+  // registered still owns its selectors, so a deprecated facet currently holding
+  // them must have them held back, not removed. Fails closed on a missing
+  // artifact (see collectActiveSelectors).
+  const activeSelectors = resolved.getActiveSelectors([...expectedNames])
```
gvladika: **[2 · correctness · high] Periphery selectors pollute the active-selector set → incomplete removal, dangling selectors.**

`[...expectedNames]` here is **all** `LiFiDiamond` target-state keys, which include periphery/timelock entries (`GasZipPeriphery`, `Executor`, `LiFiDEXAggregator`, `ReceiverAcrossV4`, …), not just facets. Any selector a deprecated facet actually registers on-chain that shares a signature with one of those periphery ABIs is wrongly *held back* instead of removed.

Concrete case: `GasZipFacet` shares `GAS_ZIP_ROUTER()` (`0x194c869f`) and `getDestinationChainsValue(uint8[])` (`0x46fd98e2`) with `GasZipPeriphery`. The proposed cut removes only the facet's *other* selectors; after execution the dead facet keeps routing those two, and the tool falsely reports "an active facet is expected to own them" when no diamond facet does.

_Fix:_ restrict the active-name set to real facets before calling `getActiveSelectors` (same root cause as finding 5).

script/deploy/safe/diamondRemovalDiff.ts
```diff
@@ -0,0 +1,551 @@
...
+    diamondAddress,
+    removals: [],
+    notFoundOnChain: [],
+    protectedSkipped: [],
+  }
+
+  const foundOnChain = new Set<string>()
+  for (const facet of onChainFacets) {
+    const name = addressToName[lower(facet.address)]
+    if (!name || !requestedNames.has(name)) continue
```
gvladika: **[3 · correctness · high] Named path silently no-ops a facet whose on-chain address isn't in the deploy log.**

Resolution here is solely on-chain-address → deploy-log-name. A facet registered on-chain at an address the log doesn't list (redeploy drift, a pruned/stale entry, a name mismatch) gets `name === undefined`, is skipped, and ends up reported as `notFoundOnChain` ("not registered on <network>"). Unlike the `--auto` path there is **no `unresolved` bucket**, so an on-chain-but-unmapped facet is misreported as "not on chain" with no signal to investigate — the operator believes cleanup succeeded while the deprecated facet stays registered and callable.

The old artifact-name-based `--facets` path (`getFunctionSelectors`) could remove it regardless of the log, so this is a capability regression.

_Fix:_ add an `unresolved`/on-chain-but-unmapped signal to `INamedRemovalResult` and surface it in `printNamedRemoval`.

script/tasks/cleanUpProdDiamond.ts
```diff
@@ -429,6 +457,346 @@ const command = defineCommand({
...
+  removals: IFacetRemoval[],
+  confirmMode: 'yes' | 'prompt' | 'dry-run'
+): Promise<void> {
+  if (removals.length === 0) return
+
+  assertGovernedProductionRemoval(network, environment)
+
+  displayEnvironmentConfiguration(environment, network)
+
+  const removalCalldata = buildDiamondCutRemoveCalldata(removals)
```
gvladika, line 573: **[4 · correctness · medium (plausible)] Timelock race: a stale `Remove` deletes a re-pointed live selector.**

The removal calldata is a loupe **snapshot** taken at proposal time, wrapped in a `scheduleBatch` that executes ≥48h later and is never re-validated at execution. If an intervening rollout re-points one of those selectors onto a new facet inside the delay window, the queued `Remove` (`facetAddress = address(0)`) still deletes that selector → `FunctionDoesNotExist` on every live call until a corrective cut ships.

Inherent to the deferred-removal model, but worth an execution-time guard or a re-validation step before the timelock op is executed.

script/deploy/safe/diamondRemovalDiff.ts
```diff
@@ -0,0 +1,551 @@
...
+export function collectActiveSelectors(
+  names: string[],
+  selectorsOf: (name: string) => `0x${string}`[] = getFunctionSelectors
+): Set<string> {
+  const selectors = new Set<string>()
+  for (const name of names)
+    try {
+      for (const sel of selectorsOf(name)) selectors.add(lower(sel))
+    } catch (error) {
+      throw new Error(
```
gvladika, line 233: **[5 · correctness · medium] One stale target-state entry aborts the entire fleet sweep.**

`collectActiveSelectors` fails closed (throws) when any passed name lacks a built artifact — but it is fed every target-state key (finding 2), including periphery/util names. A single renamed/removed contract left in `_targetState.json`'s `LiFiDiamond` block makes `computeFacetRemovalDiff` throw for **every** network that lists it; `--all-networks --auto` then marks them all failed and `exitOnFleetFailures` exits 1, blocking all intended removals until target state is reconciled — even though the intended removals were unaffected.

Same root cause as finding 2: scoping the active-name set to real facets fixes both.

script/tasks/cleanUpProdDiamond.ts
```diff
@@ -204,56 +259,17 @@ const command = defineCommand({
...
-      await sendOrPropose({
-        calldata: finalCalldata,
+      await runNamedRemoval(
         network,
-        environment: typedEnv,
-        diamondAddress: targetAddress,
-      })
+        typedEnv,
+        parseFacetNames(facets),
+        yes ? 'yes' : 'prompt'
```
gvladika, line 271: **[6 · correctness · medium] Headless `--facets` no longer auto-submits without `--yes`.**

The replaced old `--facets` path called `sendOrPropose` immediately (no confirm, no dry-run). This path defaults to `'prompt'`, and `proposeRemovals` degrades `'prompt'` → `'dry-run'` in a non-TTY. So a cron/script following the old headless runbook — `cleanUpProdDiamond.ts --facets '["X"]'` without `--yes` in a non-interactive shell — silently proposes nothing, with **no non-zero exit** to signal the no-op.

_Fix:_ document the `--yes` requirement for headless use, and/or exit non-zero when a non-TTY prompt degrades to dry-run so the missed submission is visible.

gvladika (commented):
Automated high-effort code review — 6 findings from the removal engine + CLI paths, ranked most-severe first. Each is posted inline below. Findings 1–4 can produce a wrong on-chain proposal (remove-live, incomplete-remove, silent-miss, or a timelock race); 2 and 5 share one root cause (the full target-state key set, periphery included, is fed into facet-only logic).

script/deploy/safe/diamondRemovalDiff.ts
0xDEnYO: Fixed in 05fc99eb1. `getExpectedFacetNames` now returns `undefined` for a network absent from target state (distinct from a present-but-empty block), and `computeFacetRemovalDiff` throws in that case rather than diffing — a fleet loop catches it and marks that network failed. `src/` is now resolved absolutely from the module (`fileURLToPath(import.meta.url)`) instead of `process.cwd()`, so the drift guard can't silently disable itself.

script/deploy/safe/diamondRemovalDiff.ts
0xDEnYO: Fixed in 05fc99eb1. Before computing the held-back set, the expected names are scoped to real facets only (`getFacetSourceNames()` = sources under `src/Facets/`), so periphery/util names (`GasZipPeriphery`, `Executor`, …) never enter `getActiveSelectors`. A deprecated facet's selectors that merely share a signature with a periphery ABI (e.g. `GAS_ZIP_ROUTER()`) are now removed, not held back. Verified against `_targetState.json`: all 29 facet-like keys resolve under `src/Facets/`, so no real facet loses protection.

script/deploy/safe/diamondRemovalDiff.ts
0xDEnYO: Fixed in 05fc99eb1. `INamedRemovalResult` gains an `unresolved: 0x…[]` bucket; `diffNamedFacets` now pushes on-chain-but-unmapped addresses there instead of dropping them, and `printNamedRemoval` surfaces them with a warning to investigate before assuming cleanup succeeded.

script/tasks/cleanUpProdDiamond.ts
0xDEnYO: Fixed in 05fc99eb1. Added the pure `filterRePointedRemovals(snapshot, currentLoupe)` + `revalidateRemovalsOnChain(...)` guard: it re-reads the loupe and drops any snapshotted selector that no longer routes to the doomed facet address (`re-pointed` → live facet, or `already-gone`). Per the deferred-cleanup design, the guard lives at the drain/execute boundary — the generic timelock executor is the wrong layer for removal-specific logic — so the drain consumer (spec §13 follow-up) MUST call it immediately before executing a queued removal and abort/re-propose on any stale selector. Documented as a first-class acceptance criterion in DeferredDiamondCleanupQueue.md.

script/deploy/safe/diamondRemovalDiff.ts
0xDEnYO: Fixed in 05fc99eb1 (same root cause as #2). Scoping the active-name set to real facets keeps a stale periphery/util entry in target state from fail-closing the whole fleet — those names no longer reach `collectActiveSelectors`. A stale *facet* name (renamed facet still in target state, no artifact) still throws, but that's a genuinely broken target state and the throw is isolated per-network by the fleet loop's try/catch, so survivors continue.

script/tasks/cleanUpProdDiamond.ts
0xDEnYO: Fixed in 05fc99eb1. Headless `--facets` in a non-TTY without `--yes` now exits non-zero with a clear message (re-run with `--yes` or in an interactive terminal) instead of silently degrading `prompt`→`dry-run`. The intentional auto/sweep dry-run path is untouched. Updated the `proposeRemovals` docstring and `--yes` help text to match.

gvladika (approved):

PR #2048, opened 2026-07-13
feat(FraxFacet): add Frax HopV2 (LayerZero OFT) bridge facet (EXSC-382) [FraxFacet v1.0.0, IFraxHopV2 v1.0.0]

0xDEnYO:

[EXSC-382](https://linear.app/lifi-linear/issue/EXSC-382/frax-sc-implementation) (sub-task [EXSC-383](https://linear.app/lifi-linear/issue/EXSC-383/create-fraxfacet)) — Frax bridge SC implementation.

`FraxFacet` bridges through **Frax HopV2**, a LayerZero V2 OFT hub-and-spoke bridge (Fraxtal chainId 252 is the hub; every other chain is a spoke sharing one CREATE2 address). It calls `HopV2.sendOFT(oft, dstEid, recipient, amountLd, 0, "")` for the six Frax OFTs (frxUSD, sfrxUSD, WFRAX, frxETH, sfrxETH, FPI).

**One facet source for every chain, incl. Tempo.** Tempo's LayerZero `EndpointV2Alt` rejects native `msg.value` and charges the messaging fee in a TIP20 ERC20 gas token. Rather than a second contract, the fee mode is selected by a constructor immutable (`TIP_FEE_MANAGER`, non-zero only on Tempo, with `PATH_USD` as the default gas token) read from `config/frax.json` at deploy time. This keeps one `@custom:version`, one audit surface, and one Deploy script; the standard-chain runtime cost is an immutable read + one branch (≈ free), the only price being a few-k gas of one-time deploy bloat for the unused branch on 26 chains.

**Money flow (the crux of the review):**
- **Excess native fee** — the LI.FI/Frax API quotes the native fee with a ~10% buffer; HopV2 refunds the excess to the caller (the diamond) synchronously, and the facet forwards it to `refundRecipient` via `refundExcessNative`. This is a hot path, not an edge case.
- **Dust** — HopV2 floors the amount to the OFT's `decimalConversionRate` multiple and only pulls the floored amount. The facet floors up front, bridges exactly that, returns the dust remainder to `refundRecipient` in-tx, and emits `LiFiTransferStarted.minAmount` = the bridged (floored) amount. The diamond retains **zero** token and zero native after every bridge (asserted in tests).
- **No positive slippage / no intent semantics** — `minAmountLD == amountLD`; destination funds go straight to the recipient. No LI.FI contract ever holds funds on the destination.
- **Tempo** — `msg.value == 0`; the fee token is resolved (`TIP_FEE_MANAGER.userTokens(diamond)` → `PATH_USD`), pulled from the caller, approved to the hop, and any unpulled remainder is swept back to `refundRecipient` (invariant holds even across a HopV2 proxy upgrade).

**Plain transfers only.** `hasDestinationCall == true` reverts (`doesNotContainDestinationCalls`); no receiver periphery, and `CalldataVerificationFacet` is unchanged (no compose message to validate on-chain — same as `PaxosTransitFacet`).

**Trust assumptions (documented in `docs/FraxFacet.md`):**
- `dstEid` (the LayerZero routing target) is cross-checked against `bridgeData.destinationChainId` via an owner-governed on-chain `chainId → EID` mapping (`initFrax`/`setFraxChainIdToEid`/`getFraxChainIdToEid`, diamond storage `com.lifi.facets.frax`, mirroring `SupersetFacet`): the facet requires `getFraxChainIdToEid(destinationChainId) == dstEid`, reverting `UnsupportedChainId` when the destination is unmapped and `InformationMismatch` on eid mismatch. Every destination a route targets must be seeded before it is usable — EIDs live in the shared `config/layerzero.json` (single source of truth for all LZ facets), from which the `frax.json` `mappings` are generated and applied by the `UpdateFraxFacet` cut init. This is stricter than the analytics-only model of `AcrossFacetV4`/`PaxosTransitFacet`.

**EID single-sourcing (config).** `config/layerzero.json` is now the one place `chainId → LayerZero EID` is defined; `frax.json`/`superset.json` `mappings` are generated from it (`tasks/syncLayerZeroEids.ts`) and `AcrossV4SwapFacet`'s hardcoded values are pinned to it by `script/tasks/layerZeroEids.test.ts` — so a wrong/stale EID fails CI, not on-chain. The reconciliation caught and corrected two wrong Frax EIDs (sei `30300→30280`, xLayer `30196→30274`, which disagreed with the audited AcrossV4 values); unverified `abstract` was dropped pending confirmation. No facet `.sol` changed (SupersetFacet/AcrossV4 bytecode unchanged — config/test/docs only).
- HopV2 is an upgradeable proxy with an admin `recover()`; the facet grants it the standard `maxApproveERC20` allowance.

**⚠️ Open item for BE / SC review — Tempo fee funding (EXP-514).** On Tempo the diamond must possess the TIP20 fee token for the hop's `transferFrom` pull; the facet pulls it from the caller via `depositAsset`, so the caller must make the fee token available to the diamond. This is my best reversible inference from the verified `RemoteHopV2Tempo` source and needs BE confirmation — it does **not** change the facet ABI, so it can be adjusted without a redesign.

**Testing:** `test/solidity/Facets/FraxFacet.t.sol` — 53 tests, **100% function + effectively full line/statement coverage** on `FraxFacet.sol` (the only uncovered lines under `forge coverage --ir-minimum` are the inline-assembly `_getStorage()` body and the `TIP_FEE_MANAGER` branch that is exercised by both suites — known instrumentation artifacts, same as `SupersetFacet`). Standard path uses an Arbitrum mainnet fork against the real HopV2 + frxUSD self-OFT with live fee quoting; the Tempo path uses `MockFraxHopV2Tempo` (rule 400: Tempo's TIP20/endpoint are precompile-backed and cannot be forked). Facet + deploy scripts compile clean under the deploy toolchain (solc 0.8.17 + london).

**Follow-ups (out of scope here):** SC peer review → BE integration (EXP-514) → external audit → staging then production rollout (multisig + diamondCut + diamond-log update). Aurora is excluded (absent from `config/networks.json`; needs `/add-network`) — tracked separately.

# Peer review round (@gvladika)

Three review findings resolved (commit `7a19f30bb`), plus the staging diamond brought in line with the rename (`4daf2f1e9`):

1. **Selector collision with `SupersetFacet`** — `setChainIdToEid`/`getChainIdToEid` shared byte-identical selectors with `SupersetFacet`, so cutting one facet onto a diamond holding the other silently uninstalled the other's selectors. Renamed to `setFraxChainIdToEid`/`getFraxChainIdToEid` (+ event `FraxChainIdToEidSet`); added rule `102-facets` `[CONV:FACET-SELECTORS]`. The collision had already fired on arbitrum staging (Superset was at 0 selectors), so I redeployed the renamed FraxFacet (`→ 0xCDFE2fc28F6f824F9E619728A52363bd558aB4d0`, old selectors removed from the diamond) and re-cut SupersetFacet (`→ 0x962125b0b3c2329EAd744a0ad383D37F78C608e9`). Loupe-verified: zero selector overlap, both bridges live.
2. **Tempo `unusedFee` underflow** — guarded the fee-sweep subtraction so a HopV2 over-pull degrades to "no refund" instead of an arithmetic panic; added regression test `test_Tempo_FeeOverPullDegradesToNoRefundInsteadOfReverting`.
3. **`_bridgeData` `memory` → `calldata`** on the non-swap `startBridgeTokensViaFrax` (swap entrypoint stays `memory` — it mutates `minAmount`).


0xDEnYO:
### Bot-review triage

**Applied (pushed):**
- **CodeRabbit — demoFrax.ts routing:** fixed. The source-chain Hop is always the arbitrum spoke for both routes (arbitrum is the source), and the demo now validates `oft`/`dstEid`/`recipient`/`amount`/`approvalAddress`/`transactionRequest.to` and **throws** on mismatch instead of warning.
- **CodeRabbit — docs "BridgeData is strictly analytics":** fixed. Clarified that `sendingAssetId`/`receiver`/`minAmount` are enforced; only `transactionId`/`integrator`/`referrer`/`destinationChainId` are descriptive.

**Rejected (with reason):**
- **CodeRabbit — show `_fraxData.dstEid`/`refundRecipient` in the clear-signing entries:** `config/clearSigningProposal.json` is generated by `tasks/buildClearSigningProposal.ts`, which emits only the standard `_bridgeData`/`_swapData` descriptors — no facet displays facet-specific struct fields. A manual edit would be inconsistent with every other facet, get overwritten on regeneration, and **fail the `verify-clear-signing` CI** that asserts committed == generated. The clear-signing surface is intentionally generator-driven.
- **Olympix — "reentrancy: event emitted after external call" (FraxFacet.sol, `LiFiTransferStarted`):** intentional and mandated by `[CONV:EVENTS]` (emit at the end of `_startBridge` after the bridge call), and the entrypoints are `nonReentrant`. Same pattern as every LI.FI facet.
- **Olympix — "unsafe downcast":** false positive. The flagged expression is `bytes32(uint256(uint160(receiver)))` — all widening; there is no narrowing cast in the facet.

Note: `audit-verification` failing + the auto-`draft` / `AuditRequired` label are the expected new-facet gate; the external audit is a downstream step (SC review → BE → audit → prod rollout).

src/Facets/FraxFacet.sol
```diff
@@ -0,0 +1,317 @@
...
+            HOP.sendOFT{ value: _fraxData.nativeFee }(
+                _fraxData.oft,
+                _fraxData.dstEid,
+                recipient,
+                flooredAmount,
+                0,
+                ""
+            );
+        } else {
+            _sendViaTempo(_fraxData, recipient, flooredAmount);
```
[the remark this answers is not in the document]
0xDEnYO: Resolved in `_sendViaTempo` (commit 9efefa0): the facet now rejects the collision up front — `if (feeToken == _sendingAssetId) revert InformationMismatch();` — before the balance snapshot, so the underflow path can't be reached. A Frax gas token is never a bridged Frax OFT, so rejecting is the correct handling rather than special-casing the baseline. Also covered by a new test: `testRevert_Tempo_WhenFeeTokenEqualsBridgedToken`.

tasks/syncLayerZeroEids.ts
```diff
@@ -0,0 +1,97 @@
...
+
+// Facets that seed a chainId -> EID mapping, and the config key whose network
+// names define the destinations that facet supports.
+export const CONSUMERS: { file: string; networksKey: string }[] = [
+  { file: 'frax.json', networksKey: 'hop' },
+  { file: 'superset.json', networksKey: 'poolManager' },
+]
+
+export function readJson<T>(file: string): T {
+  return JSON.parse(fs.readFileSync(path.join(CONFIG, file), 'utf8')) as T
```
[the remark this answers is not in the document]
0xDEnYO: @AikidoSec ignore: False positive (path_traversal_scripts) — internal build-time task under tasks/, not a network service and never HTTP-exposed. The `file` argument to readJson is only ever hardcoded literals ('layerzero.json', 'networks.json') and the version-controlled CONSUMERS allow-list ('frax.json', 'superset.json') derived from the repo itself — never untrusted external input. Runs locally / in CI on version-controlled repo files only.

src/Facets/FraxFacet.sol
```diff
@@ -0,0 +1,445 @@
...
+        }
+
+        s.chainMappingsInitialized = true;
+
+        emit FraxChainMappingsInitialized(_chainIdConfigs);
+    }
+
+    /// @notice Adds or updates chainId -> LayerZero EID entries (owner-only).
+    /// @param _chainIdConfigs Batch of `{chainId, lzEid}` entries.
+    function setChainIdToEid(
```
gvladika: **Selector collision with `SupersetFacet` — deploying this facet silently uninstalls a live bridge facet**

`setChainIdToEid((uint256,uint32)[])` (`0x5b1ee840`) and `getChainIdToEid(uint256)` (`0xedc3f41e`) use the exact same signatures `SupersetFacet` already exposes. This whole chainId→EID subsystem is copy-pasted from `SupersetFacet`.

Both facets are registered on the arbitrum staging diamond. When the second is cut in, `UpdateScriptBase.buildDiamondCut` sees these two selectors already owned by the other facet, pushes them to `selectorsToReplace`, and sets `oldFacet=<other facet>`. The block at `UpdateScriptBase.sol:162-177` then loads **all** of that other facet's selectors and `Remove`s every one not present in the new facet's set. Net effect: deploying `FraxFacet` onto a diamond that has `SupersetFacet` reroutes these two selectors to `FraxFacet` **and** removes `startBridgeTokensViaSuperset` / `swapAndStartBridgeTokensViaSuperset` / `initSuperset` — silently uninstalling a live bridge facet (and vice-versa depending on deploy order). `UpdateFraxFacet.getExcludes()` only excludes `initFrax`, so it doesn't guard against this.

Recommend renaming these selectors to be Frax-specific (e.g. `setFraxChainIdToEid` / `getFraxChainIdToEid`) so the two facets can coexist on one diamond.

src/Facets/FraxFacet.sol
```diff
@@ -0,0 +1,445 @@
...
+            _fraxData.oft,
+            _fraxData.dstEid,
+            _recipient,
+            _amount,
+            0,
+            ""
+        );
+
+        uint256 unusedFee = IERC20(feeToken).balanceOf(address(this)) -
+            feeTokenBalanceBefore;
```
gvladika: **Potential underflow revert in the `unusedFee` sweep**

`feeTokenBalanceBefore` is snapshotted *before* depositing the quoted `feeAmount`. This subtraction assumes the post-`sendOFT` balance is always `>= feeTokenBalanceBefore`, i.e. that HopV2 never pulls more of the fee token than `quoteStatic` reported.

The comment at lines 400-403 explicitly anticipates HopV2's fee logic changing on a proxy upgrade. If `HOP.sendOFT` ever pulls **more** fee token than was quoted, and the diamond holds any incidental pre-existing balance of `feeToken` to cover the excess, then `balanceOf(this) < feeTokenBalanceBefore` and this checked subtraction underflows — reverting a Tempo bridge that HopV2 would otherwise complete, with an arithmetic panic.

Consider guarding the subtraction (e.g. only sweep when `balanceAfter > feeTokenBalanceBefore`) so an over-pull degrades gracefully instead of reverting.

src/Facets/FraxFacet.sol
```diff
@@ -0,0 +1,445 @@
...
+        if (lzEid == 0) revert UnsupportedChainId(_chainId);
+    }
+
+    /// External Methods ///
+
+    /// @notice Bridges tokens via Frax HopV2
+    /// @param _bridgeData The core information needed for bridging
+    /// @param _fraxData Data specific to Frax HopV2
+    function startBridgeTokensViaFrax(
+        ILiFi.BridgeData memory _bridgeData,
```
gvladika: **`_bridgeData` should be `calldata`, not `memory`**

The non-swap entrypoint declares `_bridgeData` as `ILiFi.BridgeData memory`, whereas the analogous `SupersetFacet.startBridgeTokensViaSuperset` takes it as `calldata`. Nothing mutates `_bridgeData` on this path before it reaches `_startBridge` (which takes its own `memory` copy), so every non-swap bridge pays for an unnecessary full calldata→memory copy of the `BridgeData` struct — extra gas on the hot path with no functional benefit.

Suggest changing to `ILiFi.BridgeData calldata _bridgeData` to match the sibling facet.

src/Facets/FraxFacet.sol
0xDEnYO: Fixed in 7a19f30bb (rename) + verified on arbitrum staging in 4daf2f1e9.

Renamed the two colliding functions and the event to Frax-specific names so both facets coexist:
- `setChainIdToEid` → `setFraxChainIdToEid` (`0x5b1ee840` → `0xa8e13b68`)
- `getChainIdToEid` → `getFraxChainIdToEid` (`0xedc3f41e` → `0xc2974580`)
- event `ChainIdToEidSet` → `FraxChainIdToEidSet`

You were right that it had already fired on staging: SupersetFacet was sitting at **0 selectors** (fully uninstalled) on the arbitrum staging diamond. I redeployed the renamed FraxFacet (`0x3B76…` → `0xCDFE2fc28F6f824F9E619728A52363bd558aB4d0`), whose diamondCut removed the old `0x5b1ee840`/`0xedc3f41e` from the diamond, then re-cut SupersetFacet (`0x54AE…` → `0x962125b0b3c2329EAd744a0ad383D37F78C608e9`). Verified via the diamond loupe: the two facets now share **zero** selectors and both bridge entrypoints resolve. Also added rule `102-facets` `[CONV:FACET-SELECTORS]` so facet-scoped admin/config functions must carry the facet identifier.

src/Facets/FraxFacet.sol
0xDEnYO: Fixed in 7a19f30bb. Guarded the subtraction so an over-pull degrades to "no refund" instead of an arithmetic panic:

```solidity
uint256 feeTokenBalanceAfter = IERC20(feeToken).balanceOf(address(this));
uint256 unusedFee = feeTokenBalanceAfter > feeTokenBalanceBefore
    ? feeTokenBalanceAfter - feeTokenBalanceBefore
    : 0;
```

Added a regression test `test_Tempo_FeeOverPullDegradesToNoRefundInsteadOfReverting` that seeds a pre-existing fee-token balance and configures the mock hop to pull more than quoted, asserting the bridge does not revert and nothing is swept.

src/Facets/FraxFacet.sol
0xDEnYO: Fixed in 7a19f30bb — changed the non-swap `startBridgeTokensViaFrax` `_bridgeData` from `memory` to `calldata`, matching `SupersetFacet`. Left the swap entrypoint as `memory` since it mutates `minAmount` via `_depositAndSwap`.

gvladika (dismissed):

gvladika (dismissed):

gvladika (approved):

PR #2052, opened 2026-07-14
fix(safe): reach Safe Mongo via lifi-connect, drop retired-VPN check (EXSC-615)

0xDEnYO:

[EXSC-615](https://linear.app/lifi-linear/issue/EXSC-615/safedeploy-scripts-reach-safe-mongodb-via-lifi-connect-legacy-vpn)

The legacy OpenVPN was retired. `getSafeMongoCollection()` gated all Safe/deploy
Mongo access on the caller's **public IP matching the old VPN egress address**
(`18.195.61.255`). With the VPN gone that check can never pass, so it
hard-blocked every script that reads/writes the Safe proposal database
(`propose-to-safe`, `confirm-safe-tx`, `list-pending-proposals`, the rollout
skills) — for the whole SC team, not one machine.

The database is now reachable only through the **`lifi-connect`** port-forward
tunnel. So instead of an IP check I:

- **Removed the VPN IP check** and connect with a **10s `serverSelectionTimeoutMS`**
  so a down tunnel fails fast (vs. the driver's default 30s hang) with an
  **actionable** error that names `lifi-connect prod smart-contracts` and links
  the DevOps setup guide.
- **Updated `list-pending-proposals.ts`** so its "recoverable misconfig" exit
  code (2) still fires — keyed on `lifi-connect` instead of the old
  `VPN connection required` string.
- **Documented `SC_MONGODB_URI`** in `.env.example` (tunnel + 1Password creds).
- **Added `script/deploy/safe/with-safe-tunnel.sh`** and routed the Mongo-touching
  Safe package scripts (`confirm-safe-tx`, `propose-safe-tx`,
  `add-safe-owners-and-threshold`, `unpause-all-diamonds`, `execute-timelock`)
  through it (`with-safe-tunnel.sh && <cmd>`, so user args pass through). It
  ensures the tunnel is up before the command runs — auto `aws sso login` when
  the session expired, starts `lifi-connect` (visible Terminal window on macOS,
  else detached), waits for the port — so `bun confirm-safe-tx` just works.
  Tunnel-starting lives in that one wrapper and is always **visible** (announced +
  its own window); the TS signing logic never opens a tunnel itself. Also guards
  against a CA-bundle env var pointing at a missing file, and documents the
  required `directConnection=true` in `SC_MONGODB_URI` (the server-selection-timeout fix).
- **Consolidated setup docs.** Added `docs/Setup.md` as the single setup guide
  (toolchain → Cursor IDE → resource access), moved the README's Getting Started
  detail into it (README keeps a minimal quick-start + pointer), and split the
  agent-specific guidance into `docs/Setup-agents.md`. The Notion doc remains the
  canonical DevOps source for resource access.

No behaviour change for a correctly-configured environment: the happy path is
unchanged, only the pre-flight and its failure mode.

- [ ] This pull request is as small as possible and only tackles one problem — the code fix is small; the PR also bundles setup-doc consolidation and the opt-in `safe:tunnel` wrapper (folded in at the author's request)


script/deploy/safe/with-safe-tunnel.sh
```diff
@@ -0,0 +1,147 @@
...
+  activate
+end tell
+EOF
+    echo "  → opened a new Terminal window running the tunnel"
+  else
+    local log="${TMPDIR:-/tmp}/lifi-connect-safe.log"
+    nohup "$bin" prod smart-contracts >"$log" 2>&1 &
+    echo "  → started in background (PID $!, logs: ${log}); stop with: kill $!"
+  fi
+}
```
gvladika, line 76: This wrapper contradicts the agent policy this same PR introduces. `docs/Setup-agents.md` says opening a **production** tunnel is human-only, "gated behind explicit human approval." But `with-safe-tunnel.sh` now sits in the default path of `bun propose-safe-tx` / `confirm-safe-tx` / `execute-timelock` / etc., and in a non-interactive shell — precisely an agent's environment, where `osascript`/Terminal isn't available — `start_tunnel` falls to the `nohup … &` branch and opens a **detached** prod tunnel. That's the opposite of the "always visible" guarantee stated in the header comment and in the docs, and it means any agent running a wrapped bun script auto-opens the prod tunnel.

Suggestion: gate auto-*start* on an interactive TTY. Keep the "port already up" pass-through (that's free), but when stdout isn't a TTY, don't start — print the same fail-fast guidance `getSafeMongoCollection()` gives and let the human start it:

```bash
if ! [ -t 1 ]; then
  echo "Safe Mongo tunnel (localhost:${port}) is down and this is a non-interactive shell." >&2
  echo "Start it yourself: lifi-connect prod smart-contracts (see docs/Setup.md)." >&2
  exit 1
fi
```

That makes the wrapper *enforce* the policy the docs only describe, rather than quietly violating it.

0xDEnYO: Fixed in 9442ae6 — auto-start is now gated on an interactive TTY.

The "port already up" pass-through stays free for everyone (agents included). When stdout is not a TTY and the tunnel is down, the wrapper no longer falls to the detached `nohup` branch — it prints the same fail-fast, human-only guidance `getSafeMongoCollection()` surfaces (`lifi-connect prod smart-contracts`, see docs/Setup.md) and exits 1, so an agent can never auto-open the prod tunnel. Verified: in a non-interactive shell with the tunnel down, `start_tunnel` is never invoked and it exits 1; with the tunnel already up it still passes through. Header comment updated to match.

gvladika (approved):

package.json
```diff
@@ -69,15 +69,16 @@
     "build:typechain-and-abi": "bun typechain:incremental && bun abi:generate:incremental",
     "abi:removeDuplicateEvents": "bunx tsx ./script/removeDuplicateEventsFromABI.ts",
     "add-network-rpc": "bunx tsx ./script/mongoDb/add-network-rpc.ts",
-    "add-safe-owners-and-threshold": "bunx tsx ./script/deploy/safe/add-safe-owners-and-threshold.ts",
+    "add-safe-owners-and-threshold": "bash script/deploy/safe/with-safe-tunnel.sh && bunx tsx ./script/deploy/safe/add-safe-owners-and-threshold.ts",
     "codegen": "plop",
     "compile": "forge build",
-    "confirm-safe-tx": "bun build:typechain-and-abi && bunx tsx ./script/deploy/safe/confirm-safe-tx.ts",
+    "confirm-safe-tx": "bash script/deploy/safe/with-safe-tunnel.sh && bun build:typechain-and-abi && bunx tsx ./script/deploy/safe/confirm-safe-tx.ts",
+    "safe:tunnel": "bash script/deploy/safe/with-safe-tunnel.sh",
```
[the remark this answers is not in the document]
0xDEnYO: Keeping `safe:tunnel` as-is — this is a false positive for this path.

The snake_case config-key rule applies to JSON **config** keys (e.g. `config/*.json`), not to `package.json` npm run-script identifiers. Colon-namespacing is the idiomatic npm grouping convention and is the dominant style in this file: 28 of 56 script keys already use it, including the directly analogous Mongo tooling namespace (`mongo-logs:add`, `mongo-logs:sync`, …), plus `lint:fix`, `test:ts`, `abi:generate`, `typechain:incremental`, etc. Renaming to `safe_tunnel` would make this one key inconsistent with the rest of the file.

`safe:tunnel` intentionally groups the opt-in tunnel helper alongside the other `safe`-domain scripts, so it stays.

PR #2053, opened 2026-07-15
feat(PolymerCCTPFacet): add Stellar as CCTP destination (EXSC-616)

0xDEnYO:

[EXSC-616](https://linear.app/lifi-linear/issue/EXSC-616/add-stellar-as-cctp-destination-to-polymercctpfacet-smart-contract)

Adds **Stellar** (non-EVM, CCTP domain 27) as a destination to `PolymerCCTPFacet`, porting the change prototyped in polymerdao/lifi-contracts@[545cac6](https://github.com/polymerdao/lifi-contracts/commit/545cac6f2b54faac8e4c7bb42aff8ed58ca7034d) ("add stellar support to polymerfacet").

> **⚠️ Stacked on #1652.** The fork commit is built on top of the HyperCore work (EXSC-169 / #1652), and reuses `depositForBurnWithHook` + the `hookData` field introduced there. This PR therefore targets `polymer-hypercore-ex-286` (the #1652 branch), **not `main`** — it cannot compile against `main`. Rebase onto `main` once #1652 merges.

## How Stellar works here

A Stellar `G...` account can never be a CCTP `mintRecipient` directly: CCTP stores only the raw 32 bytes without the strkey type prefix, so the protocol assumes the recipient is a contract, and USDC minted to a bare account is unrecoverable. All Stellar deposits therefore mint to Circle's pinned `CctpForwarder` (`mintRecipient == destinationCaller == STELLAR_CCTP_FORWARDER`), which alone may execute the message and forwards the USDC to the strkey recipient carried in the hook data. The raw account is emitted via `BridgeToNonEVMChainBytes32` for off-chain relayer tracking. Ref: https://developers.circle.com/cctp/references/stellar-contracts

- `validatePolymerData` corridor dispatch: a new `LIFI_CHAIN_ID_STELLAR` arm requires `receiver == NON_EVM_ADDRESS`, `nonEVMReceiver != 0`, `hookData.length >= 32`, and that the hook's self-declared payload length (`hookData[28:32]`) equals `hookData.length - 32` (rejects truncated/padded hooks).
- `_startBridge`: a new Stellar arm calls `depositForBurnWithHook` with the pinned forwarder and emits the non-EVM event.

## Deviations from the fork prototype (deliberate)

1. **Forwarder pinned to MAINNET.** The fork hardcoded the Stellar **testnet** forwarder (`CA66Q2…` = `0x3de8…b93e`). On a production deploy that would mint Stellar-bound USDC to a nonexistent recipient on Stellar mainnet → **fund loss**. This PR pins the **mainnet** forwarder `CBZL2IH7…DF5T` = `0x72bd20ff…f291`, verified against Circle's docs and by base32-decoding the strkey. No testnet value is carried in the facet.
2. **No version bump.** This change merges into the in-flight #1652 development line (facet `3.0.0`, `LiFiData 1.0.1`), which is versioned when that line finalizes — so no bump here.
3. **Config minimalism.** Only the `{ chainId: 1201081091099710, domainId: 27 }` mapping is added. The fork's ~130-line `testnets` block is fork-specific and not brought upstream.

## Trust assumption to review (⚠️ for audit)

Unlike HyperCore — where the facet validates `hookData[32:52] == bridgeData.receiver` so events can't misreport the beneficiary — Stellar's strkey recipient **cannot** be validated on-chain against `bridgeData.receiver` (it isn't a 20-byte EVM address), and it is **not** cross-checked against the emitted `nonEVMReceiver`. The facet only enforces the hook's internal length consistency. Correctness of the Stellar recipient therefore relies on trusted (LI.FI API) calldata generation.

This is called out as a dedicated **`SECURITY — UNVALIDATED RECEIVER (Stellar)`** note in the facet NatSpec (per rule 102 [CONV:FACET-REQS], opaque receiver). Suggested off-chain invariant: the relayer/backend should verify the strkey in `hookData` corresponds to the emitted `nonEVMReceiver`.

## Follow-ups (out of scope)

- `LIFI_CHAIN_ID_STELLAR = 1201081091099710` decodes to `"xlm"` + the shared non-EVM suffix (consistent with `SOLANA = "sol"` + suffix); please confirm it matches the backend chain-id registry.
- A Stellar path in `demoPolymerCCTP.ts` (the TS demo already carries `hookData`; no Stellar example added here).

### DIFF
--- a/src/Facets/PolymerCCTPFacet.sol
+++ b/src/Facets/PolymerCCTPFacet.sol
@@ -27,12 +27,16 @@ import { CannotBridgeToSameNetwork, InvalidAmount, InvalidCallData, InvalidConfi
 ///      BridgeData.receiver is the NON_EVM_ADDRESS sentinel and the real recipient travels as a
 ///      Stellar strkey inside hookData, because a Stellar account can never be a CCTP mintRecipient
 ///      directly (see STELLAR_CCTP_FORWARDER). USDC is minted to the pinned Stellar forwarder,
-///      which forwards it to that strkey. Unlike HyperCore, the strkey is not a 20-byte EVM address,
-///      so it CANNOT be validated on-chain against BridgeData.receiver, nor against the
-///      nonEVMReceiver emitted in BridgeToNonEVMChainBytes32; the facet only enforces hookData's
-///      internal length consistency. Correctness of the Stellar recipient therefore relies on
-///      trusted (LI.FI API) calldata generation — integrators must treat it as not enforced on-chain.
-/// @custom:version 3.1.0
+///      which forwards it to that strkey.
+/// @dev SECURITY — UNVALIDATED RECEIVER (Stellar): unlike the HyperCore corridor, the Stellar
+///      strkey is not a 20-byte EVM address, so the facet CANNOT verify it on-chain against
+///      BridgeData.receiver, and it is NOT cross-checked against the nonEVMReceiver emitted in
+///      BridgeToNonEVMChainBytes32. The facet only enforces hookData's internal length consistency
+///      (see _startBridge). The strkey that actually receives the funds and the nonEVMReceiver used
+///      for off-chain tracking can therefore diverge. Correct routing relies entirely on trusted
+///      (LI.FI API) calldata generation — integrators MUST treat the Stellar receiver as NOT
+///      enforced on-chain.
+/// @custom:version 3.0.0
 contract PolymerCCTPFacet is
     ILiFi,
     ReentrancyGuard,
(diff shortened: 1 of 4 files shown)


src/Facets/PolymerCCTPFacet.sol
```diff
@@ -303,6 +331,29 @@ contract PolymerCCTPFacet is
             ) {
                 revert InvalidReceiver();
             }
+        } else if (_bridgeData.destinationChainId == LIFI_CHAIN_ID_STELLAR) {
```
gvladika, line 334: nit: the corridor-dispatch comment just above this arm (lines 310–312) is now stale:

> Corridor dispatch: HyperCore requires a forwarder hook; hooks toward any other destination are unsupported. Add new corridors (e.g. Stellar) as additional arms.

With this PR Stellar *is* an implemented arm that accepts hooks, so both sentences no longer match the code ("hooks toward any other destination are unsupported" and "e.g. Stellar" as a future example). Suggest updating to something like:

```solidity
// Corridor dispatch: HyperCore and Stellar require a forwarder hook; hooks
// toward any other destination are unsupported. Add new corridors as
// additional arms.
```

0xDEnYO: Good catch — updated the corridor-dispatch comment to name Stellar alongside HyperCore and dropped the now-implemented "e.g. Stellar" example. Fixed in 949339db5.

gvladika (approved):

PR #2055, opened 2026-07-15
feat(LibBytes): add address<->bytes32 conversion helpers (EXSC-618) [LibBytes v1.1.0]

0xDEnYO:

[EXSC-618](https://linear.app/lifi-linear/issue/EXSC-618/sc-libbytes-v110-addressbytes32-conversion-helpers)

The `bytes32(uint256(uint160(addr)))` / `address(uint160(uint256(b)))` conversion pattern is copy-pasted across ~10 facets, which triggers Olympix `unsafe-downcast` findings and risks divergent semantics. The only prior implementation lived in the vendored `OFTComposeMsgCodec` (a LayerZero copy), so it was not a suitable shared dependency.

**LibBytes v1.1.0** now centralises it:

- `toBytes32(address)` — widening, always lossless.
- `toAddress(bytes32)` — **checked** narrowing: reverts `NotAnAddress` if the top 96 bits are set. This is the default; a checked narrowing silences the detector honestly rather than hiding a truncation. New overload alongside the existing `toAddress(bytes memory, uint256)`.
- `toAddressUnchecked(bytes32)` — raw truncation, for intentional / pre-guarded cases.
- Drive-by: `toHexString`'s `require`-string → custom error `HexLengthInsufficient` (removes the `gas-custom-errors` solhint-disable).
- Unit tests: round-trip, zero, checked-revert, unchecked-truncation.

> **Note on the `toHexString` drive-by (transitive bytecode change):** `toHexString` is `internal pure` and inlined into its sole caller `SquidFacet` (`SquidFacet.sol:187`), so swapping the `require`-string for `revert HexLengthInsufficient()` changes SquidFacet's compiled bytecode and revert-encoding even though `SquidFacet.sol` isn't edited here. The affected branch is **unreachable for a correctly-sized `length`** (after the loop `value` is always 0, and SquidFacet always passes a correct length), so there is **no reachable behavior change** — hence no `@custom:version` bump or re-audit for SquidFacet (per `[CONV:VERSIONING]` this is a PATCH-class change that does not cascade). Flagged so whoever next deploys SquidFacet knows its 1.0.0 bytecode will differ from the currently-deployed 1.0.0.

**Call sites are intentionally NOT migrated in this PR.** Each of the 21 existing sites gets a breadcrumb comment only (`// TODO(EXSC-626): migrate to LibBytes.…`, tracked in [EXSC-626](https://linear.app/lifi-linear/issue/EXSC-626/sc-migrate-the-21-addressbytes32-conversion-sites-to-libbytes-v110)) — no logic change, no facet `@custom:version` bump, no re-audit. Each is swapped to the helper the next time its facet is touched. Audit of the sites:

- **9 widening** → `toBytes32` (zero behavior change).
- **6 checked narrowing** → `toAddress`: all are validation/compare paths that already revert on a bad value, so checked only changes *which* revert fires — never turns a revert into success.
- **6 → `toAddressUnchecked`**: `CalldataVerificationFacet` (a `view` returning `bool` — checked would turn a legitimate `return false` into a revert) and the `AcrossFacetPackedV4` packed hot paths (two of which already manually guard the high bits before branching to `NON_EVM`).

Numeric downcasts (`uintN` — the larger source of findings) are tracked separately in [EXSC-617](https://linear.app/lifi-linear/issue/EXSC-617/sc-adopt-safecast-for-numeric-downcasts-add-convdowncast-rule) (SafeCast + a `[CONV:DOWNCAST]` rule).


src/Libraries/LibBytes.sol
```diff
@@ -121,8 +123,33 @@ library LibBytes {
             buffer[i] = _SYMBOLS[value & 0xf];
             value >>= 4;
         }
-        // solhint-disable-next-line gas-custom-errors
-        require(value == 0, "Strings: hex length insufficient");
+        if (value != 0) revert HexLengthInsufficient();
```
gvladika, line 126: **`toHexString` change silently alters `SquidFacet` bytecode with no version bump.**

`LibBytes` functions are `internal pure` and get inlined into consumers. `SquidFacet.sol:187` is the sole caller of `toHexString`, so swapping the `require`-string for `revert HexLengthInsufficient()` changes SquidFacet's compiled bytecode and its revert data — yet SquidFacet stays at `@custom:version 1.0.0` and isn't listed among the "call sites intentionally not migrated". The PR body implies zero facet behavior changes, but this one does change.

Per `[CONV:VERSIONING]`, "new revert on existing function" is a MAJOR-type change, and this is a revert-encoding change on a shipped function. In practice the branch is an unreachable invariant (after the loop `value` is always 0 for a correctly-sized input), so behavioral/security risk is effectively nil — but it should be surfaced rather than folded silently into a "no facets touched" PR.

Suggestion: either split the `toHexString` refactor into its own PR acknowledging the SquidFacet impact, or note in the PR body that SquidFacet's bytecode changes and confirm with audit/deploy that an unbumped version is acceptable here.

src/Facets/AcrossFacetV4.sol
```diff
@@ -265,6 +265,7 @@ contract AcrossFacetV4 is
     function _convertAddressToBytes32(
         address _address
     ) internal pure returns (bytes32) {
+        // TODO: migrate to LibBytes.toBytes32 — see LibBytes v1.1.0
```
gvladika: **Bare `// TODO` breadcrumbs.**

The 21 breadcrumbs reference "LibBytes v1.1.0" but not the tracking ticket (EXSC-618). Un-ticketed `TODO`s tend to become permanent — consider referencing the ticket ID so they're greppable and closable, e.g. `// TODO(EXSC-618): migrate to LibBytes.toBytes32`. Also worth confirming the repo's solhint config doesn't flag `TODO`, and noting this runs against the repo's "minimal comments" default (though the migration-breadcrumb rationale is defensible).

src/Libraries/LibBytes.sol
0xDEnYO: Not bumping SquidFacet's version here — but you're right that the transitive bytecode change should be surfaced, so I've added a note to the PR description.

Reasoning: `[CONV:VERSIONING]` says a library MAJOR bump should be *evaluated* against consumers, not that any library change bumps them. The `toHexString` edit isn't a "new revert on existing function" — the revert already existed (`require(value == 0, …)`); only the revert *encoding* changed (`Error(string)` → `HexLengthInsufficient()`). And that branch is unreachable for a correctly-sized `length` (after the loop `value` is always 0, and SquidFacet always passes a correct length), so there's **zero reachable behavior change at the ABI level** → PATCH-class, which doesn't cascade. So SquidFacet stays 1.0.0 and LibBytes' MINOR bump (the three new functions dominate) is correct.

The only real action item was your "surface it rather than fold it silently" point — done: the description now flags that SquidFacet's compiled bytecode/revert-encoding changes (dead-code only) so whoever next deploys it expects the 1.0.0 bytecode to differ.

src/Facets/AcrossFacetV4.sol
0xDEnYO: Done in 0fbe4b5f9 — created a dedicated tracking ticket ([EXSC-626](https://linear.app/lifi-linear/issue/EXSC-626/sc-migrate-the-21-addressbytes32-conversion-sites-to-libbytes-v110)) and rewrote all 21 breadcrumbs to `// TODO(EXSC-626): migrate to LibBytes.…` so they're greppable and closable.

Used a new ticket rather than EXSC-618 since EXSC-618 closes with this PR, whereas the migrations outlive it (each rides its facet's next version bump). `.solhint.json` has no `max-line-length` rule, so the longer comments are fine.

gvladika (dismissed):

gvladika (dismissed):

0xDEnYO:
Resolved merge conflicts with `main` in `f2e2ad9`.

- **Conflict:** `audit/auditLog.json` — kept both `audit20260720` (LibBytes v1.1.0) and `audit20260722` (MayanFacet v2.0.0).
- **Pre-commit unblocker:** `script/demoScripts/demoMayan.ts` (from #2042 on main) had invalid `await` + a 9th `apiKey` arg against the current `@mayanfinance/swap-sdk` signature; aligned calls so lint/tsc pass.

Auto-merge is still enabled — needs a fresh core-dev approval after the merge commit dismissed prior reviews.

gvladika (dismissed):

gvladika (approved):

PR #2056, opened 2026-07-15
feat(skills): add manage-wallet-funds, replace send-deployer-funds

0xDEnYO:

Fixes [EXSC-619](https://linear.app/lifi-linear/issue/EXSC-619/sc-manage-wallet-funds-skill-bridgeswapsend-from-any-env)

The design is organised around one invariant: custody. A move that keeps funds inside a single wallet cannot lose them to a third party, so `bridge` (cross-chain native) and `swap` (same-chain native↔ERC-20) run autonomously, gated only by a mechanical `from == to == source wallet` assertion; `send` changes custody and therefore refuses to broadcast without an explicit human `--confirm`, folding in the old `send-deployer-funds` behaviour so there is one tool instead of two overlapping ones. Routing goes through the LI.FI API only (dogfooding our own product), with an up-front supported-chain check so a not-yet-indexed network aborts with a clear message rather than hanging — the deliberate trade-off of the API-only choice. Wallets resolve by role or raw address via a new *additive* `config/global.json` `walletKeys` registry (the existing role→address strings are untouched, so no consumer breaks) plus an `.env` scan fallback for scratch keys, and the key-derived address is cross-checked against `global.json` so a misplaced key surfaces before funds move. The decision logic (resolution, same-wallet gate, slippage, chain guard) lives in a separate helper module so it is unit-tested without ever broadcasting; the engine reuses the existing viem/gas patterns from `moveNativeFundsToNewWallet.ts`. Value loss is capped at 3% (USD-measured) and bridge delivery is tracked asynchronously via the LI.FI status endpoint.


script/tasks/manageWalletFunds.ts
```diff
@@ -0,0 +1,668 @@
...
+  // predeploy (e.g. arc), a value transfer would move the wrong thing, so refuse.
+  const nativeAddr = (net.nativeAddress ?? '').toLowerCase()
+  const nativeSentinels = new Set([
+    '',
+    '0x0000000000000000000000000000000000000000',
+    NATIVE_SENTINEL.toLowerCase(),
+  ])
+  if (nativeAddr && !nativeSentinels.has(nativeAddr))
+    throw new Error(
+      `send does not yet support ERC-20-predeploy gas on "${networkName}" (native asset at ${net.nativeAddress}). ` +
```
gvladika: **`send` guard misses tempo-style ERC-20-gas chains.**

This blocks ERC-20-predeploy gas by inspecting `net.nativeAddress`, which correctly refuses **arc** (`nativeAddress: 0x3600…`). But **tempo** signals ERC-20 gas differently: `nativeCurrency: "N/A"`, `nativeAddress: 0x0000…0000` (which is in `nativeSentinels`, so it passes this guard), and gas is actually paid via `feeTokenAddress`. So `send` on tempo falls through to a plain native `value` transfer on a chain with no native currency — a wasted/failing broadcast. The same blind spot lets `bridge`/`swap` request a `native` quote on such chains (less severe — the API rejects it remotely).

Suggest also refusing when `nativeCurrency === 'N/A'` or `feeTokenAddress` is set. Better: centralize "does this chain have a real native asset?" into one helper in `walletFundsHelpers.ts` and unit-test it, since three code paths need the same answer.

script/tasks/manageWalletFunds.ts
```diff
@@ -0,0 +1,668 @@
...
+      slippage: maxSlippage / 100,
+    })
+
+    // Same-wallet boundary: the quote is requested with toAddress == fromAddress == our
+    // wallet; verify the returned route echoes that, so any recipient drift is caught
+    // before signing.
+    if (quote.action.fromAddress)
+      assertSameWallet(wallet.address, quote.action.fromAddress)
+    if (quote.action.toAddress)
+      assertSameWallet(wallet.address, quote.action.toAddress)
```
gvladika: **Same-wallet gate is weaker than the docstring claims.**

The gate only fires when the API echoes the field, so if LI.FI omits `action.toAddress` the check is silently skipped — and it never inspects the actual `transactionRequest.data`, where a bridge's real destination recipient lives. It trusts that requesting `toAddress == fromAddress` produced matching calldata. That's a defensible trust model for internal tooling, but the helper's docstring ("Any drift … aborts before broadcast") oversells it. Either soften that comment, or fail closed when the echoed fields are absent for a mode that requires them.

script/tasks/manageWalletFunds.ts
```diff
@@ -0,0 +1,668 @@
...
+function requireNetwork(name: string) {
+  const net = networks[name]
+  if (!net)
+    throw new Error(`Unknown network "${name}" (not in config/networks.json).`)
+  return net
+}
+
+const ERC20_MIN_ABI = erc20Abi
+
+interface IResolvedToken {
```
gvladika: **Security-critical logic here is untested.**

`recordedAddressForRole` + `assertKeyMatchesRole` are the "key derives to the wrong address → refuse" net, but unlike the rest of the decision logic (deliberately extracted to `walletFundsHelpers.ts` to be testable) they live in the task file and aren't covered by any test. The nested-role path (`backendSignerProduction` → `backendSigner.production`) is fiddly enough to deserve one. Consider moving both into `walletFundsHelpers.ts` (injecting the config) so they're covered like everything else.

script/tasks/manageWalletFunds.ts
0xDEnYO: Fixed in a574782b6. Centralized the check into `chainUsesErc20Gas(net)` in `walletFundsHelpers.ts` (refuses when `feeTokenAddress` is set, `nativeCurrency === "N/A"`, or `nativeAddress` is a non-sentinel predeploy), unit-tested with arc + tempo + normal-chain cases, and `send` now uses it — tempo is caught.

I did **not** add the guard to bridge/swap: their leg is always native, and `chainUsesErc20Gas` is deliberately broad (it also flags celo/metis, whose `nativeAddress` is set but which have a genuinely spendable native coin that bridges fine via LI.FI). A local guard there would block those legitimate routes, so I left bridge/swap to the remote API rejection you noted, and kept the broad-but-fail-safe blocking only on `send` (where refusing and telling the user to move manually loses nothing).

script/tasks/manageWalletFunds.ts
0xDEnYO: Fixed in a574782b6 — chose fail-closed over softening: the gate now aborts if the quote omits either `fromAddress` or `toAddress`, then asserts both equal our wallet. Also rewrote `assertSameWallet`'s docstring to stop overselling — it now states it checks the quote's *declared* addresses, not the per-tool `transactionRequest.data` (decoding calldata per bridge tool is out of scope), and that callers fail closed when the fields are absent.

script/tasks/manageWalletFunds.ts
0xDEnYO: Fixed in a574782b6. Moved both `recordedAddressForRole` and `assertKeyMatchesRole` into `walletFundsHelpers.ts` with the config injected as a parameter, and the task file now calls `assertKeyMatchesRole(role, address, globalConfig)`. Added tests covering: direct role match, mismatch throws, scratch key with no recorded address passes, non-address values ignored, and the nested `backendSignerProduction → backendSigner.production` path (both resolution and the mismatch abort).

script/tasks/manageWalletFunds.ts
```diff
@@ -0,0 +1,624 @@
...
+        'Bridge reported FAILED — check the source tx and LI.FI status.'
+      )
+      return
+    }
+    await sleep(stepSeconds * 1000)
+  }
+  consola.warn(
+    `Still pending after ${deadline}s. Destination arrival is asynchronous — track it at ` +
+      `https://scan.li.fi/tx/${txHash}`
+  )
```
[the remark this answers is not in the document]
0xDEnYO: The FAILED branch already throws with substatus details (done in 93a6e65). On the timeout suggestion — I kept it as a warn+return rather than an error: by the time we poll, the source tx is already confirmed on-chain (funds have left), and bridge delivery is legitimately asynchronous and often exceeds the 180s window. Throwing there would give automation a false failure for a healthy, still-settling bridge. The warning points at scan.li.fi to track completion, which is the correct signal for a pending-but-not-failed route.

gvladika (approved):

PR #2064, opened 2026-07-16
feat(clear-signing): surface bridge-specific destination recipient for non-EVM bridges (EXSC-627)

0xDEnYO:

Fixes [EXSC-627](https://linear.app/lifi-linear/issue/EXSC-627/clear-signing-surface-bridge-specific-destination-recipient-for-non).

Follow-up from a CodeRabbit finding on [#2046](https://github.com/lifinance/contracts/pull/2046), which flagged AcrossV4 specifically. Investigation showed the issue is systemic, so it is fixed here in a dedicated PR rather than folded into that deprecation/audit PR.

**The problem.** `config/clearSigningProposal.json` (generated by `tasks/buildClearSigningProposal.ts`) surfaces only `_bridgeData.receiver` as the "Recipient" for every bridge. But `_bridgeData.receiver` is **not** the real recipient for non-EVM destinations: the facets pass the `NON_EVM_ADDRESS` sentinel (`0x1111…01`) in `_bridgeData.receiver` and carry the actual recipient — a `bytes32`/`bytes` non-EVM address (Solana / Bitcoin / NEAR / …) — in their own bridge-specific struct. So a hardware-wallet signer bridging to a non-EVM chain saw a magic sentinel constant where the recipient should be, instead of the address the funds actually go to.

This was verified against each facet's `_startBridge`: e.g. `AcrossFacetV4` passes `_acrossData.receiverAddress` as the `recipient` arg to `SPOKEPOOL.deposit` in all cases; `_bridgeData.receiver` is only a validation handle and is the sentinel for non-EVM.

**Why alongside, not replacing.** ERC-7730 `display.formats` fields are static per selector — a descriptor cannot express "show `_bridgeData.receiver`, unless it's the sentinel, then show the bridge-specific field". So the generator now emits the bridge-specific destination recipient **in addition to** the generic "Recipient" for every non-EVM-capable facet. The signer sees both.

**Why `format: raw`.** A non-EVM receiver is not an EVM address, so `addressName` (which resolves against EVM address books / ENS) cannot render it. `raw` shows the literal bytes the user is authorizing — the honest representation.

**Two label groups**, per how the field behaves:
- **"Destination Recipient"** — the field is authoritative in *all* cases (equals `_bridgeData.receiver` for EVM, holds the real recipient for non-EVM): `AcrossV4`, `AllBridge`, `DeBridgeDln`, `GasZip`, `Glacis`, `LiFiIntentEscrow`, `LiFiIntentEscrowV2`.
- **"Non-EVM Recipient"** — the field is populated only for non-EVM destinations (zero/empty for EVM, where the generic "Recipient" is authoritative): `Chainflip`, `Eco`, `Garden`, `LayerSwap`, `Mayan`, `NEARIntents`, `PolymerCCTP`.

14 non-EVM-capable facets × `startBridgeTokensVia…` + `swapAndStartBridgeTokensVia…` = 28 descriptor entries gain the extra field. Every path is verified against the facet's bridge-specific struct and its `_startBridge` receiver handling. This is a **behavioral-only change to a generated config** — no Solidity changes, external ABIs/selectors untouched, no audit or version bump required.

**Known limitations (deliberately out of scope):**
- The one-line `interpolatedIntent` summary still interpolates `{_bridgeData.receiver}`, so the headline for a non-EVM bridge still shows the sentinel; the authoritative recipient is now in the detailed fields. Reworking the summary per-bridge (interpolating a raw `bytes32`) is a larger, separate change.
- Packed/Min entrypoints and `AcrossV4Swap` are untouched — they either have no decodable params or carry no bridge-specific receiver field.

- [ ] I have added tests that cover the functionality / test the bug — n/a: the generator has no unit-test harness; `verifyClearSigning` CI regenerates and diffs the committed output, which covers correctness of this change.
- [ ] For new facets: I have checked all points from this list — n/a (no facets; generator + generated config only)
- [x] I have updated any required documentation — the rationale lives in an in-code comment block on the new `BRIDGE_RECEIVER_FIELDS` map.

### DIFF
--- a/config/clearSigningProposal.json
+++ b/config/clearSigningProposal.json
@@ -204,6 +204,12 @@
           "visible": "always"
         },
         {
+          "path": "_acrossData.receiverAddress",
+          "label": "Destination Recipient",
+          "format": "raw",
+          "visible": "always"
+        },
+        {
           "path": "_bridgeData.transactionId",
           "label": "Transaction Id",
           "visible": "never"
@@ -351,6 +357,12 @@
               "ens"
             ]
           },
+          "visible": "always"
+        },
+        {
+          "path": "_allBridgeData.recipient",
+          "label": "Destination Recipient",
+          "format": "raw",
           "visible": "always"
         },
         {
@@ -638,6 +650,12 @@
           "visible": "always"
         },
         {
+          "path": "_chainflipData.nonEVMReceiver",
+          "label": "Non-EVM Recipient",
+          "format": "raw",
+          "visible": "always"
+        },
+        {
           "path": "_bridgeData.transactionId",
           "label": "Transaction Id",
           "visible": "never"
@@ -702,6 +720,12 @@
               "ens"
             ]
           },
+          "visible": "always"
+        },
+        {
+          "path": "_deBridgeData.receiver",
+          "label": "Destination Recipient",
+          "format": "raw",
           "visible": "always"
         },
         {
@@ -772,6 +796,12 @@
           "visible": "always"
         },
         {
+          "path": "_ecoData.nonEVMReceiver",
+          "label": "Non-EVM Recipient",
+          "format": "raw",
+          "visible": "always"
+        },
+        {
           "path": "_bridgeData.transactionId",
           "label": "Transaction Id",
           "visible": "never"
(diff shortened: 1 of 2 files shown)


0xDEnYO:
Superseded by #2063 (same CodeRabbit finding / same fix, tracked under EXSC-627). #2063 is the PR we're continuing with: it adds the docs update, differentiated TYPE-A/TYPE-B labels, exact ABI-type validation in the generator, and erc7730 reference-tool verification. Closing this to avoid a duplicate. Note: #2063 currently excludes Garden (its nonEvmReceiver is an off-chain HTLC order with no on-chain guarantee — surfacing it as a recipient risks false assurance); can be revisited.

PR #2078, opened 2026-07-17
feat(healthcheck): assert Executor↔ERC20Proxy binding + declarative invariant registry + daily live sweep (bug bounty #292)

0xDEnYO:

[EXSC-655](https://linear.app/lifi-linear/issue/EXSC-655/prevention-layer-for-bug-bounty-292-assert-executorerc20proxy-binding)

Prevention layer for **bug bounty #292** — the `Executor` on Optimism + opBNB was bound to a *retired* `ERC20Proxy` while the published proxy authorized the Executor, so `healthCheck.ts` (which only checked the published proxy) stayed green, and the health check only ran on new-network onboarding so existing networks were never re-checked. This PR closes the detection gap, then makes the checker fast, trustworthy, and self-maintaining.

### 1. The invariant that catches #292 (bidirectional binding)

New `executor-erc20proxy-binding` (production): asserts `Executor.erc20Proxy()` (checksum) **==** the deployed `ERC20Proxy` **and** the bound proxy authorizes the Executor. Plus `receiver-executor-binding` (every deployed `Receiver*` points at the deployed `Executor`). Verified live: this flags Optimism + opBNB today (their Executors are bound to the retired v1.0.0 proxy) — remediation is tracked separately.

### 2. Declarative invariant registry

The inline checks became an ordered `IHealthCheckInvariant[]` (`healthCheckInvariants.ts`) that a runner iterates; `healthCheck.ts` builds the per-network context and maps the result to an exit code. Adding a check is a registry edit. A non-fatal `warning` severity tier was added; all pre-existing checks stay `error`.

### 3. Recurring live tier (new workflow)

`healthCheckAllNetworks.ts` + `.github/workflows/healthCheckAllNetworks.yml` fan the check across production networks and post one consolidated Slack status. Triggers: **daily cron** (backstops non-deploy drift) + **push-to-main touching `deployments/**`** (post-deploy) — never `pull_request` (live checks must not gate PRs; the static PR tier is `deploymentAddressConsistency.yml`).

### 4. Per-network exclusions

`HEALTH_CHECK_EXCLUSIONS` — a declarative `{ invariant, network, reason }` table to carve out an eval on a chain where it genuinely doesn't apply. `reason` is mandatory and printed on skip (never silent); a test asserts every entry references a real invariant + real network (stale carve-outs fail CI). Empty by default.

### 5. Efficiency (resource-efficient by design)

- **Whitelist source-of-truth check batched** into a single viem multicall (auto-chunked) instead of a round-trip per pair — a full `base` run drops from **281 → 63 RPC round-trips**.
- **`cast` subprocess removed** for `facets()` → in-process viem read (joins the batched multicall client).
- **Two-phase concurrent runner** (deps respected) so viem's multicall batching actually fires; **fleet fans out in one process** via `runHealthCheckForNetwork()` instead of a `bunx tsx` subprocess per network.
- **Push trigger scoped** to only the networks whose deploy logs changed (`deploymentPathsToNetworks`); the daily cron sweeps the full fleet.

### 6. Trustworthy over time

- **Re-verify a fatal failure once** before recording it (transient-RPC guard); per-invariant isolated error accounting makes concurrency + re-verify safe.
- **Remediation hints** printed on failure; **`scriptEval()`** adapter to reuse external check scripts as evals (exit 0 = pass).
- **`pauser-funded` (lightweight floor)** — asserts a non-zero native balance for the pauser on EVM **and Tron** (`production`, `skipTestnet`). `verifyEmergencyPauseReadiness` (`checkPauserFunds.sh`) remains the authoritative "can afford a `pauseDiamond()`" gate, but it is EVM-only and scheduled-only; this floor closes the two gaps that left (Tron mainnet pauser checked nowhere; a freshly deployed EVM mainnet pauser not caught until the next readiness run). [Re-added in review — was previously removed.]
- **Maintenance rule** (`.agents/rules/601-healthcheck-invariants.md`) + a non-blocking CI reminder so contract changes prompt an invariant review — keeps the eval set alive as the protocol evolves.

### Full invariant set

`diamond-deployed` (halt gate), `core-facets-deployed`, `non-core-facets-deployed`, `facets-registered`, `core-periphery-deployed`, `periphery-registered`, `whitelist-integrity`, `erc20proxy-owner`, `diamond-owner`, `feecollector-owner`, `receiver-owner`, `refund-wallet-access`, `pauser-funded`, `safe-config`, **`executor-erc20proxy-binding`** (#292), **`receiver-executor-binding`**, **`no-duplicate-selectors`**, **`no-unexpected-facets`** (warn), **`no-unexpected-erc20proxy-callers`** (warn).

### Verification

- `bunx tsc-files` / `bunx eslint` / `prettier --check` / `actionlint` — all clean.
- `bun test` — **41 pass / 0 fail** (pure helpers, registry + exclusion-table integrity, and runner regression tests: concurrent-merge, transient-vs-persistent re-verify, `onChainFacets` phase-1→phase-2 propagation, halt, out-of-scope skip, `scriptEval`).
- Live read-only runs against `base` / `optimism` confirmed detection + the 281→63 round-trip drop + in-process fan-out.
- Adversarial review (7 attack vectors: concurrency race, transient re-verify, whitelist-batching semantics, in-process isolation, regex coverage, pauser removal, type soundness) — **all refuted**, with the concurrency logic exercised at runtime.
- `deploy-smoke-test.yml` exercises `healthCheck.ts` on localanvil in CI once marked ready.


script/deploy/healthCheckAllNetworks.ts
```diff
@@ -0,0 +1,272 @@
...
+      status: result.status,
+      // Keep only the tail so a single network cannot flood the consolidated report.
+      detail:
+        result.status === 'failed'
+          ? result.errors.slice(-5).join('\n')
+          : result.status === 'skipped'
+          ? result.skipReason ?? 'skipped'
+          : '',
+    })
+  )
```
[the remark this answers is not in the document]
0xDEnYO: Fixed in 695dac4b1 — \`runOneNetwork\` is now an async function wrapped in try/catch (rejection → failed result), so it can never bypass \`Promise.race\` or reject the concurrent queue; the \`.then()\` chain and nested ternaries are gone.

script/deploy/healthCheck.ts
```diff
@@ -174,28 +101,24 @@ const main = defineCommand({
...
 
     // For staging, skip targetState checks as targetState is only for production
     let nonCoreFacets: string[] = []
     if (environment === 'production') {
       const networkTarget = targetState[networkLower]?.production
-      if (!networkTarget?.LiFiDiamond) {
+      if (!networkTarget?.LiFiDiamond)
         consola.warn(
           `Network '${networkLower}' not in target state; skipping non-core facet comparison`
         )
```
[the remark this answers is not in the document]
0xDEnYO: Deliberate: a production network missing target state is reported as a **non-fatal warning**, not a hard failure — hard-failing would false-page active networks not yet in target state. Its invisibility was the real gap, now closed in 695dac4b1: the warning count is surfaced through \`IHealthCheckResult\` → \`summarizeHealthChecks.warned\` → the consolidated runner output and Slack message, so reduced coverage is visible in the fleet report.

script/deploy/healthCheck.ts
```diff
@@ -174,28 +101,24 @@ const main = defineCommand({
...
 
     const coreFacetExclusions = supportsGasZip ? [] : ['GasZipFacet']
-    const coreFacetsToCheck = getCoreFacets({
-      exclude: coreFacetExclusions,
-    })
+    const coreFacetsToCheck = getCoreFacets({ exclude: coreFacetExclusions })
 
     // For staging, skip targetState checks as targetState is only for production
     let nonCoreFacets: string[] = []
     if (environment === 'production') {
```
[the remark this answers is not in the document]
0xDEnYO: Addressed: the CLI validates \`--environment\` up front, and 695dac4b1 adds a defensive guard inside \`runHealthCheckForNetwork\` (returns a failed result on any value other than production/staging) so selection can never run against production data under a typo.

script/deploy/healthCheckAllNetworks.ts
```diff
@@ -0,0 +1,198 @@
...
+): Promise<IHealthCheckResult> {
+  try {
+    await spawnAndCapture('bunx', [
+      'tsx',
+      'script/deploy/healthCheck.ts',
+      '--network',
+      network,
+      '--environment',
+      environment,
+    ])
```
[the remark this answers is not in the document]
0xDEnYO: Superseded by the in-process refactor (43f0304dc): the fan-out no longer spawns a subprocess — it calls \`runHealthCheckForNetwork\` in-process bounded by a \`Promise.race\` deadline, and 695dac4b1 hardened that path so a stalled/rejected check yields a failed result and never blocks the consolidated report.

.github/workflows/healthCheckAllNetworks.yml
```diff
@@ -0,0 +1,155 @@
...
+          PASSED_COUNT: ${{ steps.healthcheck.outputs.passed_count }}
+          FAILED_COUNT: ${{ steps.healthcheck.outputs.failed_count }}
+          FAILED_NETWORKS: ${{ steps.healthcheck.outputs.failed_networks }}
+        run: |
+          # Deep-link straight to THIS job so the Slack link lands on the failing step; fall back to
+          # the run overview if the jobs API can't be read.
+          RUN_OVERVIEW="${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}"
+          RUN_URL=$(gh api "repos/${{ github.repository }}/actions/runs/${{ github.run_id }}/jobs" \
+            --jq '[.jobs[] | select(.name == "${{ github.job }}") | .html_url][0] // .jobs[0].html_url' 2>/dev/null)
+          RUN_URL="${RUN_URL:-$RUN_OVERVIEW}"
```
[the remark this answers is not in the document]
0xDEnYO: Addressed earlier on this branch: the job-URL lookup is now guarded with \`if ! RUN_URL=$(gh api ...); then RUN_URL=""; fi\` before the \`${RUN_URL:-$RUN_OVERVIEW}\` fallback, so a non-zero \`gh api\` exit under \`set -e\` no longer aborts the step / skips Slack.

0xDEnYO:
@coderabbitai resolve

All review comments have been addressed in commits 43f0304dc / d5a33c9a3 / 695dac4b1 (replies on each thread). Summary: runOneNetwork hardened (async + try/catch, no bypass of Promise.race); per-network warnings surfaced in the consolidated report + Slack; defensive env guard added; the cast-subprocess timeout note is superseded by the in-process refactor; and the gh-api fallback guard was added earlier on this branch.

gvladika (commented):
Automated high-effort review (finder → independent-verifier → synthesis). 10 verified findings, all in the two TS entry points. The serious ones share a theme: an invariant reports **green without ever verifying its target** — for a health check, a false pass is worse than a false alarm. Blockers in my view are the two `consola.error`-vs-`ctx.logError` slips (one-line fixes). Inline comments below.

script/deploy/healthCheckInvariants.ts
```diff
@@ -0,0 +1,1695 @@
...
+        if (ctx.deployedContracts.LiFiTimelockController)
+          await checkOwnership(
+            'LiFiDiamond',
+            ctx.deployedContracts.LiFiTimelockController,
+            ctx,
+            ctx.publicClient
+          )
+        else
+          consola.error(
+            'LiFiTimelockController not deployed, so diamond ownership cannot be verified'
```
gvladika, line 1273: **🔴 correctness / false pass (blocker):** when `LiFiTimelockController` is missing from the deploy log, this branch logs a bare `consola.error(...)` instead of `ctx.logError(...)` (same on the Tron branch ~L1241). `executeInvariant` only inspects the `errors[]` array, so it records **zero errors and marks `diamond-owner` passed** — the diamond's actual owner is never compared to the expected one. A diamond left owned by an EOA would pass the very check meant to catch it.

`core-periphery-deployed` would independently redden the run, but this invariant still yields a false pass in isolation.

**Fix:** use `ctx.logError(...)`.

script/deploy/healthCheckInvariants.ts
```diff
@@ -0,0 +1,1695 @@
...
+              address: f.facetAddress,
+              selectors: [...f.functionSelectors],
+            }))
+          )
+          registeredFacets = ctx.onChainFacets
+            .map((f) => configFacetsByAddress[f.address.toLowerCase()])
+            .filter((name): name is string => typeof name === 'string')
+        }
+      } catch (error: unknown) {
+        facetCheckSkipped = true
```
gvladika, line 808: **🔴 correctness / silent coverage loss (blocker):** on a failed `facets()` read this catch sets `facetCheckSkipped` and calls only `consola.warn` — never `ctx.logError` or `ctx.logWarn`. So no error triggers the re-verify, nothing lands in `ctx.warnings`, and `ctx.onChainFacets` stays empty. The phase-2 consumers `no-duplicate-selectors` (error severity) and `no-unexpected-facets` then hit their `onChainFacets.length === 0` guard and **skip**.

Net effect: the network reports `passed` with `warnings=0`, so `summarizeHealthChecks` omits it from the `warned` list and the daily sweep posts a **green** Slack status — while the selector-drift / rogue-facet checks (the #292 drift class this PR exists to catch) silently didn't run, and their absence is invisible in the consolidated report.

**Fix:** record `ctx.logWarn` (or an error) so the skipped facet check surfaces in the report.

script/deploy/healthCheckAllNetworks.ts
```diff
@@ -0,0 +1,309 @@
...
+    .sort()
+}
+
+/**
+ * Map changed `deployments/**` file paths to the production network keys they belong to.
+ * Only `deployments/<network>.json` counts — `<network>.staging.json`, `<network>.diamond.json`
+ * and non-network files (e.g. `_deployments_log_file.json`) are ignored. Pure; used by the
+ * push-to-main trigger to check only the networks a deploy actually touched.
+ */
+export function deploymentPathsToNetworks(paths: string[]): string[] {
```
gvladika, line 55: **🔴 correctness / trigger mismatch:** the cron path (`getProductionNetworkNames`) filters `type === 'mainnet' && status === 'active'`, but this push path applies only the `^deployments/([a-z0-9]+)\.json$` regex with **no mainnet/active filter**. A merge touching e.g. `deployments/basesepolia.json` returns `['basesepolia']`, which `run()` feeds into `runHealthCheckForNetwork('basesepolia', 'production')` — running production-scoped invariants against **testnet** on-chain state.

Any production invariant that fails on the testnet marks it `failed` → the workflow exits 1 and posts a red `🚨 ACTION NEEDED` Slack alert for a change that touched no production network. The two triggers assert different fleets.

**Fix:** intersect `deploymentPathsToNetworks(...)` output with `getProductionNetworkNames(getAllActiveNetworks())`.

script/deploy/healthCheckInvariants.ts
```diff
@@ -0,0 +1,1695 @@
...
+      else if (ctx.publicClient)
+        await checkOwnership(
+          'Receiver',
+          ctx.refundWallet,
+          ctx,
+          ctx.publicClient
+        )
+    },
+  },
+  // NOTE: pauser-wallet-funded was intentionally removed. Pauser funding is owned by
```
gvladika: **🟠 correctness / lost coverage vs. old monolith (2 gaps, worth confirming intent with the author):**

1. **Tron pauser funding is no longer checked anywhere.** The old `healthCheck.ts` checked the Tron pauser's TRX balance (`tronWeb.trx.getBalance`) and errored on zero. That's delegated to `checkPauserFunds.sh`, which is **EVM-only and explicitly skips Tron**; `verifyEmergencyPauseReadinessGitHub.sh` only verifies the Tron pauser *matches config*, not that it's *funded*. If Tron pauser TRX hits zero, every check stays green until an incident needs `pauseDiamond()` and it can't broadcast.

2. **New EVM networks lose deploy-time pauser-funding verification.** The removed invariant ran at deploy time (`healthCheckForNewNetworkDeployment.yml`); its replacement `checkPauserFunds.sh` runs only on the scheduled/manual readiness workflow, so a freshly deployed network's unfunded pauser surfaces up to ~24h later instead of at deploy.

**Fix:** keep a lightweight balance assertion in the registry, or extend `checkPauserFunds.sh` to cover Tron **and** wire it into the deploy-time workflow.

script/deploy/healthCheckInvariants.ts
```diff
@@ -0,0 +1,1695 @@
...
+
+      const expectedAuthorized = new Set([
+        getAddress(executorAddress as Address).toLowerCase(),
+      ])
+
+      // ERC20Proxy exposes no enumerator for authorizedCallers, so reconstruct the
+      // current set from AuthorizationChanged events. Full-range getLogs is rejected by
+      // some RPC providers — treat any failure as "cannot verify" (warning, non-fatal).
+      try {
+        const logs = await ctx.publicClient.getContractEvents({
```
gvladika, line 1527: **🟠 correctness + 🟡 perf (PLAUSIBLE):** `no-unexpected-erc20proxy-callers` reconstructs the authorized-caller set from a single full-range `getContractEvents({ fromBlock: 'earliest', toBlock: 'latest' })`. Two failure modes on long-lived mainnet proxies:

- **Silent truncation (false pass):** providers that cap results *without throwing* return a partial log set; the catch (~L1510) never trips, so an `AuthorizationChanged(rogue, true)` in an omitted range is absent and the check reports "Only the Executor is authorized" — hiding the exact security drift this invariant exists to surface.
- **Cap-and-throw (silent skip + waste):** providers that throw get downgraded to a skip, so the signal isn't verified, while a heavy full-history `getLogs` is still attempted daily for every EVM production network.

**Fix:** bound `fromBlock` to the ERC20Proxy deployment block (or paginate) rather than `'earliest'`.

script/deploy/healthCheckInvariants.ts
```diff
@@ -0,0 +1,1695 @@
...
+        error instanceof Error ? error.message : String(error)
+      localCtx.logError(`[${invariant.name}] threw: ${errorMessage}`)
+    }
+  }
+
+  await runOnce()
+
+  // Re-verify a fatal failure once before recording it — guards against transient RPC errors
+  // paging on a green fleet. Only error-severity failures are re-checked (warnings are non-fatal).
+  if (invariant.severity === 'error' && errors.length > 0) {
```
gvladika, line 1671: **🟡 cleanup / perf:** the re-verify reruns `invariant.run()` **in full** whenever `severity === 'error'` and any error was recorded. For read-heavy invariants (`whitelist-integrity` fires a multicall over every expected pair; `refund-wallet-access` loops every approved selector) one genuine on-chain drift re-issues the whole read batch. On the daily all-networks sweep a truly-broken network then does 2× the RPC per failing invariant — extra load precisely when it's already failing and likely near provider rate limits.

script/deploy/healthCheckAllNetworks.ts
```diff
@@ -0,0 +1,309 @@
...
+        network,
+        status: 'failed',
+        warnings: 0,
+        detail: `unexpected error: ${message}`,
+      }
+    }
+  }
+
+  try {
+    return await Promise.race([check(), timeout])
```
gvladika, line 176: **🟡 cleanup / concurrency:** `Promise.race([check(), timeout])` abandons but does not cancel the losing `check()` (the comment says so: "abandoned, not killed"). When the 5-min timer wins, `mapWithConcurrency` immediately starts the next network, but the abandoned promise keeps issuing RPC reads — no `AbortSignal` is threaded through `runHealthCheckForNetwork`. On a slow-RPC day where several networks time out, in-flight calls climb **above** `--concurrency` (8), amplifying rate-limiting exactly when the fleet is already struggling (root cause also at L106).

**Fix:** thread an `AbortController`/signal into the viem client so a timed-out network's reads are actually cancelled.

script/deploy/healthCheckInvariants.ts
```diff
@@ -0,0 +1,1695 @@
...
+ * network + environment, and the invariant passes iff it exits 0. Non-zero exit records a
+ * failure (error/warning per `severity`) with the script's output tail as the detail. Use
+ * sparingly — a subprocess per (eval × network) loses the shared viem client/batching, so
+ * it is for reusing authoritative external checks, not for checks that belong in-process.
+ *
+ * @param scriptPath - Repo-relative path to an executable check (exit 0 = pass).
+ * @param severity - How a non-zero exit is reported.
+ * @param argsFromCtx - Builds the script args from the network context.
+ */
+export function scriptEval(
```
gvladika: **🟡 cleanup / dead code:** `scriptEval` is exported, pulls in the `spawnAndCapture` import, and carries 3 dedicated tests — but **no entry in `HEALTH_CHECK_INVARIANTS` (nor anywhere in `script/`) calls it**; `git grep` finds only its definition and its own tests. Its docstring says "Use sparingly" with zero uses. This contradicts the repo's "no speculative features / proactively flag dead code" rules.

**Fix:** remove `scriptEval` + `spawnAndCapture` + their tests, or land the invariant that needs them in this PR.

script/deploy/healthCheckInvariants.ts
0xDEnYO: Fixed in b3d0004d8 — both branches (EVM ~L1274, Tron ~L1241) now call `ctx.logError(...)` instead of `consola.error(...)`, so a missing `LiFiTimelockController` records an error and fails `diamond-owner` rather than passing in isolation.

script/deploy/healthCheckInvariants.ts
0xDEnYO: Fixed in b3d0004d8 — the `facets()` catch now records via `ctx.logWarn(...)` (both the 429 and generic branches) instead of a bare `consola.warn`. The network now lands in `summarizeHealthChecks.warned` → consolidated report + Slack, so the skipped phase-2 selector/facet-set checks are visible instead of posting a silent green. Left as a warning (not error) so a genuinely transient RPC failure does not hard-page, while the reduced coverage is never invisible.

script/deploy/healthCheckAllNetworks.ts
0xDEnYO: Fixed in b3d0004d8 — the push path now intersects `deploymentPathsToNetworks(...)` with `getProductionNetworkNames(getAllActiveNetworks())`, so a change touching e.g. `deployments/basesepolia.json` resolves to zero production networks (→ the existing "nothing to check" success exit) rather than running production invariants against testnet state. The two triggers now assert the same fleet.

script/deploy/healthCheckInvariants.ts
0xDEnYO: Fixed in b3d0004d8 — re-added a lightweight `pauser-funded` invariant to the registry (`severity: error`, `scope: { environments: ["production"], skipTestnet: true }`) that asserts a non-zero native balance on both EVM and Tron. This closes both gaps: the Tron mainnet pauser is now checked, and a freshly deployed EVM mainnet pauser is caught at deploy time via the push trigger. `verifyEmergencyPauseReadiness` (checkPauserFunds.sh) remains the authoritative "can afford a pauseDiamond()" gate — this is only the non-zero floor. `skipTestnet` keeps it off the localanvil smoke-test sandbox (unfunded pauser, testnet); both gaps you cited are mainnet so coverage is unaffected.

script/deploy/healthCheckInvariants.ts
0xDEnYO: Fixed in b3d0004d8 — `fromBlock` is now bounded to the ERC20Proxy deployment block via a `findDeploymentBlock` binary search (~log2(latest) `getCode` calls) instead of `fromBlock: "earliest"`, so the range-cap/full-history-scan failure mode is gone. `AuthorizationChanged` events are sparse, so one bounded query returns the full set. I also changed the catch from `consola.warn` to `ctx.logWarn` so a provider that still rejects the query surfaces a visible "cannot verify" warning rather than a silent pass.

script/deploy/healthCheckInvariants.ts
0xDEnYO: Kept by design — the full re-run is the transient-RPC guard: a read-only invariant is idempotent, so a failure that does not reproduce on a second pass was an RPC blip, not real drift, and re-running the whole `run()` is what lets us distinguish them without per-invariant bookkeeping. It only fires on an already-failed `error`-severity invariant (so the common path is single-pass), is capped at exactly one extra pass, and the 2× reads happen only on a network that is already red. I traded that bounded extra load on failing networks for not paging on transient blips, which for a daily prod sweep is the right call. Happy to add a `noReVerify` opt-out flag on the heaviest invariants (`whitelist-integrity`, `refund-wallet-access`) if you would prefer to exclude them.

script/deploy/healthCheckAllNetworks.ts
0xDEnYO: Fixed in b3d0004d8 — `runOneNetwork` now creates an `AbortController`, and `runHealthCheckForNetwork(network, environment, signal)` threads the signal into the viem HTTP transport `fetchOptions`. When the deadline wins the `Promise.race`, the timer calls `controller.abort()`, so the timed-out network's in-flight reads are actually cancelled and the fleet stays within `--concurrency`. (Tron reads go through TronWeb, which does not take a signal, but Tron is a single network and not part of the multicall fan-out, so it is not the concurrency-amplification path.)

script/deploy/healthCheckInvariants.ts
0xDEnYO: Fixed in b3d0004d8 — removed `scriptEval` and its 3 tests, and dropped its `spawnAndCapture` import from this file. Kept `script/utils/spawnAndCapture.ts` itself: it is a shared util still used by `script/deploy/tron/tronUtils.ts` and `script/utils/utils.ts`, so it is not dead — only its use here was.

gvladika (approved):

maxklenk (approved):

PR #2083, opened 2026-07-20
feat(healthcheck): per-network invariant exclusions with mandatory reason (stacks on #2078)

0xDEnYO:

[EXSC-655](https://linear.app/lifi-linear/issue/EXSC-655/prevention-layer-for-bug-bounty-292-assert-executorerc20proxy-binding) (same effort as #2078)

**Stacks on #2078** (base branch `healthcheck-binding-invariants`, not `main`) — it extends the invariant registry added there, and is intended to consolidate into that PR.

Adds a per-network carve-out layer so a specific eval can be excluded on a specific chain **when it genuinely doesn't apply** (e.g. a deprecated integration on one chain), without weakening the check everywhere. Chosen shape: a single declarative `HEALTH_CHECK_EXCLUSIONS` table of `{ invariant, network, reason }` rather than flags scattered across invariants or `skipHealthcheck`-style booleans in `networks.json`, because:

- **One auditable place** to see every carve-out across all chains.
- **`reason` is mandatory** and printed when the eval is skipped — a carve-out is never a silent skip.
- **Validated in tests**: every entry must reference a real invariant name AND a real network AND a non-empty reason, so a stale carve-out (renamed/removed invariant, deprecated chain) fails CI. This keeps the table honest as the registry evolves.

The table starts **empty** — the default response to a failing invariant is to fix the on-chain/config drift, not to exclude the check. (Alternative considered: a `healthCheckExclusions` map in `config/networks.json`, consistent with `skipHealthcheck`; rejected because it couples network config to invariant names, scatters carve-outs across networks, and can't be type/name-validated as cheaply.)

Runner change: `runHealthCheckInvariants` consults `getInvariantExclusion(name, network)` after scope gating and skips + logs the reason.

# Verification

- `bunx tsc-files --noEmit` — pass; `bunx eslint` — pass; `bun test` — 24 pass / 0 fail (adds `getInvariantExclusion` match/case-insensitivity/no-match + table-integrity: valid invariant name, valid network, non-empty reason).
- Module remains side-effect-free on import (`--help` loads).

### DIFF
--- a/.github/workflows/healthCheckAllNetworks.yml
+++ b/.github/workflows/healthCheckAllNetworks.yml
@@ -55,6 +55,7 @@ jobs:
         uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
         with:
           persist-credentials: false # read-only check; no git push/tag/submodule operations
+          fetch-depth: 0 # full history so a push run can diff which deployments/ changed
 
       - name: Set up Bun
         uses: oven-sh/setup-bun@3d267786b128fe76c2f16a390aa2448b815359f3 # v2.1.2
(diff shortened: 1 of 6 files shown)


0xDEnYO:
Folded into #2078 (all three commits — per-network exclusions, the efficiency/robustness refactor, and the runner regression tests — were rebased onto the #2078 branch and fast-forwarded in). Closing; the work now lives in the single consolidated PR.

PR #2092, opened 2026-07-20
feat(EXSC-423): integrate Vault Wrapper V1 subsystem [LiFiVaultWrapper v1.0.0, LiFiVaultWrapperFactory v1.0.0, LiFiVaultWrapperTypes v1.0.0, ReferenceAccessGate v1.0.0, ERC4626Adapter v1.0.0, IAccessGate v1.0.0, ILiFiVaultWrapper v1.0.0, ILiFiVaultWrapperFactory v1.0.0, IYieldAdapter v1.0.0, LibVaultWrapperMath v1.0.0]

gvladika:

Ref [EXSC-423](https://linear.app/lifi-linear/issue/EXSC-423/s15-security-review-and-audit-support) — S15, Security review & audit support (the audited slice; ticket stays open through audit engagement, so this does not auto-close it).

This is the integration PR that lands the assembled **LI.FI Earn Vault Wrapper V1** subsystem (`src/VaultWrapper/`) on `main`, opened as one reviewable diff ahead of the external audit rather than re-deriving it from the individual S1–S14 sub-PRs.

The Vault Wrapper is a **standalone product** per `[CONV:ARCH-VAULTWRAPPER]` — not a facet, not periphery, not called by the Diamond, with no `diamondCut` or shared selector/storage involvement. A factory deploys per-integrator ERC-4626 vaults as deterministic beacon proxies; each wraps a governance-curated yield source through a pluggable adapter and takes a four-type fee split between the integrator and LI.FI. It has its own governance (a dedicated 48h timelock, an emergency pauser, and an onboarding manager).

For the full picture — architecture, contract set, roles and governance, fee model, trust assumptions, and system invariants — see the subsystem overview and the per-contract docs:

- **[docs/VaultWrapper/README.md](https://github.com/lifinance/contracts/blob/dev-vault-wrapper/docs/VaultWrapper/README.md)** — start here.
- [LiFiVaultWrapper](https://github.com/lifinance/contracts/blob/dev-vault-wrapper/docs/VaultWrapper/LiFiVaultWrapper.md) · [LiFiVaultWrapperFactory](https://github.com/lifinance/contracts/blob/dev-vault-wrapper/docs/VaultWrapper/LiFiVaultWrapperFactory.md) · [ERC4626Adapter](https://github.com/lifinance/contracts/blob/dev-vault-wrapper/docs/VaultWrapper/ERC4626Adapter.md)


config/vaultWrapper.json
```diff
@@ -0,0 +1,32 @@
+{
+  "mainnet": {
+    "create3Factory": "0x93FEC2C00BfE902F733B57c5a6CeeD7CD1384AE1",
+    "multisig": "0xE3C8121DF9b1c5A7d383Ab4923fF848a6510F357",
+    "emergencyPauser": "0x0000000000000000000000000000000000000000",
+    "onboardingManager": "0x0000000000000000000000000000000000000000",
+    "lifiFeeRecipient": "0x0000000000000000000000000000000000000000",
```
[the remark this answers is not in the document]
gvladika: Acknowledged — these zero addresses and empty `allowedUnderlyings` arrays are intentional placeholders while the PR is in draft. They'll be populated with the real `mainnet`/`basesepolia` role addresses and allowed underlyings just before deployment, so the deploy-script validation won't be exercised until then.

src/VaultWrapper/LiFiVaultWrapper.sol
```diff
@@ -0,0 +1,1390 @@
...
+import { ILiFiVaultWrapperFactory } from "./interfaces/ILiFiVaultWrapperFactory.sol";
+import { IYieldAdapter } from "./interfaces/IYieldAdapter.sol";
+import { FeeConfig, FeeType, FeeReceiver, FEE_TYPE_COUNT } from "./LiFiVaultWrapperTypes.sol";
+import { LibVaultWrapperMath } from "./libraries/LibVaultWrapperMath.sol";
+
+// One over solhint's 15-state default: the 16th declaration is the write-once
+// `shareDecimalsOffset` (inflation protection), which packs into the `accessGate`
+// slot rather than widening the storage layout.
+// solhint-disable max-states-count
+
```
reednaa: I do not believe in this.

I had an agent trace this and this is from a prior version of the SC were `factory` was state not `FACTORY` immutable

src/VaultWrapper/LiFiVaultWrapper.sol
```diff
@@ -0,0 +1,1390 @@
...
+            address wallet = receivers[i].wallet;
+            // we use trySafeTransfer here so a failed transfer doesn't block the distribution
+            if (SafeERC20.trySafeTransfer(IERC20(_token), wallet, share)) {
+                continue;
+            }
+
+            // the failed transfer amount stays in the wrapper as still-owed integrator fees
+            retained += share;
+            emit IntegratorPayoutRetained(wallet, _token, share);
+        }
```
reednaa: All statements in this paragrah are correct and bounded.
i is bounded by count (is at most `(count - 1) + 1)`).
Share is bounded by `_integratorTotal`. Trivially in `true` and in `false` by 2 word multiplication on `mulDiv`.
`distributed` is bounded by `_integratorTotal` through the condition in `_setIntegratorFeeReceivers`
`retained` is bounded by `_integratorTotal - distributed`.

Can be unchecked; Math checks out, no unexpected reverts.

src/VaultWrapper/LiFiVaultWrapper.sol
```diff
@@ -0,0 +1,1390 @@
...
+        uint256 count = receivers.length;
+        uint256 distributed;
+
+        for (uint256 i; i < count; ++i) {
+            uint256 share;
+            if (i + 1 == count) {
+                // last receiver gets the whole remainder to avoid rounding dust
+                share = _integratorTotal - distributed;
+            } else {
+                // other receivers get their bps share, rounded down
```
reednaa: for gas purposes, i + 1 special case should be implemented outside of jump. This also makes it easier to audit.

src/VaultWrapper/LiFiVaultWrapper.sol
```diff
@@ -0,0 +1,1390 @@
...
+    ///      check, so a plain checked add could itself overflow uint256 and revert.
+    /// @param _accrued The current accumulator value.
+    /// @param _delta The amount to add.
+    /// @return The clamped sum.
+    function _saturatingAddUint128(
+        uint128 _accrued,
+        uint256 _delta
+    ) private pure returns (uint128) {
+        uint256 accrued = uint256(_accrued);
+        if (_delta > type(uint128).max - accrued) return type(uint128).max;
```
reednaa: My desire to write this in assembly is immense but I can understand. Consider inline-if to give the solidity optimiser more to work with.

src/VaultWrapper/LiFiVaultWrapper.sol
```diff
@@ -0,0 +1,1390 @@
...
+    ///      blacklisted wallet) has its share returned as `retained` and left in the wrapper,
+    ///      so one hostile wallet can never block the distribution.
+    /// @param _token The fee-pool token to distribute.
+    /// @param _integratorTotal The integrator's portion of the fee pool.
+    /// @return retained The sum of shares whose transfer failed (left in the wrapper).
+    function _payIntegrators(
+        address _token,
+        uint256 _integratorTotal
+    ) private returns (uint256 retained) {
+        FeeReceiver[] memory receivers = integratorFeeReceivers;
```
reednaa: Gas: Storage but copy into memory before slot is accessed. Unnesseary memory inflation.

src/VaultWrapper/LiFiVaultWrapperFactory.sol
```diff
@@ -0,0 +1,447 @@
...
+    /// @param _nonce Caller-supplied disambiguator.
+    /// @return The CREATE2 salt.
+    function _salt(
+        bytes32 _namespace,
+        address _adapter,
+        address _underlying,
+        uint256 _nonce
+    ) internal pure returns (bytes32) {
+        return
+            keccak256(abi.encode(_namespace, _adapter, _underlying, _nonce));
```
reednaa, line 375: Contract is upgradable. Consider versioning or using a domain separatior.

This could be done through the nonce; It depends on what the upgradability featureset will be. (You reserved storage)

src/VaultWrapper/libraries/LibVaultWrapperMath.sol
```diff
@@ -0,0 +1,241 @@
...
+        if (pps <= _hwmPps) return 0;
+
+        uint256 gainAssets = _totalSupply.mulDiv(
+            pps - _hwmPps,
+            PPS_SCALE,
+            Math.Rounding.Floor
+        );
+        feeAssets = feeOnRaw(gainAssets, _rateBps);
+        if (feeAssets >= _totalAssets) feeAssets = _totalAssets - 1;
+    }
```
reednaa, line 163: INFO: An underlying vault misrepresenting pps for even 1 block can lead to significant fees and later free riders. The most immediate underlying cause could be some broken or misconfigured oracle. In those situation `withdraw` would ordinarly also be an exploit but if we assume that entire path is broken, then `deposit` may still work.

Ultimately this would be a bug in the underlying vault but never the less exaggerated here.

reednaa (commented):
The codebase is compact and consise. It is a well written defensible implementation.

I could not find any issues with the code. Total report is 1 INFO with a couple GAS.

Review session consisted of 2 sets of 2 hours. Tools used: `solidity-test-overhaul`, `weird ERC20 tokens`, and manual review. 

Ideally I want to give it another 2 sets of 2 hours but I will not be able to make it in time. Apologies for being too late.

src/VaultWrapper/LiFiVaultWrapper.sol
gvladika: You're right on the cause, removed in #2296

src/VaultWrapper/LiFiVaultWrapper.sol
gvladika: True, measured on the real contract it's ~260–400 gas at realistic receiver counts. Won't apply it, prefer compiler-enforced invariant for now

src/VaultWrapper/libraries/LibVaultWrapperMath.sol
gvladika: Made it more explicit in docs, governance should curate underlying vault where this should not happen

src/VaultWrapper/LiFiVaultWrapperFactory.sol
gvladika: Ack,  abi.encode is fixed-width

src/VaultWrapper/LiFiVaultWrapper.sol
gvladika: Done

src/VaultWrapper/LiFiVaultWrapper.sol
gvladika: Done in fb31b9e75 — loop runs to count - 1, last receiver paid after it

gvladika:
F-02 and F-03 confirmed resolved. On F-01: two corrections, then why no change is needed.

**The quoted developer response isn't about this finding.** "~260–400 gas at realistic receiver counts" comes from [this comment](https://github.com/lifinance/contracts/pull/2092#discussion_r3915189880) on @reednaa's `unchecked`-in-`_payIntegrators` thread. F-01 had no response until now.

**That figure can't apply here anyway.** This contract uses OZ v5's core `ReentrancyGuard` on an ERC-7201 slot (`LiFiVaultWrapper.sol:9` and `:61`). Guard entry is a cold SLOAD plus a zero→non-zero SSTORE — on the order of 24k gas, not ~300. Adding the suggested NatSpec verbatim would have put a wrong figure in the contract right before audit.

**Why `nonReentrant` isn't needed.** The stated route — a yield-source hook calling back into `setIntegratorFeeReceivers` — cannot pass `onlyOwner`. A callback arrives with `msg.sender` set to the source vault or the fee token, not the owner. Round 1's variant ("integrator uses the same EOA as both admin and receiver") is not reachable either: an EOA executes no callback.

If the owner *were* the callback contract, there is still nothing to take. `_payIntegrators` fans out only the integrator's own portion and never reads or writes `lifiFeeAssets`/`lifiFeeShares`, so a rotation moves the integrator's money between the integrator's own wallets — which the owner can already do by direct call at any time. And if that rotation shrinks the set mid-`distributeFees`, the index cached before the loop reads out of bounds and panics, reverting the whole call rather than mispaying it. Worst case is an owner-inflicted revert: no fund loss, no state corruption.

**No NatSpec added, deliberately.** This thread is the record for the auditor. `setFeeRate` carries a `nonReentrant` note because there the guard is load-bearing — it is the only `_accrueFees` caller outside the entry/exit/`distributeFees` guard, so a payout hook could otherwise book fees that the in-flight `distributeFees` then erases. That note is not a template for documenting every setter that lacks a guard it doesn't need.

Re-triggering the review.

src/VaultWrapper/LiFiVaultWrapper.sol
```diff
@@ -1342,32 +1342,29 @@ contract LiFiVaultWrapper is
...
     ) private returns (uint256 retained) {
-        uint256 count = integratorFeeReceivers.length;
+        // `_setIntegratorFeeReceivers` is the sole writer and rejects an empty set, and
+        // `initialize` always runs it, so the count is never zero and this cannot underflow
+        uint256 lastIndex = integratorFeeReceivers.length - 1;
         uint256 distributed;
 
-        for (uint256 i; i < count; ++i) {
+        // every receiver but the last gets its bps share, rounded down
+        for (uint256 i; i < lastIndex; ++i) {
```
reednaa: I know you might have had _enough_ of me but considering starting loop at `i = 1` and then giving the **0'th** integrator the rounding diff.

This way:
- You save on 1 safe sub operation.
- Your lookup operation `integratorFeeReceivers[lastIndex].wallet;` get simplified from stack switch to `PUSH0`.

gvladika: Good point, done in 1de5ee67f. Appreciate the tips!

PR #2097, opened 2026-07-21
fix(VaultWrapper): address #2092 review findings (2, 3, 8, 9, 10)

gvladika:

_TODO: link the EXSC ticket for the VaultWrapper review follow-ups._ Addresses findings from the code review of #2092 (dev-vault-wrapper).

Five independent fixes from the #2092 review, one commit each, targeting the `dev-vault-wrapper` branch.

**Fix 2 — bind `initialize` to the deploying factory (counterfeit-vault prevention).** `initialize` was permissionless, so anyone could deploy a `BeaconProxy` against the official beacon and initialize a counterfeit "LI.FI Earn" token — delegating to the genuine implementation while draining deposits — because the factory-side validation (adapter approval, underlying allowlist, fee bounds) lived only in `deploy()`. The implementation now carries an `EXPECTED_FACTORY` immutable set at construction; `initialize` reverts `NotFactory` for any other caller. The deploy script predicts the factory's CREATE3 address and binds it into the implementation before the factory exists, then asserts the wiring post-deploy.

**Fix 3 — guard the LI.FI fee payout.** `distributeFees` paid the LI.FI share with a plain `safeTransfer` while integrator payouts used `trySafeTransfer`. A blocked LI.FI recipient (e.g. a USDC blacklist) reverted the whole call, freezing the independent integrator payouts behind the factory's 48h timelock. The LI.FI share now uses `trySafeTransfer` too; on failure it is left in the wrapper, re-booked as still-owed, and `LifiPayoutRetained` is emitted. Neither side can hold the other's payout hostage.

**Fix 8 — correct the false pragma floor.** Every subsystem file declared `pragma solidity ^0.8.17` while building on OpenZeppelin v5.6.1, whose `ERC4626Upgradeable` requires `^0.8.24` and whose codegen emits `mcopy` (cancun) — a provably false floor, and the PR had silenced the only guard by adding `src/VaultWrapper/**` to the `solc_floor` skip list. Since the subsystem is only ever deployed by LI.FI with the repo's default compiler, all 36 subsystem files are pinned to that exact version (`^0.8.29`) rather than a lower "honest minimum": a lower floor buys nothing (verification and bytecode key off the exact compiler version, not the pragma range) and would need its own floor build to police. Because the pinned pragma equals the default compile version, the ordinary `forge build` is itself the floor check — VaultWrapper stays excluded from the 0.8.17 `solc_floor` profile (it can't compile there) and needs no profile of its own. The subsystem pragma override is documented in rule 108.

**Fix 9 — drive full-range withdrawals in the invariant handler.** The withdraw handler capped assets at `maxWithdraw/2` on a stale premise that `maxWithdraw` ignores the withdrawal fee. With the pinned OZ, `maxWithdraw = previewRedeem(maxRedeem(owner))`, which the fee-deducting `previewRedeem` makes fully fee-aware, so the halving kept the fail-on-revert campaign out of the upper half of the range — never exercising near-max/full exits. Now bounded at the full ceiling; the campaign runs with zero reverts.

**Fix 10 — correct the inverted fee-bound error.** `_validateFees` reverted with the parameterless `FeeRateAboveBound()` for both the below-minimum and above-maximum branches, giving integrators a backwards diagnostic for a below-minimum rate. Replaced with `FeeRateOutOfBounds(rateBps, minBps, maxBps)` — same signature and selector as the wrapper's `setFeeRate` path — so both paths report identically.

Findings 1, 4, 5, 6, and 7 from the review are intentionally out of scope for this PR.

### DIFF
--- a/.agents/rules/108-vault-wrapper.md
+++ b/.agents/rules/108-vault-wrapper.md
@@ -71,13 +71,17 @@ and counter-intuitive against the repo's Diamond-era conventions.
   silently picks up v4 — keep VaultWrapper code under the scoped paths.
 - The repo does not enable `via_ir`; v5's namespaced-storage accessors raise stack
   pressure, so keep `initialize`/large functions shallow rather than enabling it.
-- **Pragma floor is `^0.8.24`, not the repo-wide `^0.8.17`** (a deliberate override of
+- **Pragma is `^0.8.29`, not the repo-wide `^0.8.17`** (a deliberate override of
   `[CONV:LICENSE]`'s pragma rule for this subsystem only): OZ v5.6.1's
-  `ERC4626Upgradeable` itself requires `^0.8.24` and the toolchain emits `mcopy`, so
-  declaring `^0.8.17` would be a provably false floor. Every file under `src/VaultWrapper/`,
-  `test/solidity/VaultWrapper/`, and `script/deploy/vaultWrapper/` pins `^0.8.24`. The
-  honesty of this floor is enforced by the `solc_floor_vaultwrapper` Foundry profile
-  (solc 0.8.24 + evm cancun), a second build step in `solc-floor-build.yml`.
+  `ERC4626Upgradeable` requires `^0.8.24` and the toolchain emits `mcopy` (cancun), so
+  `^0.8.17` would be a provably false floor. The subsystem is only ever deployed by LI.FI
+  with the repo's default compiler, so it pins that exact version (`^0.8.29`) rather than a
+  lower "honest minimum" — a lower floor would buy nothing (verification and bytecode key
+  off the exact compiler, not the pragma range) at the cost of a separate floor build. Every
+  file under `src/VaultWrapper/`, `test/solidity/VaultWrapper/`, and `script/deploy/vaultWrapper/`
+  pins `^0.8.29`. Because that equals the default compile version, the ordinary build is the
+  floor check; the subsystem is excluded from the `solc_floor` profile (which only compiles
+  the Diamond's 0.8.17 floor) and needs no profile of its own.
 
 ## Identity and onboarding ([CONV:VW-IDENTITY])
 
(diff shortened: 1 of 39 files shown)


.agents/rules/108-vault-wrapper.md
```diff
@@ -71,6 +71,13 @@ and counter-intuitive against the repo's Diamond-era conventions.
   silently picks up v4 — keep VaultWrapper code under the scoped paths.
 - The repo does not enable `via_ir`; v5's namespaced-storage accessors raise stack
   pressure, so keep `initialize`/large functions shallow rather than enabling it.
+- **Pragma floor is `^0.8.24`, not the repo-wide `^0.8.17`** (a deliberate override of
```
0xDEnYO: why do we not use 0.8.29 (to use the same solc for all) instead of introducing a third version?

0xDEnYO: why do we not use 0.8.29 (to use the same solc for all) instead of introducing a third version?

gvladika: Good call — done in 87fa25b. I've pinned the whole subsystem to `^0.8.29` (the default compile version) instead of `^0.8.24`, so there's no third solc anymore.

The `0.8.24` was never a *compile* version — production always builds VaultWrapper at 0.8.29 like everything else. It was the OZ-imposed floor (`ERC4626Upgradeable` is `^0.8.24`), and the extra `solc_floor_vaultwrapper` profile only existed to police that lower floor. But an "honest minimum" floor only pays off for code third parties recompile; since we're the sole deployer and always use 0.8.29, verification and bytecode key off the exact version, not the pragma range — the low floor bought nothing and cost a second CI build.

So: all 36 files → `^0.8.29`, the `solc_floor_vaultwrapper` profile and its CI step are removed, and VaultWrapper stays excluded from the 0.8.17 `solc_floor` profile (it can't compile there). Its floor now equals the default compile version, so the ordinary `forge build` is the floor check. Rule 108 updated to match.

PR #2110, opened 2026-07-23
feat(commands): eip7702-atomic-batch skill + engine (EXSC-660)

0xDEnYO:

[EXSC-660](https://linear.app/lifi-linear/issue/EXSC-660/remediate-erc20proxy-ownership-transfer-from-compromised-legacy) (tooling spun out of the remediation)

The EXSC-660 remediation had to transfer `ERC20Proxy` ownership away from compromised legacy deployer EOAs whose native gas is drained by a sweeper bot the instant it arrives. On the aggressively-swept chains (arbitrum/bsc/polygon, all `_OLD_V1`) the standard fund-then-`transferOwnership` flow **lost the race every time** — even on Arbitrum, whose public sequencer feed lets the bot react. Pre-signing didn't help either (the node won't accept the transfer until the key is funded, and by then the sweep is in flight).

The reliable fix is **EIP-7702 sponsored execution**: the compromised EOA signs an off-chain authorization delegating its code to the canonical **Multicall3** (`0xcA11…`, same address on every EVM chain), and the deployer submits and pays for a single atomic type-4 tx running `aggregate3` in the EOA's context — so `transferOwnership(refundWallet)` executes with `msg.sender == the compromised owner`, **with no fundable balance ever landing on the key**. The sweeper stays idle; there's nothing to race. Verified live on Arbitrum (owner is now `refundWallet`).

This PR generalizes that into a reusable skill + engine, since it's useful well beyond this one remediation:

- **`.agents/commands/eip7702-atomic-batch.md`** — the skill (source of truth) + `.cursor`/`.claude` symlinks + README entry, following `add-rule-or-skill` conventions.
- **`script/tasks/atomicBatch7702.ts`** — config-driven engine. Two shapes: **sponsored** (rescue a gas-starved or compromised/swept key — sponsor pays, key only signs) and **self-batch** (one EOA runs several calls atomically). Dry-run by default; simulates every call as the authority before signing/spending; `--broadcast` to execute.

No Solidity, no contracts, no governance surface touched — `script/` + agent-config only.


script/tasks/atomicBatch7702.ts
```diff
@@ -0,0 +1,274 @@
...
+    console.log('\nDRY-RUN: no tx sent. Re-run with --broadcast to execute.')
+    return
+  }
+
+  console.log('\n→ sending EIP-7702 sponsored transaction ...')
+  const hash = await wallet.sendTransaction({
+    authorizationList: [authorization],
+    to: authority.address,
+    data: aggregateData,
+  })
```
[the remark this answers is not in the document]
0xDEnYO: Addressed in c83547d16 (docs). Added a Critical safety note: a public-mempool `--broadcast` exposes the sponsor-unbound 7702 authorization to front-running, so it must not be used for funded/privileged authorities — those need protected orderflow / a private bundle and/or a restricted executor instead of open `aggregate3`. The EXSC-660 authority is an empty, already-compromised key, so there's nothing to steal there. Also fixed the related sim (#2) and undelegate-on-revert (#4) issues in code, and dropped the false idempotency claim (#1).

script/tasks/atomicBatch7702.ts
```diff
@@ -0,0 +1,294 @@
...
+  // Clear the delegation unless asked to keep it. A 7702 authorization is applied BEFORE
+  // execution and is NOT rolled back if the batch reverts, so a mined-but-reverted tx also
+  // leaves the delegate active — undelegate on any mined batch, not just successful ones.
+  // Leaving an open delegate (Multicall3) lets anyone drive the authority afterwards.
+  if (!KEEP_DELEGATION) {
+    console.log('\n→ clearing delegation (authorization -> zero address) ...')
+    const clearAuth = await wallet.signAuthorization({
+      account: authority,
+      contractAddress: zeroAddress,
+    })
```
gvladika: 🔴 **Self-batch auto-undelegate uses the wrong authorization nonce → delegation is never cleared.**

In self-batch mode the authority *is* the sponsor, so this undelegate tx is sent by the same account that signs the clearing authorization. Per EIP-7702, when the authority also sends the tx the authorization nonce must be `txNonce + 1`, and viem only applies that `+1` when you pass `executor: 'self'` (see `prepareAuthorization.js:76-78`).

The batch step above correctly passes `executor: 'self'`, but this clear step does not. Walking the nonces in self mode: after the batch the authority's nonce is `N+2`; the clear tx consumes `N+2`, so the embedded authorization needs nonce `N+3`, but viem signs it with `N+2`. The authorization is invalid, the code stays delegated, and the `getCode` check below then prints "delegation still present" (a manual re-run hits the same bug).

This fails exactly where it matters: self-batch is the funded-account path, and auto-undelegate is the only thing protecting that account from being left delegated to open, access-control-free Multicall3. Sponsored mode (authority ≠ sponsor) is unaffected.

Fix — mirror the batch call:

```ts
const clearAuth = await wallet.signAuthorization(
  isSelf
    ? { account: authority, contractAddress: zeroAddress, executor: 'self' }
    : { account: authority, contractAddress: zeroAddress }
)
```

script/tasks/atomicBatch7702.ts
```diff
@@ -0,0 +1,294 @@
...
+    },
+    rpcUrls: { default: { http: [rpc] } },
+  })
+  const pub = createPublicClient({ chain, transport: http(rpc) })
+
+  const sponsor = privateKeyToAccount(
+    keyFromEnv(cfg.sponsorKeyEnv || 'PRIVATE_KEY_PRODUCTION')
+  )
+  const authorityEnv = cfg.authorityKeyEnv ?? ''
+  const isSelf = !authorityEnv || authorityEnv === cfg.sponsorKeyEnv
```
gvladika: 🟡 **`isSelf` compares env-var *names*, not resolved keys/addresses.**

If `authorityKeyEnv` points at a different env var that happens to hold the *same* private key as the sponsor, `isSelf` is `false`, so the batch takes the sponsored path with `executor` unset — and even the batch authorization then gets nonce `N` instead of `N+1`, making it invalid.

Cheap hardening: resolve both accounts first and derive `isSelf` from the addresses, e.g.

```ts
const authority = authorityEnv
  ? privateKeyToAccount(keyFromEnv(authorityEnv))
  : sponsor
const isSelf = authority.address === sponsor.address
```

script/tasks/atomicBatch7702.ts
0xDEnYO: Fixed in ebf6125d6 — applied your exact fix. The clear-delegation `signAuthorization` now mirrors the batch call and passes `executor: 'self'` when `isSelf`, so viem applies the `txNonce+1` and the clearing authorization is valid in self-batch mode. Without it the auth carried the already-consumed nonce and the delegation was never cleared — exactly the funded-account path auto-undelegate is meant to protect.

script/tasks/atomicBatch7702.ts
0xDEnYO: Fixed in ebf6125d6 — applied your suggested restructure. `authority` is now resolved first (`authorityEnv ? privateKeyToAccount(keyFromEnv(authorityEnv)) : sponsor`) and `isSelf = authority.address === sponsor.address`, so a distinct env var holding the sponsor's key is correctly treated as a self-batch and takes the `executor: 'self'` path.

gvladika (approved):

PR #2088, opened 2026-07-20
feat(scripts): ERC20Proxy ownership remediation toolkit (EXSC-660)

0xDEnYO:
# Which Linear task belongs to this PR?

[EXSC-660](https://linear.app/lifi-linear/issue/EXSC-660/remediate-erc20proxy-ownership-transfer-from-compromised-legacy) (related: EXSC-655)

# Why did I implement it this way?

Production `ERC20Proxy` on all prod chains is still owned by **legacy deployer EOAs that are compromised** (native gets swept by bots), instead of the `refundWallet` `0x156CeBba59DEB2cB23742F70dCb0a11cC775591F` that the health-check eval (PR #2078) expects. This PR adds the operational toolkit to remediate that safely and verifiably. **No contracts change — `script/tasks/` only.**

Sweep across all production networks found **68 EVM `ERC20Proxy`** needing transfer (every Receiver is already `refundWallet`-owned; tron is separate). Split by ownership mechanism:

- **38 two-step** (`TransferrableOwnership`): `transferOwnership` sets `pendingOwner`, then `refundWallet` calls `confirmOwnershipTransfer`. Reversible until confirmed.
- **30 single-step** legacy OZ `Ownable` (incl. moonbeam): `transferOwnership` finalises immediately — **irreversible, no step-2**.

Three keys (all now in `.env`) own them: `_OLD_V2` (`0x11F1022c…`, 51), `_OLD_V3` (`0xb1376839…`, 8 + tron), `_OLD_V1` (`0x11F11121DF…`, 9).

**Files (all `script/tasks/`, viem, dry-run by default):**

- `ownershipRemediation.ts` — **step 1**: per chain, fund the compromised owner minimally → `transferOwnership(refundWallet)` → sweep leftover gas back to the current deployer. Sequential per-network to minimise exposure; idempotent + resumable (reads on-chain state, `--only` to resume).
- `confirmOwnershipForMax.ts` — **step 2** (refundWallet key): `confirmOwnershipTransfer()` on the 38 two-step chains. Aborts if the key doesn't derive to `refundWallet`. Pre-checks readiness so it can be looped while SC works through step 1.
- `verifyOwnership.ts` — **read-only** status (`DONE` / `AWAITING-MAX` / `NOT-STARTED`); source of truth for progress. Complete when `68/68`.
- `ownership-remediation-manifest.json` — the 68-proxy list + mechanism split.

[...]

PR #2120, opened 2026-07-27
fix(healthCheck): eliminate false failures + refresh target state (68/71 networks red -> green)

0xDEnYO:

[EXSC-680](https://linear.app/lifi-linear/issue/EXSC-680/diamond-health-check-eliminate-false-failures-6871-networks-red)

> [!NOTE]
> **BE confirmation (2026-07-28):** Backend has switched over to `LiFiIntentEscrowFacetV2`. The V2-as-core + target-state couple in this PR is unblocked for merge.

The nightly `Diamond Health Check (all networks)` reported **68 of 71 networks failing** ([run 30192494683](https://github.com/lifinance/contracts/actions/runs/30192494683)). Bucketing every per-network reason showed it was 6 root causes, and most of the volume was the check disagreeing with config rather than real on-chain drift. This PR fixes the parts that are ours to fix in code; the genuine drift is called out below and left alone.

| Root cause | Networks | Fixed here |
|---|---|---|
| `LiFiIntentEscrowFacet` in `coreFacets` but deployed on 13 chains | 58 | ✅ grandfathered |
| Composer pairs reported "stale, not in config" — but they **are** in config | 22 | ✅ real bug |
| `FeeCollector` owner mismatch | 5 | ✅ check removed (deprecated) |
| `ReceiverAcrossV3.executor()` mismatch | 3 | ✅ entry removed (deprecated) |
| `ERC20Proxy` owner mismatch | 6 | ❌ real drift — see #2088 |
| Individually missing facets | 8 | ❌ real, needs deploys |

**54 of the 68 go green from this PR alone.**

### 1. `whitelist-integrity` was resolving periphery addresses by the wrong key

`getExpectedPairs` looked periphery contracts up as `deployedContracts[peripheryContract.name]`, ignoring the `address` field that `config/whitelist.json` carries (and that `IWhitelistConfig` declares as required). Two consequences:

- `Composer` is whitelisted but **never deployed by this repo**, so it is absent from every `deployments/*.json`. `contractAddr` came back `undefined`, the entry was silently skipped, and all of its on-chain pairs were then reported as stale.
- A name-keyed lookup returns one address, but a network can list **several** distinct `Composer` addresses — e.g. `PERIPHERY.arbitrum` has three, which is exactly the "3 stale pairs" in the log.

Config is the right source of truth for a check that asks *"does the diamond's whitelist match config"*. I also made the two failure modes visible instead of silent: a warning when an entry resolves to no address at all, and a warning when a config address disagrees with a contract we did deploy (that means the whitelist entry is stale and `diamondSyncWhitelist` would whitelist the wrong address).

### 2. Making a facet core "going forward" — `CORE_FACET_EXEMPTIONS`

`LiFiIntentEscrowFacet` was added to `coreFacets` in #1859, which makes the check demand it on **all** networks, but it was only deployed to the chains in #1997. Rather than drop it from `coreFacets`, this adds a narrow, per-network exemption table:

- the facet **stays** in `coreFacets`, so any **newly onboarded** network is absent from the table and therefore enforced by default — the safe direction, and what "core going forward" should mean;
- the listed pre-existing networks are exempt until the facet is backfilled.

Deliberately narrower than the existing `HEALTH_CHECK_EXCLUSIONS`, which disables a whole invariant on a network and would have cost coverage for every *other* core facet. This drops one facet from the expected set and leaves `core-facets-deployed` / `facets-registered` otherwise fully enforced.

The `networks` list is a shrinking to-do, not a steady state: delete a network the moment the facet lands there, and delete the entry when the list empties. Four tests guard it — the facet must really be in `coreFacets`, every network must exist in `config/networks.json`, the reason must be non-empty, no duplicates — so a stale exemption fails CI instead of quietly hiding a gap. Exemptions apply to the health check only; `deployCoreFacets.sh` still reads `coreFacets`, so deploying it everywhere stays a one-command job.

### 3. Deprecated checks

`feecollector-owner` removed (FeeCollector is deprecated; its owner is no longer maintained against `config.feeCollectorOwner`, which the deploy scripts still read). `ReceiverAcrossV3` dropped from `RECEIVER_EXECUTOR_GETTERS` — superseded by V4, which stays checked along with Chainflip, OIF and StargateV2. The now-unused `feeCollectorOwner` context plumbing is removed rather than left dead.

### 4. The log was overstating severity

`executeInvariant` printed its banner as `` `[${invariant.severity}] ${name} — ${description}` ``. That banner runs once per invariant per network, so ~1.3k **passing** checks rendered as `[error] diamond-deployed — LiFiDiamond is deployed`, each immediately followed by `[success] LiFiDiamond deployed`. Grep-indistinguishable from real failures, and most of why the run reads as catastrophic. Now `name — description (severity: error)`.

### Verification

Run against production RPCs from this branch:

- **katana** (previously failed on stale pairs) → `✔ Pair Array (getAllContractSelectorPairs) is synced. (20 pairs)`, `✔ Deployment checks passed`
- **flow** (previously failed on the missing facet) → exemption printed with its reason, `✔ Deployment checks passed`, exit 0
- **xdc** (previously failed on FeeCollector owner) → check no longer runs

`bun test script/` → 570 pass / 0 fail. `tsc --noEmit`, eslint and prettier clean.

### Deliberately not in scope

- **`ERC20Proxy` owner** on bob, etherlink, jovay, morph, nibiru, swellchain — real drift, needs the two-step ownership transfer via governance. #2088 looks like it covers exactly this.
- **Missing facets** on blast, boba, bsc, mainnet, metis, plume, sei, zksync — real gaps, need deploys.
- **`Could not enumerate ERC20Proxy authorized callers (RPC log range limit?)`** — one unchunked `eth_getLogs` silently skipping that sub-check on 45 networks. Real coverage loss, pre-existing, worth its own PR.
- **Tron 429s** — Tron needs its own throttle below the shared concurrency of 8. Pre-existing.

- [ ] For new facets: n/a


script/deploy/healthCheckInvariants.ts
```diff
@@ -384,16 +501,43 @@ const getExpectedPairs = async (
...
+              `Whitelist config lists ${peripheryContract.name} at ${configAddr} but deployments has ${deployedAddr}; config/whitelist.json may be stale.`
+            )
+
+          for (const selectorInfo of peripheryContract.selectors || []) {
+            expectedPairs.push({
+              contract: isTron
+                ? String(contractAddr)
+                : getAddress(contractAddr as Address).toLowerCase(),
+              selector: selectorInfo.selector.toLowerCase() as Hex,
+            })
```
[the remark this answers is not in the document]
0xDEnYO: Good catch — real bug, fixed in 7e4b185d4.

Verified the scope before fixing: across `config/whitelist.json`, `Composer` is the *only* duplicated periphery name (23 networks), and it appears in no `deployments/*.json`, so `deployedAddr` was always falsy and this branch was unreachable today. Latent rather than live — but it would fire the moment any shared name got deployed, which is exactly the noise class this PR removes, so it's worth closing.

Implemented as an `entriesPerName` Map built once per network (rather than a per-entry reduce with spread) and gated the comparison on `entriesPerName.get(name) === 1`.

Also exported `getExpectedPairs` and added 5 unit tests for the resolution rules, since this bug class had no coverage: config address wins when the contract is undeployed, all same-name entries are kept, no staleness warning on shared names, warning on a unique-name mismatch, and a warning rather than a silent skip when nothing resolves.

script/deploy/healthCheckInvariants.ts
```diff
@@ -172,6 +171,127 @@ export function getInvariantExclusion(
...
+ * Every entry is validated in `healthCheckInvariants.test.ts`: the facet must really be in
+ * `coreFacets`, every network must exist in `config/networks.json`, and the reason must be
+ * non-empty — so a stale exemption fails CI rather than silently hiding a real gap.
+ */
+export const CORE_FACET_EXEMPTIONS: ICoreFacetExemption[] = [
+  {
+    facet: 'LiFiIntentEscrowFacetV2',
+    reason:
+      'LiFiIntentEscrowFacetV2 supersedes LiFiIntentEscrowFacet and is core going forward (V2-227, #1997), deployed to the chains named there. The networks below predate that decision and are exempt until the facet is backfilled — remove a network here once the facet is deployed and registered on it.',
+    networks: [
```
melianessa, line 212: The exemption list covers 63 networks; the 13 active networks left enforced are arbitrum, arc, base, bsc, jovay, katana, mainnet, megaeth, optimism, pharos, polygon, robinhood, tempo. Twelve of those have LiFiIntentEscrowFacetV2 in deployments/*.json. tempo does not, and it's status: active, type: mainnet, with LiFiIntentEscrowFacetV2: 1.1.2 in its target state.

So tempo turns red on core-facets-deployed — a network the PR wasn't accounting for (it isn't in the body's "missing facets, needs deploys" list either). Please, either add it to the exemption table with a reason, or list it alongside blast/boba/etc. as known real drift.

0xDEnYO: Good catch — confirmed and fixed in 31c373554.

Verified: `tempo` is `status: active` / `type: mainnet`, carries `LiFiIntentEscrowFacetV2: 1.1.2` in `_targetState.json`, and has neither V1 nor V2 in `deployments/tempo.json` — so `core-facets-deployed` would have gone red. That's the identical not-yet-backfilled situation as the other exempted networks, not a distinct real-drift case, so I added it to `CORE_FACET_EXEMPTIONS` rather than the "needs deploys" list. Keeps the check green with a printed exemption until the facet lands on tempo, which is exactly what the exemption table is for.

`bun test script/deploy/healthCheckInvariants.test.ts` → 43 pass / 0 fail; tsc + eslint clean.

melianessa (approved):

PR #2127, opened 2026-07-27
fix(safe): refuse execution of future-nonce proposals by default (EXSC-690)

0xDEnYO:

[EXSC-690](https://linear.app/lifi-linear/issue/EXSC-690)

`confirm-safe-tx.ts` printed a GS026 warning for a proposal whose nonce is ahead of the Safe's
on-chain nonce and then still offered `Proceed anyway? (will revert with GS026)`. Taking that
option broadcasts a transaction that cannot succeed — gas is spent on a guaranteed revert and the
resulting failure is confusing to read back. Five lines above, the stale-nonce branch (the same
class of guaranteed revert) already hard-refuses with no override, so the two branches disagreed
about the same problem.

**Consistency with the stale-nonce branch.** The future-nonce execute path now refuses by default,
matching the stale branch. The stale branch itself is unchanged in behaviour.

**Escape hatch for the stale-RPC case.** The one legitimate reason to broadcast anyway is a lagging
RPC under-reporting the Safe's on-chain nonce, which makes an already-executable proposal look like
a future one. `ALLOW_FUTURE_NONCE_EXECUTION=true` re-enables execution for that case only; it does
not unlock the stale branch, because a stale reading would require an RPC reporting a nonce *ahead*
of consensus. The flag follows the existing `DRAIN_PARKED_TASKS` / `isDrainEnabled()` convention —
a small predicate helper (`isFutureNonceExecutionAllowed()`), not an inline `process.env` read — and
is documented in `.env.example`.

**Predicate placed in safe-utils for testability.** `confirm-safe-tx.ts` has no test file and is not
unit-testable as written (interactive prompts, MongoDB and RPC side effects). The decision is
therefore a pure exported function in `safe-utils.ts`, which has a colocated
`safe-utils.test.ts`:

```ts
canExecuteWithNonceStatus(status, { allowFutureNonce })
  // -> { canExecute: true;  reason: 'nonce-current' | 'future-nonce-override' }
  // -> { canExecute: false; reason: 'stale-nonce'   | 'future-nonce' }
```

The console UX (messages, colours, the blocking-proposal lookup) stays in `confirm-safe-tx.ts`,
which just switches on `reason`. `safe-utils.ts` gains no import from `safe-decode-utils.ts`
(rule 201 unaffected).

**Signing is still permitted — only execution is refused.** The action list is built before the
gate and always offers `Sign` for an unsigned proposal, independent of nonce status. Only the four
execute actions consult the predicate, so signatures can keep accumulating on a future-nonce
proposal while the blocking one is still pending. The refusal message says so explicitly and points
at the override flag.

**Governance impact (rule 105): none.** This does not change any Safe threshold, timelock delay,
role, owner set, or transaction authorization path. It only prevents broadcasting a transaction
that would revert on-chain; which transactions are authorized, and by whom, is untouched.

Verification (worktree, `origin/main` @ 358c2b996):

- `bun test script/deploy/safe/safe-utils.test.ts` -> 42 pass, 0 fail (8 new)
- `bunx eslint` on the three changed `.ts` files -> exit 0
- `bunx tsc-files --noEmit` on the three changed `.ts` files -> exit 0


0xDEnYO:
Absorbed into #2129 and closing here.

#2127's four files (`.env.example`, `safe-utils.ts`, `safe-utils.test.ts`, `confirm-safe-tx.ts`) were a strict subset of #2129's, and both PRs change what a signer is shown and allowed to do at the moment of execution — two open drafts on the same four files would have conflicted whichever merged first, and reviewing them apart meant splitting one decision across two reviews.

The branch was **merged, not cherry-picked**, so the commits and authorship are preserved on `feat/exsc-692-safe-proposal-provenance`. Both conflicts were purely additive and resolved as unions; verified afterwards that `canExecuteWithNonceStatus` and `isFutureNonceExecutionAllowed` are still covered — `bun test script/deploy/safe/safe-utils.test.ts` gives 69 pass / 0 fail.

EXSC-690 is now listed on #2129. Review continues there.

PR #2129, opened 2026-07-27
feat(safe): record proposal provenance and refuse future-nonce execution (EXSC-690/691/692/693)

0xDEnYO:

One PR covering four tickets — they are one change, split only by concern:

- Fixes [EXSC-691](https://linear.app/lifi-linear/issue/EXSC-691) — fail-soft git provenance helper (`script/deploy/shared/git-provenance.ts`)
- Fixes [EXSC-692](https://linear.app/lifi-linear/issue/EXSC-692) — capture provenance at the proposal storage funnel
- Fixes [EXSC-693](https://linear.app/lifi-linear/issue/EXSC-693) — show provenance to the signer in `confirm-safe-tx`
- Fixes [EXSC-690](https://linear.app/lifi-linear/issue/EXSC-690) — refuse execution of future-nonce proposals by default (**absorbed from #2127**)

# Absorbed #2127 (EXSC-690)

#2127 is folded in here and closed. It was not a merely adjacent PR: its four files were
`.env.example`, `safe-utils.ts`, `safe-utils.test.ts` and `confirm-safe-tx.ts` — a strict subset of
this PR's files. Two open drafts editing the same four files in the same Safe-execution path would
have conflicted whichever merged first, and a reviewer would have had to hold both in their head
anyway: both change what a signer is shown and allowed to do at the moment of execution.

What came across: `canExecuteWithNonceStatus` gates broadcast on where a proposal's nonce sits
relative to the Safe's expected nonce, refusing future-nonce execution by default (such a broadcast
reverts with GS026), plus the `ALLOW_FUTURE_NONCE_EXECUTION` operator escape hatch read by
`isFutureNonceExecutionAllowed` for the case where the configured RPC is known to report a stale
on-chain nonce.

The branch was merged, not cherry-picked, so #2127's commits and authorship are preserved in the
history here. Both conflicts were purely additive (both sides adding to `.env.example` and to the
test file's import list) and were resolved as unions — nothing was dropped. Verified after the fold:
`bun test script/deploy/safe/safe-utils.test.ts` → **69 pass / 0 fail**, with #2127's own
`canExecuteWithNonceStatus` and `isFutureNonceExecutionAllowed` describes intact; `tsc-files
--noEmit` clean on all three shared TypeScript files.

**The problem.** A pending Safe proposal today says what it does but not where it came from. A signer asking "what is this, and can I trust the code behind it?" has to go and ask. Every proposal now records who created it, from which commit and branch, whether that commit is fetchable, whether the working tree was dirty, and optionally why.

**Captured at the single storage funnel, not at call sites.** All ten call sites — including the five bespoke task scripts and the Tron route — pass through `storeTransactionInMongoDB`, so capture lives there and every caller inherits it with **zero changes**. The new parameter is appended last and optional, so nothing else had to be touched.

**Fail-soft, because this sits on the deploy path.** A Mongo write failure here already aborts a deployment, so a git helper that throws would be able to take a production deploy down. Every probe swallows its own failure, returns an `'unknown'` sentinel and records a one-line reason in `captureErrors`; a whole-capture backstop catches anything unexpected. There are three near-duplicate git snippets in the repo today with inconsistent behaviour (one of them throws) — this module is the single fail-soft policy, but the existing call sites are deliberately **not** refactored here.

**Injectable override seam, in the same commit.** `safe-utils.test.ts` calls `storeTransactionInMongoDB` directly. Without a seam that suite would start shelling out to real `git` and asserting against whatever checkout CI happens to have. `provenanceOptions.override` is what keeps it deterministic — a requirement of this change, not polish.

**Optional field, for backward compatibility.** `provenance?` is optional and old rows simply have none. The signer view renders one explicit *"not recorded"* line for them rather than a silent gap — a gap reads as "clean and authored by nobody", which is the one impression this block must never give.

**Why the three parts ship together.** The capture (EXSC-692) does not compile without the helper (EXSC-691), and the helper alone is dead code. The display (EXSC-693) could have been split out, but it is the only thing that makes the captured data visible, so reviewing it separately would mean reviewing a feature nobody can see. I have left *"as small as possible"* unticked rather than claim otherwise.

## Design details worth a reviewer's attention

- **Scoped dirty tree.** `git status --porcelain` minus the paths the deploy pipeline rewrites during its own run (`deployments/**`, `script/deploy/_targetState.json`). `config/whitelist.json` and `config/networks.json` are deliberately **not** excluded — a dirty whitelist at proposal time is exactly what a reviewer wants to know about. Capped at 20 entries with a truncation flag.
- **Memoized per process.** The multi-network task scripts store one proposal per network in a loop; without the memo a 50-network run would spawn several hundred git processes. Measured on this branch: ~120 ms for the git probes, ~870 ms including the `gh` lookup, then 0 ms for every subsequent proposal in the run.
- **PR-URL lookup is best-effort.** 5 s timeout, non-interactive `gh` environment, skipped for `main`/detached/unknown branches, skippable via an option, and every failure (missing `gh`, unauthenticated, no PR, timeout) is swallowed *without* recording a capture error — otherwise every proposer without `gh` would see a spurious "capture incomplete" marker on every proposal.
- **`commitOnRemote` is honest rather than clever.** It reads local remote-tracking refs (`git branch --remotes --contains`), so a stale checkout can report `false` for a commit that is in fact pushed. The CLI says `NOT PUSHED (per local refs)` instead of pretending, and no network fetch is added to the hot path.
- **Never reads `safeTx`.** The Tron flow fabricates that object through a cast and its shape is not trustworthy, so capture reads ambient git state only. Covered by a test that stores a Tron-shaped document.
- **The CI and bot branches are forward-looking.** No workflow creates Safe proposals today, so those paths are unit-tested via `process.env` stubs only. This does not close an existing CI gap.
- **`SAFE_PROPOSAL_REASON`** is read from the environment — no CLI plumbing in this PR, that is EXSC-694 — and is optional, with no warning spam. `SAFE_PROPOSAL_ACTOR=bot` is the opt-in a future unattended job sets. Both are in `.env.example`.

## Provenance is context, not a security control

Worth stating plainly, because the field names invite the opposite reading: **this data is self-reported by the proposing machine.** It makes honest mistakes visible — an unpushed commit, a dirty whitelist, a proposal nobody can explain — and it gives later checks something concrete to verify against. It is **not** a defence against a proposer who is deliberately lying, and a signer should not read a green "clean / pushed" line as authentication of anything.

## Governance impact (rule 105)

**None.** This is additive metadata on a MongoDB document. No change to Safe thresholds, owner sets, timelock delays, roles, proposal authorization, signing, or execution. No Solidity is touched and no on-chain behaviour changes; the only user-visible difference is a few extra informational lines in the `confirm-safe-tx` prompt. Nothing was added to `DEPLOYMENT_QUERY_EQ_KEYS` and `mongo-log-utils.ts` is untouched, so deployment-record identity and upsert behaviour are unchanged too.

## Verification

- `bun test script/` — **633 pass, 0 fail** (35 files). `script/deploy/safe/` + `script/deploy/shared/` alone: **458 pass, 0 fail**. 70 of those tests are new: 44 in `git-provenance.test.ts`, 13 in `provenance-display.test.ts`, 13 added to `safe-utils.test.ts`.
- `bunx eslint` and `bunx tsc-files --noEmit` on all seven changed/added files: exit 0. Also typechecked all ten call sites of the changed signature: exit 0.
- Smoke-checked the real capture path against this worktree. That is how the one real bug in the first draft surfaced: trimming `git status` output ate the leading status column and turned `.env.example` into `env.example`. Fixed, with a regression test for an unstaged first entry.

- **EXSC-694** — `--reason` CLI flag plumbed through `propose-to-safe.ts` and the bash chain.
- **EXSC-695** — the deploy-log twin (`gitBranch` / `dirtyTreeScoped` / `actor` on `IDeploymentRecord`), reusing this module. Note for whoever picks it up: those fields must **not** go into `DEPLOYMENT_QUERY_EQ_KEYS`, or an upsert becomes branch-sensitive and starts duplicating records.
- Exposing `provenance` through `IProposalSummary` for `list-pending-proposals --json`, and hashing governance config into the block, are both deliberately deferred.


0xDEnYO:
@coderabbitai review

0xDEnYO:
## Review gate on `25c810d42` — escalated findings

Four review passes over the fix commit. The control-character fix **closes the CodeRabbit finding as written** (every `Cc` character is stripped; verified by execution), but the gate found the guard incomplete for its own stated threat model. Nothing below was auto-fixed — all of it touches the signing display path, which the gate escalates by policy.

### 1. CRITICAL — `\p{Cc}` is the wrong character class; a proposer can forge a line in the signer's prompt

`\p{Cc}` covers C0/C1/DEL but **excludes category `Cf`**, so `U+202E` (RLO), `U+2066`–`U+2069` (bidi isolates) and zero-width characters survive both sanitizers. `U+2028` (LINE SEPARATOR, category `Zl`) survives `sanitize` too.

`git config user.name` and dirty filenames are proposer-controlled, and `sanitize()` performs no whitespace collapse, so padding plus `U+2028` injects a whole fake line. Rendered output from the real module:

```text
Proposed by:   Alice<U+2028>    Working tree:  clean   Reason:  reviewed by security (human)
Working tree:  ⚠ 1 dirty: src/Facets/<U+202E>los.doGtsurTtEvil.sol
```

The forged line claims a clean tree and a security review while the real line below it reports the tree dirty, and RLO reverses the displayed filename (Trojan Source). `U+2028` is a hard line break in VS Code's integrated terminal and in any JSON/HTML re-render of the record.

Note the asymmetry: `normalizeProposalReason` neutralises `U+2028` **by accident**, because `\s` happens to match it. `sanitize` has no `\s` pass, so the display path does not.

- `script/deploy/safe/provenance-display.ts:39-40`
- `script/deploy/safe/safe-utils.ts:1394-1401`

A fix needs `\p{Cf}` (plus `U+2028`/`U+2029`, or an allowlist) **and** a whitespace collapse in `sanitize`. Worth deciding deliberately: a naive `\p{Cf}` strip also breaks ZWJ emoji, which a current test asserts as desired behaviour.

### 2. MAJOR — the tests cannot detect finding 1, and one enshrines it

Every new assertion is `expect(CONTROL.test(text)).toBe(false)` where `CONTROL = /\p{Cc}/u` — **the same predicate the implementation applies**, so it can only ever pass. This is the third tautological-assertion bug of the family already caught twice in this commit's own development.

Worse, `provenance-display.test.ts` (~line 228) asserts `'déployer 日本語 — naïve 👨‍👩‍👧'` survives intact as a *feature*. That emoji is held together by `U+200D` ZWJ — category `Cf` — which is precisely the class that lets RLO through. The test that proves unicode is preserved is the same test that guarantees the bypass.

- `script/deploy/safe/provenance-display.test.ts:180-236`
- `script/deploy/safe/safe-utils.test.ts:475-491`

### 3. MAJOR — sanitization is display-only; the stored record keeps raw control characters

Only `reason` is normalised at capture. `proposerHandle`, `gitBranch`, `dirtyTreeScoped`, `prUrl` and `captureErrors` are written unsanitised (`git-provenance.ts:347` `git config user.name`, `:523` `gh` output, stderr into `captureErrors`). Any second reader — a Mongo shell dump, a log line, a future dashboard — renders raw ESC. The guard protects exactly the one consumer that happens to exist today.

### 4. MAJOR — `formatProvenanceLines` throws on a malformed row and aborts the whole signing session

Its own comment (`provenance-display.ts:81-82`) states a hand-edited or half-migrated document "must degrade to 'unknown', never abort the signing session". The fix commit added `String()` guards for `path` and `captureErrors[0]` but left `reason`, `prUrl`, `proposerHandle`, `actor`, `gitCommit`, `gitBranch` unguarded — an internal inconsistency inside one commit. Executed against the real function:

```text
dirtyTreeScoped: 'config/whitelist.json'  -> TypeError: (provenance.dirtyTreeScoped ?? []).map is not a function
reason: 42                                -> TypeError: text.replace is not a function
prUrl: {}                                 -> TypeError: text.replace is not a function
```

The call site at `confirm-safe-tx.ts:393` sits inside `processTxs`, awaited at `:974` with **no** try/catch — so one bad row kills every remaining network.

### 5. MAJOR — a failed capture is memoised for the process lifetime and poisons every later proposal

`captureGitProvenance` caches the **failed** result (all-sentinel + `captureErrors`) with no invalidation, so one transient git/spawn hiccup on the first network stamps all subsequent proposals in a 50-network run as `unknown`.

CodeRabbit raised this exact shape twice on #2133: *"A failed init is cached permanently and poisons every later call"* / *"A transient failure permanently pins that rejection"*.

Related: `capturedAt` is stamped fresh per call (`safe-utils.ts:1428`) while the git fields come from the memo — so proposal #50 carries a fresh timestamp over proposal #1's git state.

Also related: the doc comment on `resetGitProvenanceCache` justifies the memo with "git state cannot meaningfully change within one script run", which `PROVENANCE_DIRTY_EXCLUDES` in the same file contradicts ("Paths the deploy pipeline itself writes **mid-run**").

### 6. MEDIUM — sentinel diverges from the repo's deliberate choice

This PR introduces `PROVENANCE_UNKNOWN = 'unknown'` (lowercase). EXSC-330 / #2017 (`1746692b3`) deliberately standardised on uppercase `'UNKNOWN'` for the same "capture failed" meaning, with the rationale that an ambiguous sentinel is indistinguishable from a pre-field default. Since this module's header states it intends to absorb the deploy-log call site next, audit queries keyed on `'UNKNOWN'` would silently miss provenance rows.

### 7. Lower-confidence — human judgment

- `slice(0, MAX_PROPOSAL_REASON_LENGTH)` can cut a surrogate pair and store a lone surrogate; BSON/JSON round-trips then replace it or throw. `safe-utils.ts:1403`
- `normalizeProposalReason` collapses whitespace *before* stripping `Cc`, so a control char between words silently joins them (`'word' + NUL + 'next'` → `"wordnext"`) and leaves double spaces. Re-collapsing after the strip fixes both. Not exploitable — no control char survives.
- `MAX_PROPOSAL_REASON_LENGTH` is enforced only at capture, so a hand-edited row renders uncapped and can scroll the transaction details off screen.
- The `catch` in `buildProposalProvenance` (`safe-utils.ts:1437-1448`) is unreachable — `captureGitProvenance` already wraps its body and `cloneGitProvenance` cannot throw.
- `SAFE_PROPOSAL_REASON` appears in no operator runbook. The comparable `DRAIN_PARKED_TASKS` is documented in `.agents/commands/multisig-rollout.md` and `docs/DeferredDiamondCleanupQueue.md`; operators following `/multisig-rollout` will never set the new var, so every proposal renders `— none given —`.

### Verified clean

All 10 `storeTransactionInMongoDB` call sites read individually — the appended optional param lands correctly everywhere, nothing in the `parkedTaskRefs` slot. The merge against main's pooled-Safe-client refactor is coherent (no dangling references to the four deleted helpers, no double init). `startupReconciledKeys` survives the #2133 prefetch boundary intact. Provenance reaches the display unprojected. Provenance is correctly excluded from `computeProposalIntentHash`. 120 tests pass, eslint clean, `tsc --noEmit` clean in every touched file.

---

**Escalated items need a decision before this PR is ready.** Findings 1–5 are behavioural changes on the Safe signing path, which this gate does not auto-fix.

0xDEnYO:
@coderabbitai review

0xDEnYO:
## CodeRabbit round closed, PR flipped to ready

Both open CodeRabbit threads are fixed in `ba25a900a` and resolved:

- `.env.example:87` — `ALLOW_FUTURE_NONCE_EXECUTION=""` → bare `=`, matching the two keys above it. Behaviourally inert: the consumer is `process.env.ALLOW_FUTURE_NONCE_EXECUTION === 'true'` (`safe-utils.ts:1016`), a strict compare under which empty-string and unset are both false, so the fail-closed default is unchanged.
- `docs/DeferredDiamondCleanupQueue.md:552` — the claim that `ISafeTxDocument` carries no cleanup-origin field contradicted Fact 6, which this PR updated to list `parkedTaskRefs?`. Reworded to state it as the pre-change position and point at the field the spec adds.

Re-gated the fix commit. The diff is one prose sentence plus an unquoted empty default — no new check, guard or code path — so the falsification pass had nothing to prove beyond the strict-compare check above.

**Also re-verified the CRITICAL escalation from the 2026-08-18 gate round** (`\p{Cc}` alone left bidi overrides and `U+2028` intact, allowing a forged line in the signer's prompt). It is genuinely closed, verified by executing `sanitizeProvenanceText` against the attack payloads rather than trusting the same-PR tests:

```text
PASS  RLO Trojan Source    -> "los.doGtsurTtEvil.sol"
PASS  U+2028 line forge    -> "Alice Working tree: clean"
PASS  U+2029 para sep      -> "Bob Reason: approved"
PASS  bidi isolates        -> "xy"
PASS  zero-width space     -> "ab"
PASS  ESC repaint          -> "Carol[2K[1Gfake"
PASS  ZWJ emoji kept       -> "dev 👨‍💻"
```

The class is now `[\p{Cc}\p{Cf}\p{Zl}\p{Zp}]` with U+200D exempted, and whitespace collapses first — which is what flattens the `U+2028` forgery onto a single line instead of letting it break one. The ESC case keeps the literal `[2K[1G` text, which is inert once the escape prefix is gone.

Branch updated onto current `main` before flipping to ready.

0xDEnYO:
@coderabbitai review

gvladika (approved):

PR #2130, opened 2026-07-28
docs(aikido): correct SAST rule name in false-positive catalog

0xDEnYO:

None — doc-only correction to an agent reference file, labelled `trivial` per the ticket-linkage carve-out in `.agents/context.md`.

While triaging Aikido findings on #2075, the `path_traversal_scripts` entry in the false-positive catalog told the operator to open a rule named **"Path Traversal"** in the Aikido UI and add Custom Code Context to it. No such rule exists — the actual rule is titled **"Potential file inclusion attack via reading file"**, and it appears once per language (TS and JS rows carry the identical title), so the pointer was ambiguous as well as wrong. This sent a real triage session down a dead end.

Two changes to the one `sast_context` pointer line:

- Name the rule as it actually appears, and specify the **TS** row.
- Mark the click path from the SAST rules list to "Custom Code Context" as **unverified** rather than asserting it. Rows on that list did not appear to open a detail panel, and I could not verify the correct path myself (no access to the Aikido web app). Better to flag the gap than to encode another guess — the step is optional and gates nothing, so a wrong instruction costs more than an honest unknown.

No behavior change: `ignore_reason` and the `sast_context` body text are untouched, so `/aikido-address-findings` triage output is unaffected.

`main` has since rewritten the `ignore_reason` and the `sast_context` blockquote body (the
containment-guard work in #2162). This branch has been merged up to `main` and the pointer-line
edit re-applied on top of that newer text, so none of it is reverted — the diff is still the
single header line.

## Known inconsistency, follow-up intended

This PR marks the click path to "Custom Code Context" as unverified, but
`.agents/commands/aikido-address-findings.md:227` still documents that same path as a working
recipe (`Aikido UI → Repositories → Checks → View SAST Rules → [rule] → Custom Code Context →
Scope: global`). So after this merges the repo briefly holds both claims. Reconciling that file is
deliberately out of scope here — this PR is one line in the catalog — and is left as a follow-up
once someone with Aikido web access can confirm the real path.

## Overlap with #2131

PR #2131 originally carried this identical hunk. It has been dropped there, so that PR no longer
touches the catalog and the two are independent — they can merge in either order.


0xDEnYO:

PR #2131, opened 2026-07-28
chore(agents): consolidate five agent-config PRs — deploy routing, Aikido rule pointer, Bun test rule, Slack MCP removal, companion periphery

0xDEnYO:
# Which Linear task belongs to this PR?

EXSC-770 (Slack MCP removal), EXSC-687 (Bun TS-test rule), EXSC-826 (companion periphery) — plus two
Aikido/agent-config fixes with no dedicated ticket.

# Why did I implement it this way?

**This PR consolidates five separate agent-config PRs into one review.** All five touched only
`.agents/**`, `.claude/**` and `.mcp.json`, none had a reviewer yet, and four of them had been idle
since 2026-08-18. Six review cycles for 35 changed lines was the wrong trade with one SC reviewer
left, so they are folded here and closed as absorbed.

Net effect is 6 files, +35/-10. Nothing was dropped in the fold — the branches were merged, not
cherry-picked, so each original commit and its authorship survive in the history.

| Absorbed | What it does |
| --- | --- |
| this branch | Route production deploys through `multisig-rollout` rather than `deploy-contract`, and sharpen the Aikido SAST-rule pointer |
| #2130 | Correct the SAST rule name in the false-positive catalog — there is no rule literally named "Path Traversal"; the real one is the **TS** row "Potential file inclusion attack via reading file" (a JS row carries an identical title). The click path to Custom Code Context is explicitly marked unverified rather than asserted |
| #2134 | Note the Bun process-level assertion gotchas in the TypeScript test rule (`402-typescript-tests.md`) |
| #2198 | Drop the unusable Slack MCP server from `.mcp.json`, and carve out non-DCR servers from the add-a-server rule in `.claude/README.md` |
| #2253 | Resolve required companion periphery at target resolution, in `deploy-contract.md` and `multisig-rollout.md` |

The two Aikido halves are complementary, which is part of why they belonged together: this branch
sharpens the *pointer* to the rule, #2130 corrects the *name* it points at. Reviewing them apart
meant reviewing half a statement.

**Deliberately NOT folded in — #2223** (`docs(agents): align /deprecate-network with real practice`,
EXSC-797). It looks like a sibling docs PR but it is not: it conflicts semantically with the deferred
diamond-cleanup queue that has since merged. Its rewrite of the `/deprecate-network` step list drops
the queue-cancellation step that is now real practice, while adding genuinely needed steps (scrubbing
[...]

PR #2134, opened 2026-07-28
docs(agents): note Bun process-level assertion gotchas in TS test rule (EXSC-687)

0xDEnYO:

[EXSC-687](https://linear.app/lifi-linear/issue/EXSC-687) — this note is a byproduct of writing
`script/deploy/github/verify-approvals.test.ts` in #2128. No separate ticket was cut; the finding
belongs to the same work.

Two Bun-specific traps cost real debugging time while writing tests for the approval-check exit-code
gate, and both are invisible until you hit them:

1. `spyOn(process.stdout, 'write')` does **not** capture writes made by the code under test — the
   assertion reports "was not called" even though the write demonstrably happens. The working
   pattern is to inject the output/exit surface as a parameter defaulting to `process`
   (`reportApprovalResult(failures, target: IReportTarget = process)` in #2128) and pass a recorder
   in the test.
2. `expect(...).rejects` trips `@typescript-eslint/await-thenable`, because Bun's matcher isn't a
   real Promise. `script/deploy/safe/parked-tasks.test.ts` already documents this inline and works
   around it with a local `expectRejects` helper.

Both are "asserting on process-level behaviour" problems, so they're captured as one
`[CONV:TEST-PROCESS-SURFACE]` section rather than two scattered bullets. Landed in
`.agents/rules/402-typescript-tests.md` (globs `**/*.test.ts`) because that's the narrowest owning
rule — a test author hits both traps, and the rule auto-loads exactly when they do. The inline
comment in `parked-tasks.test.ts` stays where it is; the rule generalises it so the next author
doesn't have to have read that file.

Edited in `.agents/` per the repo's symlink convention; `.cursor/rules/402-typescript-tests.mdc`
and `.claude/rules/402-typescript-tests.md` already point at it, so no symlink work was needed.
`.agents/README.md` is unchanged — `name`, `description` and `globs` are all untouched, which is the
only trigger for a README update.

**One thing for the reviewer:** the `verify-approvals.ts` reference points at code that only exists
on #2128's branch — it resolves once that PR merges. Everything else cited (`expectRejects` in
`parked-tasks.test.ts`) is on `main` today. The Bun version is called out as *observed on 1.3.8*
(the local toolchain); the repo pins `bun@1.3.13` in `packageManager` and I did not re-verify the
spy behaviour there, so the guidance is stated unconditionally rather than as version-gated.

### DIFF
--- a/.agents/rules/402-typescript-tests.md
+++ b/.agents/rules/402-typescript-tests.md
@@ -18,6 +18,13 @@ paths:
 - **Restore after each test**: when stubbing globals (e.g. `globalThis.fetch`), save the original in `beforeEach`/`beforeAll` and restore it in `afterEach`/`afterAll` to avoid cross-test pollution.
 - Prefer true unit tests that isolate the code under test; use mocks for any outbound calls (fetch, contract calls, file system if needed) so failures reflect logic bugs, not environment or network issues.
 
+## Asserting on process-level behavior ([CONV:TEST-PROCESS-SURFACE])
+
+Don't reach for a spy or a Bun matcher when asserting that code wrote to stdout, exited, or rejected — both routes fail for reasons unrelated to the code under test.
+
+- **Never assert via `spyOn(process.stdout, 'write')`** (same for `process.exit`). The spy does not capture writes made by the code under test — the assertion reports "was not called" even though the write happens (observed on Bun 1.3.8). Instead **inject the output/exit surface as a parameter defaulting to `process`** and pass a recorder in the test. `reportApprovalResult(failures, target: IReportTarget = process)` in `script/deploy/github/verify-approvals.ts` is the reference shape: production callers pass nothing, the test passes a fake and asserts on what it captured.
+- **Never assert rejections via `expect(...).rejects`** — awaiting Bun's matcher trips `@typescript-eslint/await-thenable` because it isn't a real Promise. Use a local `async function expectRejects(promise, match)` that catches and matches the error message, as in `script/deploy/safe/parked-tasks.test.ts`.
+
 ## Post-Change Actions
 
 - After TS test changes, run Bun tests (or state which suites remain).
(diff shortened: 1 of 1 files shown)


0xDEnYO:
**Light-triage gate note: bullet 1's factual premise fails to reproduce on Bun 1.3.8** (the exact version the observation cites).

Falsification attempt before this ships as a rule:

- `spyOn(process.stdout, 'write')` **did** capture writes in all three variants tried: direct call in the test file, `process.stdout.write` from an imported module, and a write followed by `process.exit` (both spied). `spy.mock.calls` contained the payloads every time.
- What does **not** get captured is `console.log` output — in Bun it bypasses `process.stdout.write` entirely (0 recorded calls), unlike Node where console routes through the stream. If the failing assertion in the #2128 test session was over console-based output (or a `console.error` → stderr mixup), that would explain the observation.
- Bullet 2 verified true: `bun-types/test.d.ts:840` types `rejects: Matchers<unknown>` (not a Promise), so awaiting it trips `@typescript-eslint/await-thenable`; the `expectRejects` reference in `script/deploy/safe/parked-tasks.test.ts` exists as described.

Suggestion: keep the inject-the-surface advice (good design regardless), but reword bullet 1's justification — as written it bans a working technique on a premise that doesn't repro. If the original failure is reproducible, the rule should name the actual trigger (console vs stream write) instead of the blanket claim. Repro snippet:

```ts
import { describe, it, expect, spyOn } from 'bun:test'
import { writeViaStdout, writeViaConsole } from './mod' // process.stdout.write(...) / console.log(...)

it('stdout.write from another module IS captured', () => {
  const spy = spyOn(process.stdout, 'write').mockImplementation(() => true)
  writeViaStdout()
  expect(spy.mock.calls.length).toBeGreaterThan(0) // passes on 1.3.8
  spy.mockRestore()
})
```

0xDEnYO:

PR #2137, opened 2026-07-28
feat(tron): deliver diamondCut init calldata (EXSC-650)

0xDEnYO:

EXSC-650 (Tron leg — the EVM side is in #2075)

**Problem.** The Tron cut path cannot deliver an init call.
`script/deploy/shared/propose-diamond-cut.ts` hardcoded `address(0)` as the
diamondCut init address and `0x` as its calldata, and `proposeDiamondCut()`
exposed no way to override either. On EVM the init call comes from
`UpdateScriptBase.update()`, which passes the new facet address plus
`getCallData()` as the cut's `_init`/`_calldata` pair.

For AllBridgeFacet v2.2.0 that gap is a liveness bug, not a cosmetic one:
deploying it to Tron unchanged leaves `chainMappingsInitialized == false` with
every mapping zero, so **every `startBridgeTokensViaAllBridge` from Tron would
revert `UnsupportedAllBridgeChainId`**. It is not repairable after the fact —
`setChainIdToAllBridgeChainId` itself reverts `NotInitialized`, so only the
owner-only `initAllBridge` can bootstrap the storage.

**Approach.** Thread the init through the cut rather than batching a second
timelock call. Both were viable (`propose-calls-tron.ts` already supports
multi-call `scheduleBatch`), but riding inside the cut is atomic by
construction, needs no batch plumbing, and produces the same on-chain shape as
every EVM deployment — the diamond delegatecalls the initializer within
`diamondCut`, so there is no window where the facet is live but uninitialised.

- `IDiamondCutInit` groups `initAddress` + `initCalldata` in one object so an
  address can never be passed without its calldata; a mismatched pair either
  silently skips the init or reverts the whole cut.
- `excludeSelectors` also threads through, and the AllBridge script excludes
  `initAllBridge` — mirroring `UpdateAllBridgeFacet.getExcludes()`. Without it
  `getFacetSelectors()` would register the initializer on the diamond, a
  divergence from every EVM chain.
- The initializer is encoded **before** anything is deployed, so a malformed
  `mappings` array aborts up front rather than after a facet is on-chain.

**Chain ids are revived from the JSON source text.** `config/allbridge.json`
carries `9270000000000000`, above `Number.MAX_SAFE_INTEGER`, so `JSON.parse`
alone cannot be trusted to reproduce these ids exactly. The reviver's
`context.source` argument gives the untouched number literal for exact `BigInt`
conversion (and rejects fractional ids for free). The Foundry path has no such
hazard — `stdJson.parseRaw` decodes straight to `uint256`.

**Verification.** The encoded `initAllBridge` calldata for all 15 real mappings
is byte-identical to an independent Foundry encoding
(`cast calldata "initAllBridge((uint256,uint256)[])" ...`), which is how the
`MAX_SAFE_INTEGER` issue was caught in the first place. 9 new unit tests cover
precision, the zero sentinel, fractional/missing/non-numeric ids, duplicates,
and the encode round-trip. Error paths were exercised end-to-end; notably
today's `config/allbridge.json` (no `mappings` array yet) fails loud with
`no non-empty "mappings" array` rather than encoding an empty init.

**Sequencing.** This lands the deploy-tooling plumbing only; it does not touch
`config/allbridge.json` or any contract. The `mappings` array it consumes
arrives with #2075 (AllBridgeFacet v2.2.0), so the two PRs are independent to
merge in either order, but the Tron deployment itself needs both. Everything
here reaches `lifinance/contracts-tron` via the weekly upstream sync — this PR
supersedes lifinance/contracts-tron#25, which is closed in favour of landing
the change at its canonical home so the fork copies never diverge.

### DIFF
--- a/script/deploy/shared/propose-diamond-cut.ts
+++ b/script/deploy/shared/propose-diamond-cut.ts
@@ -13,21 +13,70 @@ import { encodeFunctionData, type Address, type Hex } from 'viem'
 import { getFacetSelectors } from '../../utils/utils'
 import type { TronTvmNetworkName } from '../tron/types'
 
-import { DIAMOND_CUT_ABI } from './constants'
+import { DIAMOND_CUT_ABI, ZERO_ADDRESS } from './constants'
+
+/**
+ * Post-cut initializer, delegatecalled by the diamond in the same transaction
+ * as the cut. Mirrors the `_init`/`_calldata` pair that
+ * `UpdateScriptBase.update()` passes on EVM. Grouped in one object so an
+ * address can never be supplied without its calldata (or vice versa) — a
+ * mismatched pair either silently skips the init or reverts the whole cut.
+ */
+export interface IDiamondCutInit {
+  /** Contract to delegatecall — normally the facet being added. */
+  initAddress: Address
+  /** Encoded initializer call, e.g. `initAllBridge(ChainIdConfig[])`. */
+  initCalldata: Hex
+}
 
 /**
  * Encode a `diamondCut` calldata for adding a facet.
  * Resolves selectors from Forge artifacts automatically.
+ *
+ * @param facetName - Facet whose selectors are read from the Forge artifact
+ * @param facetAddressHex - Deployed facet address (EVM hex form)
+ * @param options.init - Optional post-cut initializer (see {@link IDiamondCutInit})
+ * @param options.excludeSelectors - Selectors to leave unregistered, mirroring
+ *   `getExcludes()` on the EVM update scripts (e.g. an owner-only `init*`
+ *   function that is delegatecalled by the cut and must not be reachable
+ *   through the diamond afterwards)
  */
 export async function encodeDiamondCutCalldata(
   facetName: string,
-  facetAddressHex: Address
+  facetAddressHex: Address,
+  options: {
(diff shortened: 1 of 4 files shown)


0xDEnYO:
@coderabbitai Regarding the Major finding ("Gate `initAllBridge` so redeploys do not submit an initialization calldata"): rejected after verifying against the v2.2.0 facet source in #2075.

`initAllBridge` has **no already-initialized guard** — it is owner-gated and idempotent by design: it overwrites `sm.allBridgeChainIds[chainId]` for each entry and (re-)sets `chainMappingsInitialized = true`. It reverts only on an empty array or a zero id, both of which `parseAllBridgeMappings` rejects before anything is deployed. So a redeploy that carries the init cannot "revert on a second initialization" — it re-seeds the mappings from `config/allbridge.json`, which keeps the config as the source of truth and matches what the EVM `UpdateAllBridgeFacet.getCallData()` path does on every update run.

The scenario that *would* block a redeploy cut is the FacetCutAction.Add colliding with already-registered selectors — but that is pre-existing behavior of this script's cut construction, independent of the init calldata, and out of scope for this PR.

0xDEnYO:
@coderabbitai review

melianessa (approved):

PR #2139, opened 2026-07-28
docs(tron): document the fork-delta guard and refresh the sync section

0xDEnYO:

[EXSC-716](https://linear.app/lifi-linear/issue/EXSC-716/guard-tron-fork-delta-against-silent-version-mislabelling) — companion to [contracts-tron#26](https://github.com/lifinance/contracts-tron/pull/26), which adds the machinery.

`docs/TronFork.md` lives in this repo by design (it says so itself) and reaches the fork through the normal sync, so the write-up belongs here even though every file it describes lives in `contracts-tron`. **Merge this before contracts-tron#26** so the fork's CI messages point at a section that exists.

## What changed

**Refreshed the sync section.** It still described the PR-based process, and carried a "Planned / target state (WIP)" block for a scheduled sync job that has since shipped as `syncUpstreamContracts.yml`. Both replaced with what actually runs, as a merge-outcome table.

**New section: the fork-delta guard.** The gap it closes is subtle enough to be worth writing down:

- This repo's `versionControlAndAuditCheck.yml` forces a version bump on any audit-relevant change to a `src/**/*.sol` file. So when upstream changes an overlaid contract, our `-tron` line and upstream's plain version line are the *same line* — git conflicts every time, and such a change can never merge silently. Good.
- But the conflict lands on a human, and nothing checked the resolution. Taking upstream's side on that line leaves a file that still carries the Tron bypass while claiming to be plain `2.2.0` — and it passes CI, because the audit check finds upstream's audit for `2.2.0`, which covered the contract *without* our delta.

The section documents what the guard enforces, all nine violation codes, and the runbook for resolving a firing — including the graded audit call (internal SC-core review when upstream did not touch the overlaid functions, full external audit when it did) and why pointing the rebased version at the previous overlay's audit ID is rejected.

**Two accuracy fixes.** The delta inventory said "CI (3)"; it is five files plus the guard's scripts. And the App-bypass description asserted the bypass is in place — softened to match the requirement the workflow itself states, since that is one-time manual setup.

Prose only; no code, config or deploy logs touched.

- [ ] I have added tests that cover the functionality / test the bug — n/a, documentation only
- [ ] For new facets: n/a

- [x] I have checked that any arbitrary calls to external contracts are validated and or restricted — n/a, docs only
- [x] I have checked that any privileged calls (i.e. storage modifications) are validated and or restricted — n/a
- [x] I have ensured that any new contracts have had AT A MINIMUM 1 preliminary audit conducted — n/a, no contracts

### DIFF
--- a/docs/TronFork.md
+++ b/docs/TronFork.md
@@ -66,7 +66,7 @@ of the fork rather than an unanswerable question in `main`.
 ### What actually differs in the fork (the delta)
 
 The fork stays close to upstream, but the delta is more than one line:
-**~17 files across 6 categories** (verified by a clean upstream→fork merge
+**~20 files across 7 categories** (verified by a clean upstream→fork merge
 of 69 commits with **zero conflicts**, then diffing the fully-synced tree
 against `main`). Everything here is Tron-enablement, CI, test, config or
 audit — **no product features**.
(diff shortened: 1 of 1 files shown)


0xDEnYO:
### Review disposition (Bugbot)

**Fixed**
- Local `checkTronForkDelta` runbook now says to run from a `contracts-tron` checkout
- Part 2 sync wording aligned with direct-push / exception-PR model
- Delta inventory header updated to ~20 files / 7 categories (Scripts counted as 3)

**Declined**
- Audit carve-out “skips the whole sync PR”: incorrect — `versionControlAndAuditCheck.yml` skips commit-association only for audit IDs that already exist upstream (per-entry), and fork-authored entries on the same sync PR are still verified. Doc wording matches the code.

melianessa (approved):

PR #2140, opened 2026-07-29
chore(deploy): record CalldataVerificationFacet v2.0.0 on zksync (EXSC-680)

0xDEnYO:

[EXSC-680](https://linear.app/lifi-linear/issue/EXSC-680)

Follow-up to #2120, which deployed this facet and explicitly deferred both records
it left behind — its own description notes "the tracked master log does not yet
contain this entry" and that the facet was "NOT yet registered in the diamond — that
needs a governance diamondCut, proposed separately". That diamondCut has now
executed; this PR closes out the deferred bookkeeping.

`CalldataVerificationFacet` v2.0.0 was deployed to zksync at
`0x2E09CeC715CB767F15c46aF8c33ff780eaAbFa74` — the last production network that was
missing this core facet — and the registering `diamondCut` has now executed through
the timelock. Verified on-chain before touching any records: all 8 selectors
(`7f99d7af 103c5200 c318eeda ee0aa320 df1c3a5b 070e81f1 d53482cf f58ae2ce`) resolve
to that address, and `facetFunctionSelectors` returns exactly those 8.

`deployments/zksync.json` already carried the address (committed in #2120). Two
records were still behind, because the deploy wrote to MongoDB only and
`update-deployment-logs.ts` syncs JSON → MongoDB, never the reverse:

1. **`deployments/_deployments_log_file.json`** had no
   `CalldataVerificationFacet.zksync.production` entry. This is the git-tracked
   historical record of what was deployed where, and the JSON side of the comparison
   `validate-deployment-logs.ts` performs. (Worth stating explicitly because the name
   invites the opposite conclusion: `findContractInMasterLog`
   (`script/helperFunctions.sh:237`) — the CREATE2 collision guard used by
   `deploySingleContract.sh` — does *not* read this file. It hard-fails unless
   `MONGODB_URI` is set and queries MongoDB. So the stale JSON never weakened that
   guard; the cost of the gap was purely to the in-repo record.)
2. **`deployments/zksync.diamond.json`** had no facet entry at all.
   `checkDeploymentAddressConsistency.loadSources` only indexes diamond facets that
   have a `Name`, so both an absent entry and an empty-named one make drift
   invisible to the consistency checker — the failure mode that hid an avalanche
   `DiamondCutFacet` mismatch for ~3 years.

**Provenance of the values.** Every field is transcribed from the canonical MongoDB
record (`_id 6a671a4b…`, `gitCommitHash 6e231e8a…`), applying the same mapping
`DeploymentLogger.updateLocalJsonFile` uses
(`script/deploy/shared/deployment-logger.ts:332-343`). The diamond-log entry was
written by the repo's own `script/deploy/updateDiamondLog.ts`.

I deliberately did **not** reach for `logDeployment({ updateLocalJson: true })`, the
one code path that writes this file: it stamps `gitCommitHash` with the *current*
HEAD, which would have overwritten both the JSON record and the MongoDB row with a
commit that has nothing to do with the deployment. There is no sanctioned
MongoDB → JSON CLI; that gap is noted below.

`OPTIMIZER_RUNS 1000000` / `SOLC_VERSION 0.8.29` / `ZK_SOLC_VERSION 1.5.15` were
cross-checked against `foundry.toml` (`[profile.default].optimizer_runs`,
`[profile.zksync].solc_version`, `[external.zksync].zksolc`), and
`EVM_VERSION zkevm` matches the 9 other zksync records carrying the current
toolchain. Diff is insertions-only; both files keep their existing
`JSON.stringify(…, 2)` + trailing-newline formatting.

## Verification

- `bunx tsx script/tasks/checkDeploymentAddressConsistency.ts` → **passed**
- `bunx tsx script/deploy/healthCheck.ts --network zksync --environment production`
  → `CalldataVerificationFacet deployed` + `registered in Diamond`. The previous
  `CalldataVerificationFacet not registered` error is gone.
- On-chain facet set now diffs clean against the diamond log (25 on-chain, 25 logged,
  no entries on either side only).

## Out of scope — flagged, not decided

- **`CBridgeFacet` on zksync.** healthCheck still errors with `CBridgeFacet not
  deployed`. The premise that it needs deploying is wrong: v1.0.0 is **already live**
  at `0x79e05dDE041E64cc13C3280B7d083F3AE976fe70` (27552 bytes of code, 3 selectors
  registered in the diamond, named in `zksync.diamond.json`). It is missing from
  `deployments/zksync.json` and from the master log, so healthCheck's "deployed"
  check — which reads `zksync.json` — fails, and the same absence raises the
  `registered on-chain but absent from the deploy log` warning. MongoDB already has
  the record, backfilled as an orphan under EXSC-508. So this is a records-only gap
  of exactly the class this PR fixes, not a deployment decision. Left alone
  deliberately; happy to do it in a follow-up.
- **The master log is behind for 14 other networks.** Same root cause. Comparing each
  `deployments/<network>.json` against the master log, `CalldataVerificationFacet`
  addresses are unrecorded on `0g`, `arbitrumnova`, `arbitrumsepolia`, `arc`,
  `arctestnet`, `basesepolia`, `jovay`, `morph`, `optimismsepolia`, `pharos`,
  `robinhood`, `somnia`, `telos`, `tempo`. No v2.0.0 entry exists in the JSON master
  log for *any* network — the whole rollout is MongoDB-only.
- **`validate-deployment-logs.ts` is currently non-functional.** It reports
  `JSON: 0 records` and flags all 4238 MongoDB rows as missing.
  `ValidationUtils.isValidRawDeploymentData` requires `SALT`, `SOLC_VERSION`,
  `EVM_VERSION` and `ZK_SOLC_VERSION` to be present strings on every record, but
  almost every historical entry omits them, so all 3199 parsed records are dropped as
  invalid. Until that is relaxed, the validator cannot detect the drift above.


melianessa (approved):

0xDEnYO (commented):
Re-triggering `protect-critical-code` — it only runs on `pull_request_review` and had not reported on the current head commit.

PR #2145, opened 2026-07-29
chore(deploy): add SquidFacet v1.0.0 on tempo (EXSC-684)

0xDEnYO:
# Which Linear task belongs to this PR?

https://linear.app/lifi-linear/issue/EXSC-684

# Why did I implement it this way?

Closes the health-check gap surfaced by #2125: tempo's `_targetState.json` lists `SquidFacet` 1.0.0, but it was never deployed there.

- **Deployed `SquidFacet` v1.0.0 to tempo (production)**: `0x58EBB75C51da4e2d19329B571dC75d815cd59915` (CREATE3, cancun, solc 0.8.29). Sourcify-verified; Mongo master log updated with `verified: true`.
- **Fixed `config/squid.json` tempo router**: #1660 copy-pasted the default Squid router `0xce16F69375520ab01377ce7B88f5BA8C48F8D666`, which has **no code on tempo**. Squid's [contracts docs](https://docs.squidrouter.com/additional-resources/contracts) list tempo (chain 4217) under the alternate router `0x2B4d4Cf15dAD79D3426D19674Bd237C1dc9144aa` — confirmed on-chain: identical proxy bytecode and identical EIP-1967 implementation (`0x55df48aaa9306d890ec177507015a4c7c7999d13`) on tempo, gnosis, sonic, and berachain. The facet was deployed with the corrected address as `_squidRouter`.
- **diamondCut proposal created** (tempo diamond is owned by `LiFiTimelockController`): Safe nonce **20**, timelock `scheduleBatch` (min delay 10800s), safeTxHash `0x64999cf9e028f79a0d065ae22ef927096b3dcde2662899322fe6842c52e1b853`, 1/N signatures. Note: a pre-existing pending proposal occupies nonce **19** and must clear first.

**Update (2026-07-30): the timelock cut executed** — execution tx `0x6289ae8a5dd612f7f92b9419638ebf776dd982f06281aeb80d570f88196c766b`. `deployments/tempo.diamond.json` was regenerated from on-chain state (commit `36a6969df`). That sync also picks up the **`ReceiverAcrossV4` periphery registration** (`0xac6ab3D8026Bfd31eDeA055deFedF61956439d0f`) from the second timelock op of the same rollout batch (execution tx `0xf25f4dd3170cf1dba42907e85b1cc73bbc6b4b9ddc58f7c581b8683d456c94b3`): that contract was already deployed on tempo and listed in `deployments/tempo.json` — only its diamond registration was pending, so no new deployment rides in this PR.

[...]

PR #2150, opened 2026-07-30
fix(deploy): apply customDeployFlags on the zkEVM deploy path (EXSC-684)

0xDEnYO:

EXSC-684 — found by `/gate-review` while working on #2145 (SquidFacet on tempo). No dedicated
ticket could be created from this session (Linear connector unavailable here); happy to move it
under its own ticket if preferred.

`deploySingleContract.sh` has two branches for the `forge script` deploy command. The standard
branch interpolates `$ADDITIONAL_FLAGS` (a network's `customDeployFlags` from
`config/networks.json`); the zkEVM branch built its own command string and never did, so
`customDeployFlags` was **silently dropped** on zkEVM chains. No zkEVM network sets the field
today (`abstract`, `lens`, `zksync` all leave it unset), so nothing is broken in production —
this is a latent trap for the next zkEVM network that needs a custom deploy flag.

I chose to **make the field work on zkEVM** rather than warn that it is ignored: a silent-drop
bug and a warn-on-drop bug both still leave the field unusable on those chains, and a warning
only helps someone who happens to read the log.

**Why the flags are composed instead of appended verbatim.** The zkEVM command pins two flags
that `customDeployFlags` values realistically also carry — `--skip-simulation` and
`--gas-limit 50000000`. Appending verbatim would pass one of them twice (all three networks that
use `customDeployFlags` today set `--gas-limit`, and `megaeth` also sets `--skip-simulation`).
Rather than rely on `forge`'s duplicate-flag semantics — which I could not verify here, since the
`foundry-zksync` binary is not vendored into a fresh checkout — the change guarantees each flag is
emitted exactly once:

- a `--gas-limit` in `customDeployFlags` **replaces** the pinned default (with an `echoDebug` note)
- the redundant `--skip-simulation` token is stripped, since the zkEVM path always skips simulation
- every other flag is appended verbatim, matching the standard path

This also resolves the related dead branch noted in review: the `--skip-simulation` override at
L306-309 (which blanks `SKIP_SIMULATION_FLAG`) is structurally a no-op on the zkEVM path because
that branch hardcodes the flag. Net behaviour is now correct there — a network asking for
`--skip-simulation` gets it — so I left those lines alone rather than widen the diff.

**Verification.** No unit-test harness exists for this script, so I dry-ran the flag composition
by extracting the actual command line from the file and expanding it with stub values, over every
`customDeployFlags` value present in `config/networks.json` plus the empty case:

| `customDeployFlags` | `--gas-limit` count | `--skip-simulation` count | effective tail |
| --- | --- | --- | --- |
| _(unset — all zkEVM networks today)_ | 1 | 1 | `--gas-limit 50000000` |
| `--gas-limit 40000000` (tempo) | 1 | 1 | `--gas-limit 40000000` |
| `--gas-limit 50000000 --gas-price 2000000 --skip-simulation` (megaeth) | 1 | 1 | `--gas-limit 50000000 --gas-price 2000000` |
| `--with-gas-price 110000000` (robinhood) | 1 | 1 | `--gas-limit 50000000 --with-gas-price 110000000` |

The unset case — which is every zkEVM network as of today — reproduces the pre-change command
exactly, so this is a no-op for current deploys. `bash -n` and a `set -euo pipefail` source both
pass.

## Two follow-ups deliberately left out of scope

1. **The `INetwork` JSDoc for `customDeployFlags` is not on `main`.** The review note asked me to
   update it, but that field and its docstring are added by **#2145**, which is still open — so
   there is nothing to edit on this base. The sentence it lands with ("Only applied on the standard
   deploy path; the zkEVM path builds its own command and ignores this field") becomes wrong once
   this PR merges. Rather than stack on an unmerged PR or pre-declare the field and guarantee a
   conflict, I've flagged it on #2145; if that merges first I'll rebase and correct the wording
   here.
2. **`megaeth`'s `devNotes` in `config/networks.json` is inaccurate** — it claims
   `customDeployFlags` "are applied by deploySingleContract.sh, updateFacetConfig.sh, and
   diamondUpdateFacet.sh". Neither `updateFacetConfig.sh` nor `diamondUpdateFacet.sh` reads the
   field at all (`grep customDeployFlags` matches only `deploySingleContract.sh`). Left alone to
   keep this diff to one file and avoid `networks.json` CI churn.

### DIFF
--- a/script/deploy/deploySingleContract.sh
+++ b/script/deploy/deploySingleContract.sh
@@ -311,11 +311,14 @@ deploySingleContract() {
     # The zkEVM command pins --skip-simulation and --gas-limit; appending customDeployFlags
     # verbatim would pass either flag twice, so drop the redundant --skip-simulation and let a
     # custom --gas-limit replace the pinned default.
-    local ZK_ADDITIONAL_FLAGS="${ADDITIONAL_FLAGS//--skip-simulation/}"
+    local ZK_ADDITIONAL_FLAGS=""
     local ZK_GAS_LIMIT_FLAG="--gas-limit 50000000"
-    if [[ "$ZK_ADDITIONAL_FLAGS" == *"--gas-limit"* ]]; then
-      ZK_GAS_LIMIT_FLAG=""
-      echoDebug "customDeployFlags specify --gas-limit; not applying the zkEVM default of 50000000"
+    if isZkEvmNetwork "$NETWORK"; then
+      ZK_ADDITIONAL_FLAGS="${ADDITIONAL_FLAGS//--skip-simulation/}"
+      if [[ "$ZK_ADDITIONAL_FLAGS" == *"--gas-limit"* ]]; then
+        ZK_GAS_LIMIT_FLAG=""
+        echoDebug "customDeployFlags specify --gas-limit; not applying the zkEVM default of 50000000"
+      fi
     fi
 
     # Use a local effective multiplier so per-network overrides don't leak into subsequent calls.
--- a/script/deploy/deploySingleContract.sh
+++ b/script/deploy/deploySingleContract.sh
@@ -308,6 +308,16 @@ deploySingleContract() {
       SKIP_SIMULATION_FLAG=""
     fi
 
+    # The zkEVM command pins --skip-simulation and --gas-limit; appending customDeployFlags
+    # verbatim would pass either flag twice, so drop the redundant --skip-simulation and let a
+    # custom --gas-limit replace the pinned default.
+    local ZK_ADDITIONAL_FLAGS="${ADDITIONAL_FLAGS//--skip-simulation/}"
+    local ZK_GAS_LIMIT_FLAG="--gas-limit 50000000"
+    if [[ "$ZK_ADDITIONAL_FLAGS" == *"--gas-limit"* ]]; then
+      ZK_GAS_LIMIT_FLAG=""
+      echoDebug "customDeployFlags specify --gas-limit; not applying the zkEVM default of 50000000"
+    fi
+
     # Use a local effective multiplier so per-network overrides don't leak into subsequent calls.
     local NETWORK_GAS_MULTIPLIER
     NETWORK_GAS_MULTIPLIER=$(jq -r --arg NETWORK "$NETWORK" '.[$NETWORK].gasEstimateMultiplier // empty' "$NETWORKS_JSON" 2>/dev/null || true)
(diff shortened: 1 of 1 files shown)


0xDEnYO:
## `/gate-review` — residual findings (escalated, not auto-fixed)

**Auto-fixed and pushed** (3498deb): the `echoDebug` about replacing the pinned zkEVM
`--gas-limit` also fired on non-zkEVM networks — both networks that set `--gas-limit` in
`customDeployFlags` today (`tempo`, `megaeth`) are standard-path, where "not applying the zkEVM
default" is simply false. The flag composition is now scoped behind `isZkEvmNetwork`, so the ZK
vars and the debug line are only computed on the path that consumes them.

The three items below were **deliberately not fixed** — each needs a scope decision rather than a
mechanical change.

### 1. The `INetwork` JSDoc this PR was asked to update does not exist on `main`

`customDeployFlags` and its docstring are introduced by #2145, which is still open. The sentence
it lands with — *"Only applied on the standard deploy path; the zkEVM path builds its own command
and ignores this field"* — becomes **incorrect** once this PR merges. There is nothing to edit on
this base, and pre-declaring the field here would guarantee a conflict with #2145.

**Needs a decision:** merge this independently and correct the wording in #2145 (or in a rebase
here once #2145 lands), or restack this PR onto `claude/deploy-squidfacet-tempo` so both land
together. Flagging on #2145 as well.

### 2. `megaeth`'s `devNotes` in `config/networks.json` is factually wrong

It claims `customDeployFlags` "are applied by deploySingleContract.sh, updateFacetConfig.sh, and
diamondUpdateFacet.sh". Neither `updateFacetConfig.sh` nor `diamondUpdateFacet.sh` reads the field
at all — `grep customDeployFlags` matches only `deploySingleContract.sh`. Left out to keep this
diff to one file and avoid `networks.json` CI churn.

### 3. The zkEVM branch drops `gasPrice` too — same latent-drop family

The zkEVM command omits `$ETH_GAS_PRICE_ENV`, which is how a network's `gasPrice` from
`networks.json` reaches the broadcast (the `--gas-price` flag only affects simulation). It is built
as an inline prefix and never exported, so it genuinely cannot reach the zkEVM `forge` invocation.
No zkEVM network sets `gasPrice` today, so this is latent — exactly the same shape as the
`customDeployFlags` bug this PR fixes, but a separate field. Deliberately out of scope; worth its
own change rather than widening this one into a zkEVM-parity audit.

(For completeness: `$CREATE3_FACTORY_ADDRESS` and `$DIAMOND_TYPE` are also absent from the zkEVM
command, but `CREATE3_FACTORY_ADDRESS` is `export`ed in `helperFunctions.sh`, so it is inherited
rather than dropped — not a finding.)

melianessa (approved):

PR #2157, opened 2026-07-30
feat(deploy): harden the deferred-cleanup queue — address-keyed removals, queue-aware stale-facet invariant, reconcile hardening

0xDEnYO:

Fixes [EXSC-723](https://linear.app/lifi-linear/issue/EXSC-723/parked-queue-drain-hardening-resolve-by-address-queue-aware-stale), fixes [EXSC-774](https://linear.app/lifi-linear/issue/EXSC-774/detect-facet-removals-that-never-executed-false-executed-parked-tasks) (folded in from #2201), fixes [EXSC-775](https://linear.app/lifi-linear/issue/EXSC-775/make-facet-removals-address-keyed-so-co-registered-versions-are), fixes [EXSC-780](https://linear.app/lifi-linear/issue/EXSC-780/parked-queue-deprecating-a-network-must-cancel-its-open-tasks) (folded in from #2203)

Combined hardening of the deferred diamond-cleanup queue (docs/DeferredDiamondCleanupQueue.md), unifying two independent investigations of the same subsystem. Every guard added here was falsified on real repo/on-chain data before merge — including one that caught a real production inconsistency during verification (below). TS-only; the governance objects (Safe + timelock `scheduleBatch`) are byte-for-byte unchanged.

**EXSC-723 (this branch):**

1. **Removal identity is the facet ADDRESS, not its name** (EXSC-775 — this supersedes the name→address *hint* this PR originally shipped). The deploy log holds exactly one address per name, so a name can never name a superseded version that is co-registered with its live successor — and asking to remove it by name resolves to the LIVE facet. Confirmed in production: SymbiosisFacet v1.0.0 and v2.0.0 are both cut into 35 diamonds ([EXSC-750](https://linear.app/lifi-linear/issue/EXSC-750/superseded-symbiosisfacet-still-cut-into-35-production-diamonds-legacy)); the drain's `removalByName` map would have folded the live facet's selectors into the Remove. So: `taskKey` is now `kind|network|environment|facetAddress`, `facetName` is a display label, and the drain resolves through the new `computeFacetRemovalsByAddress` / `diffFacetsByAddress`. Selectors still come from the live loupe, so nothing can go stale; the deploy log is reduced to supplying a label and the never-remove check (which now also runs by address, so an unlogged address cannot smuggle a protected facet past the allowlist). `prunedButRouted` is deleted — with an address identity there is no ambiguous case left to park. `cleanUpProdDiamond` gains `--facet-addresses` alongside `--facets`; the two are mutually exclusive and the address form is single-network by construction.

2. **Queue-aware stale-facet health-check invariant** (`no-stale-registered-facets`, warning, production, skipTestnet, EVM + Tron). Warns on any routed facet that is deprecated (source deleted AND absent from `_targetState.json` — the same source-gone gate the removal engine uses, so eval and remover can never disagree) with no open parked task. Open tasks log as expected-pending — no alarm flood from the queued backlog. Live facets missing from target state are drift, never warned here. Mongo is touched only when the cheap pre-pass finds something stale; an unreachable queue degrades to a visible coverage warning. Day-1 firing verified: tron GenericSwapFacet + DexManagerFacet (both true positives, see EXSC-724).
3. **Reconcile safe-to-prune report**: names `deployments/*.json` entries whose parked work is fully terminal (≥1 executed/superseded, none open; cancelled-only groups never qualify). Report-only — pruning stays a reviewed PR. Currently lists 56 entries. Address-keyed removals make one new case reachable, so the report also holds back any row whose logged address is still routed: a terminal task for a superseded version must not mark the LIVE facet's log row prunable. A network the sweep could not read contributes no routed set and is held back rather than reported against unknown chain state.
4. **Reconcile fault isolation**: per-(network,environment) try/catch — a retired network (dropped from `networks.json` with rows still parked) previously aborted the whole sweep; all 4 weekly cron runs had died this way, freezing queue statuses for 18 days. Also: an unreachable signing-store tunnel (`SC_MONGODB_URI` set, `lifi-connect` down) now degrades to loupe-only reconciliation instead of aborting — the proposal status is an optional signal by design.

**EXSC-774 (merged from #2201):**

5. **False-resolution detection**: terminal (`executed`/`superseded`) tasks are re-verified against the loupe; a facet still routed reopens to `queued` + alerts. Presence is resolved by the parked **address**, matching the address-keyed drain (item 1) — judging by name would strand every co-registered removal, since the successor keeps the name once the superseded version is cut. The name check that caught the worldchain regression (a task carrying lisk's address, which let a live facet be marked done for 18 days) survives as `isSuspectAddressSnapshot`: address gone + name still routed is now a loud warning next to the decision, never a silent resolution.
6. Reconcile failure/reopen Slack alerts with remedies (`formatReconcileFailureMessage`, `formatReopenAlertMessage`), `reopenResolvedTask` store transition, environment-scoped queue reads.

**Operational step, to run right AFTER this merges — tracked in [EXSC-790](https://linear.app/lifi-linear/issue/EXSC-790/run-the-parked-task-key-migration-immediately-after-pr-2157-merges)** (not before — against name-keyed `main` the window stays open for as long as this PR sits unmerged): the queue holds tasks parked under the old name-based `taskKey`. `bunx tsx script/deploy/safe/migrate-parked-task-keys.ts --apply` (idempotent, dry-run by default) rewrites them from their own stored fields. Without it a stale key stops colliding with a re-enqueue, the queue can hold two open tasks for one address, and the drain would fold two identical `Remove` calls into one batch — the second reverting on-chain. The script refuses to migrate a row whose new key would collide with another open row and reports it instead.

**Verification on live data (not just unit tests — 213+ of those, all green):** fleet dry-run exercised per-network isolation and loupe-only degrade end-to-end (exit 0); the reopen detection caught a **real false-supersession** (mode `AcrossFacetV3`, duplicate-address snapshot — reopened with `--yes`, will re-drain); worldchain verified clean on the live loupe (stale git `diamond.json` only); queue reconciled to truth (54 stale `proposed`→`executed` verified against loupe + signing store; retired-network rows cancelled).

**EXSC-780 (folded in from #2203)** as `de34c7418` + `783e26a68`: the opt-in single-network `--cancel-deprecated`, the fleet-wide refusal guard, the outer sweep guard so the TTL backstop survives a throw, `--ttlDays` validation, and the `getSafeMongoCollection` client close. #2203's own per-group isolation and `process.exitCode` were dropped in favour of item 4 — per-network Slack alerts carry strictly more information than a bare non-zero exit. The cron stays report-only; cancelling is opt-in via `--network <x> --cancel-deprecated --yes`, because `networks.json` is narrowed for pause rehearsals and is therefore not a durable deprecation signal.

**Hardening round (commits `514788a8f`…`8b937a34a`)** — a step-back redesign after review found the guards had grown as per-call-site patches with diverging semantics. The principle now: **one classifier, invariants at the layer that owns them, every alert names a remediation that works.**

7. **One removability classifier.** `diffFacetsByAddress` is the single gate the drain, the reconcile's reopen decision, and `cleanUpProdDiamond --facet-addresses` all run. It gained the target-state side the reconcile previously had privately: a requested address the deploy log names as an expected facet — or that routes a selector an expected facet owns (the same held-back rule as `diffFacets`, now applied to logged AND unlogged addresses) — is refused as `stillExpected`, never removed, and `unverifiable` fails closed when target state or the selector unions are unavailable. The reconcile's former name-first gate (which could judge a wrong-address task "removable" without ever looking at what the address actually was) is deleted; reopen now means literally "the removal engine would remove this address, under the same label".
8. **No contradiction is ever removed or resolved.** The drain refuses (queued + alert): an engine-resolved deploy-log name that disagrees with the parked label, a still-expected address, an unverifiable one, and a legacy row whose stored address is not a valid EVM address (previously one such row aborted the whole network's drain). The reconcile applies the identical gates, so eval and remover cannot ping-pong.
9. **Dedup survives the key migration without depending on it.** The drain's duplicate-address guard is seeded with open `proposed` rows and records lost claims; `reopenResolvedTask` recomputes legacy keys and pre-checks open rows by address. Two open tasks for one address can no longer fold duplicate `Remove`s into proposals in any interleaving, whatever their key format — `migrate-parked-task-keys.ts` is now cosmetic, not load-bearing.
10. **EVM-only enqueue.** `enqueueParkedTask` refuses any non-`0x` / invalid address: no consumer (EVM drain, viem reconcile) can process one, so admitting them only minted permanently undrainable rows.
11. **The cron can actually evaluate its gate.** `reconcileParkedTasks.yml` checks out submodules and runs `forge build` — without artifacts the reopen gate returned `undefined` for every unlogged address and EXSC-774 detection never fired on its primary vehicle; the anomaly message also misblamed a "missing target-state entry". Alerts now distinguish "engine refused (reason)" from "could not verify (run forge build)".
12. **Operator loop closed** (`cancel-parked-task.ts`): every refusal alert now names a queued-only cancel CLI that actually works (including under multi-row keys), instead of pointing at a migration that refuses collisions. Plus: dry-run reopen output no longer claims tasks were re-queued, terminal tasks suspended from re-verification by a narrowed `networks.json` are named in the Slack failure alert (scoped to recently resolved rows so deprecated-network history doesn't alert forever), `cleanUpProdDiamond` no longer prints "none of the given addresses are registered here" when they were refused, and the health check reads the queue once per process instead of once per stale network.

This round was itself adversarially reviewed (two independent reviewers: hostile-correctness + does-it-fix-it); all 4 major findings from that review are fixed in `8b937a34a`.

Deliberately NOT closed by this PR: [EXSC-790](https://linear.app/lifi-linear/issue/EXSC-790/run-the-parked-task-key-migration-immediately-after-pr-2157-merges) (runs after merge), EXSC-724 and EXSC-750 (on-chain removals; this PR is TS-only).


script/deploy/healthCheckInvariants.ts
```diff
@@ -1634,6 +1679,97 @@ export const HEALTH_CHECK_INVARIANTS: IHealthCheckInvariant[] = [
...
+            })),
+            ...(await listParkedTasks(parkedTasks, {
+              network: ctx.networkLower,
+              status: 'proposed',
+            })),
+          ]
+          openParkedFacetNames = new Set(open.map((t) => t.facetName))
+        } finally {
+          await client.close()
+        }
```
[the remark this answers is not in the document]
0xDEnYO: Fixed in 983aad5ca — `IListParkedTasksFilter`/`listParkedTasks` now accept an optional `environment` filter (applied as `$eq` in the Mongo query), and both consumers pass it: the stale-facet invariant's queued/proposed reads use `EnvironmentEnum.production` (the invariant is production-scoped), and the drain's `listQueued` passes its existing `environment` parameter so cross-environment rows can't be resolved against the production loupe. Covered by a new store test (a staging row never leaks into production reads). Theoretical in v1 — `enqueue-parked-task.ts` hard-rejects non-production — but now enforced at the query.

script/deploy/healthCheckInvariants.ts
```diff
@@ -1634,6 +1679,97 @@ export const HEALTH_CHECK_INVARIANTS: IHealthCheckInvariant[] = [
...
+        return
+      }
+      const targetFacets =
+        ctx.targetState[ctx.networkLower]?.production?.LiFiDiamond
+      if (!targetFacets) {
+        consola.info(
+          'No production target state for this network; skipping stale-facet check'
+        )
+        return
+      }
```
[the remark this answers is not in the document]
0xDEnYO: Not applied — this exact condition is already recorded as a warning once per network by the runner: `runHealthCheckForNetwork` sets `missingProductionTargetState` when `targetState[networkLower]?.production?.LiFiDiamond` is absent and calls `ctx.logWarn('Network … has no production target state; non-core facet coverage is reduced')` (script/deploy/healthCheck.ts, ctx build), which lands in `IHealthCheckNetworkResult.warnings` and the aggregated fleet report. Warning again inside this invariant would double-count the same condition in the same report; the `consola.info` here is only the local skip note for this one check. The learning's requirement (recorded, non-fatal) is satisfied at the runner level.

0xDEnYO:
## Review-gate report (ran inline — see note)

**Auto-fixed during the gate (falsification findings, commit `fix(healthcheck): … source-gone gate + skipTestnet`):**
- `no-stale-registered-facets` originally used `_targetState.json` alone as the expected set — on real repo data it would have warned on **live** facets the target state merely hasn't recorded (LayerSwapFacet, AcrossFacet, GardenFacet, LiFiIntentEscrowFacet, AcrossV4SwapFacet). Fixed with the same source-gone gate the removal engine uses (`getSourceContractNames`): stale = deprecated (source deleted), never target-state drift.
- Added `skipTestnet: true` — the parked queue is a production-mainnet construct; testnet warnings could never resolve.

**Falsification evidence (executed on real repo data):** hint resolution proven on real arbitrum log/diamond data (log-present unchanged; pruned+unambiguous → removal by address with loupe selectors; ambiguous → refused); invariant proven to fire on real tron data (GenericSwapFacet unparked — the EXSC-724 case) and to stay quiet on live-but-unrecorded facets; `computeSafeToPrune` proven against the real `getDeployments`. Day-1 fleet estimate: **3 warnings on 2 mainnets** (tron ×2, worldchain ×1) — all true positives.

**Escalated / operational findings (need human decision, not code changes):**
1. **tron `DexManagerFacet`**: deprecated (source removed), still registered on the tron diamond, no parked task — same class as EXSC-724's GenericSwapFacet; should ride the same Tron removal.
2. **worldchain `AcrossFacetV3` residue**: the parked task executed for its snapshotted address `0xB5dD…f92`, but the diamond log still lists a second address `0x08F7…1C8` under the same name with no open task. Real duplicate-address residue — needs investigation (is `0x08F7…` still routed on-chain?) and either a new parked task or log cleanup.

**Coverage note:** the gate's 6-agent fan-out (incl. git-history and prior-PR-comment passes) was killed by the org monthly spend limit; the review ran inline instead (rules adherence, collaborator/bug pass, comment-drift pass, executable falsification). The history/prior-comments dimensions got reduced coverage.

0xDEnYO:
## Incident found & fixed while verifying queue data (pushed as `fix(deploy): fault-isolate reconcile per network`)

**Symptom:** the queue reported 54 tasks `proposed` when on-chain nearly all had executed.

**Root cause:** the weekly reconcile cron (`reconcileParkedTasks.yml`) failed on **all 4 runs since it shipped**. `reconcileAll` iterated network groups with no per-network fault isolation; tasks parked for networks that were later retired from `config/networks.json` (harmony, okx, velas) made `getViemChainForNetworkName` throw, killing the entire sweep — so no status ever flipped, and the TTL alert + safe-to-prune report never ran either.

**Fix in this PR:** per-network try/catch (one dead network can no longer freeze the fleet), an explicit retired-network branch (warn + suggest `markCancelled`, skip), and a partial-failure summary that exits non-zero only AFTER the TTL alert and prune report have run for the healthy networks.

**Data remediation applied (verified line-by-line against each network's loupe + the signing store):** all 54 stale `proposed` → `executed` (facet gone on-chain AND linked proposal `executed`); mode `AcrossFacetV3` → `superseded`; harmony/okx/velas `GenericSwapFacet` → `cancelled` (networks retired — can never drain). Queue now: 67 queued / 0 proposed / 55 executed / 1 superseded / 4 cancelled. The new safe-to-prune report immediately listed **56 deploy-log entries** ready for a cleanup PR.

0xDEnYO:
Coordination note: draft PR #2201 (EXSC-774, parallel session) also fault-isolates `reconcileAll` per network and additionally detects false-`executed` tasks (never-routed address snapshots — the worldchain AcrossFacetV3 case). Overlapping files: `reconcile-parked-tasks.ts`, `healthCheckInvariants.ts`. Suggested order: land this PR first (ready, CI green); #2201 then rebases and drops its duplicate isolation hunk, keeping its false-executed detection which this PR does not cover.

0xDEnYO:
## EXSC-780 folded in — `0e959394e..783e26a68`

Handing off from the EXSC-780 session (the session that pushed `0e959394e` has ended, so this goes here rather than as a DM). #2203 is closed as folded-in, same as #2201. This PR now covers **EXSC-723 + EXSC-774 + EXSC-780**.

**`de34c7418` — the merge.** Resolution takes this branch's reconcile as the base: `IReconcileRun {reopened, failures}`, `resolveFacetPresence`, the safe-to-prune report, and the loupe-only degrade. EXSC-780's duplicate per-network isolation and its `process.exitCode` model were dropped in favour of them — per-network Slack alerts carry strictly more than a bare non-zero exit — and the workflow header plus the §8 doc paragraph that still claimed the exit code were corrected to match. What carried over: tasks on a network outside the active set are routed out before the loupe is touched; cancelling one is opt-in and single-network (`--network X --cancel-deprecated --yes`, fleet-wide refused); an outer guard around `reconcileAll` so a throw before it returns can't take the alerts and the TTL backstop with it; `--ttlDays` validation; and `getSafeMongoCollection` closing its client when the index creation fails after connecting.

**`783e26a68` — findings from gating the merge resolution.** The one worth reading: the retired-network alert recommended `--cancel-deprecated --yes` on *every line*, and under a pause-rehearsal narrowing of `config/networks.json` that alert points at 66 perfectly healthy networks. The irreversible bulk cancellation the opt-in guard exists to prevent had re-entered through the alert text. The remedy is now stated once in the footer with an explicit *CHECK FIRST — a narrowed networks.json looks identical to a deprecation here, and cancelling is irreversible*. Also fixed: the alert counted tasks while saying "network(s)" (66 vs the real 41) and rendered one line per task, so a narrowing posted 17,249 chars — now grouped per (network, environment) and capped, worst case 2,474 against `SlackNotifier`'s 2900 budget; `shouldCancelDeprecated` used `networkFilter !== undefined`, which `''` passes (citty yields `''` for a bare `--network`, which `listParkedTasks` treats as no filter at all) — both predicates are truthiness now, with a regression test; and `markCancelled`'s `null` return was discarded, so a task a concurrent drain flipped `queued`→`proposed` would still have logged as cancelled.

One deliberate behaviour note for the PR body: terminal tasks on a non-active network are reported in a single summary line rather than per task — during a narrowing that is 55 tasks temporarily outside the EXSC-774 false-resolution re-check, visible rather than silent.

**Verification on the merged system.** 655 tests pass; the reconcile suite is the exact union of both sides (49→50 tests, zero dropped from either parent — verified by diffing test names against both merge parents). eslint / tsc / prettier clean. Live fleet dry-run exit 0 with 123 decisions and the loupe-only degrade exercised. Fleet-wide `--cancel-deprecated --yes` refused (exit 1); `--ttlDays soon` refused (exit 1); the pause-rehearsal replay with the cron's own flags routed 66 tasks to the deprecated path and cancelled **0**. Parked-queue delta across all of it: none — the only queue change today is the `mode` `AcrossFacetV3` `superseded`→`queued` reopen made by the EXSC-774 work.

Still to do before review: extend the PR body to cover the EXSC-780 section.

0xDEnYO:
## Review gate — residual findings (EXSC-775 commits `375920e4c..66e53307c`)

Auto-fixable findings were fixed and pushed. Two items are escalated for a human decision.

### 1. Drain append order vs execute-side zip tie-break — `script/deploy/safe/parked-tasks.ts:408`

`listParkedTasksBySafeTxHash` sorts `{ proposedAt: 1, taskKey: 1 }`, and `buildRemovalSnapshotFromPayloads` zips the trailing Remove cuts against that order. The drain appends in `listQueued()` order (Mongo natural), relying on `proposedAt` — set sequentially at claim — to agree. Only **same-millisecond** claims fall through to the `taskKey` tie-break.

This PR changes what that tie-break sorts on (`facetName` → `facetAddress`), so a same-ms pair can now order differently than before. The failure mode is unchanged and fail-safe (mis-labelled cut → `re-pointed` abort, never a silent wrong delete), and it was already a documented accepted risk.

A one-line deterministic `.sort({ taskKey: 1 })` on the drain's `listQueued` would make the two orders agree by construction. **Not applied** — it changes the behaviour of an existing path, which is outside what this gate auto-fixes.

### 2. Operational: the 127-row key migration must run before this merges

Live dry-run against the queue: **127 rows to migrate, 0 already current, 0 collisions.**

```bash
bunx tsx script/deploy/safe/migrate-parked-task-keys.ts --apply
```

Idempotent and dry-run by default. Until it runs, a stale name-based key no longer collides with a re-enqueue of the same facet, so the queue can hold two open tasks for one address — and the drain would fold two identical `Remove` calls into a single batch, the second reverting on-chain. Timing (before merge vs. immediately after) is a human call.

### Falsification evidence (live mainnet diamond, not fixtures)

| check | result |
|---|---|
| stale SymbiosisFacet v1.0.0 `0x23Fc1b73…` routed on mainnet | **true**, and absent from the deploy log |
| deploy log maps `SymbiosisFacet` → | `0xa0353221…` (the **live** v2.0.0) |
| name path (`diffNamedFacets`) removes | `SymbiosisFacet@0xa0353221…` — **the live facet**, i.e. the bug |
| address path (`diffFacetsByAddress`) removes | `0x23Fc1b73…`, selectors `0xb70fb9a5` / `0x6e067161` |
| protected-by-address guard, real `DiamondCutFacet` address | refused, 0 removals |

The address path's selectors match the two legacy Symbiosis entrypoints documented in EXSC-750 exactly.

0xDEnYO:
## Gate follow-up — item 1 fixed, item 2 guidance corrected (`028795b8a`)

### 1. Drain/execute ordering — **fixed**

`listParkedTasks` now reads `.sort({ taskKey: 1 })`, so the drain claims (and appends Remove cuts) in exactly the order `listParkedTasksBySafeTxHash` replays them. The two agree by construction instead of by timing, and the same-millisecond `proposedAt` tie no longer depends on Mongo's natural order.

Covered by a test that was verified against a negative control — with the sort removed it fails, with it restored it passes — so it is not asserting an order the fixture already happened to have.

### 2. Migration timing — **run it AFTER merge**, not before

Correcting my earlier comment, which said "before this merges". That is the wrong way round.

Either ordering opens the same window: rows keyed one way while the running code computes the other, so a re-enqueue does not collide, the queue can hold two open tasks for one address, and the drain folds two identical `Remove` calls into a batch (the second reverting). Running the migration **before** merge opens that window against `main`, which is still name-keyed, and holds it open for as long as this PR sits unmerged. Running it **after** merge makes the window as short as you choose.

So: merge, then run it promptly — before the next enqueue or drain.

```bash
bunx tsx script/deploy/safe/migrate-parked-task-keys.ts --apply
```

Idempotent, dry-run by default. Live dry-run: 127 to migrate, 0 already current, 0 collisions.

0xDEnYO:
## Pushed: both reconcile alert sites now gated + routed (47d2480f2)

`runReconcileAlerts` posted unconditionally to `WEBHOOK_DEV_SC_MULTISIG_PROPOSALS`, so a local rehearsal against a narrowed `networks.json` paged the signing worklist — that is what produced the five bot posts in `#dev-sc-multisig-proposals` on 2026-08-18 (`⚠️ 66 network(s) could not be reconciled`, `⚠️ 41 network(s)…`). Both send sites in this file now go to `WEBHOOK_DEV_SC_GITHUB_CI_NOTIFICATIONS` and require `isUnattendedRun()`.

This branch now contains #2208 (the routing + gate change). **Merge #2208 first**, or merge this one and #2208 becomes a near-no-op. The two conflicted only on the `import` line — which is exactly why the ungated sends were invisible: resolving that conflict gives no signal that two send sites underneath it were left on the old channel.

Reproduction of the incident shape against the merged code, counting real webhook POSTs (not branch shape):

```text
local rehearsal → webhook POSTs: 0 PASS
CI cron        → webhook POSTs: 1 PASS
```

## This PR is also the fix for the reconcile cron, which has never once succeeded

`reconcileParkedTasks.yml` has failed 4/4 runs (Jul 27, Aug 3, Aug 10, Aug 17) — every run since it was added. Root cause is not RPC flakiness: `reconcileAll` on `main` calls `getSafeMongoCollection` (`safe-utils.ts:1470`) and **throws** when the signing store is unreachable. GitHub Actions has no `lifi-connect` tunnel, so it can never reach it — while the workflow header claims the job "runs in loupe-only mode" there. That degradation was documented but not implemented until `0e959394e` on this branch.

Verified by running the same command with no tunnel in both trees:

```text
main            → ERROR Could not reach the Safe MongoDB via lifi-connect … exit 1
this branch     → WARN  Signing store unreachable — reconciling loupe-only … exit 0
```

So the TTL/cold-network backstop has been dead since it shipped, and merging this PR revives it.

Worth noting separately: the workflow's `Notify Slack on failure` step fired correctly every week into `#dev-sc-general` and went unnoticed for a month — an argument for someone actually watching the new `#dev-sc-github-ci-notifications` channel, or for repeated identical failures escalating rather than repeating.

0xDEnYO:
## Merge order: #2208 goes first

This PR carries the two reconcile alert send sites that spammed `#dev-sc-multisig-proposals` on 2026-08-18 (`⚠️ N network(s) could not be reconciled`, `🚨 … STILL ROUTED — re-queued`) — both on `WEBHOOK_DEV_SC_MULTISIG_PROPOSALS` and both ungated.

[#2208](https://github.com/lifinance/contracts/pull/2208) moves scheduled-job output to `#dev-sc-github-ci-notifications` and adds the `isUnattendedRun()` delivery gate. The two branches conflict only on the `import` line, so **the merge itself gives no signal about the ungated sends** — git will resolve it silently and the incident reproduces verbatim on `main`.

So: land #2208 first, then rebase this branch onto it and route both new send sites through the same webhook + gate. Doing it in the other order means fixing the same routing decision twice, on a branch where it is no longer the topic.

script/deploy/safe/reconcile-parked-tasks.ts
```diff
@@ -219,47 +309,96 @@ async function reconcileAll(
...
-        else if (decision === 'superseded')
-          await markSuperseded(parkedTasks, task.taskKey)
-        else if (decision === 'revert')
-          await revertToQueued(parkedTasks, task.taskKey)
       }
     }
   } finally {
     await safeMongoClient?.close()
   }
+  return failed
```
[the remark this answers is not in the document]
0xDEnYO: Moot as of the current head: the `failed: string[]` accumulator this flagged no longer exists. `collectFailedNetworks` was refactored during the queue-ordering work and the routing/dedup state is now Set-backed throughout (`routedNames`, `routedAddresses`, `routedByNetworkEnv: Map<string, Set<string>>`), so a network cannot be counted twice across environment groups. Resolving.

script/deploy/safe/drain-parked-tasks.ts
```diff
@@ -187,33 +191,30 @@ export async function prepareDrainNetwork(
           continue
         }
         claimed.push({ task, removal })
-      } else if (notFound.has(name)) {
+      } else if (notFound.has(address)) {
         await deps.supersede(task.taskKey)
```
gvladika, line 300: **[medium] Drain supersedes on address-absence with no suspect-snapshot guard.**

This branch marks a queued task `superseded` purely because its stored `facetAddress` is no longer routed by the loupe. But an address can be absent for two reasons: the facet really was removed, **or** the stored snapshot was wrong to begin with (the documented worldchain/lisk contamination — a task carrying another network's address). In the second case the address was never routed here, so it resolves to `notFoundOnChain`, the task is silently marked `superseded`, and the *actual* deprecated facet stays routed with nothing tracking its removal.

The reconcile path guards exactly this class with `isSuspectAddressSnapshot` (address gone + a facet of the same **name** still routed → alert, don't resolve). The drain has no equivalent, so the same wrong snapshot that reconcile would flag, the drain accepts as done. Since `prunedButRouted` was removed in this PR, there is no longer any drain-side check at all.

**Suggested fix:** give the drain parity with reconcile. Surface the routed name set from the engine (`IAddressRemovalResult.routedNames`, cheap — `diffFacetsByAddress` already has `onChainFacets` + `addressToName`), then here: if `result.routedNames.has(name)`, alert and leave the task **queued** instead of superseding; only supersede when the name is gone too. Worth a unit test: address absent + name still routed ⇒ not superseded (still queued) + alerted.

script/deploy/safe/reconcile-parked-tasks.ts
```diff
@@ -178,77 +572,222 @@ async function resolveProposalStatus(
...
+              network,
+              facet: task.facetName,
+              prUrl: task.prUrl,
+              from: task.status,
+            })
+          continue
+        }
+        if (!apply) continue
         if (decision === 'executed')
           await markExecuted(parkedTasks, task.taskKey)
```
gvladika: **[medium] Per-task write failures can still abort the whole sweep.**

The per-group `try/catch` above only wraps the loupe **read** (it closes before the task loop begins). These per-task transition writes — `reopenResolvedTask`, `markExecuted`, `markSuperseded`, `revertToQueued` — run outside any guard. `reopenResolvedTask` re-throws everything except `E11000`, and the others propagate all errors, so one transient Mongo write failure unwinds past every remaining `(network, environment)` group to the outer guard in `run()`, which records a single generic "reconcile sweep aborted" entry.

That loses all per-network detail and any reopen alerts gathered so far, and leaves the rest of the queue unverified — the exact abort-the-sweep failure mode EXSC-780 fixed on the read path, reintroduced on the write path.

**Suggested fix:** wrap the per-task write block in a `try/catch` that pushes an `IReconcileFailure` (or logs + collects) and `continue`s to the next task, so a single bad write can't take down the networks ordered after it.

script/deploy/healthCheckInvariants.ts
```diff
@@ -1770,6 +1846,115 @@ export const HEALTH_CHECK_INVARIANTS: IHealthCheckInvariant[] = [
...
+              environment: EnvironmentEnum.production,
+              status: 'queued',
+            })),
+            ...(await listParkedTasks(parkedTasks, {
+              network: ctx.networkLower,
+              environment: EnvironmentEnum.production,
+              status: 'proposed',
+            })),
+          ]
+          openParkedFacetNames = new Set(open.map((t) => t.facetName))
```
gvladika: **[medium] `no-stale-registered-facets` matches parked-task coverage by name, not address.**

Coverage is computed as `openParkedFacetNames = new Set(open.map(t => t.facetName))` and then `deprecated.filter(f => openParkedFacetNames.has(f.name))`. A parked task with the correct `facetName` but a wrong/duplicate `facetAddress` (the same suspect-snapshot case reconcile explicitly warns about) makes this invariant report the stale live facet as "covered" and suppress the warning — even though the address-keyed drain will never actually remove that facet.

This reverts to name-as-identity, which is the premise the rest of the PR argues against (a name maps to exactly one address in the deploy log; two versions co-register under one name). The backstop meant to catch un-removed facets can be silenced by the very address mismatch it should surface.

**Suggested fix:** key the coverage check on `facetAddress` to stay consistent with the drain and reconcile, so a task whose address doesn't match the stale on-chain facet doesn't count as covering it.

script/deploy/safe/diamondRemovalDiff.ts
```diff
@@ -660,7 +732,50 @@ export async function computeNamedFacetRemovals(
...
+): Promise<IAddressRemovalResult> {
+  const resolved: IRemovalDiffIO = { ...defaultIO, ...io }
+
+  const diamondAddress = await resolved.getDiamondAddress(network, environment)
+  if (!diamondAddress)
+    return {
+      network,
+      environment,
+      removals: [],
+      notFoundOnChain: addresses,
```
melianessa: **The no-diamond early return mass-supersedes the entire queue.**

When `resolveDiamondAddress` returns undefined, every requested address is returned in `notFoundOnChain`. The drain reads that set as "already absent on-chain":

```ts
} else if (notFound.has(address)) {
  await deps.supersede(task.taskKey)
  outcome.superseded.push(name)
```

So a `deployments/<network>.json` whose `LiFiDiamond` key is missing or renamed — a mid-migration deploy log, or a partially written file — silently marks **every queued task** as `superseded` while every facet is still routed, without a single chain read. The log line even claims "already absent on-chain".

To be fair on provenance: this shape is inherited from the base (`notFoundOnChain: names`, `0ba27aa1:639`), not introduced here. But the address-keyed rewrite carried it forward unguarded, and the drain's own comment further down acknowledges the early return exists ("`computeFacetRemovalsByAddress` only omits `diamondAddress` on the no-diamond early return"). Since this PR is the hardening pass, it seems like the right place to distinguish "the chain says absent" from "we could not ask the chain" — e.g. an explicit `diamondUnresolved` flag that makes the drain bail rather than resolve.

script/deploy/safe/diamondRemovalDiff.ts
```diff
@@ -600,18 +600,89 @@ export function diffNamedFacets(params: {
...
+  )
+
+  const foundOnChain = new Set<string>()
+  for (const facet of onChainFacets) {
+    const key = lower(facet.address)
+    if (!requested.has(key)) continue
+    foundOnChain.add(key)
+
+    const name = addressToName[key]
+    if ((name && protectedNames.has(name)) || protectedAddresses.has(key)) {
```
melianessa: **This protected-facet check is provably a no-op on its second clause, and unlogged addresses fall through to `removals`.**

`protectedAddresses` is derived from `addressToName` itself:

```ts
const protectedAddresses = new Set(
  Object.entries(addressToName)
    .filter(([, name]) => protectedNames.has(name))
    .map(([address]) => lower(address))
)
```

So `protectedAddresses.has(key)` can only be true when `addressToName[key]` already resolved to a protected name — i.e. when `(name && protectedNames.has(name))` is already true. The second clause never adds coverage.

The consequence is the case the docstring claims to cover: an address absent from the deploy log gets `name === undefined`, skips both clauses, and lands in `removals` with `name: undefined`. `runAddressRemoval` then labels it `unnamed facet 0x…` and proposes it, so `cleanUpProdDiamond --facet-addresses '["0x…DiamondCutFacet"]'` with an unlogged protected facet builds and proposes the Remove instead of refusing.

Two things in the file confirm this wasn't the intent:
- the `protectedSkipped` push carries a `'unknown (address not in deploy log)'` label that is **unreachable** as written;
- the docstring states "The protected check runs on both sides … so an unlogged address cannot smuggle a never-remove facet past the allowlist."

A real second side would need protection keyed on something independent of the deploy log (e.g. matching the loupe selectors against the known `DiamondCutFacet`/`DiamondLoupeFacet`/`OwnershipFacet` selector sets), or an outright refusal to remove an address the log cannot name.

script/deploy/healthCheckInvariants.ts
```diff
@@ -1770,6 +1846,115 @@ export const HEALTH_CHECK_INVARIANTS: IHealthCheckInvariant[] = [
...
+        return
+      }
+
+      // Only stale networks touch the queue (and its mongodb dep).
+      let openParkedFacetNames: Set<string>
+      try {
+        const { getParkedTasksCollection, listParkedTasks } = await import(
+          './safe/parked-tasks'
+        )
+        const { client, parkedTasks } = await getParkedTasksCollection()
```
melianessa: **This queue read never succeeds on the fleet job — `MONGODB_URI` is not in that step's env.**

In `.github/workflows/healthCheckAllNetworks.yml` (not part of this PR, which is why it is easy to miss), `MONGODB_URI` is set only on the *Fetch RPC endpoints from MongoDB* step:

```yaml
        env:
          MONGODB_URI: ${{ secrets.MONGODB_URI }}
```

The *Run health check* step that actually invokes `healthCheckAllNetworks.ts` passes only `MAX_CONCURRENT_JOBS`, `TRONGRID_API_KEY`, `EVENT_NAME`, `BEFORE_SHA`, `AFTER_SHA`.

`getParkedTasksCollection()` calls `getEnvVar('MONGODB_URI')` (`parked-tasks.ts:291`), which throws when unset. So the catch below fires on every stale network, every run, and emits via `ctx.logWarn`:

> Parked-task queue unreachable — stale-facet coverage check skipped (N stale facet(s) unverified)

Net effect: the invariant's whole purpose — telling queued backlog apart from genuinely uncovered stale facets — never runs on the job it was written for, and instead the scheduled fleet run gains a permanent new warning on every stale network. Needs `MONGODB_URI: ${{ secrets.MONGODB_URI }}` added to the health-check step.

.github/workflows/weeklyCoverageReport.yml
```diff
@@ -190,7 +190,7 @@ jobs:
         if: ${{ !cancelled() }}
         uses: slackapi/slack-github-action@485a9d42d3a73031f12ec201c457e2162c45d02d # v2.0.0
         with:
-          webhook: ${{ secrets.SLACK_WEBHOOK_SC_GENERAL }}
+          webhook: ${{ secrets.SLACK_WEBHOOK_DEV_SC_GITHUB_CI_NOTIFICATIONS }}
```
melianessa: **`SLACK_WEBHOOK_DEV_SC_GITHUB_CI_NOTIFICATIONS` has no prior consumer in the repo, and none of the call sites guard against it being unset.**

`git grep SLACK_WEBHOOK_DEV_SC_GITHUB_CI_NOTIFICATIONS` matches only the files this PR touches; there is nothing on `main`. The six call sites are:

- `weeklyCoverageReport.yml:193` (`if: !cancelled()`)
- `ticketLinkageMetric.yml:35` and `:47`
- `updateScriptConfigReminder.yml:25`
- `reconcileParkedTasks.yml:81` and `:88`

`slackapi/slack-github-action@v2` fails the step when `webhook` resolves to an empty string, and none of these carry an `if: secrets.… != ''` guard.

If the secret is not provisioned before this merges, three currently-green scheduled workflows turn red on their next scheduled run, and the reconcile cron loses the failure notification that is its dead-man's switch. Worth confirming the secret exists at org/repo level as part of the EXSC-790 migration — and ideally guarding the sends so a missing webhook degrades to "no notification" rather than "red workflow".

script/deploy/safe/reconcile-parked-tasks.ts
```diff
@@ -53,31 +87,57 @@ const DAY_MS = 24 * 60 * 60 * 1000
...
  * @param ctx - On-chain presence + optional linked-proposal status.
  * @returns The transition to apply (`keep` = no change).
  */
 export function reconcileDecision(
   task: Pick<IParkedTask, 'status'>,
   ctx: IReconcileContext
 ): ReconcileDecision {
+  if (task.status === 'cancelled') return 'keep'
+
+  if (RESOLVED_STATUSES.includes(task.status))
```
melianessa: **The reopen path has no "should this facet still be gone?" gate, so a legitimately re-added facet re-queues its own removal.**

```ts
if (RESOLVED_STATUSES.includes(task.status))
  return ctx.facetPresentOnChain ? 'reopen' : 'keep'
```

The decision is status + presence, nothing else. Scenario:

1. `FacetV1` at `0xA` is parked, drained, marked `executed`.
2. An incident rollback re-cuts `0xA` onto the diamond and restores it in `_targetState.json`.
3. Next reconcile: status `executed` + `resolveFacetPresence` true → `reopen` → back to `queued`.
4. Next `proposeWithDrain` folds a `diamondCut` Remove for `0xA` into the timelock batch and deletes a live, target-state-expected facet.

Note the asymmetry with the removal engine: `findDeprecatedLiveFacets` gates on "source deleted **and** absent from target state" before it will call anything stale. The reopen gates on neither, and the drain's only refusal is the never-remove allowlist — which a normal bridge facet is not on.

Address reuse is not exotic here: restoring a previous deployment or re-deploying the same bytecode via CREATE2 lands on the same address, which is exactly the rollback shape. A target-state / source-still-deleted check before `reopen` would close it.

script/deploy/healthCheckInvariants.ts
melianessa: **Queue coverage is keyed on `facetName` — the identity the rest of this PR is deliberately moving away from.**

```ts
openParkedFacetNames = new Set(open.map((t) => t.facetName))
```

and then at :1930, `deprecated.filter((f) => openParkedFacetNames.has(f.name))`.

That reintroduces exactly the ambiguity `resolveFacetPresence`'s docstring spells out in this same PR — "Judging by NAME would strand every co-registered removal … once superseded SymbiosisFacet v1.0.0 is cut, v2.0.0 still routes under that name".

Concretely: SymbiosisFacet v1.0.0 (`0xA`) and v2.0.0 (`0xB`) both cut into the diamond (EXSC-750), with a parked task for `0xA` only. `openParkedFacetNames` is `{'SymbiosisFacet'}`, so once `0xB` is also deprecated-and-routed it is classified `parked`, logged as "expected-pending" via `consola.info`, never via `ctx.logWarn`, and never counted in the warning total. The uncovered facet is precisely the one the invariant exists to surface.

Should match on address, as the drain and reconcile now do: `openParkedAddresses.has(f.address.toLowerCase())` built from `t.facetAddress`.

script/deploy/safe/parked-tasks.ts
```diff
@@ -495,6 +536,52 @@ export async function setSafeTxHash(
...
+ * collides (E11000) when a fresh open task already exists for the same `taskKey`.
+ * That is a benign race — the facet is already tracked — so it returns `null`
+ * rather than throwing, mirroring {@link enqueueParkedTask}.
+ *
+ * @param parkedTasks - The queue collection.
+ * @param taskKey - The terminal task to reopen.
+ * @returns The reopened task, `null` if it was not `executed`/`superseded`, or
+ *   `null` if an open task for the same `taskKey` already exists.
+ */
+export async function reopenResolvedTask(
```
melianessa, line 596: **Reopening by `taskKey` can modify a different document than the one `reconcileAll` decided on.**

The transition matches `{taskKey, status: {$in: ['executed','superseded']}}`. Per this function's own docstring, the partial unique index covers only `{queued, proposed}` — so multiple *terminal* rows can share one `taskKey` (parked → executed → re-parked → executed).

When that happens, `reconcileAll` iterates both documents, but `findOneAndUpdate` satisfies the filter with whichever Mongo reaches first — not necessarily the one the loop is holding. Two consequences:

- `reopened.push({ facet: task.facetName, prUrl: task.prUrl, from: task.status })` reports fields from a document that was **not** the one modified, so the Slack alert names the wrong label and origin PR;
- the second iteration's reopen then collides on the now-open index (E11000 → `null`) and is dropped from the alert entirely.

Since the caller already holds the document, transitioning by `_id` would make the write target unambiguous and keep the alert fields honest.

script/deploy/safe/reconcile-parked-tasks.ts
```diff
@@ -178,77 +572,222 @@ async function resolveProposalStatus(
...
-            getAddress(task.facetAddress).toLowerCase()
-          ),
+          facetPresentOnChain: resolveFacetPresence(task, routedAddresses),
           proposalStatus,
         })
         consola.info(
           `[${network}:${environment}] ${task.facetName} (${task.status}) → ${decision}`
         )
-        if (!apply || decision === 'keep') continue
+        if (isSuspectAddressSnapshot(task, routedNames, routedAddresses))
```
melianessa: **A suspect address snapshot warns and then applies the resolution anyway — on the cron, nobody sees the warning.**

`consola.warn` here goes to the GitHub Actions job log; the code then falls through to `markSuperseded` and records the task as done. Nothing is pushed to `run.reopened` or `run.failures`, so `runReconcileAlerts` sends nothing to Slack for this case.

Replaying the worldchain incident: a task carrying lisk's `AcrossFacetV3` address → `resolveFacetPresence` correctly says "not routed" → `superseded` → one warn line buried in an unattended job log, and the still-live deprecated facet is now marked resolved with no open task tracking it.

The PR description says this case is "a loud warning next to the decision, never a silent resolution", but the only unattended channel this job has is Slack, and this path does not reach it. Options: carry these into a `run.suspect` bucket that `runReconcileAlerts` reports, or treat a suspect snapshot as a refusal (leave the task open, alert, let a human adjudicate) rather than resolving on an address the code just flagged as probably wrong.

script/deploy/safe/drain-parked-tasks.ts
0xDEnYO: Fixed in `50205d846`. The drain now has parity with the reconcile: `diffFacetsByAddress` returns `routedNames` (built from `onChainFacets` + `addressToName`, as you suggested), and on `notFoundOnChain` the drain checks it — if a facet of the same name is still routed it pushes to a new `outcome.suspectSnapshots`, alerts, and leaves the task **queued** rather than superseding.

Test: `refuses to supersede an absent address while its NAME is still routed` — verified against a negative control (guard removed → fails, restored → passes).

script/deploy/safe/reconcile-parked-tasks.ts
0xDEnYO: Fixed in `50205d846`. The per-task transition block is now wrapped: a failed write pushes an `IReconcileFailure` (`<facet>: <decision> write failed — <reason>`) and `continue`s to the next task, so neither the remaining tasks nor any later `(network, environment)` group is lost, and the Slack alert names what was skipped instead of reporting one generic abort.

`reconcileAll` is inside the unit-test-exempt live adapter, so this one has no unit test — it is covered by the live dry-run rather than a fixture.

script/deploy/healthCheckInvariants.ts
0xDEnYO: Fixed in `50205d846`, same fix as melianessa's thread on the same lines: coverage is now keyed on `facetAddress`, not `facetName`.

The split moved into an exported pure helper, `splitByParkedCoverage(deprecated, openParkedAddresses)`, so the keying is unit-tested — including the case that motivated this: two co-registered `SymbiosisFacet` versions with a task for v1 only now classifies v2 as **unparked** (it was silently "expected-pending" before). Negative control: switching the helper back to name-matching fails three of the four new tests.

script/deploy/safe/diamondRemovalDiff.ts
0xDEnYO: Fixed in `50205d846` along the lines you proposed. The early return no longer reports the addresses at all — it sets `diamondUnresolved: true` and leaves `notFoundOnChain` empty, so "the chain says absent" and "we could not ask the chain" are now distinct states rather than the same array. The drain bails on the flag before the partition loop, alerting `parked-task drain skipped: no LiFiDiamond resolved for <env> — N task(s) left queued`, and the stale comment about the early return is gone with it.

Tests: `flags an unresolvable diamond instead of reporting the addresses absent` and `bails out entirely when the diamond could not be resolved` (asserts no supersede, no claim, no calls).

script/deploy/safe/diamondRemovalDiff.ts
0xDEnYO: You are right on both halves — the second clause was derived from `addressToName` filtered by `protectedNames`, so it could only be true when the first clause already was, and an unlogged address fell straight through to `removals` with `name: undefined`, which made the `'unknown (address not in deploy log)'` label unreachable.

Fixed in `50205d846` with the selector-based side you suggested rather than by refusing every unlogged address — refusing outright would break the case this path exists for, since a superseded version (SymbiosisFacet v1.0.0) is normally *absent* from the log. So:

- `diffFacetsByAddress` now takes `protectedSelectors: Set<string> | undefined` (required, never optional, so no caller can pick a failure direction by omission).
- An address the log cannot name is matched by selector against the protected facets' union; a hit is refused with `unknown (address not in deploy log, holds protected selector 0x…)`.
- `collectProtectedFacetSelectors` builds that union from `protectedNames ∩ getFacetNames()` via `getActiveSelectors` — scoped to real facets, since protected periphery is not diamond-routed and its overlapping signatures would refuse legitimate removals.
- If the union cannot be built (missing artifact) it is `undefined`, and every unlogged address is then refused: `unknown (address not in deploy log, protected selectors unavailable)`. Fails closed.

Three new tests cover refuse-on-selector-hit, refuse-when-unavailable, and still-remove-when-not-protected. The docstring now describes what the second side actually does.

script/deploy/healthCheckInvariants.ts
0xDEnYO: Confirmed and fixed in `50205d846` — `MONGODB_URI: ${{ secrets.MONGODB_URI }}` is now on the *Run health check* step of `healthCheckAllNetworks.yml` (it was only on *Fetch RPC endpoints from MongoDB*). Added the same to the *Run Health Checks on New Networks* step of `healthCheckForNewNetworkDeployment.yml`, which had the identical gap.

Without it the invariant could only ever emit the coverage warning, so its entire purpose — telling a queued backlog apart from an uncovered stale facet — never ran on the job it was written for.

script/deploy/safe/reconcile-parked-tasks.ts
0xDEnYO: Fixed in `50205d846`. `IReconcileContext` gains `facetDeprecated?: boolean` and the resolved branch is now:

```ts
if (RESOLVED_STATUSES.includes(task.status)) {
  if (!ctx.facetPresentOnChain) return 'keep'
  return ctx.facetDeprecated === true ? 'reopen' : 'keep'
}
```

The live adapter computes it per network with the same gate `findDeprecatedLiveFacets` uses — absent from `_targetState.json` **and** source deleted (`getExpectedFacetNames` + a memoized `getSourceContractNames`) — so eval and remover cannot disagree about what is removable. `undefined` (no target-state entry for the network) also withholds the reopen rather than guessing.

Your rollback scenario now stops at step 3: routed again + present in target state → `keep`, and because a silent `keep` would hide it, the withheld reopen is pushed to the new `anomalies` bucket and alerted to Slack.

Tests: `does NOT reopen a routed facet that is no longer deprecated (incident rollback)` and `does NOT reopen when deprecation cannot be determined`; both fail against a negative control that returns `'reopen'` unconditionally.

script/deploy/healthCheckInvariants.ts
0xDEnYO: Fixed in `50205d846` — coverage is keyed on `t.facetAddress.toLowerCase()`, exactly as you spelled it out. The split is now an exported pure helper, `splitByParkedCoverage`, so it is unit-tested rather than buried in the dynamic-import `run()` path.

Your EXSC-750 example is one of the tests: both versions routed, a task for v1 only → v1 `parked`, v2 `unparked` and counted in the warning total. Negative control (helper switched back to name-matching) fails three of the four new tests.

script/deploy/safe/parked-tasks.ts
0xDEnYO: Fixed in `50205d846`. `reopenResolvedTask(parkedTasks, id: ObjectId)` now transitions on `{_id: {$eq: id}, status: {$in: ['executed','superseded']}}` instead of going through the shared `taskKey` `transition()` helper, and the reconcile passes `task._id` — the document it already holds. So the write targets the row the loop decided on, the `reopened.push({facet, prUrl, from})` fields describe that same row, and the second terminal row no longer collides itself out of the alert.

Test: `reopens the exact row it was given when one taskKey owns several terminal rows` — two terminal rows under one key, asserts the second flips and the first does not. Fails against a negative control that matches on status alone. The docstring now records why `_id` and not `taskKey`.

script/deploy/safe/reconcile-parked-tasks.ts
0xDEnYO: You are right that the console warning was the only channel, which on the cron is no channel. Fixed in `50205d846` by taking the stronger of your two options — refusal, not just reporting:

- A suspect snapshot is detected **before** any transition is computed; the task is left exactly as it is (no `markSuperseded`) and pushed to a new `run.anomalies` bucket.
- `formatReconcileAnomalyMessage` sends those to Slack through `runReconcileAlerts` (`⚠️ N … left UNCHANGED because their on-chain data is contradictory — a human has to adjudicate`, one line per task with facet, network and origin PR, capped at 15 to stay inside the 2900-char budget).
- The same bucket carries the withheld reopens from your other thread.

The drain got the matching refusal (gvladika's thread) so the two sides cannot disagree, and the PR body claim about "a loud warning next to the decision" is now true of the alert rather than the log line.

0xDEnYO:
## Review round addressed — `081a7ad77` (merge main) + `50205d846` (fixes)

All ten unresolved inline threads (@gvladika ×3, @melianessa ×7) are answered in-thread with the fix and its test. Below is what the QA report added on top, plus the two items that are not code.

### QA report items

| # | Item | Status |
|---|---|---|
| 1 | Duplicate network names inflate the failure count | **Fixed** — `formatReconcileFailureMessage` counts `new Set(failures.map(f => f.network)).size`; the per-`(network, environment)` detail lines stay. Test: two environments of `ethereum` → `1 network(s)`, both rows still listed. |
| 2 | Mongo write failures can abort the sweep | **Fixed** — same as @gvladika's thread: per-task `try/catch` → `IReconcileFailure` + `continue`. |
| 3 | Tron address case-sensitivity dedup gap | **Fixed at the root** — `enqueueParkedTask` now stores the canonical spelling (`getAddress` for `0x…`, base58 verbatim), so the stored `facetAddress` and `computeTaskKey` can no longer diverge. Tron stays case-sensitive by design; two tests pin both halves. |
| 4 | Invariant integration path untested | **Partly closed** — the coverage split moved out of `run()` into the exported `splitByParkedCoverage` and is now unit-tested (4 tests). The dynamic-import queue-open and the Mongo-unreachable fallback remain untested; that path is inside the unit-test-exempt live adapter. |
| 5 | `migrate-parked-task-keys.ts` `finally` masks the `MONGODB_URI` error | **Fixed** — `mongoClient?.close(true)`. |
| 6 | `deprecate-network.md` step 7 ordering not an explicit precondition | **Fixed** — step 7 opens with a **Precondition** block stating that running before step 2 makes the cancellation a silent no-op, with a verification command (`jq -r '."<network>".status // "absent"' config/networks.json`, must not print `active`). Command executed against the real config before documenting it. |
| 7 | Merge-order dependency on #2208 | **Moot** — [#2208](https://github.com/lifinance/contracts/pull/2208) merged 2026-08-18, and `081a7ad77` merges `main` into this branch. Both reconcile alert send sites are already on `WEBHOOK_DEV_SC_GITHUB_CI_NOTIFICATIONS` behind `isUnattendedRun()`; the merge kept `main`'s stricter `ttlAlertDelivery` model (a missing webhook on the unattended run is now a `misconfigured` failure, not a silent skip) alongside this branch's reconcile. |

### Merge resolution notes (`081a7ad77`)

The branch and `main` had both edited the reconcile since the fork. Resolution keeps `main`'s `ttlAlertDelivery` state machine and its stricter workflow/doc wording about the webhook being required, and this branch's `IReconcileRun` / `resolveFacetPresence` / safe-to-prune / loupe-only-degrade work. The reconcile test file is the union of both sides — no test dropped from either parent, and the two extra workflow-header bullets this branch added (per-network isolation, deprecated-network policy) were re-applied by hand after the conflicting hunk.

### Verification

- `bun test script/` → **851 pass / 0 fail** (up from 655; 21 new tests across five files).
- Every new assertion checked against a negative control: reverting each fix in turn (name-keyed coverage, row-count failure message, ungated reopen, missing drain suspect guard, `taskKey`-matched reopen) fails exactly the 8 new tests that should fail, and nothing else.
- eslint + `tsc-files` clean on all touched files; `markdownlint-cli2` 0 errors; all three touched workflows parse.

### Still deliberately out of scope

`reconcileAll` itself stays unit-test exempt (live Mongo/loupe/Slack adapter), so the per-task write isolation and the anomaly delivery are covered by the pure decision/formatter tests plus a live dry-run rather than by a fixture. And the key migration ([EXSC-790](https://linear.app/lifi-linear/issue/EXSC-790/run-the-parked-task-key-migration-immediately-after-pr-2157-merges)) still runs **after** merge, not before.

0xDEnYO:
## Review-gate report — `cfeb884bc` (gate of the fix commit `50205d846`)

Per the per-push gate rule, the fix commits were themselves gated before re-pinging. The agent fan-out was not used this run (session constraint), so this was the inline gate: rules adherence, collaborator/bug pass, comment-drift pass, and the falsification pass. **The falsification pass found two defects in my own fix commit**, both now fixed in `cfeb884bc`.

### 1. An unverifiable protected set was reported as "protected" — which *cancels* the task

`50205d846` refused an unlogged address when the protected-selector union could not be built, and put it in `protectedSkipped`. Reading the drain's collaborator code (not the diff) shows that set is terminal:

```ts
} else if (protectedAddresses.has(address)) {
  await deps.cancel(task.taskKey)   // terminal
  deps.alert(`… a PROTECTED facet was parked for removal — cancelling (enqueue bug)`)
```

So a missing artifact would have **terminally cancelled a legitimate removal** and blamed it on an enqueue bug. Fixed: a new `unverifiable` bucket on `IAddressRemovalResult`; the drain leaves those tasks **queued** with an alert naming `forge build`, and `cleanUpProdDiamond` prints them as a distinct refusal. Test asserts `cancel` is never called and `protectedSkipped` stays empty.

### 2. The reopen gate was name-based, dropping the co-registered case the queue exists for

Falsified on real repo data:

| probe | real result |
|---|---|
| `mainnet/SymbiosisFacet` — in target state? source exists? | `true` / `true` → name-gate says **not deprecated** |
| so a parked *superseded* v1.0.0 task would have | **withheld its reopen** — the EXSC-774 case, dropped |
| v1.0.0's routed selectors `0xb70fb9a5`, `0x6e067161` in the active union? | `false`, `false` |
| live v2.0.0's own artifact selectors | `0xe23b7a08`, `0xc46059b2` |
| active union size (36 target-state facets) | 156 selectors |

Fixed with `isParkedFacetStillRemovable` (pure, exported, 6 tests): name-deprecated → removable; otherwise an address the deploy log does not name is removable when **none** of its routed selectors belongs to a target-state facet — the removal engine's own held-back rule. A logged address, a selector an expected facet owns, or any undecidable input still withholds and reports an anomaly. Your rollback scenario stays closed: a re-added facet's address is in the log, so it returns `false`.

### Other falsification probes (all on real data, no fixtures)

- **Protected selector union actually builds and actually covers something**: 47 selectors over 12 protected facets (`DiamondCutFacet`, `DiamondLoupeFacet`, `OwnershipFacet`, `EmergencyPauseFacet`, `AccessManagerFacet`, `CalldataVerificationFacet`, `GasZipFacet`, `GenericSwapFacetV3`, `LiFiIntentEscrowFacetV2`, `PeripheryRegistryFacet`, `WhitelistManagerFacet`, `WithdrawFacet`), and it contains the real `DiamondCutFacet` selector `0x1f931c1c`. So the new second side of the protected check can genuinely fire rather than being a differently-shaped no-op.
- **The `diamondUnresolved` path is reachable, not hypothetical**: of 72 deploy logs, `deployments/tronshasta.json` has no `LiFiDiamond` key — exactly the early return that previously reported every requested address as absent.
- **The deprecation gate fires on the known true positives**: `tron/GenericSwapFacet` and `tron/DexManagerFacet` both resolve `deprecated=true` (EXSC-724), as does `mainnet/AcrossFacetV3`.
- **Negative controls**: reverting each new guard in turn fails exactly the new tests that should fail (10 across five files) and nothing else.

### Verification after the gate fixes

`bun test script/` → **858 pass / 0 fail**. eslint clean on every file I touched (the 13 remaining warnings are pre-existing `no-explicit-any` in `deploy-safe.ts` / `ledger.ts`, untouched). `tsc-files` clean; markdownlint 0 errors.

### Escalated / nothing to decide

No escalations this round — both findings were mine, unambiguous, and outside the always-escalate change classes (no storage, selector-layout, governance, event-signature or amount-math changes; the governance objects remain byte-for-byte unchanged). Coverage note: the git-history and prior-PR-comment dimensions got reduced coverage without the fan-out.

0xDEnYO:
**/gate-review** (per-push gate) on `10cad6e9f..725987a19` — 7 review agents (rules, bugs+collaborators, history, prior-PR comments, comment-guidance, falsification, removed-keys grep). 15 findings auto-fixed in `725987a19`; the falsification agent **executed every new guard against real repo data** (real mainnet deploy log, real `out/` artifacts, the real 156-selector active union) and confirmed each partition fires.

**Escalated — needs a human decision (not changed):**

- `script/deploy/safe/safe-utils.ts:~1428` — `ensurePendingProposalIndex` still swallows any index-creation error other than Mongo code 85 (`consola.warn` + return), so the new close-on-throw guard in `getSafeMongoCollection` never fires for a real index-creation failure and the signing-store dedup index could silently be missing. This is CodeRabbit's finding from #2203 which was claimed fixed there but never independently re-verified (its re-review was rate-limited). Tightening it changes the signing-store connection's failure semantics (availability vs consistency), so it's deliberately left for a maintainer call — either rethrow non-85 errors, or document the warn-and-continue as intended.

**Accepted residuals (reviewed, deliberate):**

- A legacy `executed`/`superseded` row with no `resolvedAt` on a permanently non-active network alerts weekly until hand-retired — fail-loud by design; no code path can stamp it there.
- A rare cross-drain race can report a lost claim as `duplicateAddresses`; the alert text now says when no action is needed.
- The widened fail-closed surface (missing target-state entry or stale `out/` ⇒ `unverifiable`, refused) is intended: an unverifiable removal waits, and the alert names the remedy.

0xDEnYO:
@coderabbitai review

script/deploy/safe/migrate-parked-task-keys.ts
```diff
@@ -0,0 +1,156 @@
...
+          `  ${task.network}/${task.facetName} (${task.status}): ${task.taskKey} → ${newKey}`
+        )
+        if (args.apply)
+          await parkedTasks.updateOne(
+            { _id: task._id },
+            { $set: { taskKey: newKey } }
+          )
+        if (isOpen) claimedOpenKeys.add(newKey)
+        migrated++
+      }
```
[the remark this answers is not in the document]
0xDEnYO: Fixed in a1f334fec — the updateOne is wrapped per row: a rejected write is reported and the loop continues, so un-migrated rows can no longer be stranded by one failure.

script/deploy/safe/reconcile-parked-tasks.ts
```diff
@@ -201,88 +706,453 @@ async function resolveProposalStatus(
...
+    })
+  for (const g of deprecatedByNetwork.values())
+    if (g.blocked.length > 0)
+      failures.push({
+        network: g.network,
+        environment: g.environment,
+        reason: `outside the active set in networks.json — ${
+          g.blocked.length
+        } parked task(s) not reconciled: ${g.blocked.join(', ')}`,
+      })
```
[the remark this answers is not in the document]
0xDEnYO: Fixed in a1f334fec — the per-network reason now lists at most 10 blocked entries with an '… and N more (see the job log)' tail, so a fleet-wide narrowing stays inside the Slack budget.

script/deploy/safe/reconcile-parked-tasks.ts
```diff
@@ -201,88 +706,453 @@ async function resolveProposalStatus(
...
+  const send = async (message: string): Promise<void> => {
+    if (apply && webhookUrl && isUnattendedRun())
+      await new SlackNotifier(webhookUrl).sendNotificationWithRetry({
+        text: message,
+      })
+  }
+  if (apply && webhookUrl && !isUnattendedRun())
+    consola.info(
+      'Local run: reconcile alerts logged only. Set CI=1 to deliver them to Slack.'
+    )
```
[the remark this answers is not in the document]
0xDEnYO: Fixed in a1f334fec — runReconcileAlerts now shares ttlAlertDelivery: an unattended applied run with alerts to post but no webhook throws (after logging the alerts), matching runTtlAlert, so the run can no longer stay green while dropping the reopen signal.

script/deploy/safe/safe-utils.ts
```diff
@@ -1480,8 +1480,15 @@ export async function getSafeMongoCollection(): Promise<{
...
-  await ensurePendingProposalIndex(pendingTransactions)
+  // Ensure the partial unique index exists for duplicate prevention. The client is
+  // already connected here, so a failure must close it before rethrowing — an
+  // orphaned client is unreachable to the caller and keeps the process alive.
+  try {
+    await ensurePendingProposalIndex(pendingTransactions)
+  } catch (error) {
+    await client.close().catch(() => undefined)
+    throw error
+  }
```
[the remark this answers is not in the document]
0xDEnYO: Fixed in a1f334fec — ensurePendingProposalIndex now rethrows anything other than code 85 (exists with same options) and code 13 (unauthorized, warn-and-continue for permission-limited roles, mirroring the parked-tasks store), so real connection failures reach getSafeMongoCollection's close-and-rethrow guard.

0xDEnYO:
Update on the gate escalation: CodeRabbit's re-review independently flagged the same `ensurePendingProposalIndex` swallow as **Major**, so with two reviewers converging it is now **fixed in `a1f334fec`** — code 85 and code 13 (unauthorized, warn-and-continue for permission-limited roles) stay tolerated, everything else rethrows so `getSafeMongoCollection`'s close-and-rethrow guard can actually fire. No escalated items remain open.

0xDEnYO:
Per-push gate on `a1f334fec` (falsification-grade review of the CodeRabbit-round fix commit): callers of `getSafeMongoCollection` all traced and tolerate the tightened rethrow; no critical/major findings. Its two minor prescriptions are implemented verbatim in `1c07981e6`: migration write failures are bucketed separately from collisions (a transient tunnel error no longer tells the operator to cancel a healthy task — it says re-run), and index code 86 (definition drift) is tolerated with a warning, mirroring the parked-tasks store, so drift can't hard-fail the whole signing pipeline. `bun test script/`: 875 pass, 0 fail.

gvladika (approved):

gvladika (dismissed):

melianessa (dismissed):

gvladika (approved):

melianessa (approved):

melianessa (approved):

PR #2171, opened 2026-08-03
fix(FeeForwarder): make native refund invocation-local (EXSC-737) [FeeForwarder v2.0.0]

gvladika:

https://linear.app/lifi-linear/issue/EXSC-737/feeforwarder-v200-derive-native-refund-from-msgvalue-instead-of

`forwardNativeFees` paid each recipient via `LibAsset.transferNativeAsset` (which forwards all gas) and only then read `address(this).balance` to size the caller's refund. A recipient could call back into `forwardNativeFees` from its receive hook; the nested invocation's refund was sized from a balance that still contained the outer invocation's undistributed funds, so it paid them to the attacker while the outer call completed normally with a zero refund. The loss was absorbed silently — no revert. Reported through the Cantina bug bounty.

The refund is now `msg.value` minus the sum of the distributions. Both inputs belong to the current invocation and no callback can reach either, so a nested call can only ever be paid what its own caller funded.

**Re-entry is not blocked, it is made unprofitable.** A recipient can still call back in. What changes is that a nested call claiming more than it funded is rejected — it reverts with `InsufficientBalance(required, balance)` — so the attack fails at its source.

Re-entry that has nothing to claim still succeeds, which `test_ForwardNativeFeesAllowsHarmlessReentryWhileRefundOutstanding` pins down: the outer call still owes a 4.9 ether refund when the callback fires, the nested call takes nothing, and the refund still reaches the original caller.

## Why a named error rather than the checked subtraction

Letting `msg.value - totalDistributed` underflow would enforce the same rule, but it surfaces as a bare `Panic(0x11)`. `LibSwap.swap` bubbles returndata verbatim, so through the Diamond that is byte-identical to an arithmetic overflow anywhere else in the route — unattributable for the backend and for tx-analysis tooling. An explicit `if (totalDistributed > msg.value) revert InsufficientBalance(...)` costs nothing extra, because the subtraction can then run `unchecked`.

Note this error is reachable only when the contract's own balance could have covered the shortfall. With an empty contract the failing transfer reverts first with `ETHTransferFailed`. Both cases are documented in `docs/FeeForwarder.md`.

## Alternatives considered

**`nonReentrant`** — the textbook fix, and it closes the hole. Measured on this contract it costs **+7,266 gas** for a single distribution (38,824 vs 31,558), roughly a quarter of the whole operation, because the repo's guard writes a storage slot and Solidity 0.8.17 has no transient storage. Transient storage is not an option here: FeeForwarder is deployed across networks with mixed `targetEvmVersion`, so a Cancun-only opcode would not be portable.

**Snapshotting `address(this).balance` before the loop** — preserves v1.0.0's refund semantics exactly, including returning a pre-existing balance to the caller. Rejected because it is the one option where the nested call *succeeds* and the outer refund then fails, so any re-entering recipient reliably reverts the whole route. That trades a theft vector for a liveness vector, which is the wrong trade for the threat model in EXSC-737 (a compromised fee wallet or an EIP-7702 delegation on a configured recipient).

**Checking the budget before the loop instead of after** — a separate summing pass would refuse to spend before any external call, and would make the revert reason the same regardless of contract balance. It costs +252/+435/+618 versus v1.0.0, roughly 4x this version, for no change in outcome: the over-budget frame reverts either way, and a revert unwinds every transfer and callback in it. Not worth the gas.

## Deliberate behavioural change

A balance the contract already held is no longer returned to the caller. That was never designed — it fell out of reading the global balance, and it made any caller of `forwardNativeFees` able to collect whatever dust was sitting there. Stray funds now stay put and the owner recovers them through `WithdrawablePeriphery`, which is what that inheritance is for.

This is not separable from the fix: if the contract is willing to refund a balance it did not receive in `msg.value`, then a nested call has an equally legitimate-looking claim on its parent's funds, and telling them apart requires a guard.

## Cost and versioning

Gas, measured against v1.0.0: **+61** for one distribution, **+84** for two, **+107** for three, **+38** for an empty array. Under 0.3% of a 22–30k native transfer.

Version bump is **MAJOR**: `forwardNativeFees` gains a revert path, and the stray-balance refund behaviour changes.

## Tests

Six cases added. Five cover the accounting directly — a re-entering recipient cannot claim the refund, the same holds when the nested call is funded by the recipient so it looks legitimate, re-entry with nothing to claim still succeeds while a refund is outstanding, a pre-existing balance is not refunded, and over-distribution reverts even when a stray balance would cover it. All five fail against v1.0.0 (the mutation output shows the attacker taking 5 ETH where it should get 0.1, and 5.5 where it should get 0.6), while all 21 pre-existing tests still pass there — so nothing in the legitimate path changed.

The re-entry harness counts callbacks and captures the nested revert data, so the PoC asserts that re-entry actually fired and that it was rejected by the `msg.value` accounting rather than by an unrelated failure.

The sixth is `test_CanForwardNativeFeeViaFeeForwarderThenSwapToERC20` in `GenericSwapFacetV3.t.sol`. `forwardNativeFees` is whitelisted as a Diamond call target on ~50 networks but was only ever exercised from an EOA; this covers the production shape, where the fee step is funded with the full swap amount and the refund must come back to the Diamond for the next step to have anything to swap. Removing the refund transfer makes it fail.

## Notes

`GasZipPeriphery.depositToGasZipNative` uses the same balance-based refund shape and is **not** fixed here. It is not exploitable the same way — the only external call before its balance read is to the immutable `GAS_ZIP_ROUTER` — but it is `public payable` with no access control and that contract does have a `receive()`, so a balance can accumulate and any caller can sweep it. Tracked separately rather than widened into this PR.

FeeForwarder is deployed on Tron, which is not reached by the EVM rollout (Foundry cannot deploy there and FeeForwarder is not part of the `contracts-tron` source delta). The Tron redeploy needs to be tracked as part of closing this out.

`script/deploy/_targetState.json` is untouched — per the repo's split, the version cut lands with the deploy commit, not the source PR. The audit entry for v2.0.0 follows at the audit stage, so `VersionControlAndAuditVerification` stays red until then.


gvladika:
@coderabbitai review

src/Periphery/FeeForwarder.sol
```diff
@@ -3,12 +3,16 @@ pragma solidity ^0.8.17;
...
+import { InvalidConfig, InsufficientBalance } from "../Errors/GenericErrors.sol";
 
 /// @title FeeForwarder
 /// @author LI.FI (https://li.fi)
 /// @notice Forwards various fee amounts to designated recipients
-/// @custom:version 1.0.0
+/// @dev This contract is not intended to custody funds: a normal invocation forwards
+///      everything it receives and leaves no net balance behind. A balance that does
+///      accumulate anyway (e.g. a forced transfer) is never paid out as a caller refund
+///      and is recoverable only by the owner via WithdrawablePeriphery.
```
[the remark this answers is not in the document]
gvladika: @coderabbitai The tag is already there. `src/Periphery/FeeForwarder.sol:15` is `/// @custom:version 2.0.0`, immediately above the `contract FeeForwarder is WithdrawablePeriphery {` declaration on line 16 — which is exactly what the path instruction asks for.

I think the anchor slipped: this comment is pinned to line 14, the last line of the `@dev` block, and the suggested diff proposes inserting the tag right where it already exists. The `@dev` block grew by a line in this PR, so the version tag moved from 14 to 15.

No change needed — please confirm.

src/Periphery/FeeForwarder.sol
```diff
@@ -3,12 +3,16 @@ pragma solidity ^0.8.17;
...
+/// @custom:version 2.0.0
```
0xDEnYO, line 15: imho this is a bugfix and doesnt change any interfaces so I would have bumped a minor version but I am not too fuzzy about it

src/Periphery/FeeForwarder.sol
```diff
@@ -73,39 +77,49 @@ contract FeeForwarder is WithdrawablePeriphery {
...
 
     /// @notice Forwards native token fees to the specified recipients
-    /// @dev Any excess native tokens sent will be refunded to the caller. Transaction will revert if insufficient funds.
+    /// @dev The unspent part of msg.value will be refunded to the caller. An invocation may
+    ///      never distribute more than its own msg.value: it reverts with InsufficientBalance
+    ///      if the contract balance covered the shortfall, otherwise the failing transfer
+    ///      reverts first with ETHTransferFailed.
     ///      The tx will not revert if the array is empty, but will still emit the FeesForwarded event.
     /// @param _distributions Array of fee distributions containing recipients and amounts
     function forwardNativeFees(
```
0xDEnYO, line 86: we could have also just added reentrancy protection, right?
did you consider that and weigh the different options? If yes, why did you decide for this one (seems more gas intensive to me on first glimpse)

src/Periphery/FeeForwarder.sol
gvladika: Repo convention forces MAJOR here even though the ABI signature is unchanged — [CONV:VERSIONING] lists "new revert on an existing function" as a MAJOR trigger, and this adds the `InsufficientBalance` revert path plus a small behavioral change (stray balance no longer refunded).

src/Periphery/FeeForwarder.sol
gvladika: `nonReentrant` was the first thing considered and it does close the hole with the same observable behavior. But it's actually gas intensive: measured +7,266 vs v1.0.0 (the repo's guard writes a storage slot, and Solidity `0.8.17` has no transient storage). The current `msg.value`-accounting fix is +61 for the same outcome. Full comparison (snapshot-before-loop and check-before-loop too) is in the PR description under "Alternatives considered."

src/Periphery/FeeForwarder.sol
0xDEnYO: ok, got it. 
resolving

src/Periphery/FeeForwarder.sol
0xDEnYO: awesome, thanks for clarification

0xDEnYO (dismissed):

gvladika:
Thanks for the review. Status on the flagged items:

**Medium #1 & #2 (open reviewer threads) — resolved.** Both `0xDEnYO` threads were replied to and marked resolved on 2026-08-16:
- **Line 15 (version bump):** replied citing `[CONV:VERSIONING]` — a new revert on an existing function is a MAJOR trigger, plus the stray-balance behavioural change. Reviewer acknowledged and resolved.
- **Line 86 (reentrancy):** replied with the gas comparison (`nonReentrant` +7,266 vs the `msg.value`-accounting fix's +61/+84/+107). Reviewer acknowledged and resolved.

The review was anchored to a snapshot from before those replies landed.

**Low #3 (msg.value=0 + stray-balance test):** acknowledged, leaving as-is. `testRevert_ForwardNativeFeesDistributingMoreThanMsgValue` already exercises the same `InsufficientBalance` guard; the zero-`msg.value` Patcher path is a strict subset of that same check, so a dedicated case adds coverage of the input but not of a distinct code path.

Re-triggering the QA run against the current state.

0xDEnYO (approved):

PR #2179, opened 2026-08-06
chore(scripts): cap cast RPC calls and add castRead fallback helper

gvladika:

Fixes [EXSC-746](https://linear.app/lifi-linear/issue/EXSC-746/cap-cast-rpc-calls-with-eth-rpc-timeout-castread-fallback-helper)

During the LiFi Intents settler switchover rollout, on-chain read calls via `cast` repeatedly hung for the full 2-minute shell timeout whenever an RPC endpoint was unreachable or slow, because `cast` had no client-side request cap and `getRPCUrl` returns a single endpoint with no fallback. Over a multi-network rollout that added up to a lot of dead waiting.

Two layered fixes, both low-risk and opt-in for existing callers:

- **`ETH_RPC_TIMEOUT`** — documented in `.env.example`. `cast` (Foundry ≥ 1.7) reads this env var as a per-request timeout, so a dead or blocked endpoint fails in seconds instead of hanging. It flows into every `cast` invocation automatically, including the existing `universalCall`/`universalCode` paths, with zero code changes to them.
- **`castRead()`** in `helperFunctions.sh` — a read-only `cast` wrapper that applies a hard timeout (a coreutils `timeout`/`gtimeout` kill when present, otherwise `ETH_RPC_TIMEOUT` alone) and falls back across the env RPC (`getRPCUrl`) then the public `rpcUrl` from `networks.json`, returning the first endpoint that answers. It reuses the two RPC sources that already exist rather than introducing a new curated endpoint list.

`castRead` is deliberately standalone and has no callers in this diff — I did not rewire the deploy-critical `universalCall`/`universalCode` path, since adding fallback there changes behavior for every existing caller (silent failover could mask a primary-RPC problem and shift timing). **Design question for the reviewer:** keep `castRead` as an opt-in helper adopted incrementally, or fold the timeout+fallback into `universalCall`/`universalCode` so it applies everywhere? I lean standalone first, but it's a call for whoever owns that path.

Verified locally: `ETH_RPC_TIMEOUT=12` aborts a non-routable endpoint in 12s (vs hanging); `castRead katana chain-id` with a deliberately dead primary falls back to the `networks.json` RPC and returns `747474`; `bash -n`, strict-mode source, and `shellcheck` are clean on the new function (the one SC2155 in `helperFunctions.sh` is pre-existing on `main`).

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -21,6 +21,10 @@ PRIVATE_KEY="<your private key>"
 # To populate all RPCs, please run "bun fetch-rpcs"
 # To add a new RPC for a specific network, run: "bun add-network-rpc --network {networkName} --rpcUrl {rpcUrl}"
 
+# Hard cap (seconds) on every `cast` RPC request. Without it, a dead or blocked endpoint
+# hangs until the caller's own timeout, blocking deploy/verify/proposal scripts for minutes.
+ETH_RPC_TIMEOUT=12
+
 # Mainnet Explorer API Keys
 MAINNET_ETHERSCAN_API_KEY=
 ABSTRACT_ETHERSCAN_API_KEY=
--- a/script/helperFunctions.sh
+++ b/script/helperFunctions.sh
@@ -3912,6 +3912,73 @@ function getRpcUrlFromNetworksJson() {
   echo "$RPC_URL"
 }
 
+# castRead: Run a read-only `cast` call with a hard timeout and RPC fallback, so a dead or
+# blocked endpoint fails fast instead of hanging the caller. Tries the env RPC (getRPCUrl)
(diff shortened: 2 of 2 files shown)


gvladika:
Closing — the diagnosis was wrong. The multi-minute RPC waits during the rollout came from the **viem** transport in the TS Safe scripts (default `timeout=10s` + `retryCount=3`, single endpoint, sequential reads), not from `cast`. `cast` already self-caps at ~12s, and `ETH_RPC_TIMEOUT` is cast-only (viem ignores it), so this PR would not have saved any of the time it claimed to. Re-scoping a fix at the viem layer instead.

PR #2183, opened 2026-08-07
fix(deploy): check escrow settler is a contract; demote off core

gvladika:

Fixes [EXSC-740](https://linear.app/lifi-linear/issue/EXSC-740/deploylifiintentescrowfacetv2-does-not-verify-the-settler-address-is-a)

`DeployLiFiIntentEscrowFacetV2.s.sol` read its settler constructor arg with a bare `json.readAddress(".lifiEscrowInputSettler")`, with no check that the address holds code. On chains where the settler was not deployed the facet deployed anyway and bound its `immutable LIFI_INTENT_ESCROW_SETTLER_V2` to a codeless address; because the binding is immutable the only remedy is a redeploy. The fix swaps the bare read for the inherited `ScriptBase._getConfigContractAddress` helper — the same checked read `DeployReceiverOIF.s.sol` already uses — which reverts `"<key> in file <path> is not a contract"` when the address has no code, failing the deploy loudly at the right moment.

That check surfaced a deeper mismatch. `LiFiIntentEscrowFacetV2` was listed in `config/global.json` → `coreFacets`, so `deployCoreFacets.sh` deploys it unconditionally on every network — but its constructor requires an external settler contract that is deployed on only 12 chains. That mismatch is exactly how the facet came to be deployed against a codeless settler on chains that lack it (e.g. jovay, whose own `CORE_FACET_EXEMPTIONS` entry documented "Intent escrow settlers are not deployed on Jovay"). The new check turns that latent over-deploy into a hard failure — first caught by `deploy-smoke-test`, which runs a full core-facet deploy on a throwaway anvil where no settler can exist.

The right home for a facet that can only deploy where its external prerequisite exists is the per-network `_targetState.json` path, not the deploy-everywhere core path. The facet is already pinned in `_targetState.json` on exactly the 12 chains where the settler is deployed, so moving it off `coreFacets` changes nothing on those chains: the standard non-core deploy-and-cut flow (its existing `UpdateLiFiIntentEscrowFacetV2.s.sol` + target state) maintains it there, identical to every other prerequisite-gated facet (AcrossFacetV4, StargateFacetV2, ReceiverOIF, …). localanvil and the previously grandfathered chains simply never deploy it. Concretely this PR removes the facet from `coreFacets` and removes the two now-obsolete `CORE_FACET_EXEMPTIONS` entries — those existed only to suppress the *core* health check on chains without the settler, and `healthCheckInvariants.test.ts` enforces that every exemption targets a facet that is in `coreFacets`, so they are removed together. Health-check coverage on the 12 production chains is unchanged: the facet is now verified as a non-core facet via target state.

This reverses the "core going forward" designation from V2-227/#1997: a new intent chain must now be added to `_targetState.json` explicitly rather than picking the facet up automatically — the correct model for a facet whose constructor can only bind where its settler is deployed. Two things intentionally left out of scope: jovay's target state still lists the facet even though its settler is codeless there (pre-existing data from the original over-deploy — a separate backfill/cleanup), and the audit of other `script/deploy/facets/` scripts for the same unchecked-`readAddress` pattern found the escrow facet was the only offender (the remaining bare reads resolve EOAs or sentinel addresses, not contracts).


gvladika:
Closing — superseded by the actual deployment model. The intent settlers are reserved, deterministically-deployed vanity addresses (same address on every chain), so deploying LiFiIntentEscrowFacetV2 against a not-yet-deployed (codeless) ref is the intended pre-deploy workflow, not a bug (see the 'new LI.FI Intent Contracts' thread in #dev-sc-review). EXSC-740's check would block that workflow, and the demotion here was only introduced to make the smoke test pass after the check broke it. EXSC-740 marked Cancelled; a follow-up ticket will track the inverse (allow deploying against codeless reserved refs).

PR #2191, opened 2026-08-11
fix(EcoFacet): route Tron via non-EVM receiver convention [EcoFacet v2.0.0]

gvladika:

Fixes [EXSC-755](https://linear.app/lifi-linear/issue/EXSC-755/fix-ecofacet-tron-destination-receiver-handling)

EcoFacet v1.1.0 routed a Tron destination through its EVM-compatible branch, which requires `bridgeData.receiver` to equal the recipient decoded from the route's final TRC-20 `transfer` and rejects `NON_EVM_ADDRESS` for anything but Solana. The backend's generic `buildBridgeData` sends the `NON_EVM_ADDRESS` sentinel for every non-EVM destination (confirmed on-chain in the failed Base tx `0xa08781049ac3075fb06c598430a1624cb2d7fbbd486672c358d5d8b15080d5a0`, where `bridgeData.receiver == 0x11f1…f1`), so every EVM→Tron route reverted with `InvalidConfig()` before deposit. This change makes Tron follow the same non-EVM convention as the other 11 facets: the sentinel receiver is accepted for Tron, the real recipient is carried in `nonEVMReceiver` and cross-checked against the address decoded from the route (the authoritative source, mirroring how the Solana path cross-checks `solanaATA`), and `BridgeToNonEVMChainBytes32` is emitted. The route-receiver decode is extracted into `_decodeRouteReceiver` and reused by the EVM path with no behavior change there. Solana keeps the legacy `BridgeToNonEVMChain` (bytes) event because its 44-byte base58 pubkey does not fit in `bytes32`.

`nonEVMReceiver` for Tron is expected as a 32-byte left-padded address (`abi.encode(address)`), matching AllBridgeFacet's `bytes32` recipient convention; a wrong value cannot pass because it is cross-checked against the route. The `_targetState.json` bump to 1.2.0 and the on-chain rollout are intentionally left to the separate deploy step.


src/Facets/EcoFacet.sol
```diff
@@ -254,54 +261,89 @@ contract EcoFacet is ILiFi, ReentrancyGuard, SwapperV2, Validatable, LiFiData {
...
-                revert InvalidReceiver();
-            _validateSolanaReceiver(_ecoData);
         } else {
             if (_ecoData.encodedRoute.length == 0) revert InvalidConfig();
 
-            // If receiver is not NON_EVM_ADDRESS but destination is Solana, reject
-            if (isSolanaDestination) {
+            // A concrete receiver is only valid for EVM destinations; non-EVM
+            // chains must use the NON_EVM_ADDRESS sentinel path above.
+            if (isSolanaDestination || isTronDestination) {
```
0xDEnYO, line 319: does this check here really make sense? 

Isnt the flow

if(isSolanaDestination) ....
else if(isTronDestination) ....
else ( and here we check again `if (isSolanaDestination || isTronDestination) ` which cannot really be the case, can it?

gvladika: Branching is a bit confusing due to nesting, but this check in line 291 lives in the outer branch's `else`, reached when `receiver != NON_EVM_ADDRESS`. So it is reachable in the misconfiguration case where the backend sends a real EVM-looking receiver but a Solana/Tron destination chain id.

0xDEnYO: got it, thanks

0xDEnYO (dismissed):

0xDEnYO (approved):

PR #2192, opened 2026-08-11
fix(parked-tasks): reject enqueue for networks not active in networks.json

gvladika:

Fixes [EXSC-764](https://linear.app/lifi-linear/issue/EXSC-764/parked-task-enqueue-reject-networks-not-active-in-networksjson)

The scheduled `reconcile-parked-tasks` workflow crashed with `Chain evmos does not exist in config/networks.json` and aborted the entire run. The cause is a source-of-truth mismatch: the parked-task **enqueue** path (driven by `/deprecate-contract`) enumerates target networks from the **deploy logs** (`deployments/<network>.json`), which outlive a network's removal from `config/networks.json`. A deprecated network like `evmos` (a 2022 deployment) keeps its deploy log, so a facet-removal task was parked for it — but the **reconcile/drain** path resolves chains via `networks.json`, where `evmos` no longer exists, so it threw and blocked reconciliation for every other network. `evmos` is one of 9 orphaned networks that have deploy logs but no `networks.json` entry.

The fix makes `networks.json` authoritative at write time. The guard lives in `enqueueParkedTask` — the single enqueue chokepoint that already validates `prUrl`/`facetName` — so every caller (CLI, `/deprecate-contract`, the future drain) is covered; it throws unless the network is `status: 'active'` in `config/networks.json`. A matching fail-fast pre-check in the `enqueue-parked-task.ts` CLI gives a clean exit before the Mongo connect, mirroring the existing `prUrl`/`facetName` belt-and-suspenders pattern. The predicate reuses the canonical `networks` map from `viemScriptHelpers` (the same `status === 'active'` definition `getAllActiveNetworks` uses), so no new config reader is introduced. The already-parked `evmos` task was separately cancelled to unblock the current run; two follow-ups (reconcile skip-and-warn resilience, and cleaning up the 9 orphaned deploy logs) are noted on the ticket and left out of this PR to keep it single-concern.

### DIFF
--- a/.agents/commands/deprecate-contract.md
+++ b/.agents/commands/deprecate-contract.md
@@ -106,10 +106,14 @@ The command performs these steps in order:
    and [docs/FacetRemovalReconciliation.md](../../docs/FacetRemovalReconciliation.md).
 
    - **What to park**: for each deprecated facet, one parked task per PRODUCTION
-     network whose **deploy log** (`deployments/<network>.json`) lists it — that
-     log is the authoritative facet → address map. Read the diamond address and
-     the facet address from that log; do **not** delete those entries yet
-     (see step 7).
+     network that (a) is **active** in `config/networks.json` **and** (b) whose
+     **deploy log** (`deployments/<network>.json`) lists the facet — that log is
+     the authoritative facet → address map. Read the diamond address and the facet
+     address from that log; do **not** delete those entries yet (see step 7).
+     Skip any network absent from `networks.json`: deprecated networks keep their
+     deploy logs but are gone from `networks.json`, so a task parked for one can
+     never be resolved by the reconcile/drain (which route via `networks.json`).
+     The enqueue CLI enforces this too and rejects a non-active network.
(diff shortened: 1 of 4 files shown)


gvladika:
@coderabbitai review

gvladika:
Addressed in b915de590. Extracted a shared `assertActiveNetwork(network)` helper (exported from `parked-tasks.ts`) that both `enqueueParkedTask` and the CLI now call — single source for the normalization, the active-network rule, and the message. It also distinguishes the two cases you flagged: an **absent** network ("not in config/networks.json" → gone, clean up via `/deprecate-network`) vs. a **present-but-inactive** one ("is \"<status>\" (not active)" → exists in config but not parkable). The CLI validates before the MongoDB connect as before. Added a direct unit test covering all three branches.

@coderabbitai review

gvladika:
Superseded by #2202. Rather than guard the enqueue path against orphaned networks, we're removing the 9 orphaned deploy logs at the source (the cleanup EXSC-764 already listed as its follow-up), which is the cleaner fix. Closing in favor of #2202.

PR #2197, opened 2026-08-16
feat(VaultWrapper): cost-aware exits charge source exit fee to the exiter

gvladika:

This PR makes the wrapper's exit paths **and** exit views correct against a real yield source: the exiter bears the source's exit cost (no dilution of stayers, no bricked last exit), and the `max*` views never over-report what the source can actually honor right now. It folds in two findings that were originally split across two branches — the exit-cost fix and the source-liquidity-aware `max*` views (finding #4, previously PR #2206, now merged in here).

## Cost-aware exits (exit-cost)

The wrapper previously priced exits off the shares' face valuation. When the underlying yield source charges an exit fee, that fee was socialized onto the remaining holders (dilution), and a last holder's full redeem could revert on a fee-charging source. This PR prices exits off the source's own EIP-4626 previews (its true exit cost), so the exiting user bears the source's exit fee instead of the remaining holders.

- Added three views to `IYieldAdapter` (implemented in `ERC4626Adapter`): `previewWithdrawCost` (exact-out: position value the source consumes to deliver `N` assets, `>= N` on a fee source), `previewWithdrawUpTo` (exact-in realizable, net of the source fee), and `withdrawUpTo` (exact-in execution, delegatecall).
- `previewWithdraw` is now cost-aware exact-out: the exiter's share burn is priced off the source's true exit cost, clamped at the full-position floor value so source-preview rounding cannot overshoot near a full exit. The exiter bears the source exit fee; remaining holders are not diluted.
- `redeem` is now exact-in and loss-tolerant: it realizes the burned shares' proportional slice via `withdrawUpTo` and pays out what the source actually delivered, net of the wrapper withdrawal fee (charged on actual proceeds). A source paying under valuation (exit fee or loss) reduces this caller's payout rather than diluting stayers or bricking the last exit. `previewRedeem` is the realizable mirror.
- Exits always floor-value (no raw-position drain). Draining the raw position for the last holder would bypass the virtual-offset inflation protection (a donation-inflation attacker who is the last holder could recover their donation); floor-valuing keeps the small virtual-offset residue in the source as the inflation buffer, matching standard OpenZeppelin behaviour.

On a fee source, `maxWithdraw`/`maxRedeem` are fee-aware (they route through the realizable `previewRedeem`), so attempting to exact-out `withdraw` the full nominal deposit reverts `ERC4626ExceededMaxWithdraw`; `redeem` is the guaranteed full exit.

## Source-liquidity-aware max views (finding #4)

The exit-cost work above made the `max*` views fee/loss-aware but not **source-liquidity**-aware. When a yield source can only release part of the wrapper's position right now (an Aave-style utilization cap, a paused source, an illiquid Morpho market), the views over-reported and a `redeem`/`withdraw` within the reported max could revert — breaking the EIP-4626 guarantee that an exit within `maxRedeem` never reverts.

- Added one static adapter view `IYieldAdapter.maxWithdrawableValue(underlying, holder)` (implemented in `ERC4626Adapter`): the source's floor valuation (`convertToAssets`) of `min(holder position, source.maxRedeem)`. It is the **gross** valuation (deliberately not the fee-netted `previewRedeem` the sibling `previewWithdrawUpTo` uses), because the wrapper feeds it straight into its own gross-denominated share math to clamp the views; any source exit fee is applied separately by the realizable redeem path.
- `LiFiVaultWrapper.maxRedeem` now clamps the owner's balance to `_convertToShares(maxWithdrawableValue, Floor)` — but only when the source is genuinely liquidity-limited. A full-position short-circuit (`realizable >= totalAssets()`) returns the plain balance on fully-liquid sources, avoiding the ~1-share under-report the virtual-offset share round-trip would otherwise introduce. `maxWithdraw` inherits the clamp because OZ derives it as `previewRedeem(maxRedeem(owner))`, so the `withdraw` entrypoint's `ERC4626ExceededMaxWithdraw` guard is liquidity-aware too.
- `previewRedeem`/`previewWithdraw` stay liquidity-agnostic (EIP-4626 requires previews to ignore withdrawal limits). Exit execution paths are unchanged — they already fail on a short source; finding #4 was about the views.

Accepted tradeoff (documented): on OZ-style sources whose own `maxRedeem` floors ~1 share below `balanceOf` even at full liquidity (e.g. MetaMorpho), the clamp lands a sub-`1e-12`-token dust below `balanceOf`, so a one-call `redeem(balanceOf)` can hit `ERC4626ExceededMaxRedeem`. Callers should exit via `redeem(maxRedeem(owner))` (the EIP-4626-recommended pattern); a follow-up call clears the dust. `redeem` remains the guaranteed exit.

## Scope, versioning, health-check

Scope is the exit paths and exit views only. Out of scope and documented as the only remaining gaps: deposit entry-fee awareness, and fee-on-transfer / non-compliant sources (unsupported — they need a dedicated adapter). `@custom:version` is held at `1.0.0` (subsystem is pre-release). `script/deploy/healthCheckInvariants.ts` was reviewed: no change needed (this adds interface methods and internal view logic only — no deployed-contract topology, constructor, selector, or storage change; the Vault Wrapper is outside the Diamond the health-check covers).

# Testing

- Exit-cost: added `MockFeeERC4626` (a compliant ERC-4626 whose exit fee leaves the source vault, so dilution is observable), exit-cost unit and regression tests, and extended the stateful invariant suite to run over both a no-fee and a fee-charging source with a new `invariant_ExitsNeverCreateValueForExiter`. Two pre-existing full-supply-redeem assertions were widened from 1 to 2 wei (the realizable path adds one bounded layer of source-preview rounding); the `LossyVault` test mock gained standard ERC-4626 views now reached through the exit-cost path.
- Source-liquidity (finding #4): added `MockLiquidityCappedERC4626` (models an Aave-style withdrawal cap — assets stay on the vault, only a policy-limited amount may leave), adapter and wrapper unit tests, and a third stateful invariant suite that fuzzes the source's liquidity cap up and down while a `redeem` bounded to `maxRedeem` runs under `fail-on-revert` (0 reverts, with the clamp actively binding on most executed redeems), plus `invariant_MaxRedeemWithinSourceLiquidity`. The Morpho Arbitrum fork scenario was updated to exit via `redeem(maxRedeem)` and to assert the drain residue is dust (`convertToAssets(totalSupply()) <= 10 wei`) instead of exactly 0.

Full subsystem suite: 349 tests passing (347 non-fork re-verified locally after merging latest `main` via `dev-vault-wrapper`; fork suite unchanged).

### DIFF
--- a/docs/VaultWrapper/ERC4626Adapter.md
+++ b/docs/VaultWrapper/ERC4626Adapter.md
@@ -21,6 +21,30 @@ asked. It does **not** catch share-side dilution (a vault that consumes the full
 asset but credits fewer shares via an internal deposit fee); such non-standard
 sources are unsupported and require a dedicated adapter.
 
+## Exit cost views
+
+Three methods give the wrapper cost-aware and loss-tolerant exits without
+assuming anything about the source beyond its own standard EIP-4626 previews:
+
+- `previewWithdrawCost` — the exact-out cost: the position value the source
+  consumes to deliver exactly `assets` out (`>= assets` when the source charges
+  an exit fee). Backs the wrapper's cost-aware `previewWithdraw`, so an exiting
+  holder's share burn is priced off the source's true exit cost instead of
+  diluting remaining holders.
+- `previewWithdrawUpTo` — the exact-in realizable amount: what the source would
+  deliver if the holder realizes up to `assets` of position value now, net of
+  any source exit fee. `type(uint256).max` means "the whole position" (the
+  drain affordance). It is the static mirror of `withdrawUpTo` and backs the
+  wrapper's realizable `previewRedeem`.
+- `withdrawUpTo` — the exact-in execution. DELEGATECALL only; realizes up to
+  `assets` of position value into the wrapper and returns the measured asset
+  delta. Backs `redeem`.
+
+These views read cost/realizable amounts straight from the source's own
+EIP-4626 `previewWithdraw`/`previewRedeem`, so they carry the same
+standard-source assumption as the rest of the adapter: a non-compliant or
+fee-on-transfer source is unsupported and needs a dedicated adapter.
+
 ## Functions
 
 ```solidity
@@ -35,6 +59,15 @@ function deposit(address _asset, address _underlying, uint256 _assets) external 
 
 /// Redeem `_assets` from the yield source; returns the asset amount received.
 function withdraw(address _asset, address _underlying, uint256 _assets) external returns (uint256 withdrawn)
+
+/// Position value the source consumes to deliver exactly `_assets` out (exact-out cost).
+function previewWithdrawCost(address _underlying, uint256 _assets) external view returns (uint256 cost)
+
+/// Assets delivered if `_holder` realizes up to `_assets` of position value now (exact-in preview).
+function previewWithdrawUpTo(address _underlying, address _holder, uint256 _assets) external view returns (uint256 delivered)
+
+/// Realizes up to `_assets` of position value from the source; returns the measured asset delta.
+function withdrawUpTo(address _asset, address _underlying, uint256 _assets) external returns (uint256 withdrawn)
 ```
 
 ## Related contracts
--- a/docs/VaultWrapper/ERC4626Adapter.md
+++ b/docs/VaultWrapper/ERC4626Adapter.md
@@ -38,6 +38,12 @@ assuming anything about the source beyond its own standard EIP-4626 previews:
 - `withdrawUpTo` — the exact-in execution. DELEGATECALL only; realizes up to
   `assets` of position value into the wrapper and returns the measured asset
   delta. Backs `redeem`.
+- `maxWithdrawableValue` — the source-liquidity signal: the floor valuation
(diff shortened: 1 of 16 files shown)


gvladika:
Closing in favor of #2214, which carries only the source-liquidity-aware `maxRedeem`/`maxWithdraw` half of this PR (originally #2206, stacked on this branch).

The cost-aware exit machinery (`previewWithdrawCost`, `previewWithdrawUpTo`/`withdrawUpTo`, exact-in loss-tolerant `redeem`) is deliberately dropped: its benefit only materializes on sources charging exit fees, which the `ERC4626Adapter` declares unsupported anyway, and comparable products (e.g. Kiln's DeFi Vault) ship without it by scoping to fee-less sources. Exits stay exact-out with the `AdapterWithdrawShortfall` guard as the tripwire if a fee source is ever onboarded by mistake. Cost-awareness can be added in a future release — instances are beacon-upgradeable and a fee-charging source needs a dedicated adapter regardless. The tradeoff is documented in `docs/VaultWrapper/LiFiVaultWrapper.md` and `ERC4626Adapter.md` on the new PR.

PR #2201, opened 2026-08-17
fix(deploy): detect facet removals that never executed (EXSC-774)

0xDEnYO:

[EXSC-774](https://linear.app/lifi-linear/issue/EXSC-774/detect-facet-removals-that-never-executed-false-executed-parked-tasks)

A facet removal that never executed could be recorded as done and then stay invisible indefinitely. On **worldchain** the parked `AcrossFacetV3` task was marked `executed` on 2026-08-03 while `AcrossFacetV3` at `0x08F7800449ad6681bd607EF21d3cc9C9dDDaF1C8` is still routed by the diamond today — 18 days later. Neither existing safety net could see it:

- `no-unexpected-facets` compares on-chain facets against `deployments/<net>.json` and warns only on addresses *absent* from that log. Deprecated facets are never pruned from the deploy log, so a deprecated-but-logged facet reads as "known" and passes. That invariant answers *"is this facet known?"*, not *"should this facet still be here?"*.
- `reconcile-parked-tasks` only reconciled `queued`/`proposed` tasks. `executed` was terminal and never re-verified, so a false-`executed` was permanent by construction.

## The root cause is not what the ticket first assumed

The worldchain task's stored `facetAddress` is `0xB5dD83183fD7CCF859b227CA83663a034d5B2f92` — **lisk's** `AcrossFacetV3` v1.0.0, not a worldchain address at all. That address genuinely is not routed on worldchain, so the reconcile's address-based presence check correctly concluded "gone" **while checking the wrong address entirely**. The facet actually routed under the name `AcrossFacetV3` on worldchain was never examined.

So re-verifying the *stored address* — the fix as originally proposed — would **not** have caught this. Proven against live data:

```text
task: status=executed storedFacetAddress=0xB5dD83183fD7CCF859b227CA83663a034d5B2f92
address-only presence : false   → decision: keep      ← the proposed fix misses it
name-primary presence : true    → decision: reopen    ← this PR catches it
routed under the name AcrossFacetV3: 0x08F7800449ad6681bd607EF21d3cc9C9dDDaF1C8
```

Presence therefore resolves by facet **name** first, with the stored address kept only as a fallback (so a pruned deploy-log entry still can't false-resolve a live facet). This also aligns the reconcile with the drain, which already removes by name via `computeNamedFacetRemovals` — otherwise the two disagree about what "done" means.

## What changed

**1. New `no-deprecated-facets` health-check invariant.** Diffs the on-chain facet set against `_targetState.json` and warns on facets that are routed, name-resolvable in the deploy log, unprotected, and whose `.sol` source no longer exists (i.e. genuinely deprecated). Reuses `diffFacets` — the same diff `cleanUpProdDiamond --auto` computes, which was never wired to a scheduled alarm. Severity `warning`, so it surfaces without failing bring-up on partially-deployed networks.

Deliberate scope choices:

- Target-state **drift** (facet live, source still present) is excluded. It is a different failure class and far noisier — the probe found it on ~40 networks, mostly benign.
- Findings are aggregated into **one warning per network** rather than one per facet, because the fleet-wide backlog is large (see below) and per-facet lines would drown the report.
- `activeSelectors` is passed empty: this is detection only, so nothing needs holding back, and populating it requires compiled artifacts and *throws* when they are stale — unacceptable inside a health check. The real removal path still computes the true held-back set.

**2. Name-primary presence + terminal re-verification in `reconcile-parked-tasks`.** `executed`/`superseded` are re-verified against the loupe on every run; a task whose facet is still routed is reopened to `queued` (so the next drain re-proposes it) and alerted to `#dev-sc-multisig-proposals`. `superseded` is included because it makes the identical "we believe it's gone" claim. `cancelled` is deliberately excluded — it records an operator's decision, not a claim about on-chain state.

**3. Per-network isolation in the reconcile sweep** — this one is load-bearing, not incidental. The fleet sweep was **aborting** at `harmony`, a retired network that kept its parked tasks but left `networks.json`, so the loupe read threw and killed the run. Every network ordered after it was never reconciled — **worldchain included**. Without this fix both changes above would have been inert in the scheduled job for the very network that motivated the ticket. Skipped networks are now alerted with a remedy rather than silently dropped.

**4. Alert delivery is no longer silently optional.** All three reconcile alerts degrade to log-only when `WEBHOOK_DEV_SC_MULTISIG_PROPOSALS` is unset (pre-existing behaviour for the TTL alert, inherited by the new ones). The job now says so explicitly rather than degrading in silence — the same failure class this PR closes.

## Fleet-wide scope of the underlying problem

worldchain is one instance of a systemic backlog, not an isolated incident. A probe of the loupe ∖ target-state diff across all active production networks found **~50 networks with at least one deprecated-but-live facet** (~100 findings), dominated by `GenericSwapFacet` and `CBridgeFacet`/`CBridgeFacetPacked`. The new invariant will be loud on its first run; that volume is a real cleanup backlog, and draining it is tracked separately (see follow-ups).

## Verification: proven to fire on real data, not just in unit tests

Same-PR unit tests are not evidence a new check works, so each was falsified against the live queue and live chains (read-only; **no Safe proposals created**).

*Invariant fires, and discriminates:*

```text
severity=warning
worldchain: 19 facets routed
worldchain warnings: ["1 deprecated facet(s) still routed by the diamond: AcrossFacetV3 (0x08F7800449ad6681bd607EF21d3cc9C9dDDaF1C8)"]
✔ POSITIVE (worldchain names AcrossFacetV3): PASS
✔ warning-severity only (no errors logged): PASS
✔ NEGATIVE (jovay quiet, 13 facets): PASS
✔ NEGATIVE (pharos quiet, 14 facets): PASS
```

*Reconcile fires on the real false-`executed`, and discriminates within the same network* — all three worldchain tasks had an executed proposal, only the one still routed reopened:

```text
[worldchain:production] GenericSwapFacet (proposed)     → executed
[worldchain:production] AcrossFacetV3 (executed)        → reopen
[worldchain:production] AcrossFacetPackedV3 (proposed)  → executed
🚨 1 deferred diamond-cleanup task(s) were marked done but their facet is STILL ROUTED — re-queued for removal
```

*Fleet sweep now completes instead of aborting, with no false-reopen storm* — 123 tasks decided (previously the run died after ~47), exactly 1 reopen across the whole fleet, and the 3 unreconcilable retired networks reported rather than fatal:

```text
exit=0
  53 → executed
  69 → keep
   1 → reopen        ← worldchain AcrossFacetV3, the only one
⚠️ 3 network(s) could not be reconciled — harmony, okx, velas (left networks.json)
```

## Operational impact a reviewer should weigh

**This changes the daily healthcheck's Slack behaviour.** `healthCheckAllNetworks.yml` posts when `warned_count != 0`, and its own comment states that a green run "stays silent to keep the channel signal-only." Because the invariant is `warning` severity, the ~50 affected networks all land in `warned` — so the daily job will post every day, listing them, until the backlog is drained (weeks of Safe proposals across ~50 chains).

Two readings, both defensible:

- **Land as-is.** The alert is true, it self-clears per network as each is drained, and it gives the cleanup visible progress. Risk: sustained ~50-network daily posts invite the team to mute the channel, which would be worse than not having the check.
- **Drain first, then land.** Keeps the channel signal-only, at the cost of leaving the detection gap open while the backlog is worked.

Worth noting the channel likely already posts on many days: `facets-registered` warns on any RPC failure, and the fleet probe for this PR hit 404/429/403/revert on arbitrum, bsc, moonbeam and gravity. So this may be incremental rather than a regime change — but I did not measure the current baseline, and that is the number that decides it.

**Coverage of the new invariant is 69/72 active networks.** The 3 skips — `arbitrumsepolia`, `basesepolia`, `tronshasta` — have no production target-state `LiFiDiamond` block, which is correct for testnets. This is deliberately the inverse of the failure that got `no-unexpected-erc20proxy-callers` deleted in #2156 (silently no-oped on 38 of ~62 networks while the sweep still reported green): here the skip set is small, is testnet-only, and is logged per network.

## Follow-ups (deliberately out of scope)

1. The enqueue-time bug that wrote a **lisk** address onto a **worldchain** parked task.
2. `deployments/worldchain.json` and `_deployments_log_file.json` disagree on `AcrossFacetV3`'s address (flat log `0x08F7…`; master log `0xAd99…` v1.0.0 / `0x5052fc…` v1.1.0).
3. Draining the ~100-item deprecated-but-live backlog.
4. Cancelling the parked tasks for retired networks (harmony, okx, velas) so that alert clears.
5. Target-state drift, e.g. `AcrossFacetV4`/`AcrossFacetPackedV4` live on linea but absent from its target state.
6. Measure the current `warned_count != 0` rate of the daily sweep, to settle whether this PR's Slack impact is incremental or a regime change.

### DIFF
--- a/docs/DeferredDiamondCleanupQueue.md
+++ b/docs/DeferredDiamondCleanupQueue.md
@@ -586,12 +586,20 @@ origin-PR lines to the signer.
         │                    │
         │       linked proposal executed + loupe confirms facet absent
         │                    ▼
-        │                executed  (terminal, = done)
-        │
-        └── operator CLI (deprecation reverted / obsolete) ─► cancelled (terminal)
+        │       executed / superseded  (terminal, = done)
+        │                    │
+        └────────────────────┘ reconcile: facet NAME still routed → reopen (EXSC-774)
+
+             operator CLI (deprecation reverted / obsolete) ─► cancelled (terminal)
 ```
 
-All five transitions ship as helpers in `parked-tasks.ts` (#2051, Fact 15):
+`executed`/`superseded` are terminal but **not trusted**: every reconcile re-verifies
+them against the loupe, because a removal recorded as done that never actually landed
+was otherwise never re-checked and stayed invisible indefinitely (EXSC-774 — worldchain's
+`AcrossFacetV3` sat live for 18 days behind an `executed` record). `cancelled` is the only
+truly final state.
+
+All six transitions ship as helpers in `parked-tasks.ts` (#2051, Fact 15):
 
 - **queued → proposed**: the drain, via the atomic `claimForProposal(parkedTasks,
   taskKey)` (`:324`) filtered on `status:'queued'` (§6 step 3). This is the dedup gate
@@ -608,6 +616,11 @@ All five transitions ship as helpers in `parked-tasks.ts` (#2051, Fact 15):
 - **queued/proposed → superseded**: `markSuperseded` (`:360`, accepts both open states)
   — the facet is already absent on-chain (removed via another route); self-healing
   reconcile.
+- **executed/superseded → queued**: `reopenResolvedTask` — the reconcile finds the facet
+  still routed despite a terminal status, so the removal never landed. Clears
+  `resolvedAt`/`proposedAt`/`safeTxHash` and alerts the multisig-proposals channel.
+  `cancelled` is deliberately excluded (it records an operator's decision, not a claim
+  about on-chain state).
 - **→ cancelled**: `markCancelled` (`:383`) — an operator explicitly abandons the intent
   (deprecation reverted, facet re-added, or a protected facet queued in error). **Merged
   behaviour: restricted to `queued`** — cancelling a `proposed` task would orphan its
@@ -723,7 +736,8 @@ only the PR-link surfacing (§6) and drops the manual `--auto` invocation.
 | No double-enqueue / no double-propose | Partial unique index `unique_open_task_key` on `taskKey`; the atomic `claimForProposal` flip — independent of the salt-nondeterministic `intentHash` (Facts 8, 9, 15; §7). |
 | Never park/remove a protected facet | Enqueue and drain both call `getProtectedNames()` (`diamondRemovalDiff.ts:119`); a queued protected facet is `cancelled` + alerted (§6). Inherits every #2047 guardrail (drift gate is N/A — named path). |
 | Deferred ≠ orphaned | Cold-network backstops: `--auto --all-networks` sweep + TTL Slack alert + observability CLI (§8). No silent truncation — the TTL alert names what's still queued. |
-| Deploy-log longevity | Address snapshot + loupe-by-address check so pruning the log doesn't false-`superseded` a live facet; strengthened `/deprecate-contract` warning (§8). |
+| Deploy-log longevity | Address snapshot as a *fallback* so pruning the log doesn't false-`superseded` a live facet; strengthened `/deprecate-contract` warning (§8). Presence is resolved by facet **name** first — a snapshot address can be flat wrong, and an address-only check then reports "gone" about an address nobody asked about (EXSC-774). |
+| A resolution can be wrong | `executed`/`superseded` are re-verified against the loupe on every reconcile and reopened when the facet is still routed (§7). A stored terminal status is never taken as proof. |
 | Opt-in in v1 | `DRAIN_PARKED_TASKS` **default off; ON for rollouts, OFF for emergencies** (§6) — an urgent pause/break-glass proposal never drags unrelated removals into its signing set; reentrancy-guarded (§6). |
 | Direct-send safety | Drain no-ops on staging/testnet/`SEND_PROPOSALS_DIRECTLY_TO_DIAMOND` (Fact 13; §12). |
 | Rule compliance | TS/Bash, no Python (`000:15`); viem (`200:14`); reuse helpers (`:24`); new helpers 100%-covered colocated tests (`:120`); `citty`/`consola`/`getEnvVar` CLI (`:116`); `I`-prefixed interfaces; injectable I/O + dry-run-default per #2047 convention (Fact 14). |
--- a/script/deploy/healthCheckInvariants.test.ts
+++ b/script/deploy/healthCheckInvariants.test.ts
@@ -4,14 +4,17 @@ import {
   it,
   // eslint-disable-next-line import/no-unresolved
 } from 'bun:test'
+import { type Hex } from 'viem'
 
 import globalConfig from '../../config/global.json'
 import networksConfig from '../../config/networks.json'
+import { EnvironmentEnum } from '../common/types'
 
 import {
   CORE_FACET_EXEMPTIONS,
   HEALTH_CHECK_EXCLUSIONS,
   HEALTH_CHECK_INVARIANTS,
+  findDeprecatedLiveFacets,
   findDuplicateSelectors,
   getExemptCoreFacets,
   getExpectedPairs,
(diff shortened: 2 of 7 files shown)


0xDEnYO:
Coordination note: PR #2157 (EXSC-723, ready for review) now also ships per-network fault isolation in `reconcileAll` (pushed 2026-08-17 after the same 4/4 cron-failure investigation) plus queue remediation: 54 stale `proposed`→`executed` (verified loupe + signing store), mode superseded, harmony/okx/velas cancelled. On merge order: if #2157 lands first, rebase and drop the duplicate isolation hunk here — the false-executed detection in this PR remains its unique value.

0xDEnYO:
Folded into #2157 (commit 2c0536b53 merges this branch verbatim; all EXSC-774 work — reopen detection, name-primary presence, per-network isolation — ships there, verified on live data: the reopen pass caught a real false-supersession on mode). Single combined PR per Daniel.

0xDEnYO:
Addressed CodeRabbit's merge-risk note ("skipped networks may still be reported only in logs when alerting is not configured") in f26ee242d.

The three reconcile alerts all degrade to log-only when `WEBHOOK_DEV_SC_MULTISIG_PROPOSALS` is unset — pre-existing behaviour for the TTL alert, which the new alerts inherited. Rather than change the optional-webhook contract, the job now says so explicitly:

```text
WARN  ⚠️ 1 network(s) could not be reconciled — their parked tasks were NOT verified this run:
   - harmony:production — Chain harmony does not exist. …
   → transient RPC/config problem: no action needed. Retired network: cancel its parked tasks …

WARN  WEBHOOK_DEV_SC_MULTISIG_PROPOSALS is not set — the alert(s) above were logged only, not delivered to Slack.
```

Silent degradation is the exact failure class this PR exists to close, so it seemed wrong to leave it implicit.

Verified against the live queue with `--network harmony --yes`, which mutates nothing: a retired network is skipped before any task transition is reached, so the run exercises the failure-alert and log-only-warning paths without writing to the queue or touching a chain.

0xDEnYO:
Reopened in error by me a few minutes ago — I was chasing what looked like a stale PR head (pushes to a closed PR don't advance it) and hadn't yet read the fold-into-#2157 note above. Re-closing to restore the intended state. Apologies for the noise.

For the record, one commit here is **not** in #2157: `f26ee242d` "surface reconcile alerts that were logged but not delivered". It addresses CodeRabbit's merge-risk note on this PR — the reconcile alerts degrade to log-only when `WEBHOOK_DEV_SC_MULTISIG_PROPOSALS` is unset, and the job now says so instead of degrading silently. If #2157 doesn't already cover that, it wants porting there rather than being lost with this PR.

Two findings from this branch that may be worth carrying into #2157's description as well:

1. **The new stale-facet invariant changes the daily healthcheck's Slack behaviour.** `healthCheckAllNetworks.yml` posts whenever `warned_count != 0`, and its own comment notes that a green run stays silent "to keep the channel signal-only." At `warning` severity the ~50 affected networks all land in `warned`, so the daily job will post every day until the backlog drains. Unmeasured: the current baseline rate of `warned_count != 0` (RPC failures already trip `facets-registered`), which is the number that decides whether this is incremental or a regime change.
2. **Invariant coverage is 69/72 active networks.** The 3 skips — `arbitrumsepolia`, `basesepolia`, `tronshasta` — have no production target-state `LiFiDiamond` block, correct for testnets. Deliberately the inverse of the failure that got `no-unexpected-erc20proxy-callers` removed in #2156 (silent no-op on 38 of ~62 while the sweep reported green).

0xDEnYO:
Correction to my previous comment — I checked #2157 afterwards and all three carry-over items are already handled there, better. Nothing needs porting:

1. **Log-only alerts:** #2157 already distinguishes an unset webhook from a local run (`WEBHOOK_DEV_SC_GITHUB_CI_NOTIFICATIONS is unset — alert logged only.` / `Local run: reconcile alerts logged only. Set CI=1 …`). `f26ee242d` is redundant; it can die with this PR.
2. **Daily-Slack impact:** solved by design. `no-stale-registered-facets` only warns on a deprecated routed facet with **no open parked task**, so the queued backlog logs as expected-pending instead of inflating `warned_count`. That removes the alert-flood risk my `no-deprecated-facets` had, which is the better fix.
3. **Testnet skips:** #2157 scopes the invariant `skipTestnet` deliberately, rather than skipping implicitly on a missing target-state block.

So this branch has no unique value left. Closed correctly.

PR #2203, opened 2026-08-17
fix(deferred-cleanup): abandon parked tasks on deprecated networks; isolate reconcile failures (EXSC-780)

0xDEnYO:

Fixes [EXSC-780](https://linear.app/lifi-linear/issue/EXSC-780/parked-queue-deprecating-a-network-must-cancel-its-open-tasks)

The weekly `reconcile-parked-tasks` cron has failed **every run since it shipped** in #2076: run #1 on the Mongo grant (fixed via DO-723), then runs #2–#4 on `Chain <x> does not exist ... in 'config/networks.json'` — `evmos` twice, then `harmony` ([run #4](https://github.com/lifinance/contracts/actions/runs/32010652627)). The job aborts on the first unresolvable network, so the TTL alert — the cold-network backstop the cron exists for — has **never fired**.

## 1. One bad network must not take out the fleet

The `try` in `reconcileAll` existed only to close the Mongo client in `finally` — there was no `catch`. Run #4 decided 27 tasks and dropped everything after `harmony`, TTL alert included. Now each `(network, environment)` group, the optional proposal-store connection, each cancellation, and `reconcileAll` itself are isolated: failures are counted and logged, the run continues, the TTL alert always fires, and the process exits non-zero **afterwards** so a real failure still reaches Slack. `drain-parked-tasks.ts` was already defensive per item — reconcile was the outlier.

A network whose deploy log has no LiFiDiamond stays a *legitimate skip*, warned about and not counted.

## 2. A task on a non-active network is reported; cancelling it is opt-in

Such a task can't be reconciled — `getViemChainForNetworkName` throws for an absent key, and no `ETH_NODE_URI_<NETWORK>` exists for an inactive one — so it is routed out of the on-chain path before the loupe is touched.

**It is not auto-cancelled, and this is the important design point.** I originally had the `--yes` cron cancel these, and the review gate falsified it: `config/networks.json` is *not* a deprecation signal in this repo. It is deliberately narrowed to three networks for emergency-pause rehearsals — `f99db1607` (2026-05-27, reverted next day) and `216bad0e4` (2026-06-15, **a Monday**, reverted three days later) — and `status` is toggled back to `active` (`51a04fc64`). Both commits *delete* entries, so "absent" and "inactive" are both unreliable.

Replaying `216bad0e4`'s config against the live queue, with the exact flags CI uses (`--yes`, no `--network`):

```text
active set: boba, rootstock, sophon
open tasks: 67  ->  live: 2  |  routed to the deprecated path: 65
```

The Monday cron would have terminally cancelled 65 of 67 queued removals during a pause rehearsal. `cancelled` is terminal, there is no undo, and re-enqueue needs origin-PR context.

So cancellation now requires a human naming one network: `--network <x> --cancel-deprecated --yes`. Running `--cancel-deprecated` fleet-wide is refused outright, and the unattended cron only ever reports. That same command is what `/deprecate-network` step 7 now runs — the step previously told operators to "cancel each one" with **no executable path at all**, since no CLI exposes `markCancelled` (EXSC-715).

A `proposed` task is never cancelled on either path: it carries a live Safe proposal and `markCancelled` is restricted to `queued`, so cancelling would cut it from its origin-PR linkage (spec §6/§7). It needs `revertToQueued` first, which has no CLI yet — so it is surfaced, and the doc says so instead of implying a fix exists.

Deliberately **not** re-adding the enqueue-side guard: that approach was closed with #2192 in favour of #2202's deploy-log removal. This is the case #2202 cannot cover — park while the network is active, deprecate it later — and the two don't overlap at file or behaviour level.

# Verification

10 new unit tests over the pure decisions, including the gate that makes a fleet-wide cancel impossible. The live adapter is unit-test exempt per this file's carve-out, so I exercised it against the real queue in dry-run:

| Scenario | Result |
|---|---|
| `216bad0e4`'s 3-network config + `--yes` (the blast radius) | 65 tasks routed to the deprecated path, **0 cancelled**, exit 0 |
| `--cancel-deprecated --yes` without `--network` | refused, exit 1, nothing touched |
| `zksync` set to `status: inactive`, single-network opt-in | routed and reported with the exact remediation command |
| `ETH_NODE_URI_METIS` blanked | group logged and skipped, **66 other tasks still decided**, TTL alert ran, exit 1 |
| `--ttlDays soon` | rejected, exit 1 (previously `NaN`, which flags *every* open task as stale) |
| unmodified config | all 67 open tasks decided, TTL alert ran, exit 0 |

The queue was byte-identical before and after every run (`queued 67 / executed 55 / cancelled 4 / superseded 1`) — no `--yes` write path was exercised against production, on purpose.


0xDEnYO:
## Review-gate residual — items for a human decision

The gate's falsification pass killed this PR's original design (an auto-cancelling cron); that is fixed in 72780d4ef and the PR body now describes the shipped behaviour. Two things remain that I did **not** decide unilaterally:

**1. Should the cron ever auto-cancel? (design call)**
Today it only reports, and cancelling is `--network <x> --cancel-deprecated --yes`. That is the safe default given `networks.json` is narrowed for pause rehearsals. If we want the queue to self-heal unattended, the honest prerequisite is a durable deprecation signal — an explicit deprecated-networks list, or a `deprecatedAt` field — rather than inferring it from `status`/absence. Worth its own ticket if wanted.

**2. A `proposed` task on a non-active network has no remediation path.**
It is reported every run and never resolvable, because `revertToQueued` is library-only — no CLI exposes it (**EXSC-715**, still Backlog). The queue has **0** `proposed` tasks today, so this is latent, not live. Left as-is rather than widening this PR into the operator-CLI work.

## CodeRabbit findings — all six triaged

| Finding | Action |
|---|---|
| `deprecate-network.md` — only `queued` should be cancelled, state it consistently (Major) | Fixed — Overview, How-to-Use, step 7 and Key-Files now all say `queued`, with `proposed` explicitly reported-not-cancelled |
| `reconcile-parked-tasks.ts` — TTL alert skipped when reconcile setup rejects (Major) | Fixed in 91e831db5 — `reconcileAll` is wrapped, the alert always runs, the failure still fails the job. This was the PR's own thesis violated in its own code |
| `safe-utils.ts` — proposal-store client leaks if `ensurePendingProposalIndex` fails post-connect (Major) | Fixed in 91e831db5 — closes before rethrowing, matching the connect path two lines above. Newly reachable because the reconcile no longer dies on that throw, so it could have hung the job |
| `docs` §7 — rationale for inactive networks contradicted itself (Minor) | Fixed — the mechanical reason (unresolvable chain / no RPC var) is now separated from the claim that inactive ≠ deprecated |
| `--ttlDays` accepts `NaN`/negative/fractional (Minor) | Fixed in 91e831db5 — `NaN` made `ageDays < ttlDays` false for every task, so a typo would have posted the whole open queue to Slack as stale. Pre-existing; fixed because it is one flag away from the alert this PR is protecting |
| `deprecate-network.md:110` (Minor, empty body) | No actionable content |

Verification for all of the above is in the PR body's table; the parked-tasks queue was byte-identical before and after every run.

0xDEnYO:
@coderabbitai review

Three commits landed after your first pass (72780d4ef, 91e831db5) — the design changed materially: the cron no longer cancels anything, and all six of your findings are addressed. Please re-review the fix commits; your earlier round was rate-limited.

0xDEnYO:
Folded into #2157 — closing this in favour of the combined PR, same as #2201.

The EXSC-780 work is now on `feature/exsc-723-parked-queue-drain-hardening` as `de34c7418` (merge) + `783e26a68` (gate findings on the resolution). The conflict resolution takes #2201's reconcile as the base — `IReconcileRun {reopened, failures}`, name-primary presence, safe-to-prune, loupe-only degrade — and drops this branch's duplicate isolation and `process.exitCode` approach in favour of it, since per-network Slack alerts carry strictly more information than a bare non-zero exit.

What carried over: the opt-in single-network `--cancel-deprecated`, the fleet-wide refusal guard, the outer sweep guard so the TTL backstop survives a throw, `--ttlDays` validation, and the `getSafeMongoCollection` client close.

Review the combined change in #2157.

PR #2205, opened 2026-08-17
ci(clear-signing): normalize registry sync with erc7730 format

gvladika:

Fixes [EXSC-781](https://linear.app/lifi-linear/issue/EXSC-781/wire-erc7730-format-into-clear-signing-registry-sync-to-eliminate)

The clear-signing registry stores `calldata-LIFIDiamond.json` in the compact shape produced by the ERC-7730 registry's own formatter (`erc7730 format`, run by its `format.yml` bot), but our sync generator writes fully-expanded `JSON.stringify(data, null, 2)`. The two disagree on every line, so each sync re-expands the registry's file and their bot re-compacts it — the file oscillates between formats and every sync PR carries a ~20k-line formatting diff that buries the real ABI/deployment change (e.g. the CBridge removal in registry PR #2901 was ~250 real lines out of ~20,700). Rather than reimplement the registry's serializer in TypeScript (a moving target that would drift the moment they change it), the sync now runs the registry's exact formatter on the generated descriptor before diffing/committing, so our output already matches the canonical shape. `erc7730` pins `Requires-Python >=3.12,<3.13`, so Python 3.12 is pinned explicitly instead of trusting the runner default. Verified locally under 3.12: feeding our generator output through `erc7730 format` reproduces the registry's format exactly, dropping the raw diff against the committed file from ~20,700 to 250 lines, all of which are the genuine change. The first sync PR after this merges is still a one-time reformat (adopting their shape as our baseline); every sync after is clean.

### DIFF
--- a/.github/workflows/syncLedgerClearSigning.yml
+++ b/.github/workflows/syncLedgerClearSigning.yml
@@ -1,7 +1,7 @@
 # Sync Clear Signing (ERC-7730) Registry
 # - Purpose: Keep `registry/lifi/calldata-LIFIDiamond.json` in the Ethereum Foundation-hosted ERC-7730 registry up to date with this repo's Diamond ABI + deployments AND with the `display.formats` proposal at `config/clearSigningProposal.json`.
 # - Triggers: push-to-main on ABI/deployment/proposal-relevant paths (fast path), a monthly cron safety net, and manual workflow_dispatch (with optional dry_run).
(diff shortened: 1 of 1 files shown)


gvladika:
@coderabbitai review

0xDEnYO (approved):

maxklenk (approved):

PR #2206, opened 2026-08-17
feat(VaultWrapper): source-liquidity-aware exit views (finding #4)

gvladika:

**Stacked on #2197** (`fix/vault-wrapper-exit-cost-awareness`) — this PR targets that branch and should be reviewed/merged after it. Once #2197 merges to `dev-vault-wrapper`, rebase this onto `dev-vault-wrapper`.

This is finding #4 from the exit-cost PR (#2197): the wrapper's `maxWithdraw`/`maxRedeem` were fee/loss-aware but not **source-liquidity**-aware. When a yield source can only release part of the wrapper's position right now (an Aave-style utilization cap, a paused source, an illiquid Morpho market), the views over-reported and a `redeem`/`withdraw` within the reported max could revert — breaking the EIP-4626 guarantee that an exit within `maxRedeem` never reverts.

- Added one static adapter view `IYieldAdapter.maxWithdrawableValue(underlying, holder)` (implemented in `ERC4626Adapter`): the source's floor valuation (`convertToAssets`) of `min(holder position, source.maxRedeem)`. It is the **gross** valuation (deliberately not the fee-netted `previewRedeem` the sibling `previewWithdrawUpTo` uses), because the wrapper feeds it straight into its own gross-denominated share math to clamp the views; any source exit fee is applied separately by the realizable redeem path.
- `LiFiVaultWrapper.maxRedeem` now clamps the owner's balance to `_convertToShares(maxWithdrawableValue, Floor)` — but only when the source is genuinely liquidity-limited. A full-position short-circuit (`realizable >= totalAssets()`) returns the plain balance on fully-liquid sources, avoiding the ~1-share under-report the virtual-offset share round-trip would otherwise introduce. `maxWithdraw` inherits the clamp because OZ derives it as `previewRedeem(maxRedeem(owner))`, so the `withdraw` entrypoint's `ERC4626ExceededMaxWithdraw` guard is liquidity-aware too.
- `previewRedeem`/`previewWithdraw` stay liquidity-agnostic (EIP-4626 requires previews to ignore withdrawal limits). Exit execution paths are unchanged — they already fail on a short source; finding #4 was about the views.

Accepted tradeoff (documented): on OZ-style sources whose own `maxRedeem` floors ~1 share below `balanceOf` even at full liquidity (e.g. MetaMorpho), the clamp lands a sub-`1e-12`-token dust below `balanceOf`, so a one-call `redeem(balanceOf)` can hit `ERC4626ExceededMaxRedeem`. Callers should exit via `redeem(maxRedeem(owner))` (the EIP-4626-recommended pattern); a follow-up call clears the dust. `redeem` remains the guaranteed exit. The exact-out `withdraw` share-rounding boundary near `maxWithdraw` is unchanged and remains the one tolerated artifact.

Scope is the exit **views** only. `@custom:version` is held at `1.0.0` (subsystem is pre-release). `script/deploy/healthCheckInvariants.ts` was reviewed: no change needed (this adds an interface method + internal view logic only — no deployed-contract topology, constructor, selector, or storage change; the Vault Wrapper is outside the Diamond the health-check covers).

Testing: added `MockLiquidityCappedERC4626` (models an Aave-style withdrawal cap — assets stay on the vault, only a policy-limited amount may leave), adapter and wrapper unit tests, and a third stateful invariant suite that fuzzes the source's liquidity cap up and down while a `redeem` bounded to `maxRedeem` runs under `fail-on-revert` (0 reverts; ~80% of executed redeems occur while the clamp is actively binding). The Morpho Arbitrum fork scenario was updated to exit via `redeem(maxRedeem)` and to assert the drain residue is dust (`convertToAssets(totalSupply()) <= 10 wei`) instead of exactly 0. Full subsystem suite: 349 tests passing.

### DIFF
--- a/docs/VaultWrapper/ERC4626Adapter.md
+++ b/docs/VaultWrapper/ERC4626Adapter.md
@@ -38,6 +38,12 @@ assuming anything about the source beyond its own standard EIP-4626 previews:
 - `withdrawUpTo` — the exact-in execution. DELEGATECALL only; realizes up to
   `assets` of position value into the wrapper and returns the measured asset
   delta. Backs `redeem`.
+- `maxWithdrawableValue` — the source-liquidity signal: the floor valuation
+  (`convertToAssets`) of the GROSS worth of `min(holder position, source.maxRedeem)`.
+  Gross by design (not `previewRedeem`) — it does NOT net the source's exit fee, because
+  the wrapper feeds it into its own gross share math to clamp its `max*` views to what the
+  source can currently honor. Equals the full position value on a fully-liquid source (one
+  with no active liquidity limit).
 
 These views read cost/realizable amounts straight from the source's own
 EIP-4626 `previewWithdraw`/`previewRedeem`, so they carry the same
@@ -67,6 +73,9 @@ function previewWithdrawUpTo(address _underlying, address _holder, uint256 _asse
 
 /// Realizes up to `_assets` of position value from the source; returns the measured asset delta.
 function withdrawUpTo(address _asset, address _underlying, uint256 _assets) external returns (uint256 withdrawn)
+
+/// Gross floor valuation of `min(holder position, source.maxRedeem)` — the source-liquidity signal (not exit-fee-netted).
+function maxWithdrawableValue(address _underlying, address _holder) external view returns (uint256 assets)
 ```
 
 ## Related contracts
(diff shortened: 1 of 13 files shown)


gvladika:
@coderabbitai review

PR #1002, opened 2025-02-18
Deprecate stargate v1 and amarok

0xDEnYO:
# Which Jira task belongs to this PR?
https://lifi.atlassian.net/browse/LF-12447

# Why did I implement it this way?

# Checklist before requesting a review

- [x] I have performed a self-review of my code
- [x] This pull request is as small as possible and only tackles one problem
- [ ] I have added tests that cover the functionality / test the bug
- [ ] For new facets: I have checked all points from this list: https://www.notion.so/lifi/New-Facet-Contract-Checklist-157f0ff14ac78095a2b8f999d655622e
- [ ] I have updated any required documentation

# Checklist for reviewer (DO NOT DEPLOY and contracts BEFORE CHECKING THIS!!!)

- [ ] I have checked that any arbitrary calls to external contracts are validated and or restricted
- [ ] I have checked that any privileged calls (i.e. storage modifications) are validated and or restricted
- [ ] I have ensured that any new contracts have had AT A MINIMUM 1 preliminary audit conducted on <date> by <company/auditor>

PR #2212, opened 2026-08-18
feat(healthcheck): periphery chain→log parity, registry-first receivers, selector-set facet identity (EXSC-783)

0xDEnYO:

Fixes [EXSC-783](https://linear.app/lifi-linear/issue/EXSC-783) — subticket T1 of EXSC-782.

Extracted and hardened from the reference implementation on `claude/busy-solomon-53e1c2` (draft PR #2125). Deliberately excludes its immutable-bindings work (EXSC-784), deploy-time reminders (EXSC-785) and `skipHealthcheck` removal (EXSC-786) — those ship separately.

Facets have had two-direction log↔chain coverage for a while (`facets-registered` for log→chain, `no-unexpected-facets` for chain→log). Periphery had only one direction, and receivers had a hole big enough that a live `ReceiverOIF` sat unlogged on mainnet without a single check noticing. This closes all three gaps.

## 1. Chain→log parity: a new `periphery-registry-log-sync` invariant, not a fold

The ticket preferred folding the reverse assertion into `periphery-registered`. I went with a separate invariant, because the fold **does** muddy it, on one decisive point: **an invariant carries exactly one `severity`.**

`periphery-registered` is `error` — a contract we deployed and did not register is a broken diamond. The reverse direction has to be `warning`: a deploy log that lags the chain is a bookkeeping failure, not a broken diamond, and it is common enough that an error gate would turn the whole fleet sweep red over drift nobody can fix in the same change. Folding forces one of the two directions to the wrong severity. The separate invariant also gets its own `remediation` string and its own exclusion entries, both of which are per-invariant.

This also makes it the exact analog of `no-unexpected-facets`, which is a `warning` for the same reason.

Two scope calls inside it:

- **The flat deploy log is the only comparison target**, but *both* logs contribute candidate names. `no-unexpected-facets` reconciles against `deployments/<network>.json`, and that is the log every other invariant resolves through (`ctx.deployedContracts`) — so it is the log whose gaps cause the silent exemptions this ticket is about. Reconciling `<network>.diamond.json` as a second target is a genuinely separate concern (snapshot vs. history semantics) and is left as a follow-up rather than doubling the warning surface. Its *names* are read, though: the review gate caught that `FeeCollector`, `LiFiDEXAggregator` and `ReceiverAcrossV3` are recorded in `<network>.diamond.json` while absent from the flat log on some chains, so a candidate list built from the flat log alone would never probe them — the check would have been blind to exactly the contracts it claims to cover. The description is also "every **known** periphery contract", since a non-enumerable registry can only ever be probed by name.
- **Facet names are not probed.** The registry is a `mapping(string => address)` with no enumerator, so the on-chain side can only be discovered by probing names. The candidate set is core periphery ∪ whitelist periphery ∪ coupling companions ∪ the receiver list ∪ both deploy logs' keys, minus facets and the diamond — probing facet names per network would multiply RPC reads for lookups that can only ever return zero. Excluding by target state alone was not enough: target state names only *current* facets, so retired ones lingering in the flat log were still probed (21% of all probes on a real network). The exclusion is name-based as well, which is safe because no periphery contract is named `*Facet`.

## 2. Registry-first receiver resolution

`receiver-executor-binding` and `receiver-owner` resolved receivers from the deploy log and `continue`d when absent, so a log-absent receiver was silently exempt from the only checks covering it. Both now go through `resolvePeripheryAddress()` — on-chain registry first, deploy log as fallback. A failed registry read falls back to the log rather than aborting: a read failure is not evidence of absence.

`receiver-owner` is **repointed, not just widened.** On `main` it checks exactly one contract — the generic `Receiver`, deleted from `src/Periphery` in #1002 and present in no network's target state. So the invariant guarding receiver ownership was, in practice, guarding only a deprecated contract while the four receivers actually in service had **no owner check at all**. It now checks those four and nothing else.

### Deprecated receivers are out of scope for these checks, deliberately

Review raised two structural gaps here — that the sync candidate list could only find receivers a log already named, and that `ReceiverAcrossV3` was excluded from ownership by reusing a binding-scoped list. Both were real, but the fix is not to check more receivers; it is that neither deprecated receiver belongs in a health check at all:

| | `Receiver` | `ReceiverAcrossV3` |
|---|---|---|
| Source in repo | deleted (#1002, 2025-02-26) | present, but superseded by V4 |
| Networks in target state | 0 | 0 |
| Registered on chain | **35** | **17** |
| In flat deploy log | 36 | 18 |
| In diamond deploy log | 0 | **66** |

The candidate-discovery gap resolves on its own: `RECEIVER_EXECUTOR_GETTERS` is exactly the set of coupling companions in `config/global.json`, so every receiver in service already reaches the candidate union twice over and none of them depends on a log naming it first. A static receiver list would have added only the two deprecated names — turning a stale on-chain registration into a permanent warning instead of a thing to clean up. A test now asserts the checked set equals the coupling companions, so neither name can creep back in.

`receiver-owner` also becomes `evm-only`, matching `receiver-executor-binding`. Its Tron branch looked up `Receiver` in `deployments/tron.json`, which has never carried that key, so it could only ever no-op.

Those registrations are left in place: a registry entry nothing routes through is harmless, and unwinding it across 35 + 17 chains would cost a Safe wave for no functional gain. The point of excluding the names here is that the health check should not assert anything about them, not that the chain state needs changing.

Two supporting changes fall directly out of that:

- **A run-wide registry read cache** (`ctx.peripheryRegistryCache`). Four invariants now probe overlapping name sets; uncached, that multiplies the RPC reads per network and feeds the rate limits that degrade other checks. The promise is cached before it settles so concurrent invariants share one in-flight read, and a failed read is evicted so a retry reaches the RPC again.
- **The re-verify pass gets a private cache.** A lagging RPC node can return stale registry state as a *success*, which the shared cache would replay into the re-verify and defeat its whole point. Only that pass gets a fresh map; concurrent invariants keep their shared entries.

Each receiver read is also individually guarded, so one flaky RPC no longer abandons the receivers not yet checked.

## 3. Selector-set facet identity — and why the matching rule is not the one the ticket describes

The ticket asked to identify an on-chain facet by matching its **full** compiled selector set against `facets()`. I measured that against live chains before building on it, and it does not hold: **on mainnet only 29 of 45 facets match their compiled set exactly.** 12 are strict subsets (constants and view getters are routinely left unregistered at cut time) and 3 differ in both directions (deployed build predates HEAD). A full-set rule would identify almost nothing and warn constantly.

So the implemented rule is: **on-chain selectors must be _contained_ in a compiled set, and exactly one compiled facet may fit** — ambiguous stays unidentified rather than guessed, and an exact match wins over a merely containing one (which is what separates a facet from the packed variant re-exporting its selectors).

I validated that this cannot misidentify, on real data: across 7 production networks, **221 log-named facets — 200 identified correctly by selectors alone, 21 left unidentified (the safe outcome), 0 misidentified.**

The reference branch solved the same problem differently, by parsing `getExcludes()` out of each `script/deploy/facets/Update<Facet>.s.sol` to compute the exact registered set. That is more precise in principle, but it costs ~390 lines including a Solidity-source parser with several silent-degradation paths, and the containment rule above is empirically sufficient with zero misidentifications. I took the smaller one.

`facet-required-periphery` now resolves live facets through this (log ∪ selectors) instead of the log-only `resolveLiveFacetsFromLog`, which is removed. Rather than adding a second warning for an unidentifiable on-chain facet — which would duplicate what `no-unexpected-facets` already prints for the same address — I enriched that existing warning: it now either names the facet from its selectors ("add it to `deployments/<network>.json`") or states that no compiled selector set identifies it.

Selector identity degrades to inert when `out/` is absent, which keeps unbuilt checkouts working. The health-check workflow runs `bun run typechain` (which does `forge build src`) before the check, so artifacts are present in CI.

## Verification on real chains

Every new check was proven to fire on real fleet data, not just fixtures.

**No new reds anywhere.** The two new error surfaces are (a) bridge-receiver owners, which nothing checked before, and (b) log-absent receivers, which were skipped before. Swept both across **all 66 production EVM networks**: `0` new owner failures, `0` new binding failures, `0` failed reads. That sweep was itself validated with a negative control — planting a wrong expected owner made it flag 10 receivers across 3 networks, so the zeros are a real result rather than a silent no-op.

**mainnet, before vs. after: 7 errors → 7 errors** (the 7 are pre-existing whitelist source-of-truth failures). All four bridge-specific receivers are now owner- and binding-checked there, where previously none were owner-checked.

**`periphery-registry-log-sync` across the same 66 networks: 770 entries in sync, 0 log gaps, 26 address mismatches** — i.e. nothing in service is registered on chain and missing from the deploy log, so **no deploy-log backfills are needed** in this PR. The 26 mismatches are all genuine drift this check is the first thing to see; most are retired contracts (`ServiceFeeCollector`, `AxelarExecutor`, `RelayerCelerIM`), but three look worth a follow-up ticket rather than a silent warning:

| Network | Contract | Deploy log | On-chain registry |
|---|---|---|---|
| jovay | `LiFiTimelockController` | `0xDDB12A9303F7bdDa8C7c39E58d42af3896330fEd` | `0x707F90DFb87F88690B6a4fC74107a820af646c47` |
| metis | `ServiceFeeCollector` | `0x877ea756597A822a0A0269819c843dA59A43Fe2C` | `0x0000000000000000000000000000000000001234` |
| soneium | `Receiver` | `0x614Fe86aF0b9740946A94a5Cb8f07C1Dc9d77eBC` | `0x65B6093ff981113AC6efb5b77B2D74D96edF4593` |

26 warnings across 66 networks is a reviewable volume; excluding the `*.diamond.json` direction (see above) is what keeps it there.

**Selector identity, on facets the log CAN name:** 221 log-named facets across 7 production networks — 200 identified correctly by selectors alone, 21 left unidentified (the safe outcome), **0 misidentified**.

**Selector identity, on the orphans that actually matter** (facets the log cannot name — the case this feature exists for): across all 66 networks there are 38 such on-chain facets; the matcher names 2 and leaves 36 unnamed. **Both names have since been independently confirmed correct by other PRs merged to `main`:**

| Network | Orphan address | Named by selectors | Independently confirmed |
|---|---|---|---|
| robinhood | `0x8854…D291` | `LayerSwapFacet` | #2207 recorded exactly this address (EXSC-718) |
| sei | `0x9deD…3255` | `PolymerCCTPFacet` | #2200 recorded exactly this address (EXSC-662) |

Both were live current deployments missing from the flat deploy log — precisely the blind spot this PR closes — and in both cases a human independently reached the same conclusion the check reaches automatically. **Zero false positives among the named orphans.** Because `main` now carries both entries, this PR ships no deploy-log changes at all: it is code and tests only.

The remaining 36 unnamed orphans are the known unexecuted-removal backlog (`SymbiosisFacet` on ~40 chains, plus a few testnet `EmergencyPauseFacet`/`SupersetFacet`); the check correctly declines to name them rather than guessing.

## Known limits, stated rather than buried

- **The selector fallback currently adds nothing to `facet-required-periphery` on today's fleet.** Measured across 64 networks: there is no network where selector identity adds a live coupled facet that the deploy log did not already name. Its realized value today is in `no-unexpected-facets`, where it names 4 of the 37 log-absent on-chain facets — including the robinhood gap backfilled here. It is insurance against log drift that has not yet happened, and it is what the ticket asked for; I would rather say that plainly than imply it is load-bearing today. **Decision: kept** — the cost is already paid, it is validated at zero misidentifications across 1229 on-chain facets, and both facets it named were independently confirmed correct by other merged PRs.
- **`ChainflipFacet` is the one coupled facet selector identity cannot rescue.** Its deployed mainnet build registers a selector HEAD no longer compiles, so containment fails — for the single coupled facet whose log entry going stale is most plausible, the fallback would not help. The deploy log covers it today.
- **Receiver read failures are split by cause, not lumped together.** Each receiver read is individually guarded so one failure cannot abandon the receivers not yet checked — but the report level now depends on *why* it failed. A chain-level failure (the call reverts, or the address holds no code — i.e. the registry points at something that is not the expected contract) is deterministic, reproduces on every retry, and is reported as an **error**. Anything else — rate limits, timeouts, transport errors, unrecognised shapes — is treated as **transient** and warns, so a network blip can never redden the fleet. The classifier walks the error's `cause` chain and defaults to transient, so a misjudgement fails safe. Verified on real viem errors, not just fixtures: a call to a no-code address classifies deterministic, an unreachable host transient. Fleet dry-run over all 64 production networks: **185 receiver probes, 0 deterministic failures**, so this adds no new reds today.

**Tests:** 120 passing across the two suites; `bunx eslint` and `bunx tsc-files --noEmit` clean on all five changed files.


0xDEnYO:
## Review-gate residuals (local trial)

This PR went through the auto-fixing review gate twice — once on the initial draft, once on the fix commits. Most findings were fixed in place; these are the ones I deliberately did **not** change, plus context a reviewer should have.

### Escalated — your call, not mine

**1. The selector-set fallback adds nothing to `facet-required-periphery` on today's fleet.**
Measured across 64 production networks: there is no network where selector identity surfaces a live coupled facet the deploy log did not already name. Its realized value today is in `no-unexpected-facets`, where it names 4 of the 37 log-absent on-chain facets — including the robinhood `LayerSwapFacet` gap this PR backfills. It is what the ticket asked for and it is insurance against log drift that has not happened yet, but it is ~200 lines whose coupling-gate benefit is currently zero. Keeping it is defensible; cutting it is also defensible. Flagging rather than deciding.

Related: `ChainflipFacet` is the one coupled facet the fallback cannot rescue — its deployed mainnet build registers a selector HEAD no longer compiles, so containment fails. The deploy log covers it today.

**2. Receiver read failures are warnings, not errors.**
Each receiver read is individually guarded so one flaky RPC cannot abandon the receivers not yet checked. The trade-off: a receiver whose `owner()` reverts — e.g. a registry entry pointing at a non-Ownable address — now warns instead of failing, and warnings skip `executeInvariant`'s re-verify pass. I chose resilience over strictness because a transient RPC error reddening the fleet is a failure mode this sweep already has, but it is a deliberate severity change on an error-severity invariant and deserves a second opinion.

### Drift this PR surfaces but does not fix

`periphery-registry-log-sync` reports 26 registry/deploy-log address mismatches across 66 networks. Most are retired contracts (`ServiceFeeCollector`, `AxelarExecutor`, `RelayerCelerIM`) whose registrations were never cleaned up. Three look worth their own ticket:

| Network | Contract | Deploy log | On-chain registry |
|---|---|---|---|
| jovay | `LiFiTimelockController` | `0xDDB12A9303F7bdDa8C7c39E58d42af3896330fEd` | `0x707F90DFb87F88690B6a4fC74107a820af646c47` |
| metis | `ServiceFeeCollector` | `0x877ea756597A822a0A0269819c843dA59A43Fe2C` | `0x0000000000000000000000000000000000001234` |
| soneium | `Receiver` | `0x614Fe86aF0b9740946A94a5Cb8f07C1Dc9d77eBC` | `0x65B6093ff981113AC6efb5b77B2D74D96edF4593` |

The soneium one is why the generic `Receiver` is now resolved registry-first: before this PR, the contract the diamond actually points at there was owner-checked by nothing.

Separately, a fleet sweep shows ~44 on-chain facets absent from the flat log that are the known unexecuted-removal backlog (`SymbiosisFacet` on ~40 chains, plus a few testnet `EmergencyPauseFacet`/`SupersetFacet`). Not this PR's job — noting it so the new `no-unexpected-facets` warnings are not mistaken for a regression.

### Known test caveat

The Tron serialization test is white-box: it intercepts `ctx.peripheryRegistryCache.get` and never exercises `readPeripheryRegistryUncached`, so it would silently stop testing anything if the caching layer were refactored. Every new test in this PR was verified to fail when its corresponding fix is reverted.

0xDEnYO:
@coderabbitai review

0xDEnYO:
@coderabbitai review

0xDEnYO:
Thanks — 5 of the 6 nitpicks applied in `47231e3a9`, 1 declined with reasoning.

**Applied**

- **Selector fallback no longer shadowed by a non-candidate log name** (`facetPeripheryCouplings.ts`). Good catch, and it is the strongest of the six: a mislabelled or superseded log line naming an address something uncoupled would previously hide a live coupled facet exactly as effectively as a missing entry did — which is the bug this PR exists to fix. Covered by a new test that fails against the previous logic, and re-verified on real data: 221 log-named facets across 7 production networks, still **0 misidentified**.
- **Registry cache keyed by diamond address, not name alone.** Agreed the current per-network `Map` made it correct-but-fragile; the composite key removes the footgun.
- **Negative result memoized in `loadCompiledFacetSelectors`.** Confirmed the temp-dir tests still pass, as you noted.
- **Tron concurrency test renamed** to claim only the sequencing it actually measures — it intercepts the cache, so it does not exercise `readPeripheryRegistryUncached`. (An earlier review pass flagged the same white-box limitation.)
- **`Object.keys(...)[0]` replaced with a pinned coupling** plus an assertion that it still exists, so a reordered registry or a carve-out landing on the first entry cannot silently turn the test into a passing no-op.

**Declined — `Promise.all` around the receiver loops to enable multicall batching**

Both receiver invariants iterate at most 5 names, and `resolvePeripheryAddress` reads go through the run-wide registry cache, so the second invariant's resolutions are already served from memory rather than the RPC. The saving is a handful of round-trips per network, against restructuring two loops whose per-read `warn`-and-`continue` semantics are load-bearing — that path is what keeps one flaky RPC from abandoning the receivers not yet checked, and it was added in response to an earlier review round. Not worth the regression risk for the gain; happy to revisit if the sweep's wall-clock becomes a concern.

Verification on the fix commit: 761 TS tests pass, `bunx eslint` and `bunx tsc-files --noEmit` clean, and a real mainnet health-check run is unchanged at 7 errors / 4 warnings / 17 periphery entries in sync.

0xDEnYO:
## Both escalated items resolved

**1. Selector-set fallback — kept.**
The measurement stands (it adds nothing to `facet-required-periphery` on today's fleet), but the cost is already paid and its value in `no-unexpected-facets` is not theoretical: both facets it named have since been independently confirmed correct by other merged PRs — `LayerSwapFacet` on robinhood (#2207, EXSC-718) and `PolymerCCTPFacet` on sei (#2200, EXSC-662), each at exactly the address the matcher derived. It stays as insurance against the deploy-log drift that motivated this ticket in the first place.

**2. Receiver read failures — split by cause instead of a blanket warning.**
The previous behaviour downgraded *everything* to a warning, which lost a real signal: a registry entry pointing at something that is not the expected contract. Now the report level follows the cause:

| Failure | Reported as | Why |
|---|---|---|
| Call reverts, or address holds no code | **error** | Deterministic — reproduces on every retry; the contract is wrong, not the network |
| Rate limit, timeout, transport error, anything unrecognised | **warning** | Not evidence of anything; a blip must never redden the fleet |

Both paths still `continue`, so one failing receiver never abandons the ones not yet checked — the coverage property the guard existed for is unchanged.

`isDeterministicReadFailure` walks the error's `cause` chain (with a cycle guard) and **defaults to transient**, so a misjudgement fails safe rather than paging someone.

Verified against real viem error shapes rather than only fixtures: a call to a no-code address on mainnet classifies deterministic, an unreachable host classifies transient. Fleet dry-run across all 64 production networks: **185 receiver probes, 0 deterministic failures, 0 transient failures** — so this introduces no new reds today while making a genuinely broken receiver fail loudly.

Tests: `isDeterministicReadFailure` covered directly (revert, no-code, nested cause, cycle, rate limit, transport, unrecognised) plus both invariants on both paths; each new test verified to fail when the split is reverted. Full TS suite 825 pass / 0 fail; mainnet re-run unchanged at 7 errors / 4 warnings / 17 in sync.

script/deploy/healthCheckInvariants.ts
```diff
@@ -1289,6 +1500,131 @@ export const HEALTH_CHECK_INVARIANTS: IHealthCheckInvariant[] = [
...
+        ctx.coreFacetsToCheck.includes(name) ||
+        ctx.nonCoreFacets.includes(name)
+      const candidates = [
+        ...new Set([
+          ...(ctx.isTron ? getTronCorePeriphery() : getCorePeriphery()),
+          ...Object.keys(ctx.globalConfig.whitelistPeripheryFunctions),
+          ...Object.values(getFacetPeripheryCouplings()).map(
+            (coupling) => coupling.requires
+          ),
+          ...RECEIVER_EXECUTOR_GETTERS.map((receiver) => receiver.name),
```
gvladika, line 1693: `Receiver` is in none of the static sources here — not `corePeriphery`, not a coupling companion, and deliberately not in `RECEIVER_EXECUTOR_GETTERS`. So it is only probed when a log already names it, which is circular for a check whose job is finding names the log is missing.

Verified on real data: 36 of 72 production flat logs have no `Receiver` key, and 30 of those lack it in the diamond log too. On **ink**, `getPeripheryContract("Receiver")` returns `0x65B6093ff981113AC6efb5b77B2D74D96edF4593` while neither `deployments/ink.json` nor `deployments/ink.diamond.json` records `Receiver` — exactly the log gap this invariant exists to catch, and it stays silent.

That also caveats the "0 log gaps, so no deploy-log backfills are needed" number in the description: the zero is partly a consequence of this blind spot rather than a clean fleet.

Suggest seeding `'Receiver'` and `'ReceiverAcrossV3'` into the candidate union and re-running the 66-network sweep. Note `receiver-owner` already hardcodes `'Receiver'`, so the two checks currently disagree about which receivers exist.

script/deploy/healthCheckInvariants.ts
```diff
@@ -1588,13 +1928,42 @@ export const HEALTH_CHECK_INVARIANTS: IHealthCheckInvariant[] = [
...
+      }
+      if (!ctx.publicClient) return
+      const publicClient = ctx.publicClient
+
+      // Resolved registry-first so a receiver missing from - or stale in - the deploy log is still
+      // covered; absent from both sources means the receiver genuinely is not on this chain. The
+      // generic Receiver is included: resolving it from the log alone would check a contract the
+      // diamond no longer points at.
+      for (const { name } of [
+        { name: 'Receiver' },
```
gvladika: The list is `['Receiver', ...RECEIVER_EXECUTOR_GETTERS]`, but that constant's only exclusion is `ReceiverAcrossV3`, justified on *Executor binding* grounds ("its Executor binding is no longer kept current") — which says nothing about ownership. `ReceiverAcrossV3` is still in many production flat logs (ink included) and inherits `WithdrawablePeriphery`, so it ends up the one receiver with no owner check, by list reuse rather than by decision.

Either give ownership its own explicit list, or include `ReceiverAcrossV3` here with a note on why the binding is skipped but the owner is not.

Minor while here: wrapping a string in `{ name: 'Receiver' }` only to unwrap it immediately — `['Receiver', ...RECEIVER_EXECUTOR_GETTERS.map((r) => r.name)]` reads better.

script/deploy/healthCheckInvariants.ts
0xDEnYO: Confirmed, and fixed in 92ef67ee2 — you're right that this was circular.

Reproduced your numbers locally: 36 of 71 production flat logs have no `Receiver` key, and all 36 lack it in the diamond log too. On **ink**, `getPeripheryContract("Receiver")` returns `0x65B6093ff981113AC6efb5b77B2D74D96edF4593` with neither log naming it — the check was silent on exactly the gap it exists to find.

Fix: a new `RECEIVER_NAMES` constant (all six receivers), seeded into the candidate union in place of `RECEIVER_EXECUTOR_GETTERS.map(name)`. Deliberately a separate constant rather than a derivation, so the binding-scoped exclusion can't leak into name discovery again (same point as your other comment).

Re-swept all 67 production mainnets with `Receiver` + `ReceiverAcrossV3` seeded — **exactly one new finding**, the ink log gap:

| Network | Contract | Finding |
|---|---|---|
| ink | `Receiver` | registered `0x65B6…4593`, absent from both logs |

`ReceiverAcrossV3` adds nothing: on all 18 chains where it's registered the log entry matches. The soneium `Receiver` mismatch already in the description was found via the log, so it's unchanged. Four networks (`arc`, `moonbeam` [deprecated], plus transient `arbitrum`/`bsc` RPC failures) didn't answer; `arbitrum` and `bsc` re-checked clean via `cast`.

You're right that this caveats the "0 log gaps" number — it's 1, not 0, and it stays a warning so nothing turns red. I'll fix the description before merge.

script/deploy/healthCheckInvariants.ts
0xDEnYO: Agreed on both counts — fixed in 92ef67ee2.

You're right that this was list reuse rather than a decision: `ReceiverAcrossV3` inherits `WithdrawablePeriphery` (`src/Periphery/ReceiverAcrossV3.sol:18`), sits in 19 production flat logs, and a stale *Executor pointer* says nothing about whether its *owner* is still the refund wallet.

Ownership now has its own explicit list (`RECEIVER_NAMES`, all six receivers) instead of borrowing the binding list, and `RECEIVER_EXECUTOR_GETTERS`' exclusion comment now states that the owner is still checked. Also took the `{ name: 'Receiver' }` unwrap — the loop reads over a plain name list now.

Verified this adds no new reds before pushing: swept `ReceiverAcrossV3.owner()` across every production mainnet where it's registered — **18 networks, 0 mismatches** (`arc`/`moonbeam` have no V3; `arbitrum` and `bsc` re-checked via `cast`, both correct). Negative control: reverting the constant back to `['Receiver', ...RECEIVER_EXECUTOR_GETTERS]` makes the three new tests fail, so they're not no-ops.

While in those loops I also took CodeRabbit's batching nit — both receiver invariants now resolve their registry addresses in one `Promise.all` before iterating, so the multicall client can coalesce them. The per-read guard (one flaky receiver doesn't abandon the rest) is unchanged.

0xDEnYO:
@coderabbitai review

script/deploy/healthCheckInvariants.ts
0xDEnYO: Reversing my previous reply — Daniel pushed back that `Receiver` is long deprecated, and he's right, so seeding it was the wrong fix. Reverted in 4352e0bfd.

The structural gap you identified is real, but it resolves without a static list. `RECEIVER_EXECUTOR_GETTERS` is **exactly** the set of coupling companions in `config/global.json`:

```text
AcrossFacetV4 / AcrossFacetPackedV4 / AcrossV4SwapFacet -> ReceiverAcrossV4
ChainflipFacet                                          -> ReceiverChainflip
LiFiIntentEscrowFacetV2                                 -> ReceiverOIF
StargateFacetV2                                         -> ReceiverStargateV2
```

Those companions already enter the candidate union via `getFacetPeripheryCouplings()`, so every receiver **in service** reaches it twice over and none depends on a log naming it first. The `RECEIVER_EXECUTOR_GETTERS.map(name)` seed was already redundant — which means the static list I added contributed exactly two names, both deprecated:

| | `Receiver` | `ReceiverAcrossV3` |
|---|---|---|
| Source in repo | deleted (#1002, 2025-02-26) | present, superseded by V4 |
| Networks in target state | 0 | 0 |
| Registered on chain | 35 | 17 |
| In flat deploy log | 36 | 18 |
| In diamond deploy log | 0 | 66 |

So the ink finding is not a log gap to backfill — it's a deprecated contract still registered on a diamond, and the log is *correct* to omit it. Seeding the name would have converted stale state into a permanent warning instead of something to clean up. Description is back to "0 log gaps".

A test now asserts the checked set equals the coupling companions exactly, so neither deprecated name can creep back in, and a second asserts neither is probed.

The 35 + 17 live registrations are still real stale state — that's a `registerPeripheryContract(name, address(0))` wave plus a deploy-log purge (36 + 18 flat entries, and a `ReceiverAcrossV3` key in all 66 diamond logs). Own ticket, not this PR.

script/deploy/healthCheckInvariants.ts
0xDEnYO: Reversing my previous reply here too. You correctly spotted that ownership was inheriting a binding-scoped exclusion — but Daniel's point is that `ReceiverAcrossV3` is fully deprecated and shouldn't be checked at all, so the answer is subtraction, not a second list. Reverted in 4352e0bfd.

Checked the deprecation claim: `ReceiverAcrossV3` is in **0** networks' target state (`script/deploy/_targetState.json`), superseded by V4. Same for the generic `Receiver`, whose source was deleted in #1002 back in Feb 2025.

Which produces a better framing of the invariant than the one I had. On `main`, `receiver-owner` checks **exactly one** contract — the generic `Receiver`. So the ownership guard was, in practice, guarding only a deleted contract while the four receivers actually in service had no owner check at all. It now checks those four and nothing else. Not "widened from one to five" — repointed from a deprecated one to the live ones.

Two things fell out of that:

- **`receiver-owner` is now `evm-only`**, matching `receiver-executor-binding`. Its Tron branch called `checkOwnershipTron('Receiver', ...)`, and `deployments/tron.json` has never carried a `Receiver` key — `checkOwnershipTron` no-ops on a missing name, so that branch could only ever do nothing. It's gone rather than left looking like coverage.
- **`RECEIVER_EXECUTOR_GETTERS` is now the single receiver list** for both binding and ownership, and its doc comment states why the deprecated ones are absent outright rather than merely unbound. Your `{ name: 'Receiver' }` unwrap nit is moot — the loop iterates that list directly now.

Guard tests: one asserts the checked set equals the coupling companions exactly, one asserts a deprecated receiver registered on chain with a wrong owner *and* a wrong binding produces no errors. Both fail if either name is re-added (verified by putting them back). 120 tests pass.

Kept from the earlier round: CodeRabbit's batching nit — both receiver invariants resolve registry addresses in one `Promise.all` before iterating, per-read guards unchanged.

0xDEnYO:
@coderabbitai review

0xDEnYO:
## Review gate — residual findings (escalated, not fixed)

Ran the pre-review gate over the three fix commits from this round. Auto-fixed in `45769c63d`: a factually wrong scope comment, a fragile set-equality assertion, an over-claiming test title, and a missing positive control. Two items need a human decision.

### 1. `ReceiverOIF` is live on Tron and `receiver-owner` no longer covers it

`receiver-owner` is now `chains: 'evm-only'`, and its rationale ("every receiver in service is EVM-only") was **wrong as of `main`** — #2220 (EXSC-768) deployed `ReceiverOIF` to Tron at `TSZnVJnhPCH8PECjEY4zNbv7kK1XiBUwwb`. This branch's base predates that merge, which is why the branch's own `deployments/tron.diamond.json` still shows `ReceiverOIF: ""`.

Not a regression against `main` — `main`'s Tron branch called `checkOwnershipTron('Receiver', …)` and no `deployments/tron.json` has ever carried a `Receiver` key, so it was a provable no-op. But after merge a live Tron receiver has neither an owner nor a binding check (`receiver-executor-binding` has been `evm-only` since before this PR).

Not fixed here because it is not a one-line scope flip:

- dropping `chains: 'evm-only'` alone silently no-ops on Tron — `if (!ctx.publicClient) return` returns before any check
- a real Tron path cannot reuse the batched `Promise.all` resolution: each Tron registry read spawns a `troncast` subprocess, which is exactly why `periphery-registry-log-sync` deliberately serializes its Tron reads
- it is a new `error`-severity surface on Tron, so it wants fleet evidence before it lands

The comment now states the gap explicitly instead of asserting a false rationale. **Decision needed:** build the sequential Tron path in this PR, or ticket it.

### 2. The deprecated receivers are still reconciled where a deploy log names them

Worth being precise, because the PR body could be read as claiming more than it delivers. Removing the deprecated names from the static seed stops `periphery-registry-log-sync` from *probing them by name* — but the candidate union also seeds from `Object.keys(ctx.deployedContracts)`, so on every chain whose flat log still carries `Receiver` or `ReceiverAcrossV3` they are still probed and still reported. That is why the sweep table in the description shows a `soneium / Receiver` mismatch.

Arguably correct behaviour: a deploy-log entry is a claim, and reconciling it is this invariant's job. But if the intent is that these contracts stop appearing in health-check output altogether, the remaining step is purging them from the deploy logs — 36 flat-log `Receiver` entries, 18 flat-log `ReceiverAcrossV3` entries, and a `ReceiverAcrossV3` key in all 66 diamond logs. That is a log-only change requiring no on-chain action. **Decision needed:** purge the logs, or accept the warnings as legitimate drift reporting.

- `'probes every receiver in service without a log naming it first'` cannot detect the loss of either discovery path: coupling companions and `RECEIVER_EXECUTOR_GETTERS` are the same four names today, so deleting either source leaves the test green.

### Verified sound (not findings)

- `chains: 'evm-only'` is a typed literal union, so a typo is a `tsc` error rather than a silently disabled invariant; honored in `isInvariantApplicable`.
- The `Promise.all` batching cannot abandon un-resolved receivers: `resolvePeripheryAddress` catches registry failures internally and returns `undefined` rather than rejecting.
- Removing the static seed loses no coverage for receivers in service — the four names are hardcoded in the candidate union independently, verified against real logs on networks whose flat log omits them (`base` lacks `ReceiverChainflip`; `soneium` lacks both `ReceiverChainflip` and `ReceiverOIF`).
- The PR's fleet-sweep zeros remain valid: the checked set only got smaller, so zeros over the superset imply zeros over the subset.

script/deploy/healthCheckInvariants.ts
```diff
@@ -1427,6 +1653,133 @@ export const HEALTH_CHECK_INVARIANTS: IHealthCheckInvariant[] = [
...
+        if (!logged) {
+          ctx.logWarn(
+            `${name} is registered on chain (${onChain}) but missing from the deploy log - add it to deployments/${ctx.networkLower}.json`
+          )
+          return
+        }
+        // Tron addresses are base58 and case-sensitive; only EVM hex gets checksum-normalized.
+        const normalize = (address: string): string =>
+          ctx.isTron ? address : tryGetAddress(address) ?? address
+        if (normalize(String(logged)) !== normalize(onChain))
```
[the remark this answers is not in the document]
0xDEnYO: @AikidoSec ignore: False positive — internal developer CLI/CI tool, no HTTP surface. At this call site (loadDiamondLogPeripheryNames in script/deploy/healthCheckInvariants.ts) the path input is doubly constrained before the read: (1) isValidNetworkName() allow-lists the network segment against /^[A-Za-z0-9_-]+$/, so '..', '/', '\\' and absolute paths cannot appear; (2) the containment guard `const rel = path.relative(deploymentsDir, logPath); if (rel.startsWith('..') || path.isAbsolute(rel)) return []` confines the resolved path to deployments/. The value itself is a canonical network key from config/networks.json (a version-controlled allow-list), and any read failure returns an empty list rather than propagating.

gvladika (approved):

PR #2214, opened 2026-08-18
feat(VaultWrapper): source-liquidity-aware exit views

gvladika:

This PR makes the exit `max*` views source-liquidity-aware, replacing #2197 with a deliberately narrower scope.

`maxRedeem` now clamps the owner's balance to what the source can currently honor, read through a new adapter view `maxWithdrawableValue` (the source's floor valuation of `min(wrapper position, source.maxRedeem)`). A full-position short-circuit returns the plain balance whenever the source can release the wrapper's whole position — the common case — so the lossy share round-trip only runs when the source is genuinely liquidity-limited (an Aave-style utilization cap, a paused source) and one-call full exits keep working on healthy sources. `maxWithdraw` inherits the clamp for free because OZ derives it as `previewRedeem(maxRedeem(owner))`, and both exit entrypoints' `ERC4626Exceeded*` guards pick it up automatically — this restores the EIP-4626 guarantee that an exit within the reported max never reverts on a source liquidity limit, with zero changes to the exit paths themselves. When clamping, the conversion floors, so the view may under-report by up to one source share's value but never over-promises. Previews stay liquidity-agnostic per EIP-4626.

What this PR deliberately does **not** do (the scope cut vs #2197): no cost-aware exit pricing. Exits remain exact-out, and sources that charge an exit fee (or under-pay their reported value) remain unsupported — the existing `AdapterWithdrawShortfall` guard freezes exits on such a source instead of mispricing them, which is the enforcement mechanism for that scoping decision. This matches how comparable products handle it (e.g. Kiln's DeFi Vault supports only fee-less sources while running the same liquidity clamp), and cost-awareness can land in a future release: wrapper instances are beacon-upgradeable and a fee-charging source would need a dedicated adapter anyway. The tradeoff is documented in both affected docs pages.

Test coverage: unit tests for the adapter view and the clamp (including redeem-at-clamped-max and the revert above it), a new invariant suite running the full stateful campaign over a source whose withdrawal liquidity is fuzzed up and down (proving `maxRedeem` never over-reports and a redeem within it never reverts), and the Morpho Arbitrum fork scenario now exits at `maxRedeem` — the EIP-4626-recommended pattern — since MetaMorpho's own `maxRedeem` floors below `balanceOf`.

### DIFF
--- a/docs/VaultWrapper/ERC4626Adapter.md
+++ b/docs/VaultWrapper/ERC4626Adapter.md
@@ -19,7 +19,19 @@ asset. `deposit`/`withdraw` return the wrapper's asset balance delta and the
 wrapper reverts on a shortfall, catching a yield source that moves less than
 asked. It does **not** catch share-side dilution (a vault that consumes the full
 asset but credits fewer shares via an internal deposit fee); such non-standard
-sources are unsupported and require a dedicated adapter.
+sources are unsupported and require a dedicated adapter. The same applies to
+sources charging an **exit fee**: exits are exact-out, so a fee-on-exit source
+trips the wrapper's shortfall guard on every withdrawal — unsupported until a
+future release adds cost-aware exit pricing.
+
+## Source liquidity view
+
+`maxWithdrawableValue` is the source-liquidity signal behind the wrapper's
+liquidity-aware `maxRedeem`/`maxWithdraw`: the source's floor valuation
+(`convertToAssets`) of `min(holder position, source.maxRedeem)`. It equals the
+full position value on a source with no active liquidity limit, and shrinks to
+what the source can currently honor under a limit (an Aave-style utilization
+cap, a paused source).
 
 ## Functions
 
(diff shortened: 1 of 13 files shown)


gvladika:
@coderabbitai review

gvladika:
@coderabbitai Both nitpicks are declined as per repo conventions:

1. `maxRedeem(address owner)` keeps OZ's parameter name because it overrides `ERC4626Upgradeable.maxRedeem(address owner)` — all existing `max*` overrides in this file (`maxDeposit`, `maxMint`, `maxWithdraw`) deliberately match the OZ virtual signature rather than the underscore convention used for our own functions.
2. The test names follow the repo's testing rule (`.agents/rules/400-solidity-tests.md`): segments after the `test_`/`testRevert_` prefix are PascalCase with **no** underscores within a segment, so `test_MaxRedeemEqualsBalanceWhenSourceUnlimited` is the required form and `test_MaxRedeem_EqualsBalance_WhenSourceUnlimited` would violate it.

gvladika:
@coderabbitai review

PR #2215, opened 2026-08-18
chore(deployments): deploy logs track current state — prune injective (EXSC-723)

0xDEnYO:

https://linear.app/lifi/issue/EXSC-723

Follow-up to #2204, which retired the intent-escrow flow on injective in `_targetState.json` and `healthCheckInvariants.ts` but left the deploy logs listing the removed facet.

The convention so far was "flat `deployments/<network>.json` logs are deployment history and stay as-is". We are dropping that: **both** deploy logs describe current on-chain state, so evaluations, health checks and the target-state diff can use them as reference for what is live. History remains available from the MongoDB deployment log and this repo's git history, which is where it was always actually queried from — nothing consumes the flat log as a historical record.

**Injective cleanup.** Verified against the loupe (`facetAddresses()` on `0xeCeC…81F5`), which routes 13 facets. Two logged facets are not among them:

| Facet | Address | Why it is gone |
| --- | --- | --- |
| `LiFiIntentEscrowFacet` 1.1.2 | `0xD63D…16E6` | removed by the manually executed timelock op behind #2204 |
| `GenericSwapFacet` 1.0.0 | `0xce42…Faf6` | deprecated repo-wide (no `src/`, absent from `_targetState.json`), removal already executed |

`injective.diamond.json` was already cleaned of both by #2200 (its `Facets` block now matches the loupe exactly), so this PR closes the other half: the flat `injective.json`. No parked task covers either facet (`list-parked-tasks.ts --network injective` → 0), so nothing in the deferred-cleanup queue depended on those entries.

**Where the convention now lives.** `docs/DeploymentLogs.md` is the canonical statement; rule `001-project-structure` and the two commands that used to encode the old rule point at it:

- `/deprecate-contract` — the log-retention warning becomes two-sided: entries stay while the facet is still registered (the drain resolves its address from the flat log until #2157 lands), and are deleted by whoever records the executed removal.
- `/finish-rollout` Phase 4 — `updateDiamondLogs` rebuilds `Facets` from the loupe, so a removed facet drops out of the diamond log on its own; the append-only flat log needs the manual delete, in the same commit.
- `docs/DeferredDiamondCleanupQueue.md` §8/§10 — the "until the parked task retires" formulation was ambiguous for a task that retires as `cancelled`/`superseded` with the facet still registered. The gate is the loupe, not the task state.

**No CI gate.** "In the flat log, absent from the diamond log" is also the legitimate shape of a contract deployed ahead of its cut (propose-only registration reads exactly that), so `checkDeploymentAddressConsistency.ts` cannot decide it without loupe + pending-cut state. Documented in the new doc rather than papered over with a check that would fire on every deferred deploy.

**Known follow-up (not in this PR).** A local scan finds ~123 facet entries across ~30 networks that sit in a flat log but not in that network's diamond log — mostly `GenericSwapFacet` (21), `NXTPFacet` (11), `LiFiIntentEscrowFacet` (9), old Across/CelerIM facets. Each needs a loupe check plus a source-gone/target-state check to tell "removal executed" from "deployed, awaiting cut", so it belongs in its own reviewed sweep. That is EXSC-723's item 3 ("reconcile: safe-to-prune reporting") — with this convention, the report becomes a prune obligation rather than an FYI.


0xDEnYO:
Both CodeRabbit findings applied in 290a2ef:

- `deprecate-contract.md:279` — the "deployments only" escape hatch now protects the `*.diamond.json` entry as well, gated on loupe routing rather than on registration wording.
- `DeploymentLogs.md` / `finish-rollout.md` — dropped "append-only" for the flat log; it is a mutable current-state map with no generator that prunes it, so the deletion is a manual edit. Same wording in both files.

0xDEnYO:
Absorbed into #2252, which carries the full convention write-up (`docs/DeploymentLogs.md` plus the `.agents/` and `docs/` pointers) and applies it across the fleet — 392 stale entries pruned, 8 addresses corrected, 21 diamond-log facet entries removed, 6 orphan logs deleted, all chain-confirmed. This PR's injective prune is reproduced there byte-identically.

Two things changed on the way in: the doc's "the drain resolves its address from the flat log" justification is stale now that #2157 made the drain address-keyed — the entry is kept for the health check's stale-facet name mapping instead; and the when-to-prune table gained the two cases the sweep actually hit (correcting a stale address when the name is live elsewhere, and deleting a flat log whose diamond was never deployed).

PR #2218, opened 2026-08-19
fix(scripts): retry mongodb+srv connects via public DNS on SRV failure

0xDEnYO:

Fixes [EXSC-794](https://linear.app/lifi-linear/issue/EXSC-794/retry-mongodbsrv-connections-via-public-dns-when-the-local-resolver)

Every `mongodb+srv://` script in this repo dies immediately with `Error: querySrv EBADRESP _mongodb._tcp.<cluster>.mongodb.net` on networks whose DNS forwarder re-encodes SRV answers with a name-compression pointer in the target field. RFC 2782 forbids compression there, so Node's c-ares resolver rejects the packet — while `dig`, `host` and the macOS system resolver accept it. The result is a failure that looks like broken credentials, a dead tunnel, or an Atlas outage, while DNS appears perfectly healthy from the shell. Captured on a router at `10.0.148.1`: it returns the SRV target as `rdlength 0x24` ending in a `c02a` pointer, where 1.1.1.1 and 8.8.8.8 return the fully expanded `rdlength 0x35`. Bun's resolver tolerates it, Node's does not — and our scripts run through `bunx tsx`, which spawns node, so they're on the failing path.

The fix is one shared helper, `withSrvDnsFallback()`, wrapped around the connect call at every `MongoClient` site. It retries exactly once against public DNS servers, and only when the error is an SRV lookup failure (`syscall === 'querySrv'`); anything else rethrows untouched. Three deliberate choices: the switch is failure-triggered rather than applied at startup, because unconditionally pointing the process at public resolvers would bypass split-horizon DNS for internal endpoints; it wraps the existing connect call instead of re-creating the client, which was verified to be reusable after an SRV failure; and `MONGODB_DNS_SERVERS` is available as an override for networks where 1.1.1.1 / 8.8.8.8 are themselves unreachable. Rebuilding a plain `mongodb://` seedlist by hand was rejected — it would hardcode shard hosts, `replicaSet` and `authSource` and silently rot when Atlas rescales.

Two sites (`timelock-queue`, `parked-tasks`) never call `client.connect()` explicitly and connect lazily on their first index-ensure, so the wrapper goes around that call instead; `createIndex` is idempotent, so the retry is safe.

Verified end-to-end on the affected network: `query-deployment-logs latest --no-use-cache` previously failed at SRV resolution and now emits the warning and fetches all 4356 records. The helper's own tests cover the success path, the retry, the non-SRV rethrow, the failing-retry rethrow, the once-per-process latch and an empty override list — and were checked against a deliberately broken implementation to confirm they actually fail when the fallback is removed.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -70,6 +70,11 @@ SC_MONGODB_URI=""
 # never dragged into an urgent signing set.
 DRAIN_PARKED_TASKS=""
 
+# Optional. Only used as a fallback: if the local resolver cannot deliver a usable
+# SRV record for a "mongodb+srv://" URI, the tooling retries the connection against
+# these DNS servers (defaults to 1.1.1.1,8.8.8.8). See docs/Setup.md.
+MONGODB_DNS_SERVERS=""
+
 # MongoDB Configuration Note:
 # MONGODB_URI is required for deployment logging
 # All deployment logs are stored in MongoDB (database: contract-deployments)
--- a/docs/Setup.md
+++ b/docs/Setup.md
@@ -200,16 +200,20 @@ perfectly healthy from the shell.
 The tooling handles this on its own: the connection is retried once against public
 DNS servers, with a `WARN` naming the resolver that failed. Set
 `MONGODB_DNS_SERVERS` (comma-separated) if 1.1.1.1 / 8.8.8.8 are unreachable on
-your network, or fix it at the OS level with
+your network, or fix it at the OS level — on macOS, with `<service>` being your
+network service name as listed by `networksetup -listallnetworkservices`:
 
 ```bash
-networksetup -setdnsservers Wi-Fi 1.1.1.1 8.8.8.8
+networksetup -setdnsservers "<service>" 1.1.1.1 8.8.8.8   # e.g. "Wi-Fi"
+networksetup -setdnsservers "<service>" empty              # restore DHCP-provided DNS
 ```
 
-To confirm the diagnosis, compare the two resolvers directly:
+To confirm the diagnosis, run the same SRV query through the local resolver and
+through a public one — the first fails, the second returns the record count:
 
 ```bash
-node -e "require('dns').resolveSrv('_mongodb._tcp.<cluster>.mongodb.net',(e,r)=>console.log(e?e.code:r.length))"
+node -e "require('dns').resolveSrv('_mongodb._tcp.<cluster>.mongodb.net',(e,r)=>console.log('local:',e?e.code:r.length))"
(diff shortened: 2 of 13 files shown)


0xDEnYO:
@coderabbitai review

0xDEnYO:
Addressed all three CodeRabbit findings in e86463b9:

- **docs/Setup.md:207** — the DNS override is now labelled macOS-specific, uses a `<service>` placeholder (with `networksetup -listallnetworkservices` to find it) instead of a hardcoded `Wi-Fi`, and documents the `empty` restore. Flags and the `empty` sentinel verified against the CLI's own usage output and man page.
- **docs/Setup.md:213** — correct catch: the text promised a comparison but only ran one query. It now runs the SRV lookup through the local resolver *and* through 1.1.1.1. Verified as printed: `local: EBADRESP`, `public: 3`.
- **mongo-srv-dns.test.ts:141** — both tests now assert `rejects.toBe(srvError)` on the original object. Confirmed the strengthened assertion is real: against an implementation that rethrows `new Error(error.message)`, both tests fail (they passed before the change).

0xDEnYO:
Closing: the cost/benefit doesn't hold up on review.

The defect is real — a DNS forwarder that re-encodes SRV answers with a name-compression pointer, which RFC 2782 forbids and Node's c-ares resolver therefore rejects while `dig`/`host`/Bun all accept it. But the blast radius is one specific resolver, and the fix wraps **every** `MongoClient` connect site in the repo (13 files) in a permanent shared helper to work around it.

Two things tipped it:

- **The acute trigger is being removed anyway.** #2264 replaces the 71-network `Promise.all` fan-out — which fired ~142 concurrent SRV/TXT queries and was what made this failure routine — with a single connection and one `$in` query. The per-connection SRV resolution largely disappears rather than needing a retry path.
- **There are cheaper places to fix it** for anyone who does hit it: point the resolver at 1.1.1.1/8.8.8.8, or run the affected script under Bun, whose resolver tolerates the malformed answer. Neither costs the repo any shared surface.

If it turns out others on the team hit this on their own networks, that changes the calculus and this is worth reopening — the diagnosis in this PR is solid and worth keeping for that reason.

**Note for #2228:** that PR is based on this branch (`fix/exsc-794-mongo-srv-dns-fallback`) and imports `withSrvDnsFallback` at two sites. Closing this PR does **not** delete the branch, so #2228 is not broken — but before it can merge it must be retargeted to `main` with the helper absorbed into it. Flagged on that PR.

PR #2222, opened 2026-08-19
chore(networks): deprecate moonbeam (chain shut down) (EXSC-796)

0xDEnYO:
# Which Linear task belongs to this PR?

Fixes [EXSC-796](https://linear.app/lifi-linear/issue/EXSC-796/deprecate-moonbeam-network-chain-shut-down)

# Why did I implement it this way?

## Why moonbeam is being removed

Moonbeam (chainId 1284) has been shut down by its operator. This is **not** an RPC or tooling problem:

- Moonbeam and Moonriver entered **Maintenance Mode at 00:00 UTC on 2026-08-01**; after that time all transactions are rejected. Blocks continue during the operational wind-down — [moonbeam.network/news/what-is-maintenance-mode](https://moonbeam.network/news/what-is-maintenance-mode/)
- Wormhole announced Moonbeam officially shut down on 2026-07-31.
- Moonbeam is exiting the Polkadot ecosystem entirely and migrating GLMR 1:1 to Base — [chaincatcher](https://www.chaincatcher.com/en/article/2275443), [cryptoslate](https://cryptoslate.com/moonbeam-just-halted-all-user-transactions-leaving-late-glmr-holders-at-the-mercy-of-an-email-helpdesk-to-recover-stranded-funds/)

Observed directly on 2026-08-19: two independent RPC providers (`moonbeam.drpc.org` and `public.1rpc.io/glmr`) both return the same block `0x1004c1b` (16796699) and it does not advance across repeated reads 20s apart; [Moonscan](https://moonbeam.moonscan.io/) reports 16796694. `eth_gasPrice` is constant. Transaction submissions are accepted into the mempool then never included, surfacing as JSON-RPC `-32603 "already known"` with the deployer's pending nonce never advancing past latest.

> Note for anyone re-deriving this: there was a **different, unrelated** Moonbeam maintenance-mode event in 2022 caused by a smart-contract security incident. The relevant event here is the 2026 wind-down.

**Impact today:** `config/networks.json` still carried moonbeam as `"status": "active", "type": "mainnet"`. That is the only reason it was pulled into the 66-chain FeeForwarder v2.0.0 production rollout, where it consumed three failed deploy runs (10 attempts each). Until it is deprecated it keeps being pulled into every fleet deploy, health check and whitelist sync.

## Scope of the change

Full deprecation following the corn (#2004) and taiko/swellchain/superposition/botanix/sophon (#2141) precedents:

| File | Change |
|---|---|
[...]

PR #2223, opened 2026-08-19
docs(agents): align /deprecate-network with real practice (EXSC-797)

0xDEnYO:

Fixes [EXSC-797](https://linear.app/lifi-linear/issue/EXSC-797/update-deprecate-network-doc-to-match-real-deprecation-practice)

`.agents/commands/deprecate-network.md` described a narrower procedure than the one we
actually follow. Three real deprecations all had to go beyond it — #2004 (corn), #2141
(taiko/swellchain/superposition/botanix/sophon) and #2222 (moonbeam, EXSC-796) — and one
of its steps was actively harmful. This aligns the doc with what those PRs did.

Five changes, each grounded in those PRs:

**1. Bridge/integration configs — search, don't enumerate.** The doc named only
`config/networks.json`, `foundry.toml`, `_targetState.json`, the deployment logs and the
whitelist. Every deprecation additionally had to scrub `gaszip.json`, `permit2Proxy.json`,
`squid.json`, `stargateV2.json`, `relay.json`, `glacis.json`, `garden.json`,
`layerswap.json`, `symbiosis.json`. The new step lists those as observed examples but
explicitly says not to work from the list — it prescribes `git grep -inw` instead, because
the set grows every time an integration is added, and because `git grep` reaches tracked
files outside `config/script/src/tasks/test` that the old path-scoped `grep` missed.

**2. `CORE_FACET_EXEMPTIONS` was missing.** `script/deploy/healthCheckInvariants.test.ts`
has a test `every exempt network is a known network` that asserts every exemption network
exists in `networks.json`, so a stale entry is a CI failure, not a cosmetic leftover. Now
its own step, with the test named so the reader knows why it's mandatory.

**3. The whitelist step was wrong and is now inverted.** Step 7 told you to run
`bun update-whitelist-periphery`. That generator regenerates the `PERIPHERY` section for
*every* network from the current deployment logs, so whatever drift has accumulated since
the last whitelist sync lands in the deprecation diff alongside the removal. The step now
prescribes surgical hand-removal of `PERIPHERY.<net>` (4-space key) and every
`DEXS[].contracts.<net>` block (8-space), with `jq empty` plus a deletions-only
`git diff --numstat` check, and a warning that the generator rewrites the gitignored
`config/whitelist.staging.json` unconditionally — so `git status` can never prove that file
was left alone, and a checksum is used instead.

**4. Preservation is now explicit.** `deployments/_deployments_log_file.json` (the master
log) and `archive/config/*` must be kept — both #2004 and #2141 deliberately left the
deprecated networks' historical records in place. This was unstated, and it's the one gap
that's easy to get wrong in the *opposite* direction: the cleanup search hits both, so the
doc now classifies them as intentional keeps that get reported, never offered for deletion.

**5. Facet chainId mappings are marked, not deleted.** If the deprecated network appears in
a chainId → LayerZero eid or CCTP domain mapping under `src/Facets/`, the entry stays — the
bridge may still support the chain even though we don't. Instead a comment marks it for
removal on the next modification of the facet. The wording template is lifted from the Taiko
entry (#2141) and the Moonbeam entry (#2222) in `src/Facets/AcrossV4SwapFacet.sol`, with the
network name and ticket parameterised rather than hard-coded.

Supporting edits: the Execution Steps section was renumbered (6 steps → 15), and the
Overview, How to Use, Key Files Modified, Error Handling and Example Output sections were
updated so they no longer contradict it — the old text referenced the removed generator step
in several places. The frontmatter `description` and the `.agents/README.md` Custom Commands
row were widened to match the procedure the doc now describes.

Per `010-agents-authoring`, only the canonical `.agents/commands/deprecate-network.md` was
edited (`.cursor/commands/` and `.claude/skills/` are symlinks, both verified intact), the
`.agents/README.md` row was updated as that rule requires, and the file is kept under the
500-line cap (495).

# Verification

Every factual claim in the doc was executed against the repo rather than taken from the
ticket or from review feedback:

- `bun update-whitelist-periphery` on current `main` leaves `config/whitelist.json`
  **byte-identical**, and rewrites the gitignored `config/whitelist.staging.json`
  unconditionally. An earlier revision of this PR claimed the committed file was out of sync
  with the generator and quantified the churn at ~2100 lines; that was wrong and has been
  removed. The directive to hand-edit stands on the mechanism (regenerating every network
  pulls unrelated drift into the diff), not on a figure.
- `-w` drops `tronshasta` when searching for `tron`, but **does not** drop `eth-taiko` when
  searching for `taiko` — a hyphen is a word boundary. An earlier revision claimed it did
  (that claim came from a review comment and was folded in unverified). `-w` is now described
  as a filter that narrows candidates, with the exact-key check carrying the safety.
- A comment-only edit to `src/Facets/AcrossV4SwapFacet.sol` leaves `bytecode.object`
  identical (built both ways and compared), which is why the marker comment needs no version
  bump — consistent with the facet still being at v1.0.0 after both #2141 and #2222 added
  exactly such comments.
- `git grep -inw`, `jq empty config/whitelist.json` and `shasum -c` were each run in the
  worktree.
- The `every exempt network is a known network` test exists in
  `healthCheckInvariants.test.ts`; `config/whitelist.staging.json` is gitignored;
  `PERIPHERY.<net>` keys sit at 4-space and `DEXS[].contracts.<net>` at 8-space indent; every
  file path the doc names exists.

Doc-only change: no Solidity, no scripts, no CI. `bunx markdownlint-cli2` clean on both files
(scoped, not repo-wide, to avoid unrelated auto-fix churn).


0xDEnYO:
## Review gate — escalated finding (not fixed here)

**`.agents/commands/deprecate-contract.md:92` still prescribes `bun update-whitelist-periphery`.**

This PR establishes that the committed `config/whitelist.json` is out of sync with the generator, so running it rewrites the entire PERIPHERY section. The sibling `/deprecate-contract` skill still instructs "For periphery contracts: Run `bun update-whitelist-periphery` once after all removals" — the same generator, the same drift.

Escalated rather than auto-fixed for two reasons:

1. **Out of the stated scope** of this PR (network deprecation, EXSC-797).
2. **Genuinely ambiguous.** When you deprecate a *periphery contract*, a full PERIPHERY-section rewrite may be the intended outcome — the contract really is gone from every network. That is not the network-deprecation case, where the churn is pure noise. I have no evidence from a real contract-deprecation PR either way, so I did not guess.

Worth a separate ticket if the drift bites there too.

---

### Verification performed for this PR

The gate's falsification pass proved each command the doc now prescribes actually works on real repo data:

- `git diff --numstat <file>` — confirmed first column is additions (2 added / 0 deleted on a controlled fixture), so the doc's "additions column must be 0" check is correct.
- The step-8 repo-wide grep — runs as written and returns 14 real `moonbeam` hits in this checkout.
- `bun test:ts` / `every exempt network is a known network` — 43 tests pass; injecting a fake stale network (`fantomdeprecated`) into `CORE_FACET_EXEMPTIONS` makes exactly that test fail. **The check can fire on the condition it targets**, which is the premise of step 9.

Not verified: the ~2100-line generator-churn figure (running the generator was blocked by the local permission classifier; not worked around). Noted in the PR body.

0xDEnYO:
## Review rounds addressed — CodeRabbit (6) + lifi-qa-agent (3)

All nine findings were folded into the branch. Two of them were **wrong as stated** and are answered rather than applied — with evidence, since both had already been applied verbatim in an earlier revision and had to be backed out.

### Applied

| Finding | Where |
|---|---|
| CR: validate whitelist JSON after hand-edits | step 10 — `jq empty` + deletions-only `git diff --numstat` |
| CR: grep hits are candidates, not deletion targets | step 8 — exact-key match required before editing any config |
| CR: missing deploy log — say "warn and continue" in both places | step 7 + Error Handling (Error Handling still said "skip silently") |
| CR: `git status` can't detect a gitignored file / QA #3 | step 10 — `shasum -c`, and the reason is now the generator, not "don't touch it" |
| CR: parameterise the facet comment template | step 11 — `<Network>` / `<TICKET>` placeholders, explicitly flagged as non-literal |
| CR + QA #2: search scope excludes tracked dirs | steps 8 and 15 — `git grep` over all tracked files |

### Answered, not applied

**QA #1 / CR: "add word-boundary anchors so `taiko` doesn't match `eth-taiko`."** `-w` does not do this. A hyphen *is* a word boundary, so `grep -inw taiko` still matches `eth-taiko`:

```
$ printf 'a: eth-taiko\nb: taiko\nc: taikofoo\n' | grep -inw taiko
1:a: eth-taiko
2:b: taiko
```

`-w` is kept — it correctly drops `tronshasta` when deprecating `tron` — but the doc now describes it as a filter that narrows candidates, **not** a guarantee, and points at the exact-key check as the thing that actually prevents a wrong deletion. Documenting `-w` as sufficient would have been the more dangerous outcome.

**Its own earlier claim: "the committed whitelist is out of sync with the generator (~2100 lines of churn)."** Not true on current `main` — regenerating leaves `config/whitelist.json` byte-identical:

```
$ bun update-whitelist-periphery && git diff --numstat config/whitelist.json
(no output)
```

The hand-edit directive stands on the mechanism instead (the generator regenerates *every* network, so whatever drift exists at deprecation time lands in the diff), and no figure is asserted. The run also rewrote the gitignored `config/whitelist.staging.json` unconditionally — which is the real reason that file needs a checksum rather than `git status`.

Branch is rebuilt on current `main` (`.agents/README.md` keeps the `create-merged-typings` row that main added meanwhile), and the file is back under the `010-agents-authoring` 500-line cap.

0xDEnYO:
@coderabbitai review

0xDEnYO:
## Round 2 — both findings applied, with one correction to the proposed fix

**1. Pre/post key comparison in step 10 — applied.** The reasoning is right: zero net additions doesn't prove the deletions stayed inside the target network.

The snippet both reviews proposed doesn't run, though:

```
$ jq '[.PERIPHERY | keys[], (.DEXS[] | .contracts | keys[])] | unique | sort' config/whitelist.json
jq: error (at config/whitelist.json:9707): Cannot iterate over null (null)
```

`|` binds looser than `,`, so it parses as `.PERIPHERY | (keys[], (.DEXS[] | …))` and `.DEXS` is evaluated inside the `.PERIPHERY` object, where it's null. The parenthesised form is what landed:

```bash
jq -r '[(.PERIPHERY|keys[]), (.DEXS[]|.contracts|keys[])] | unique | .[]' config/whitelist.json
```

Verified on the real file — correct removal emits exactly one line, an unrelated network dropped alongside it emits two:

```
# del lisk only            # del lisk + polygon
38d37                      38d37
< lisk                     < lisk
                           53d51
                           < polygon
```

**2. Contradictory step 15 example — applied.** It now shows what a real post-cleanup search returns (master log, archive, the marked facet mapping, a commented-out network list) instead of files steps 7–10 had already deleted.

### Also: this round falsified a round-1 item

Testing the new check showed **`git diff --numstat config/whitelist.json` reports 1 addition for a correct single-network removal**, not 0 — removing the last key of an object rewrites the preceding line's trailing comma:

```
$ git diff --numstat config/whitelist.json
1	99	config/whitelist.json
```

So "additions column must be 0" — carried since round 1 — would fail a correct deprecation. It's dropped rather than loosened, since the key diff is strictly stronger and has no such false positive. The Error Handling and Implementation Notes references were updated to match.

0xDEnYO:
## Round 3 — applied

Correct catch, and it was real comment-vs-code drift on my side: round 2 replaced the additions-count check with the key-set diff in step 10, Error Handling and Implementation Notes, but left the Example Output still printing a line the command no longer emits.

```diff
-✓ Hand-removed whitelist blocks (PERIPHERY + 3 DEXS entries) — jq empty OK, 0 additions
+✓ Hand-removed whitelist blocks (PERIPHERY + 3 DEXS entries) — jq empty OK, key diff: fantom, harmony only
```

Swept the file for the rest of that change's blast radius while I was in there — `grep -n 'additions'` now returns nothing, and the only two remaining `--numstat` mentions are the two that are deliberate: the optional "measure the drift if you want to know" aside in step 10, and the sentence explaining why a line-count check must *not* be substituted for the key diff.

0xDEnYO:
## Round 4 — both applied

**1. Key diff: zero `<` lines is also correct.** Premise verified against the repo — two networks are in `networks.json` with no whitelist entry at all:

```
$ comm -23 <(jq -r 'keys[]' config/networks.json | sort) \
           <(jq -r '[(.PERIPHERY|keys[]),(.DEXS[]|.contracts|keys[])]|unique|.[]' config/whitelist.json | sort)
localanvil
tronshasta
```

Deprecating either produces an empty diff, which the old wording read as a failure. The rule now names the empty case as correct and states the condition that actually signals a bad edit — a `<` line naming a network you are *not* deprecating — rather than a line count.

**2. Implementation Notes contradicted the Execution Steps on `-w`.** Correct, and the same drift class as round 3: `64706ad5` fixed the claim in step 8 and step 15 but missed this one. All four `-w` mentions now agree.

While fixing it I swept every `-w` claim in the file rather than just the reported line, which is what should have happened in round 3. That trimmed the step-15 bullet down to a pointer at step 8 instead of a restatement of it — also what kept the file inside the `010-agents-authoring` 500-line cap (499).

gvladika (approved):

gvladika (approved):

PR #2225, opened 2026-08-19
fix(clear-signing): keep destinationChainId raw where a LI.FI id collides with EIP-155 (EXSC-798)

0xDEnYO:

https://linear.app/lifi-linear/issue/EXSC-798/clear-signing-keep-destinationchainid-raw-for-hypercore-capable

Follow-up to EXSC-793 (#2217).

EXSC-793 switched `_bridgeData.destinationChainId` to ERC-7730 `format: "chainId"` so wallets render an EIP-155 network name instead of a bare number. That is right for every LI.FI chain id but one.

`LIFI_CHAIN_ID_HYPERCORE` (`src/Helpers/LiFiData.sol`) is **1337** — a registered EIP-155 id. `chainid.network/chains_mini.json`, the source the Sourcify reference runner uses, resolves it to **"Geth Testnet"**. So with `chainId` a Hypercore bridge shows `Destination Chain: Geth Testnet` on the signing screen: not a degraded rendering but a confidently wrong one, which is the specific failure clear signing exists to prevent. `UnitFacet` is Hypercore-only (`onlyAllowDestinationChain(_bridgeData, LIFI_CHAIN_ID_HYPERCORE)` on both entry points), so it would do this on *every* transfer.

Every other LI.FI-internal id (Solana, BTC, Sui, Stellar, Aptos, LTC/BCH/DOGE, Tron) is far outside the allocated range and unregistered, so it degrades to digits — harmless, and not worth losing the naming win for.

So rather than reverting EXSC-793 wholesale, the generator now picks the format **per bridge**: bridges that can reach a colliding id keep `raw`, the other 62 entries keep `chainId`. That is 8 of 70 entries — Mayan, Unit, PolymerCCTP, AcrossV4Swap, each in its `Bridge`/`Swap & Bridge` variant.

Two things I deliberately did not hardcode:

- **The affected bridge set is derived by scanning the facet sources**, not a hand-maintained list. Bridge identity comes from each artifact's own function names rather than its filename, since `AcrossFacetV4.sol` and `StargateFacetV2.sol` show filenames do not reliably carry it. AcrossV4Swap writes the literal `1337` in a numeric lookup rather than using the constant, so the scan matches both forms (with comments stripped, so a passing mention cannot force a bridge to `raw`).

  To be precise about what this does NOT do: it is a source-reference scan, **not a reachability analysis**. `validateBridgeData` only rejects `destinationChainId == block.chainid`, so a facet can be pointed at a colliding chain by an owner call (`SupersetFacet.setChainIdToEid`) or a backend-signed EIP-712 payload with no code change at all — invisible to any check here. Recorded in the doc; see the escalations below.
- **The colliding id itself is parsed from `LiFiData.sol`**, so the value cannot drift out of sync with the Solidity constant.

Both new guards were falsified against real repo data rather than only covered by same-PR tests:

1. Any new `LIFI_CHAIN_ID_*` that lands on an EIP-155 reserved id fails the generator with a message naming the constant. Verified across five declaration spellings — plain decimal, `1_337`, `0x539`, `public`, and `uint64` — because the first version of this guard matched only one of them and let the other four through silently.
2. A bridge that references a colliding id but matches no emitted format fails the generator. Note this is now a backstop rather than the main line of defence: deriving names from the ABI makes the filename-mismatch failure it originally guarded structurally impossible. I did not manufacture a synthetic facet to prove it can still fire, so treat it as defence-in-depth, not a verified check.

## Downstream

`config/clearSigningProposal.json` is regenerated in the same commit, so `verifyClearSigning` stays green. Once this merges, `syncLedgerClearSigning` will push the corrected `display.formats` into the standing registry PR (ethereum/clear-signing-erc7730-registry#2901) automatically.

Related upstream context: the registry maintainer has separately flagged that ERC-7730 does not define a canonical chain-name source at all — the two reference runners disagree on names for the same id — so `chainId` rendering is loosely specified in general. This PR does not try to solve that; it only removes the case where we would be actively wrong.

Note on tests: the generator has no unit-test suite in this repo; correctness is enforced by the `verifyClearSigning` CI gate (regenerate + diff) plus the two negative controls described above, which I ran against the real repo rather than fixtures.


0xDEnYO:
## Review-gate residuals (escalated — not auto-fixed)

Six findings were auto-fixed in c62a420 (facet-scan glob asymmetry, single-spelling constant parser, the magnitude threshold, comment/hex false positives, and two stale doc statements). Generated output is unchanged by all of it. These three need a human decision:

**1. The colliding-id set is a source scan, not reachability — residual gap by construction.**
`validateBridgeData` only rejects `destinationChainId == block.chainid`. A facet can be pointed at chain 1337 with **no code change**: `SupersetFacet.setChainIdToEid` is an owner call, and NEARIntents / LayerSwap / Garden / LiFiIntentEscrowV2 take the destination from a backend-signed EIP-712 payload. No generator check and no CI gate can observe either. So the fallback protects today's four bridges, and adding such a route in future requires remembering to update `COLLIDING_LIFI_CHAIN_IDS` consumers by hand.

Options: (a) accept + document (what this PR does), (b) revert to `raw` for **all** bridges — kills the risk, loses the network name on the other 62 entries, (c) fix at the root by changing `LIFI_CHAIN_ID_HYPERCORE` off 1337, which is an on-chain constant with backend coupling and a much bigger blast radius.

**2. The collision table is EIP-155's seven ids, not the registry wallets actually use.**
ERC-7730 defines `chainId` via EIP-155 reference values, but the Sourcify reference runner resolves against `chainid.network` (~2.7k entries). Ten of those are at or above `1e12` — above our Solana (1.15e15), Stellar (1.20e15) and Tron (1.88e15) ids. None collides today (verified against the live registry), but a future LI.FI id landing on a registry-only entry would not be caught. Closing it means committing a registry snapshot and keeping it fresh; I did not want to add that dependency unilaterally.

**3. Pre-existing, out of scope: renaming a facet file silently drops it from the descriptor.**
Surfaced while testing this PR's guards. Renaming `MayanFacet.sol` → `MayanFacetV2.sol` (the repo's own versioning convention, cf. `AcrossFacetV4`, `StargateFacetV2`) makes the generator emit **176 fewer lines** — every Mayan entry gone — and still exit 0. `verifyClearSigning` only compares generated-vs-committed, so a dev who renames and regenerates gets a fully green CI with Mayan silently missing from clear signing. Not introduced here; worth its own ticket.

<sub>Posted by the local review-gate trial.</sub>

0xDEnYO:
Closing — superseded by #2261, which takes `destinationChainId` back to `raw` across the board.

This PR pinned `raw` only for ids colliding with EIP-155's reserved set (`1,2,3,4,5,42,1337`). Measuring what the two ERC-7730 reference implementations actually render showed the collision surface is wider than that list: `999` is HyperEVM to us and resolves to "Wanchain Testnet" through the public chain list one implementation uses — a mainnet destination shown as a testnet, and not caught here. Only 7 of 30 ids tested render identically across the two implementations at all.

So the per-id pinning approach cannot be made complete against a name source we do not control. All-`raw` is the honest version of the same idea, and it resolves the HyperCore/1337 case this PR was opened for. The follow-up that would actually let us show a name is tracked in EXSC-838 (upstream: `chainId` needs a defined fallback and a pinned name source, or ENUM_VALUE needs to be wider than one byte).

PR #2233, opened 2026-08-20
fix(config): scope LiFiDEXAggregator whitelisting to 7 networks (EXSC-745)

0xDEnYO:

Fixes [EXSC-745](https://linear.app/lifi-linear/issue/EXSC-745/whitelist-periphery-drift-feecollectorlifidexaggregator-de-whitelisted) — whitelist PERIPHERY drift: FeeCollector/LiFiDEXAggregator de-whitelisted on only 15 of 70 networks

## What this changes

Keeps `LiFiDEXAggregator` (LDA) whitelisted on **exactly 7 networks** — `bob`, `lens`, `lisk`, `opbnb`, `vana`, `viction`, `worldchain` — and makes the whitelist generator enforce that scope instead of re-deriving it on every run.

Parts:

1. `config/global.json`: restores the `LiFiDEXAggregator` entry (`0x2646478b` / `processRoute(address,uint256,address,uint256,address,bytes)`) to `whitelistPeripheryFunctions`, in the slot it occupied before, and adds a new **`whitelistPeripheryNetworks`** map that scopes it to those 7 networks.
2. `script/common/whitelistScope.ts` (new): single `isNetworkInScope` predicate. A contract absent from the scope map is unrestricted — so behaviour for every other periphery contract is unchanged. `assertScopeContractsEligible` makes the generator fail closed if the scope map ever names a contract absent from `whitelistPeripheryFunctions` (lookups are case-sensitive, so a typo would otherwise silently widen that contract to the whole fleet).
3. `script/tasks/updateWhitelistPeriphery.ts` and `script/tasks/checkDeploymentAddressConsistency.ts` both consume that predicate, so the generator and the consistency gate agree on where LDA belongs.
4. New `findScopeViolations` in the consistency gate **fails** on any `config/whitelist.json` entry for a network the contract is scoped away from. Without it nothing in the repo detected over-whitelisting: `findCoverageGaps` only suppresses reports, and `findMismatches` compares addresses rather than presence — so a regenerated `whitelist.json` that re-added LDA fleet-wide would have passed silently.

`config/whitelist.json` is regenerated accordingly: **+70 lines, 0 deletions**, LDA on those 7 networks and nowhere else.

`FeeCollector` stays removed — it is genuinely deprecated.

## Why the scope map, rather than just restoring the entry

`whitelistPeripheryFunctions` says *which selectors* a periphery contract may be whitelisted with, but nothing about *where*. `updateWhitelistPeriphery.ts` derives the network set from the deployment logs and then assigns `whitelistData.PERIPHERY = productionPeripheryData` — a wholesale replacement, not a merge. `script/tasks/diamondSyncWhitelist.sh:16` runs that generator **before** it diffs against on-chain state.

So there was no way to express "LDA on these 7 only":

- Leave LDA out of `global.json` (state on `main`): every sync run regenerates `whitelist.json` without it, the live on-chain pairs read as "not in config", and the sync proposes **unwhitelisting LDA on the 7 chains we want to keep**. This is what makes the change urgent.
- Restore LDA plainly: the generator puts it back on all 68 networks where it is deployed, re-whitelisting it fleet-wide — the opposite of the decision.

Hence a declarative scope map that the generator honours, so the intent survives regeneration.

## Why LDA is not deprecated

Commit `d90ed882b` (in #2120) dropped both `FeeCollector` and `LiFiDEXAggregator` under the heading "drop deprecated FeeCollector and LiFiDEXAggregator". LDA is not deprecated: #1782 removed it from `corePeriphery` so it would stop being deployed to *new* chains, and said so explicitly:

> Contract is **not** deprecated — existing deployments remain in place; the `whitelistPeripheryFunctions.LiFiDEXAggregator` entry is left untouched.

`src/Periphery/LiFiDEXAggregator.sol` is still present at v1.12.0. The "deprecated" label came from the target-state sheet refresh landed in the same PR, not from a deprecation decision in this repo.

#2120's own earlier commit `cbdc6a89a` had stated the ordering constraint that `d90ed882b` then violated:

> Deliberately NOT removing FeeCollector or LiFiDEXAggregator from `whitelistPeripheryFunctions` or `config/whitelist.json` […] They must be unwhitelisted on-chain first, then dropped from config.

## Which 7, and why

Per Ikram in Slack: LDA should be kept only on Viction, opBNB, Lens, World Chain, Lisk, Vana and Bob (chain-partner discussions in progress). This PR encodes exactly that list.

## Verification

**On-chain sweep** of `isContractSelectorWhitelisted(LDA, 0x2646478b)` across all 68 EVM networks with an LDA deployment, against the `config/networks.json` RPCs:

| state | count | networks |
| --- | --- | --- |
| whitelisted, **in scope** | 7 | `bob`, `lens`, `lisk`, `opbnb`, `vana`, `viction`, `worldchain` |
| whitelisted, out of scope | 4 | `arbitrumsepolia`, `arctestnet`, `basesepolia`, `optimismsepolia` (all testnets) |
| already de-whitelisted | 55 | incl. `mainnet`, `base`, `arbitrum`, `polygon`, `optimism`, `bsc`… |
| RPC unresolved | 2 | `arc`, `nibiru` (transient; the unresolved set varies between runs) |

All 7 in-scope networks are **still whitelisted on-chain right now** (`opbnb` re-checked individually after a transient RPC error in the batch run). The earlier removal wave has already executed on every other production chain, so scoping to 7 matches on-chain reality — this PR does not propose re-whitelisting anything on a production mainnet.

**Falsification of each new guard** on real repo data — proving it can fire, not just that tests pass:

| guard | with guard | guard removed / condition injected |
| --- | --- | --- |
| generator scope filter (`updateWhitelistPeriphery.ts`) | `LiFiDEXAggregator: 7 networks` | deleting the filter → **68 networks**; `updateWhitelistPeriphery.test.ts` fails (`Expected: 7, Received: 68`) |
| `findCoverageGaps` scope check | **0** gaps | scope ignored → **61** real coverage-gap errors (`mainnet`, `0g`, `abstract`, `apechain`, `arbitrum`, …); unit test fails |
| `findScopeViolations` (new) | passes | injecting LDA on `mainnet` at its real diamond-log address → `1 out-of-scope whitelist entr(ies)`, exit 1 (this case passed with exit 0 before) |
| `assertScopeContractsEligible` | passes | typo’d contract name in the scope map → throws |

The generator filter is the *primary* enforcement point and had **zero** test coverage in the first revision of this PR — deleting it left the whole 929-test suite green. `updateWhitelistPeriphery.test.ts` drives the real CLI (the module calls `runMain` at import time, so it cannot be imported without executing) and derives the expected count from config + deployment logs rather than hardcoding 7, so it stays correct as LDA deployments change. It also asserts the scope is genuinely narrower than the deployment set, so the test cannot pass vacuously.

**Other checks:**

- `bunx tsx script/tasks/updateWhitelistPeriphery.ts --dryRun` → `LiFiDEXAggregator: 7 networks`.
- `bunx tsx script/tasks/checkDeploymentAddressConsistency.ts` (full repo) → passes.
- `bun test script/` (full TS suite) → **939 pass, 0 fail**.
- Selector re-derived from the signature: `keccak(processRoute(...))[0:4] == 0x2646478b`. ✔
- `bunx tsc-files --noEmit` and `bunx eslint` clean on all changed files.
- `config/whitelist.json` is generator output, committed unformatted-by-hand (insertions only, no reordering).

The `global.json` LDA entry is byte-identical to the pre-`d90ed882b` entry, and sits in the same slot.

## Consequences the reviewer should expect

1. **The 4 testnets above will get LDA removal proposals** on the next whitelist sync. Intended.
2. **The health check will log stale-pair errors on those 4** until those removals execute — `checkWhitelistIntegrity` reports on-chain pairs absent from config via `logError`. This is the config-first-then-on-chain ordering that `cbdc6a89a` warned about, accepted deliberately here because the alternative (no config entry) keeps proposing removal on the 7 chains we are protecting.
3. **Not scope-aware — three deploy/propose paths:** `script/tasks/diamondUpdatePeriphery.sh`, `script/tasks/proposePeripheryWithWhitelist.ts` and `script/tasks/proposeContractToNetworks.sh:48` (`isDiamondCalledPeriphery()`) all route on bare `whitelistPeripheryFunctions` membership, so a *fresh LDA deployment* to an out-of-scope chain would whitelist it there. Unreachable in practice while LDA is absent from `corePeriphery` (verified: `ERC20Proxy, Executor, FeeForwarder, GasZipPeriphery, LiFiTimelockController, OutputValidator, Permit2Proxy, TokenWrapper` — no LDA), so I left it out rather than widen this hot-fix. Worth a follow-up if the scope map gets a second entry.

1. **Pending whitelist Safe proposals** generated from the old config bundle the LDA removal together with the FeeCollector removal and the FeeForwarder adds. Any that are still pending need to be deleted and regenerated from this branch, otherwise signing one re-removes LDA on the 7. The Safe proposal store is behind the `lifi-connect` tunnel and was not reachable from this session, so the current pending count is **not** verified here — it must be checked before signing anything in that wave.
2. A sync run after this merges still proposes the fleet-wide `FeeCollector` unwhitelist. Intended (it is deprecated), but it means the sync does not go quiet.
3. **CI path filters need widening** — separate PR. `ts-unit-tests.yml` and `deploymentAddressConsistency.yml` do not trigger on `config/global.json`, so a future edit to the scope map *alone* skips the very tests that validate it, and both aggregators report green on skip. The fix was written and verified here but removed again: touching `.github/workflows/*` requires Information Security Manager approval (`protect-critical-code`), which would gate this time-critical config fix behind a separate review. Shipping it separately.


0xDEnYO:
## Review gate — escalated findings (not auto-fixed)

Ran the auto-fixing review gate on this draft. The 6-line diff itself verified clean: selector matches the signature, regenerated LDA entries are byte-identical to the committed `config/whitelist.json` (69/69 networks, zero address deltas), and no consumer of `whitelistPeripheryFunctions` newly fails (`checkDeploymentAddressConsistency` gap set empty, `periphery-registered` unaffected since the only remaining `_targetState.json` LDA row is `pharos/staging`).

Three PR-description gaps were auto-fixed (provenance of the earlier executed removal wave, a rebuttal of `d90ed882b`'s own safety argument, and a corrected staging-file claim — `config/whitelist.staging.json` is gitignored, not committed).

Two findings are **escalated for a human decision** and this PR should not go ready until they are answered.

### E1 — This is privilege-expanding on ~15 mainnets, not a no-op (always-escalate: access control)

For the 51 networks where `processRoute` is still whitelisted, this restore is a no-op that stops the removal proposals. For the 15 where it is already de-whitelisted — `mainnet, base, arbitrum, optimism, polygon, avalanche, bsc, linea, scroll, zksync, mantle, blast, cronos, fraxtal, sonic` — the next sync will instead propose `batchSetContractSelectorWhitelist(..., true)`, re-opening an arbitrary-external-call surface that has already been revoked there.

That consequence is documented in follow-up 2 of the description, but it is a production security change on the highest-volume chains and needs an explicit decision rather than arriving as a side effect of a config fix.

### E2 — Was the removal deliberate? (ambiguous intent)

The gate argued the removal was an in-flight deliberate deprecation rather than a mistake, citing `d90ed882b`'s own text ("Both are deprecated… Order has to be diamondSyncWhitelist first, then the config removal") plus the fact that the sync has already run on the major chains.

The counter-evidence this PR rests on:

- **#1782 states the opposite in writing** — "Contract is **not** deprecated… the `whitelistPeripheryFunctions.LiFiDEXAggregator` entry is left untouched" — and no reviewer contested it.
- **The executed removals were generator-driven, not targeted.** Both batches (12 networks on 2026-07-31, 3 on 2026-08-06) bundle the LDA removal together with the FeeCollector removal in one `scheduleBatch`, exactly like the 50 currently pending ones, which additionally bundle the FeeForwarder v2.0.0 adds. A deliberate LDA deprecation rollout would not be shaped like the output of a generic whitelist sync.
- **The first batch is three days after #2120 merged** — the first sync run against the changed config.
- **EXSC-745 was opened as unexplained drift**, and its action item 1 is literally "decide whether FeeCollector and LiFiDEXAggregator should be diamond-callable at all" — i.e. the question was open, not settled.
- **The removal in #2120 went unreviewed.** Every review comment on that PR targeted `healthCheckInvariants.ts`; nobody commented on the `whitelistPeripheryFunctions` deletion.

I believe the mistake reading is correct, but the deliberate-deprecation reading is not unreasonable and the two lead to opposite actions (restore + re-whitelist vs. finish the deprecation and prune `whitelist.json`). Needs a call from someone who can speak to whether the backend still routes diamond swaps through LiFiDEXAggregator.

0xDEnYO:
## Review gate — escalated / not auto-fixed

Auto-fixed and pushed in `3af2ee7f6`: generator test for the primary enforcement point (it had zero coverage — deleting the filter left all 929 tests green), `findScopeViolations` (nothing detected over-whitelisting), `assertScopeContractsEligible` fail-closed guard, CI path filters for `config/global.json`, type-alias naming, a WHAT-narrating comment, and an unrealistic test fixture.

The following were **not** changed — they need a human decision:

**1. The 4 testnets still whitelisted on-chain but outside the scope — decision needed.**
`arbitrumsepolia`, `arctestnet`, `basesepolia`, `optimismsepolia` currently have LDA whitelisted on-chain. They are not in the 7, so the next `diamondSyncWhitelist.sh` run will propose **removing** LDA there. That removal would itself be an incidental side effect — the same class of change this PR exists to prevent. Kept at 7 because the instruction was explicitly "only these chains"; adding the 4 is a one-line change to `whitelistPeripheryNetworks` if the intent is to preserve the status quo on testnets.

**2. Pending whitelist Safe proposals — must be checked before signing.**
Proposals generated from the old config bundle the LDA removal with the FeeCollector removal and the FeeForwarder adds. Signing one after this merges would re-remove LDA on the 7. The Safe proposal store is behind the `lifi-connect` tunnel and was **not reachable from this session**, so the pending count is unverified — this must be confirmed before any signing in that wave.

**3. Three deploy/propose paths are not scope-aware** (disclosed in the PR body, consequence 3): `diamondUpdatePeriphery.sh`, `proposePeripheryWithWhitelist.ts`, `proposeContractToNetworks.sh:48`. Unreachable today because LDA is absent from `corePeriphery`, so intentionally left out of this hot-fix. Needs addressing if the scope map ever gets a second entry.

**4. No CI check that committed `config/whitelist.json` matches generator output.** A future change that drops the scope filter without regenerating would not be caught by the committed artifact. Worth a follow-up "whitelist.json is up to date" job.

**5. Latent:** `healthCheckInvariants.ts` `periphery-registered` gates on `whitelistPeripheryFunctions` membership without consulting the scope map. Inert today — LDA appears on 0 networks in `_targetState.json` — but would require registration regardless of scope if that changes.

**Unrelated CI failure:** `verify-emergency-pause-readiness` is red, and it is not from this PR. Branch `fix/redact-verify-api-key` (which touches only `script/helperFunctions.sh`) fails the identical live-state gate at the same step (`Fail if any readiness check failed`) this morning; yesterday's runs were green on both. Looks like live pauser/governance/PAT state, not code under review.

0xDEnYO:
@coderabbitai review

0xDEnYO:
Update: the CI path-filter hardening (`ts-unit-tests.yml`, `deploymentAddressConsistency.yml` triggering on `config/global.json`) has been **removed from this PR** and moved to a follow-up.

`protect-critical-code` fails on any PR touching `.github/workflows/*` until an Information Security Manager approves. That gate would hold up a time-critical config fix — LDA is still whitelisted on-chain on all 7 target chains right now, and the next whitelist sync removes it there unless this merges first. The change itself is written and verified; it ships separately.

Item 4 in the escalation list above (no CI check that committed `whitelist.json` matches generator output) belongs in that same follow-up.

melianessa (approved):

PR #2240, opened 2026-08-20
fix(VaultWrapper): mirror sub-floor perf-fee skip in fee-share previews

gvladika:

Fixes [EXSC-813](https://linear.app/lifi-linear/issue/EXSC-813/vault-wrapper-fee-share-previews-omit-the-sub-floor-performance-fee)

`_accrueFees` short-circuits below `MIN_SHARE_SUPPLY`: when the performance fee is enabled and the supply is sub-floor it re-floats the watermark and mints no performance shares. `_pendingFeeShares` — which every preview and conversion override runs through — had no matching branch, so in the sub-floor state previews priced against phantom performance-dilution shares the accrual never mints. That broke the parity invariant documented at `LiFiVaultWrapper.sol:342-344` (a preview returned before an operation equals what the caller gets after accrual): `previewMint` under-quoted assets and `previewDeposit` over-quoted shares (both EIP-4626-violating directions), and `maxRedeem`'s liquidity clamp over-reported. The fix mirrors the accrual's sub-floor branch in `_pendingFeeShares` — return the management shares alone when `supply + mgmtShares < MIN_SHARE_SUPPLY` — so the two paths cannot diverge in that state. I kept the change to the single divergent function rather than extracting a shared fee-share helper across accrual and previews; the larger unification touches the accrual hot path on every operation and is better done as its own reviewed refactor. This is a Low-severity, integration-correctness fix — the state is only reachable on a dust vault (supply pushed below the floor by floor-exempt exits) and there is no fund loss. No version bump, per the pre-deployment convention on this branch (see #2227).

### DIFF
--- a/src/VaultWrapper/LiFiVaultWrapper.sol
+++ b/src/VaultWrapper/LiFiVaultWrapper.sol
@@ -1187,9 +1187,17 @@ contract LiFiVaultWrapper is
         uint256 _assets
     ) private view returns (uint256) {
         uint256 mgmtShares = _pendingManagementFee(_supply, _assets);
+        uint256 effectiveSupply = _supply + mgmtShares;
 
-        return
-            mgmtShares + _pendingPerformanceFee(_supply + mgmtShares, _assets);
+        // Mirror the sub-floor branch in `_accrueFees`: below `MIN_SHARE_SUPPLY` no
+        // performance shares are minted (the watermark is re-floated instead), so a
+        // preview that still charged them would price against dilution the accrual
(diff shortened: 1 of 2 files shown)


gvladika:
Superseded by #2241 (EXSC-814): the `_pendingFeeShares` fix landed there unchanged in substance (commit e3e309e05), and both parity tests were ported (adapted to the floor removal's `DUST_SUPPLY_THRESHOLD` naming, commit f3095b80e). Closing per author decision to absorb this into the supply-floor-removal PR.

PR #2243, opened 2026-08-20
feat(safe): cross-check whitelist pair contracts against the deploy log (EXSC-811)

0xDEnYO:

https://linear.app/lifi-linear/issue/EXSC-811/safe-signing-display-cross-check-whitelist-pair-contracts-against-the

Follow-up to EXSC-804 / #2232.

`registerPeripheryContract` proposals already render a deploy-log verdict next to the address (`✅ matches deployments` / `❌ mismatch: expected …`) — it can, because the call carries the periphery **name** in its calldata.

`batchSetContractSelectorWhitelist` carries **only addresses**, so it had no such verdict. Its magenta `(PERIPHERY/<name>)` label comes from `config/whitelist.json`, which is *generated* by `script/tasks/updateWhitelistPeriphery.ts`. When a rollout proposes a whitelist sync built from a stale `whitelist.json`, the signer therefore sees a confident, correct-looking label on a superseded address and has nothing to catch it with. This PR turns that single-source label into a two-source agreement check.

With no name in the calldata, the check is a *reverse* lookup over the deploy log rather than a copy of the registration check, and the verdict has to be direction-aware:

| whitelist.json label | in deploy log | Adding | Removing |
| -- | -- | -- | -- |
| `PERIPHERY/N` | as `N` | `✅ matches deployments` | `⚠️ un-whitelists current 'N'` |
| `PERIPHERY/N` | `N` = other address | `❌ whitelist.json and deployments disagree` | `(superseded 'N')` |
| `PERIPHERY/N` | absent | `❌ no deployments entry for 'N'` | `(not in deployments)` |
| none | present as `N` | `(deployments: N)` | `⚠️ un-whitelists current 'N'` |
| none | absent | *silent* | *silent* |

Three constraints drove that shape:

- **Removals invert the meaning.** Absence from the deploy log is the *expected* state for a removal (the superseded address is gone from it); presence is the alarm, because removing the selectors of an address the deploy log still points at un-whitelists the contract in use. Reusing the registration check verbatim would have painted every correct removal red. The removal verdict keys off the reverse lookup alone, so it still fires when the deploy log holds the address under a name other than the one `whitelist.json` labels it with.
- **Most pairs are third-party.** DEX/aggregator contracts can never appear in the deploy log, so they stay unannotated — a permanent ❌ there would train signers to ignore the marker.
- **Not every periphery is deploy-log-backed.** `Composer`'s whitelist entries are merged from `config/composerWhitelist.json` and never reach a deploy log, so its pairs are checked against *that* file (`✅ matches composerWhitelist.json` / `⚠️ un-whitelists an address composerWhitelist.json still lists`). Checking Composer against the deploy log would have emitted a false ❌ on every correct Composer pair on all 70 networks it is configured for — see the review notes below; this was caught pre-merge, not shipped.

Because Safe calldata is decoded with viem, a Tron pair arrives as `0x` hex while both `whitelist.json` and the deploy log store base58. The pre-existing periphery match folded case instead of normalising per network, so `peripheryName` was always undefined on Tron — meaning the magenta label never appeared there and the new ✅/❌ verdicts could never fire. The match now goes through the same network-aware comparison as the rest of this file, which also restores the missing Tron periphery label.

The decision table lives in a pure exported function (`formatWhitelistDeployLogCheck`); the source-selection wrapper (`getWhitelistDeployLogCheckSuffix`) is exported too, so the whitelist/deploy-log/config resolution that picks the row is covered by tests reading the repo's real `config/` and `deployments/` files rather than synthetic fixtures.

### Verification beyond the unit tests

Rendered against the repo's own unmodified `config/whitelist.json`, `config/composerWhitelist.json` and `deployments/*.json`:

- every whitelist pair of `mainnet` (125), `arbitrum` (123), `tron` (8) and `nibiru` (11), in both directions: **0 red rows**; the only ⚠️ rows are the removal half for our own live periphery, which is the intended alarm.
- adding a live periphery → `✅ matches deployments`; removing the same address → `⚠️ un-whitelists current '<name>'` (negative control: the alarm does fire on real data).
- a real third-party DEX address (mainnet DODO) → silent in both directions.
- a Tron periphery passed in the hex form viem actually decodes → label plus `✅`; reverting the network-aware match makes that test fail, and reverting the Composer entry makes the Composer test fail, so both new guards are proven to be load-bearing.

### Known limits

- An address in **neither** source is indistinguishable from a third-party contract and stays silent. Nothing claims it is ours, so there is nothing to cross-check.
- Both sources read are the **production** ones, so a staging whitelist proposal gets no verdict — as it gets no `(PERIPHERY/…)` label today, for the same reason. Threading the environment through the decode path is a larger change than this PR; flagging rather than silently widening scope.
- The deploy-log reverse lookup takes the **first** matching name. No deployments file currently registers one address under two names, so this cannot fire today, but co-registering two versions under different names is a known pattern in this repo.

### DIFF
--- a/script/deploy/safe/safe-decode-utils.test.ts
+++ b/script/deploy/safe/safe-decode-utils.test.ts
@@ -20,12 +20,24 @@ import {
 import { consola } from 'consola'
 import { toFunctionSelector } from 'viem'
 
+import composerWhitelistJson from '../../../config/composerWhitelist.json'
+import whitelistJson from '../../../config/whitelist.json'
+import mainnetDeployments from '../../../deployments/mainnet.json'
+import tronDeployments from '../../../deployments/tron.json'
+import { normalizeAddressForNetwork } from '../../utils/normalizeAddressStringForViem'
+
 import {
   getRoleName,
   formatRoleChange,
   formatBatchSetContractSelectorWhitelist,
   formatWhitelistDeployLogCheck,
 } from './safe-decode-utils'
+
+/** Fails the test loudly rather than silently skipping when repo data moves. */
+const required = <T>(value: T | undefined, what: string): T => {
+  if (value === undefined) throw new Error(`test fixture missing: ${what}`)
+  return value
+}
 
 const DEFAULT_ADMIN_ROLE = `0x${'00'.repeat(32)}`
 // OpenZeppelin AccessControl role hashes (public, keccak256 of role names) —
@@ -265,17 +277,19 @@ describe('formatBatchSetContractSelectorWhitelist', () => {
 })
 
 describe('formatWhitelistDeployLogCheck', () => {
+  // The decision table is exercised through explicit inputs, so these are
+  // deliberately role-neutral: which address is live and which is superseded is
+  // whatever the arguments of a given case say it is.
   const NETWORK = 'mainnet'
-  const LIVE = '0xC748171a028401BfD0c0F0a757ab4A1F93b00576'
-  const SUPERSEDED = '0x5Bf8351f8C349634911965Bd000fE9c625C1A0d9'
-  const THIRD_PARTY = '0xeCCc2Ce02907c1C86B286FB50d1355384F55A298'
+  const ADDRESS_A = '0xC748171a028401BfD0c0F0a757ab4A1F93b00576'
+  const ADDRESS_B = '0x5Bf8351f8C349634911965Bd000fE9c625C1A0d9'
 
   const check = (
     overrides: Partial<Parameters<typeof formatWhitelistDeployLogCheck>[0]>
   ): string =>
     formatWhitelistDeployLogCheck({
       network: NETWORK,
-      contractAddress: LIVE,
+      contractAddress: ADDRESS_A,
       whitelisted: true,
       ...overrides,
     })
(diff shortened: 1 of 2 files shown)


0xDEnYO:
### Review-gate residual notes (not fixed in this PR)

The gate's findings were applied in 0086bb956 (Composer source-of-truth, Tron network-aware match, fixture/coverage fixes, wording). Three items were deliberately **not** fixed, and need a call rather than a quiet change:

1. **Staging proposals get no verdict** — `getDeploymentsRecord` and `getWhitelistJson` both read the production sources, so a `ENVIRONMENT=staging` whitelist sync (`diamondSyncWhitelist.sh` supports it) renders with no label and no verdict. The missing label is pre-existing; the missing verdict is new. Fixing it means threading the environment through the decode context and into `getWhitelistJson`, which touches the `confirm-safe-tx` / `execute-pending-timelock-tx` call sites. Documented as a limit in the function doc and PR body instead — say the word and I'll do it here or as a follow-up ticket.

2. **Reverse lookup takes the first matching deploy-log name** — no `deployments/*.json` currently maps one address to two names, so this cannot fire today, but co-registering two versions under different names is a known pattern in this repo. Deterministic-but-arbitrary today; would need a rule for which name wins.

3. **`❌ whitelist.json and deployments disagree` does not say which side is stale** — it names the deploy-log address, so the signer can tell which file to refresh only by inspection. Reworded to name both sources; going further would mean guessing at provenance.

PR #2246, opened 2026-08-21
chore(deployments): prune deprecated entries from every deploy log (EXSC-818)

0xDEnYO:

https://linear.app/lifi-linear/issue/EXSC-818/fleet-wide-deploy-log-prune-remove-deprecated-entries-from

Follows the convention landed in #2215 (`docs/DeploymentLogs.md`): both `deployments/<network>.json` and
`deployments/<network>.diamond.json` describe what is live on that chain **now**. History lives in the
MongoDB master log and in git.

Every deploy log was reconciled against the chain itself rather than against `_targetState.json` or the
presence of a Solidity source file, because only the chain can say what is live:

- `facetAddresses()` on each diamond (81 EVM logs; `tron.json` through TronGrid's JSON-RPC)
- `getPeripheryContract(name)` on the PeripheryRegistry for every entry that is not a live facet

An entry was removed only when the chain confirmed neither route. That keeps names that are still cut even
though their source is gone — `GenericSwapFacet` is still live on 17 chains, `AcrossFacetV3` on 1 — and it
removes names whose source still exists but whose logged address is no longer routed anywhere.

## What changed

| | |
|---|---|
| flat-log entries removed | **397** across 82 files |
| flat-log entries corrected | **3** (`arbitrum.staging`: `CalldataVerificationFacet`, `CelerCircleBridgeFacet`, `GlacisFacet` — name live at a different address, so the address was fixed rather than dropped) |
| `*.diamond*.json` facet entries removed | **138** (addresses absent from `facetAddresses()`) |
| files deleted | **6** |

Biggest contributors to the flat-log prune:

| Contract | Entries |
|---|---|
| `GenericSwapFacet` | 64 |
| `AcrossFacetV3` | 20 |
| `AcrossFacetPackedV3` | 19 |
| `CBridgeFacet` | 19 |
| `ServiceFeeCollector` | 16 |
| `CBridgeFacetPacked` | 15 |
| `LiFiDiamondImmutable` | 13 |
| `LiFiIntentEscrowFacet` | 13 |
| `AcrossFacetPacked` | 12 |
| `CelerIMFacetMutable`, `HopFacetPacked`, `NXTPFacet`, `RelayerCelerIM`, `RelayerCelerIMMutable` | 10 each |
| `HopFacet`, `HopFacetOptimized`, `LIFuelFacet` | 9 each |
| 50 further names (`StandardizedCall*`, `Amarok*`, `Multichain`, `Axelar*`, `Hyphen`, `Pioneer`, `Wormhole`, …) | 1–8 each |

Deleted files — active networks with no staging diamond, so nothing they listed was reachable through a
diamond: `tronshasta.json` (already `{}`), `xdc.staging.json`, `abstract.staging.json`,
`monad.staging.json`, `scroll.staging.json`, `soneium.staging.json`.

## Deliberately left alone

- **13 facets live on-chain but absent from every log.** Identified by selector match: 10× `LiFiIntentEscrowFacetV2`
  (avalanche, unichain, stable, ink, monad, celo, gnosis, worldchain, hyperevm, mantle) belonging to open PR #2231,
  `SymbiosisFacet` on `bsc.staging`, and one unnamed 3-selector facet on `base.staging`/`arbitrum.staging`. Their
  own rollout PRs add the log entries; adding them here would collide.
- **65 `FeeForwarder` entries whose registry address differs from the log** — expected while the FeeForwarder v2.0.0
  rollout (#2226) is unmerged. Same for `arbitrum.staging` `Permit2Proxy`/`Patcher` and jovay `LiFiTimelockController`.
- **516 empty-string `Periphery` placeholders** in the diamond logs: `saveDiamondPeriphery` regenerates the whole
  block from `src/Periphery/`, so deleting them produces churn on the next rollout, not a cleaner log.
- `script/deploy/resources/deployRequirements.json` still carries a `ServiceFeeCollector` block for a contract with
  no source — flagged in EXSC-818 as separate hygiene.

## Why this is safe for the parked-removal queue

Until #2157 lands, the parked-removal drain resolves a facet's name from the flat log, so pruning an entry that a
queued removal still needs would break the drain. It cannot happen here by construction: every pruned address owns
**zero selectors** on that diamond, so no removal proposal can target it. Facets still routed were kept, which is
exactly what the drain needs.

## Verification

- Re-ran the same on-chain reconciliation after the edits: **0** entries remain that the chain does not confirm
  (only the 62 `LiFiTimelockController` entries — neither facet nor registered periphery by design — and the
  pending-rollout registry mismatches above).
- Re-ran the diamond-log check: the only remaining drift is the 13 in-flight additions listed above.
- `bunx tsx script/tasks/checkDeploymentAddressConsistency.ts` → passed.
- All 171 files were rewritten by `JSON.parse`/`JSON.stringify(…, null, 2)`; verified beforehand that this
  round-trips every deployment log byte-identically, so the diff is deletions only (plus the 3 corrected
  addresses) with no reformatting.
- `deployments/injective.json` reproduces #2215's prune exactly; whichever merges first leaves the other's hunk
  a no-op.

- [ ] I have updated any required documentation — the convention itself is documented by `docs/DeploymentLogs.md` in #2215; this PR only applies it


0xDEnYO:
Superseded by #2252 — same change re-derived on top of current `main` (which now carries #2226, #2231 and #2157), plus two fixes: the sweep is fail-closed per network, and live facets whose diamond-log `Name` is empty are now resolved by selector match, so 5 entries that this PR would have deleted are correctly re-pointed instead.

PR #2254, opened 2026-08-21
chore(deployments): roll out FraxFacet v1.0.0 to production (EXSC-387)

gvladika:
# Which Linear task belongs to this PR?

Fixes [EXSC-387](https://linear.app/lifi-linear/issue/EXSC-387/deploy-to-prod) — Deploy to Prod (parent: EXSC-382, FraxFacet integration)

# Why did I implement it this way?

Production rollout of `FraxFacet v1.0.0` (Frax HopV2 / LayerZero OFT bridge), merged in #2048 and audited by Sujith Somraaj. This PR contains only the generated deployment-log changes from the rollout — one `deployments/<network>.json` entry per network recording the deployed `FraxFacet` address.

The facet was deployed via CREATE3 to 24 production networks (every chain with a Frax `hop` entry in `config/frax.json`; berachain/mode/scroll are intentionally excluded per the Frax team). On each network it is registered in the `LiFiDiamond` through a **timelock-wrapped Safe cut proposal** (`diamondCut` + a one-shot `initFrax` call seeding the chainId→LayerZero-EID mappings). Each proposal currently carries one signature and awaits multisig signing/execution.

Rollout driven with `DRAIN_PARKED_TASKS=true`, so parked facet-removal tasks fold into each network's cut proposal where present.

## Deployed addresses & Safe proposal nonces

| Chain | FraxFacet address | Safe nonce |
| --- | --- | --- |
| mainnet | 0x8452788daad6af88fe88BC5dFc892974C11C32Ad | 123 |
| arbitrum | 0x8452788daad6af88fe88BC5dFc892974C11C32Ad | 88 |
| avalanche | 0x8452788daad6af88fe88BC5dFc892974C11C32Ad | 65 |
| base | 0x8452788daad6af88fe88BC5dFc892974C11C32Ad | 87 |
| bsc | 0x8452788daad6af88fe88BC5dFc892974C11C32Ad | 66 |
| fraxtal | 0x8452788daad6af88fe88BC5dFc892974C11C32Ad | 29 |
| optimism | 0x8452788daad6af88fe88BC5dFc892974C11C32Ad | 76 |
| polygon | 0x8452788daad6af88fe88BC5dFc892974C11C32Ad | 76 |
| sei | 0x8452788daad6af88fe88BC5dFc892974C11C32Ad | 35 |
| sonic | 0x8452788daad6af88fe88BC5dFc892974C11C32Ad | 51 |
[...]

PR #2259, opened 2026-08-24
feat(healthcheck): group fleet failures by root cause, route alerts to #dev-sc-github-ci-notifications

0xDEnYO:

EXSC-818 (adjacent — no dedicated ticket).

Two changes to the fleet health check's Slack alert: **where** it goes, and **what it says**.

## Routing

The fleet health check was the last recurring-CI alerter still posting to `#sc-general`. Its
alerts are GitHub-Actions output, so it now uses `SLACK_WEBHOOK_DEV_SC_GITHUB_CI_NOTIFICATIONS`
— the same secret `reconcileParkedTasks`, `weeklyCoverageReport`, `ticketLinkageMetric` and
`updateScriptConfigReminder` already use. Secret confirmed present (added 2026-08-18) and the
channel's `SC_GitActionsNotifications` incoming-webhook integration is installed. Verified live:
run 32680003424 was dispatched from this branch and its alert landed in
`#dev-sc-github-ci-notifications`.

Deliberately **not** moved here: `diamondEmergencyPause` and `verifyEmergencyPauseReadiness` are
emergency signals that arguably want human eyes rather than a CI feed, and
`networkRpcsChecker` / `syncLedgerClearSigning` are a separate judgement call.

## Grouped digests — failures and warnings

The alert previously said `17 of 66 network(s) failed: avalanche, bob, celo, …` and linked to a
run containing 66 networks of interleaved concurrent output. The count tells you how many chains
are red; it does not tell you *why*, which is the part that decides whether anyone has to act.

The runner already computed per-network detail (`detail = result.errors.slice(-5)`) and threw it
away after printing. It is now grouped by normalized cause and rendered into the message. On the
real data from run 32680003424, 17 failed networks collapse to **3** causes:

```text
🚨 Diamond health check — ACTION NEEDED
17 of 66 network(s) failed:
• LiFiIntentEscrowFacetV2 live but companion ReceiverOIF not registered in Diamond - destination calls for this integration are disabled on this network (9): avalanche, celo, gnosis, hyperevm, ink, mantle, monad, stable, unichain
• Pair Array has [n] stale pairs not in config: Stale: [address] / 0x2646478b (7): bob, lens, lisk, opbnb, vana, viction, worldchain
• Pair Array has [n] stale pairs not in config: Stale: [address] / 0xe0cbc5f2 Stale: [address] / 0xeedd56e1 (1): tron
<link|View workflow run>
⚠️ 40 network(s) passed with warnings: …
```

Design notes:

- **Addresses are masked, 4-byte selectors are not.** The same cause recurs per chain with a
  different contract address, so masking is what lets the shapes collapse. A different selector,
  by contrast, *is* a different cause (`0x2646478b` is `processRoute`; `0xe0cbc5f2` is not) — which
  is why tron stays its own group rather than being folded into the 7.
- **Standalone integers are masked** so "1 stale pair" and "2 stale pairs" group together. The
  `\b` anchors keep digits inside an identifier intact — `LiFiIntentEscrowFacetV2` survives.
- **Mask tokens avoid angle brackets.** Slack parses `<…>` as a link element, so `<address>`
  would have been mangled in the very message it exists to clarify.
- **Capped on every axis** (6 groups, 12 networks per group, 160 chars per cause). Slack rejects
  an over-long incoming-webhook `text` outright rather than truncating it, so an unbounded digest
  would drop the whole alert.
- **The flat network list is kept as a fallback** for the case where the runner dies before
  producing results and there is no digest to render.
- **The heredoc delimiter is stripped from the digest body.** The digest is assembled from revert
  strings and RPC error text; a line that happened to equal the terminator would close the
  `$GITHUB_OUTPUT` block early and let the remainder be parsed as further step outputs.

### Warnings got the same treatment

Warnings were worse off than failures: the runner reduced each network's warning texts to a
`number`, so the alert could only say `40 network(s) passed with warnings: <40 names>`. No reason,
nothing to act on — a line people learn to skip. `IHealthCheckResult.warnings` is now `string[]`
and warnings group by cause exactly like failures.

On the current fleet that turns **37 warnings into 2 causes**, verified against run 32680003424's
real log output:

```text
⚠️ 40 network(s) passed with warnings:
• Facet [address] is registered on-chain but absent from the deploy log, and no compiled selector set identifies it (unexpected/rogue facet, or a retired… (20): abstract, apechain, arbitrumnova, +17 more
• Facet [address] is registered on-chain but absent from the deploy log; its selectors match FraxFacet - confirm whether this is the current build before… (17): abstract, arbitrum, avalanche, base, bsc, fraxtal, ink, katana, linea, mainnet, optimism, polygon, +5 more
```

That second line is #2254 — FraxFacet's cuts have executed on 17 chains while the deploy-log
entries recording them are still in an unmerged PR. One line, one decision, instead of 17
networks' worth of noise.

Three details worth reviewing:

- **Warning texts name their own network** (`…recording it in deployments/<network>.json…`), which
  would have put all 17 FraxFacet networks in groups of one — the exact fragmentation the grouping
  exists to prevent. Normalization therefore masks the network id, on word boundaries so a short id
  (`ink`, `sei`) cannot rewrite an unrelated word (`thinking`).
- **Warnings are collected from every network regardless of status.** A network can fail one
  invariant and warn on another; dropping that warning because the network was already red is how a
  warning goes unseen for months.
- **A network emitting several distinct warnings appears in several groups**, deduplicated within
  each. Each cause is a separate decision.

Known limitation, deliberately not changed here: grouping keys off the whole detail, which
`runOneNetwork` already trims to the last 5 errors. A network failing more than five ways groups
by its tail and can land in its own group rather than beside networks sharing its first cause.
Widening that trim changes what the console prints too, so it belongs in its own change.

- [x] I have added tests that cover the functionality / test the bug — 43 tests in
      `healthCheckAllNetworks.test.ts` plus 7 in the new `redactUrls.test.ts`, built on the real
      detail and warning strings from run 32680003424
- [x] For new facets: n/a
- [x] I have updated any required documentation — the workflow's header comment describes the new
      message shape

Verification beyond the unit tests: both digests were generated from run 32680003424's real log
output, and the compose step's rendered output was exercised end-to-end for the failure-digest
path, the warning-digest path, and the runner-died fallback. The workflow YAML and the compose
step's bash were syntax-checked.

Note for the reviewer: **CodeRabbit is rate-limited on this PR and reports `pass` without having
reviewed the last two commits** (it reviewed and I addressed two findings before that, one of them
a real credential-leak path). Treat its green as absent, not as approval.

- [x] I have checked that any arbitrary calls to external contracts are validated and or restricted — n/a
- [x] I have checked that any privileged calls (i.e. storage modifications) are validated and or restricted — n/a
- [x] I have ensured that any new contracts have had AT A MINIMUM 1 preliminary audit — n/a, no contract changes

### DIFF
--- a/.github/workflows/healthCheckAllNetworks.yml
+++ b/.github/workflows/healthCheckAllNetworks.yml
@@ -18,8 +18,11 @@
 #   deploymentAddressConsistency.yml for the static PR tier).
 # - Slack: SIGNAL-ONLY. A message is posted only when a run surfaces something actionable — a failed
 #   network or runner error (:rotating_light: ACTION NEEDED) or a network that passed with warnings
-#   (:warning:). Fully-green runs, and the staging-only no-op push that resolves zero production
-#   networks, stay SILENT so the channel carries only issues, not routine confirmations.
+#   (:warning:). Failures are reported grouped by root cause (one bullet per cause, with the
+#   networks it hit) rather than as a flat network list, so the alert is triageable without opening
+#   the run; the grouping is computed in healthCheckAllNetworks.ts. Fully-green runs, and the
+#   staging-only no-op push that resolves zero production networks, stay SILENT so the channel
+#   carries only issues, not routine confirmations.
 #   Trade-off: no green message means no positive heartbeat, so a silently-broken cron no longer
 #   shows up as a MISSING message — add a separate cron-liveness monitor if that signal is needed.
 # - Known limitations: needs RPCs (MongoDB) + deploy logs; transient RPC errors are absorbed by viem
@@ -156,6 +159,7 @@ jobs:
           WARNED_COUNT: ${{ steps.healthcheck.outputs.warned_count }}
           FAILED_NETWORKS: ${{ steps.healthcheck.outputs.failed_networks }}
           WARNED_NETWORKS: ${{ steps.healthcheck.outputs.warned_networks }}
+          FAILURE_DIGEST: ${{ steps.healthcheck.outputs.failure_digest }}
         run: |
           # Deep-link straight to THIS job so the Slack link lands on the failing step; fall back to
           # the run overview if the jobs API can't be read.
@@ -180,8 +184,17 @@ jobs:
             BODY="<${RUN_URL}|View workflow run>${WARN_LINE}"
           else
             HEADER=":rotating_light: *Diamond health check — ACTION NEEDED*"
-            BODY=$(printf '%s of %s network(s) failed: %s\n<%s|View workflow run>%s' \
-              "${FAILED_COUNT:-?}" "${TOTAL:-?}" "${FAILED_NETWORKS:-unknown (runner error)}" "$RUN_URL" "$WARN_LINE")
+            # Prefer the grouped cause digest: the flat network list says how many chains are red
+            # but not why, which is the part that decides whether anyone has to act. It is absent
+            # when the runner itself died before producing results, so keep the list as fallback.
+            if [[ -n "${FAILURE_DIGEST:-}" ]]; then
+              FAILURE_LINES=$(printf '%s of %s network(s) failed:\n%s' \
+                "${FAILED_COUNT:-?}" "${TOTAL:-?}" "$FAILURE_DIGEST")
+            else
+              FAILURE_LINES=$(printf '%s of %s network(s) failed: %s' \
+                "${FAILED_COUNT:-?}" "${TOTAL:-?}" "${FAILED_NETWORKS:-unknown (runner error)}")
+            fi
+            BODY=$(printf '%s\n<%s|View workflow run>%s' "$FAILURE_LINES" "$RUN_URL" "$WARN_LINE")
           fi
           TEXT=$(printf '%s\n%s' "$HEADER" "$BODY")
           {
--- a/.github/workflows/healthCheckAllNetworks.yml
+++ b/.github/workflows/healthCheckAllNetworks.yml
@@ -18,9 +18,9 @@
 #   deploymentAddressConsistency.yml for the static PR tier).
 # - Slack: SIGNAL-ONLY. A message is posted only when a run surfaces something actionable — a failed
 #   network or runner error (:rotating_light: ACTION NEEDED) or a network that passed with warnings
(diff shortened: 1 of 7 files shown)


0xDEnYO:
@coderabbitai Good catch — fixed in bbd323d, and it was a real hole.

A failure `detail` can be a raw viem or Mongo error message, and this PR is what newly routes that text to Slack, which sits outside the workflow log's masking. The failing test made it concrete before the fix landed: the endpoint flowed through `normalizeFailureCause` untouched.

Rather than write a second redactor, I reused the one this repo already had — `redactErrorReason` in `script/deploy/safe/reconcile-parked-tasks.ts`, whose docstring describes exactly this hazard for the parked-task alerts. It is now extracted to `script/utils/redactUrls.ts`, so the health check does not have to import the Mongo-backed reconcile module to reach it. `redactUrls` takes a configurable placeholder because the default has angle brackets that Slack parses as link syntax; the digest passes `[redacted-url]`.

Redaction runs first, before the address and count masking. Four tests cover it, including one asserting that two networks failing on the same RPC fault still group together once their differing endpoints are redacted.

One thing I did *not* do: per-component masking of query-parameter values and token-like path segments. The pattern replaces the entire `scheme://…` token up to the first whitespace, so anything sensitive inside the URL goes with it. Narrower masking would leave the host visible for no real gain on an alert line.

0xDEnYO:
@coderabbitai Agreed, and you were more right than the finding says — fixed in 21b03c2.

`redactErrorReason`'s existing consumer is the parked-task Slack alert, so the angle-bracket marker was already latently wrong there, independently of this PR. That makes the placeholder parameter pointless: every consumer of this module publishes to Slack, so there is no caller who wants the unsafe form. Rather than change the default and leave the choice available, I dropped the parameter — the marker is now a single module-level constant.

Follow-ons: updated the two `reconcile-parked-tasks` assertions that pinned the old marker, and added `script/utils/redactUrls.test.ts` with direct coverage of the default, including that the marker carries no Slack link syntax, that every endpoint in a string is replaced rather than only the first, that non-http schemes match, and that redaction happens before length-capping so a long endpoint cannot survive truncation.

0xDEnYO:
@coderabbitai review

melianessa (approved):

PR #2260, opened 2026-08-24
fix(clear-signing): render destinationChainId via a LI.FI-authored enum (EXSC-827)

0xDEnYO:

https://linear.app/lifi-linear/issue/EXSC-827

## The problem

PR #2217 (EXSC-793) switched `_bridgeData.destinationChainId` from `"format": "raw"` to `"format": "chainId"` so wallets would show a network name instead of a number. That broke the ERC-7730 registry's `testsv2` CI on both of its independent runners, and blocked https://github.com/ethereum/clear-signing-erc7730-registry/pull/2901.

The root cause is not the test harness. `format: chainId` is defined as *"Value is converted to a Blockchain name using EIP-155 reference values"* (`specs/erc-7730.md:1411-1416`) and — unlike `tokenAmount` or `nft` — **defines no fallback** for an unresolvable value. But our `destinationChainId` is not an EIP-155 id space: `src/Helpers/LiFiData.sol` mints synthetic ids for every non-EVM destination (Solana `1151111081099710`, Tron `1885080386571452`, Stellar, Sui, Aptos, the Bitcoin family, HyperCore `1337`). For those values the rendering is implementation-defined, and the two runners the registry pins genuinely diverge:

| destinationChainId | `llbartekll/clear-signing` | `sourcifyeth/clear-signing-test-runner` |
| --- | --- | --- |
| `8453` | `Base` | `Base` |
| `56` | `BNB Chain` | `BNB Smart Chain Mainnet` |
| `4663` | `Chain 4663` | `Robinhood Chain` |
| `1151111081099710` | `Chain 1151111081099710` | `1151111081099710` |

They disagree on the resolved name, on unknown-but-EIP-155 ids, and on the fallback shape itself. Registry PR #2901 (fixture expects raw numbers) fails both runners; registry PR #2903 (fixture tuned to Sourcify) passes Sourcify and fails Rust. Empirically, no fixture satisfies both.

More importantly this was never only a CI problem: **a real wallet has no test fixture**, so a user bridging to Solana was signing a screen that read *"to chain Chain 1151111081099710"*. Making CI green while leaving that on the device would have removed the signal rather than fixed the bug.

## Why `enum` rather than reverting to `raw` or fixing the test runners

`format: enum` against `$.metadata.enums.lifiChains` makes the rendered string a pure function of the descriptor: identical on every wallet and every runner, no network lookup, no provider, no undefined behaviour, and correct for non-EVM destinations. It is also entirely within our control — no third-party runner change and no registry ref bump on the critical path.

It follows the registry's own precedent for exactly this situation, a project-specific non-EIP-155 destination id space including non-EVM members:

- `registry/threshold/calldata-L2WormholeGateway.json` → `$.metadata.enums.wormholeChain` (`"21": "Sui"`)
- `registry/hyperliquid/calldata-CctpExtension.json` → `$.metadata.enums.cctpDomains`

Notably, **zero descriptors in the registry use `format: chainId`** — ours would have been the first.

Options considered and rejected:

- **Revert to `raw`.** Deterministic and safe, but gives up the network name that EXSC-793 delivered and Ambire asked for.
- **Fix the runners upstream** (add a fixture-fed chain-name data provider). Both runners route every other implementation-defined lookup through the fixture's `dataProvider`; chain names are the sole exception. Worth reporting, but it is a *test-only* fix — it would make CI green while a real wallet still showed `Chain 1151111081099710` — and it needs two third-party merges plus a registry ref bump before anything goes green.
- **Scope the fixtures to chain ids where both runners agree.** Only 7 ids qualify, none of them non-EVM, so it would have meant deleting the Solana/Tron bridge test cases.

## Why the enum has 32 entries and not ~80

`erc7730 lint`, which the registry runs in its `validate descriptors` job, rejects a larger enum:

```text
Invalid value at $.metadata.enums.lifiChains: Dictionary should have at most 32 items after validation, not 81
```

The cap lives in the linter's `EnumDefinition` model (`max_length=32`). The registry's own `specs/erc7730-v2.schema.json` sets **no** limit, and `check-jsonschema` accepts the 81-entry version — so this looks like an unintended tool-vs-spec divergence. It is raised in the registry PR, and the generator's error message points at it so the map can be expanded in one line if the cap is lifted.

The 32 chosen cover every non-EVM destination — the ids a signer has no chance of recognising, and where `chainId` was most broken — plus the EVM chains with the broadest bridge-route support across `config/*.json`. An unlisted id renders as its decimal number: identical on both runners, and no worse than the `raw` format this replaces. The generator logs which chains fall back, so the tradeoff is visible on every run rather than silent.

## What changed

- `tasks/buildClearSigningProposal.ts` — `CHAIN_FIELD` becomes `format: enum` with `params.$ref: $.metadata.enums.lifiChains`; emits the map as `enums` in the proposal; drops the literal word "chain" from both interpolated-intent templates (they otherwise render *"to chain BNB Chain"* / *"to chain Chain 4663"*).
- `tasks/generateLedgerClearSigning.ts` — merges `metadata.enums` with the same ownership contract already applied to `display.formats` (names we declare are replaced, everything else preserved verbatim). Without this the map would have to be hand-maintained in the registry, where it would silently go stale the moment we add a network.
- `config/clearSigningProposal.json` — regenerated (`bunx tsx tasks/buildClearSigningProposal.ts`), never hand-edited.

Three new generator invariants, each verified to actually fire:

| invariant | verified by |
| --- | --- |
| enum must hold ≤ 32 entries | added a 33rd → exit 1, message names the linter cap |
| every named id must exist in `networks.json` or `LiFiData.sol` | swapped in `999999999` → exit 1, message names the id |
| `LiFiData.sol` must still parse | guard throws if the constant regex matches nothing |

## Testing

The descriptor half of this change lands in the ERC-7730 registry, so it was verified there rather than only in unit tests. Ran `tasks/generateLedgerClearSigning.ts` against a real checkout of `ethereum/clear-signing-erc7730-registry@master`, then ran **both** runners at the exact SHAs the registry pins in `.github/actions/run-*-tests/action.yml`:

| check | result |
| --- | --- |
| `sourcifyeth/clear-signing-test-runner` @ `dae3cda` | **14/14 pass** |
| `llbartekll/clear-signing` @ `10605ba` | **14/14 pass** |
| `erc7730 lint` | exit 0 |
| `erc7730 format` | no-op on the generated file |
| `check-jsonschema` vs `erc7730-v2.schema.json` | ok |
| `check-jsonschema` vs `erc7730-tests-v2.schema.json` | ok |
| generator determinism | two consecutive runs byte-identical |
| `bunx tsc-files --noEmit`, `bunx eslint` | clean |

Companion registry PR (descriptor + `testsv2` expectations in one atomic commit, so `master` is never inconsistent): https://github.com/ethereum/clear-signing-erc7730-registry/pull/2914

## Follow-ups this supersedes

- #2225 (`clear-signing-hypercore-chainid`, EXSC-798) — kept `destinationChainId` raw where a LI.FI id collides with EIP-155. `1337` → `HyperCore` in the enum resolves that collision properly; the PR can be closed.


0xDEnYO:
## Review gate — escalated findings (need a decision before this goes ready)

Auto-fixed and pushed in 57ecd79: `verifyClearSigning.yml` trigger paths, a dangling-`$ref` assertion in the sync generator, the stale `docs/ClearSigningProposal.md`, `--skipDisplayMerge` help text, two comment-discipline violations. Registry PR gained an enum-miss regression case (both runners 15/15).

Three findings were **not** applied because they need a human call.

### 1. 🔴 Ledger's on-device encoder cannot represent enum keys > 255

`erc7730 calldata` — the converter that produces the on-device v1 descriptor — rejects our enum outright:

```text
$ erc7730 calldata registry/lifi/calldata-LIFIDiamond.json
🔴 error: Invalid value at $.value: Value "999" is not valid: Input should be less than or equal to 255
```

`convert/calldata/v1/enum.py` does `value=int(ordinal)` into a field constrained `ge=0, le=255`. **Every** chain id we use exceeds that. The enum primitive's own schema example is `{"1": "stable", "2": "variable"}` — small ordinals — whereas `format: chainId` is the spec's purpose-built primitive for this field.

Why it does not block CI: the registry never runs `erc7730 calldata`, and the Ledger Speculos test jobs are **explicitly disabled** in `clear-signing-tests.yml` (they depend on the private `LedgerHQ/coin-apps`, registry issue #2459). Note `erc7730 calldata` also exits 0 while emitting that error.

So both runners going green is **not** evidence this renders on a Ledger device. This is the one thing that could invalidate the approach, and it cannot be verified from outside — it needs an answer from the EF/Ledger side. Raised as a question on registry PR #2914; carrying it into the Telegram ping.

**Decision:** proceed on the assumption that a >255 enum key is acceptable on-device (and revert to `raw` if Ledger says otherwise), or hold this PR until they confirm.

### 2. 🟠 Dropping the word "chain" makes the sentence worse for the 49 unnamed destinations

The templates changed from `… to chain {destinationChainId} …` to `… to {destinationChainId} …`, which avoids reading "to chain BNB Chain" for a named destination. But 49 of ~81 destinations are not in the 32-entry enum, and for those the rendered sentence regresses against `main`:

| destination | on `main` | on this branch |
| --- | --- | --- |
| Base (named) | `to chain Base` | `to Base` ✅ |
| Celo (unnamed) | `to chain 42220` | `to 42220` ⚠️ |

Keeping the word is strictly safer — with `enum` there is no `Chain N` fallback, so it yields `to chain Base` (mildly redundant) and `to chain 42220` (fine), never anything broken. **My recommendation: revert the template change and accept the redundancy.** Not applied because it changes what appears on a signing screen.

### 3. 🟠 zkSync Era (324) is omitted from the 32 by an undocumented tiebreak

Ranking the 32 by bridge-route breadth across `config/*.json` leaves a six-way tie at 10 configs (`robinhood`, `megaeth`, `sei`, `tempo`, `worldchain`, `zksync`), and only Robinhood made the cut. zkSync Era is the more recognisable destination; it currently renders as `324`. Swapping requires dropping another entry, which is a product call.

**Decision:** leave as is, or name the swap.

- The `LIFI_CHAIN_ID_*` regex matches only single-line decimal declarations. A future constant written across two lines, in hex, or with a numeric separator would be silently omitted from the known-id set. It degrades loudly rather than dangerously (the "not declared anywhere" invariant then fires), so it was left alone.
- On an enum miss the Sourcify implementation emits an internal per-field `FORMAT_PARAM_RESOLUTION_ERROR`; the runner only escalates top-level warnings, so CI is green today. If the registry ever escalates per-field warnings, all 49 unlisted chains would surface warnings at once.

0xDEnYO:
## ⚠️ Do not merge — the enum approach does not work on a Ledger

Following up on escalation 1 in my earlier comment. I went and answered it, and the answer kills this approach.

`erc7730/convert/calldata/v1/tlv.py:255` encodes an ENUM_VALUE entry as **one byte**:

```python
out += tlv(CalldataDescriptorEnumValueTag.VALUE, obj.value.to_bytes(1))
```

two lines below `obj.chain_id.to_bytes(8)`, so the width is deliberate, and the model bound (`ge=0, le=255`) matches. `convert_enums` takes that value from the enum **key** — our chain id. This is the on-device struct the Ledger Ethereum app parses, so it is not something a package release can lift.

Measured against the real registry descriptor:

| variant | `erc7730 calldata` (on-device) | `testsv2`, both runners |
| --- | --- | --- |
| `raw` (registry master) | 25,382,949 bytes ✅ | green |
| `chainId` (#2217, what this PR replaces) | 25,385,397 bytes ✅ | **fails both** |
| `enum` (this PR) | **`[]` — descriptor dropped entirely** ❌ | green |

So this branch is the only one of the three that takes `calldata-LIFIDiamond` from producing an on-device descriptor to producing nothing at all. On a Ledger that is not a worse label — it is **no clear signing for the diamond**. Strictly worse than the bare number this set out to fix.

It also fails silently: `erc7730 calldata` exits 0 while emitting the error, `erc7730 lint` passes, `verify-clear-signing` passes, and both `testsv2` runners pass 15/15. None of them touch that path.

And the precedent I leaned on in the PR body does not hold: `threshold`'s `wormholeChain` (2, 5, 21, 23, 24, 30) and `hyperliquid`'s `cctpDomains` convert fine **because their ids are ≤ 255**. They are precedent for a small-ordinal destination enum, not for an EIP-155 chain-id enum. Restricting our enum to keys ≤ 255 would cover 19 chains and exclude Base, Arbitrum and every non-EVM id — the exact cases that motivated the change.

### Where this leaves the ticket

| option | on-device | both runners | user sees |
| --- | --- | --- | --- |
| `raw` | ✅ | ✅ | bare number |
| `chainId` | ✅ | ❌ | wallet-dependent name; undefined for our ~10 non-EVM ids |
| `enum` | ❌ | ✅ | consistent name, 32 chains — but nothing on Ledger |

`raw` is the only variant that works everywhere. My recommendation is to **revert the format change on this branch and keep the rest** — the intent-template fix, the `verifyClearSigning` trigger paths, the dangling-`$ref` guard and the docs update all stand on their own merits — then take the real problem upstream.

The smallest correct upstream fix is for the spec to **define a fallback for `chainId`**, as it already does for `tokenAmount` ("raw value plus an Unknown token warning", `erc-7730.md:1356`) and `nft` (`:1369`). That would make `chainId` deterministic across implementations, remove the `testsv2` divergence entirely, and be honest about non-EIP-155 ids — without touching any device firmware. Raised on [registry#2914](https://github.com/ethereum/clear-signing-erc7730-registry/pull/2914#issuecomment-5390414530); registry PR converted back to draft.

Holding here for a decision rather than pushing the revert unprompted, since it changes what this PR is.

0xDEnYO:
Closing in favour of #2261.

The `format: enum` approach in this PR does not work: the on-device ENUM_VALUE struct encodes an entry value in a single byte (`obj.value.to_bytes(1)` in `erc7730/convert/calldata/v1/tlv.py:255`) taken from the enum key, so no chain id above 255 is representable. `erc7730 calldata` drops the whole descriptor to `[]` while exiting 0 — no clear signing at all for the diamond on a Ledger — with `lint`, `format`, `verify-clear-signing` and both testsv2 runners green. Details in the comment above.

Probing further also showed `chainId` cannot be used either: it renders HyperEVM (999) as "Wanchain Testnet" and HyperCore (1337) as "Geth Testnet". #2261 reverts to `raw` and explains both.

Worth keeping from this PR if we revisit: the `verifyClearSigning` trigger paths for `config/networks.json` and `src/Helpers/LiFiData.sol`, and the dangling-`$ref` assertion in the sync generator — both only matter once the generator reads those files or emits an enum, so they are out of scope for the revert.

PR #472, opened 2023-07-11
LF-2894 zkSync era integration

satoshi-lifi:

PR #2262, opened 2026-08-24
chore(deps): audit all declared dependencies and remove the 13 dead ones (EXSC-839)

0xDEnYO:

https://linear.app/lifi-linear/issue/EXSC-839/audit-all-declared-dependencies-for-real-usage-remove-the-dead-ones

**Stacked on #2257** (`fix/exsc-834-aikido-dep-bumps`) — that PR removed `@uma/sdk`,
`@layerzerolabs/lz-v2-utilities` and `defender-relay-client`. Merge #2257 first, then rebase/merge this.

#2257 found three unused direct dependencies **by accident** while triaging Aikido, and removing each
cleared findings no override could fix. This PR generalises that: an exhaustive audit of all 59 entries
in `dependencies` + `devDependencies`, not just the ones a scanner happened to flag.

**Method** — for each package, check module specifiers (`from`/`require()`/`import()`, incl. subpaths)
across `*.ts *.js *.mjs *.cjs`, plus `package.json` `scripts`, `.github/workflows/**`, `.husky/**`,
`lint-staged`, `.eslintrc.cjs`, `.solhint.json`, `.prettierrc.cjs`, `tsconfig.json` `types`,
`plopfile.mjs`, `templates/**`, `preinstall.sh` and every tracked `*.sh`.

Two things a naive grep gets wrong, both of which mattered:

- **Config shorthand.** `extends: ['prettier']` → `eslint-config-prettier`; `plugins: ['import']` →
  `eslint-plugin-import`; `plugins: ['@typescript-eslint']` → `@typescript-eslint/eslint-plugin`;
  `settings['import/resolver'].typescript` → `eslint-import-resolver-typescript`; `.solhint.json`
  `plugins: ["prettier"]` → `solhint-plugin-prettier`; `typechain --target ethers-v5` →
  `@typechain/ethers-v5`; `tsconfig` `types: ["node","bun"]` → `@types/node`, `@types/bun`.
  All seven look dead to grep and are **kept**.
- **Negative-only git pathspecs match nothing.** `git grep -- ':(exclude)bun.lock'` silently returns
  zero hits for everything, which reads as "all dead". Every scan here carries a positive `.` pathspec.

## Removed (13) — each referenced nowhere outside its own `package.json` line

| Package | Why it's dead | What left the tree with it |
|---|---|---|
| `@uniswap/sdk` (dep) | No import anywhere. Distinct from `@uniswap/permit2-sdk`, which **is** used (`demoPermit2.ts`). Present since the initial commit. | `@uniswap/v2-core`, `big.js`, `decimal.js-light`, `jsbi`, `toformat`, `tiny-warning` |
| `enquirer` (dep) | Added for the Safe scripts in #518; no longer imported. | nothing — remains installed as an **optional peer** of `listr2` (`peerDependencies: {enquirer: ">= 2.3.0 < 3"}, optionalPeers: ["enquirer"]`), not a plain transitive. Nothing prompts through listr2. |
| `light-spinner` (dep) | No import. Initial commit. | — |
| `notify-send` (dep) | No import. Added in #309. | — |
| `pino` (dep) | No import; nothing logs through it (repo uses `consola`). | `atomic-sleep`, `fast-redact`, `on-exit-leak-free`, `pino-abstract-transport`, `pino-std-serializers`, `process-warning`, `quick-format-unescaped`, `real-require`, `safe-stable-stringify`, `sonic-boom`, `split2`, `thread-stream`, `duplexify`, `stream-shift` |
| `@types/pino` (dev) | **Stub package** — its own README says "pino provides its own type definitions, so you don't need @types/pino installed". Sole remaining consumer of `pino`. | — |
| `@types/minimatch` (dev) | **Stub package**, same shape. `minimatch` isn't even a declared dep or imported anywhere. | — |
| `cross-env` (dev) | Not in any `scripts` entry, workflow, or shell script. Initial commit. | — |
| `ethers6` (dev, alias `npm:ethers@^6.12.1`) | Nothing imports `ethers6`. Added in #645 for the Safe scripts, which now use v5 / viem. | `ethers@6.15.0` (the alias entry), `@noble/curves@1.2.0`, `@noble/hashes@1.3.2`, a nested `@types/node@22.7.5` |
| `fs-extra` (dev) | No import. Remains in the tree as a transitive of `patch-package`, `solhint` and `typechain`. | `fs-extra@10.1.0` only |
| `postinstall-postinstall` (dev) | Exists solely so `patch-package` survives `yarn remove`. Repo is bun-only (`bun.lock` is the sole lockfile). | — |
| `ts-generator` (dev) | Legacy `typechain` **v7** dependency. Installed `typechain@8.3.2` does not depend on it (verified against its `package.json`); nothing else in the lock does either. | `@types/mkdirp`, `@types/resolve`, `chalk@2.4.2` + its `ansi-styles`/`color-convert`/`color-name`/`escape-string-regexp`/`supports-color`/`has-flag`, `glob@7.2.3`, `ts-essentials@1.0.4`, `mkdirp@1.0.4`, nested `@types/node@22.7.5` |
| `zksync-web3` (dev) | Deprecated package; no import. Added in #472 for the zkSync Era integration, which now goes through the standard tooling. | `ws@8.21.0` (nested) |

**Net: 52 packages leave the lockfile (1265 → 1213).**

## Deliberately kept — candidate that is NOT dead

- **`node-gyp`** — grep-invisible but **load-bearing**. `node-hid` (pulled by
  `@ledgerhq/hw-transport-node-hid`, dynamically imported in `script/deploy/safe/ledger.ts:49`) has
  `"install": "prebuild-install --runtime napi || node-gyp rebuild"`. In this checkout the prebuild
  download failed and it took the fallback — `node_modules/node-hid/build/Release/HID.node` exists
  while `prebuilds/` is empty. The fallback resolves `node-gyp` from the root `node_modules/.bin`,
  which only exists because of this devDependency. Removing it would break installs exactly where it
  hurts most (a machine or runner with no prebuilt binary) and would look like a toolchain fault.

## Aikido impact

**None — this PR clears zero Aikido findings.** Cross-referencing all 53 removed lockfile entries
against the open feed for `contracts` produces no overlap. The remaining open findings
(`fast-uri`, `smol-toml`, `validator`, `undici`, `follow-redirects`, `zod`, `katex`, `lodash`,
`lodash-es`, `viem`, `yargs`) all reach the tree through dependencies that are genuinely used, and
the two AGPL/BUSL license criticals are handled by #2257. Flagging this explicitly because the
premise of the exercise was that dead deps carry findings — here the payoff is dead-weight and
supply-chain-surface reduction (52 fewer packages), not finding clearance.

## Overrides re-asserted

All 9 entries still match a real lockfile version after the removals — none went dead:

```text
live handlebars -> 4.7.9      live axios     -> 1.18.1      live fast-uri -> 3.1.5
live protobufjs -> 7.5.5      live smol-toml -> 1.7.1       live lodash   -> 4.17.23
live socks      -> 2.8.9      live validator -> 13.15.26    live katex    -> 0.18.2
```

## Lockfile version-change accounting (3 changes, all hoist reshuffles)

Every change is a hoist slot moving after its previous occupant left. Each surviving parent's range
was re-checked against what it now resolves to — **no consumer is unsatisfied, none went unresolved**:

| Package | Change | Why, and who still consumes it |
|---|---|---|
| `fs-extra` | `10.1.0` → `9.1.0` (**downward**) | The removed root devDep `^10.0.1` was the only thing wanting v10 and held the top slot. `patch-package` (`^9.0.0` → 9.1.0) inherits it. `solhint` (`^11.1.0` → nested 11.3.0) and `typechain` (`^7.0.0` → nested 7.0.1) keep their own nested copies, unchanged. |
| `mkdirp` | `0.5.6` → `1.0.4` (**upward**) | `mkdirp@0.5.6` was `ts-generator`'s. With it gone nothing wants 0.5.x, so the three `^1.0.x` consumers (`node-plop`, `tar`, `typechain`) hoist 1.0.4 and their three nested duplicates disappear — a net reduction in copies, not an added one. |
| `ws` | `8.17.1` → `8.21.0`, plus one **new** nested `ws@8.17.1` | `ethers6` (ethers 6.15.0) pinned `ws` to exactly `8.17.1` and held the slot. Gone → `8.21.0` (exact pin from `viem` and `tronweb/ethers`) hoists. `@mayanfinance/swap-sdk/ethers` still needs exactly `8.17.1`, so it gets the one nested copy — the single lockfile insertion. `@ethersproject/providers` (8.18.0), `jayson` (7.5.10), `rpc-websockets` (8.18.2), `isomorphic-ws`/`isows` (`*`) all still resolve in range. |

The `ws` movement is a net **improvement**, not a regression: top-level `ws` goes **up** 8.17.1 →
8.21.0, and the one new nested `ws@8.17.1` is the same version that was previously hoisted for the
whole tree, so no new version enters. `ws` carries no open Aikido finding for this repo.

`ethers` itself is untouched: the hoisted copy is still `ethers@5.8.0` (root `ethers: ^5.7.0`), and the
v6 consumers keep their nested copies (`@mayanfinance/swap-sdk` → 6.15.0, `tronweb` → 6.17.0).

Lockfile pruned properly — of the 13 removed names, **11 are fully gone** from `bun.lock`. Two
legitimately remain, both reachable from surviving roots:

- `enquirer@2.4.1` — via `lint-staged@13.3.0 → listr2@6.6.1`, which declares it an **optional peer**
  (`optionalPeers: ["enquirer"]`), not a dependency. Satisfied, and nothing prompts through listr2.
- `fs-extra` at 9.1.0 / 11.3.0 / 7.0.1 — via `patch-package@6.5.1` (`^9.0.0`), `solhint` (`^11.1.0`)
  and `typechain` (`^7.0.0`) respectively.

`bun install` (non-frozen) re-adds nothing and leaves `bun.lock` byte-identical.

## Flagged for a human call — used, but thinly

Not touched in this PR; listed because they are the `@layerzerolabs` shape (a dependency held alive by
one script) and the right fix may eventually be deleting the script:

- **`@arbitrum/sdk`** — sole consumer `script/demoScripts/demoArbitrumBridge.ts`, which has 5
  pre-existing type errors, one of them a config-shape mismatch (`Property 'l2GatewayRouter' does not
  exist`). It reaches into a deep dist path (`@arbitrum/sdk/dist/lib/message/L1ToL2MessageGasEstimator`).
- **`@mayanfinance/swap-sdk`** — `demoMayan.ts` is clean, but `demoMayanSolana.ts` has 5 errors that are
  clearly SDK API drift (`'mayanAddr' does not exist in type 'MayanDataStruct'`, `Expected 8-9 arguments,
  but got 7`). The dep stays; the Solana demo looks unrunnable.
- **`@uniswap/permit2-sdk`** (sole consumer `demoPermit2.ts`, 3 errors) and **`merkletreejs`**
  (sole consumer `demoGenerateMerkleTree.ts`, 3 errors) — same pattern, smaller.
- **`patch-package`** — genuinely invoked (`postinstall`), but there is **no `patches/` directory**, so
  it is currently a no-op. Kept because removing it also means removing the postinstall step.

All of the above are pre-existing and unrelated to this diff; they're reported, not fixed.

## Verification

Baselines taken on the parent commit before any change, so only deltas count.

**Important caveat, and why the numbers below are trustworthy:** `bun install` does **not** prune
`node_modules`. After removing the 13, all of them were still physically on disk in this worktree
(`node_modules/pino`, `node_modules/ethers6`, `node_modules/cross-env`, … with unchanged mtimes), so a
warm-tree test run is **not** evidence the removals are safe — a stale copy on disk happily satisfies
an import the lockfile no longer provides. The suite below was therefore re-run after moving
`node_modules` aside and reinstalling from scratch (1052 packages), with all 13 confirmed absent from
disk and only the expected survivors (`enquirer`, `fs-extra`, `node-gyp`) present.

```text
                                        baseline          after (pruned tree)   delta
bun test script/                        984 pass/0 fail   984 pass/0 fail         0
bunx eslint script/utils/utils.ts       exit 0            exit 0                  0
bunx tsc --noEmit -p tsconfig.json      47 errors         47 errors               0
bun lint:js (repo-wide)                 151 probs/1 err   151 probs/1 err         0
validateScripts --base origin/main      pass              pass                    —
bun install --frozen-lockfile           clean             clean                   —
bun install (non-frozen)                —                 no lockfile change      —
git status                              clean             clean                   —
```

- The eslint canary is the one that matters: an `ajv` override once broke eslint fatally while `tsc`
  and all 984 tests stayed green. Run explicitly, exit 0.
- Repo-wide `bun lint:js` was run because `import/no-extraneous-dependencies` is **enabled** — it would
  flag any import of a package no longer declared. Its single error
  (`Unable to resolve path to module '../../diamond.json'` in `demoEco.ts`/`demoLidoWrapper.ts`) is a
  missing generated artifact and is **byte-identical on the pristine base commit** (verified by
  stashing the change and re-running).
- `validateScripts` is the repo's own gate for exactly this failure mode — it verifies every bare
  import under `script/**` resolves against `package.json`. It passes against the reduced manifest.
- The 47 `tsc` errors are the pre-existing demo-script errors described above; count unchanged.
- `forge build` not run — no Solidity touched.

**Note for reviewers:** trust CI (fresh checkout) over any local `bun test` for this PR, for the
`node_modules`-pruning reason above. `bun-lockfile-check` (`bun install --frozen-lockfile` +
`git diff --exit-code bun.lock`) is the load-bearing CI job here and already passes. The
`run-ts-tests` / `run-unit-tests` jobs report `skipped` while this PR is a draft — that is the draft
gate, not a path-filter gap (`ts-unit-tests.yml` explicitly filters on `package.json` and `bun.lock`) —
so they execute on the flip to ready.


0xDEnYO:
@coderabbitai review

PR #2264, opened 2026-08-24
fix(timelock): pre-check the fleet over one MongoDB connection, stop silently skipping networks (EXSC-841)

0xDEnYO:

Fixes [EXSC-841](https://linear.app/lifi-linear/issue/EXSC-841/execute-pending-timelock-tx-dns-fan-out-silently-skips-networks-plus)

A single `bun execute-timelock` run produced 28 errors from three independent causes. The first is a correctness problem, not noise.

**The prefetch silently skipped 27 networks.** `fetchPendingForNetwork` was fanned out over all 71 active networks with a bare `Promise.all`, and every call opened its own `MongoClient` and closed it again. With a `mongodb+srv://` URI each client resolves its own SRV _and_ TXT record before connecting, so a run fired ~142 DNS queries in the same second at the local resolver. It rate-limited them, 27 timed out (`querySrv ETIMEOUT` / `queryTxt ETIMEOUT`), and those networks — **mainnet, base and bsc among them** — were never checked for ready operations. The run's headline ("2 have pending timelock tx(s)") was only true of the 44 that resolved; a ready op on mainnet would have gone unexecuted behind a single `warn` line.

The fix is one connection and one `$in` query for the whole fleet, so the per-network DNS resolution disappears rather than being retried. I deliberately did **not** build this on EXSC-794 / #2218: that helper retries via public DNS on a malformed SRV answer, which is a different failure. It would not have saved this run — it gates on `syscall === 'querySrv'` so the `queryTxt` timeouts rethrow untouched, its `fallbackApplied` latch is once-per-process so the 26 concurrent failures after the first skip the retry entirely, and 142 simultaneous queries would likely time out against a public resolver too. The two changes are complementary and touch disjoint files: after this lands, #2218's helper covers the single remaining connect instead of 71.

Rejected alternative: a concurrency limiter around the existing per-network client. It would have reduced the DNS burst without removing it, and left the connect-per-network cost in place for what is a read-only count.

**The prefetch logic now lives in `timelock-prefetch.ts`.** `execute-pending-timelock-tx.ts` calls `runMain` at module scope and so cannot be imported by a test. The `import.meta.main` guard used by siblings in that directory is not an option: the CLI runs under `bunx tsx`, where `import.meta.main` is `undefined` — I verified this, and adding the guard would silently turn the command into a no-op. Extracting the logic was the only way to get it under test. The two near-identical prefetch blocks in the `--executeAll` and interactive branches also collapse into one helper, so the fix cannot land in one branch and be missed in the other.

**`tronshasta` was an expected skip reported as a failure.** It is `status: active` but has no `deployments/tronshasta.json` (the only active testnet without one), so `getDeployments` threw and the prefetch logged a stack trace and counted it among the failures. A missing deployments file now classifies as a skip, distinct in both the output and the counts from a genuine fetch failure. The summary line reports "Checked N of M" so every network is accounted for.

The skip requires the file to be genuinely **absent**, which is why `getDeploymentsFilePath` is now exported: `getDeployments` reports a corrupt or unreadable file as not-found too, so treating every throw as a skip would silently drop a network that does have a timelock — the same bug class this PR exists to fix, in a narrower form. Anything other than an absent file is recorded as a per-network fetch error instead.

**Init calldata failed to decode when arguments contain nested bigints.** `formatDecodedArg` handled a top-level `bigint` but fell through to a plain `JSON.stringify` for tuples and arrays, which throws on nested bigints; both call sites swallow the throw. Hit on the FraxFacet cut on worldchain — `initFrax` carries `(chainId, eid)` pairs, so **the cut was approved at the interactive prompt with its init arguments never displayed**. That is a review-safety gap in the confirmation path, which is why it is fixed here rather than deferred.

## Verification

Run against the real fleet and the production timelock queue, before and after:

|                      | before               | after                        |
| -------------------- | -------------------- | ---------------------------- |
| networks reported on | 44 of 71             | **71 of 71**                 |
| prefetch failures    | 27                   | **0**                        |
| Mongo connections    | 71                   | **1**                        |
| `tronshasta`         | error + stack trace  | skip (`no-deployment-log`)   |

A full-fleet dry run reports `Checked 66 of 71 network(s); 1 have pending timelock tx(s): tron` with 5 skips itemised and no failures. Skips resolve correctly against real deployment logs: 4 testnets `no-timelock-deployed`, `tronshasta` `no-deployment-log`. Dry-runs also cover the single-network skip path, the missing-log path and the reached-but-empty path, and tron's queued op was confirmed untouched afterwards.

Every fix here was negative-controlled — each was reverted in turn and the corresponding test observed to fail (BigInt replacer: 2 tests; missing-deployments skip: 1; absent-vs-unreadable guard: 1) — so none of them pass vacuously.

The exit-1 path for "0 pending but some networks unreachable" is covered by unit test rather than by a live outage.

## Review round 2

Six findings from @melianessa, all answered in the threads; five changed code (`2cd33583f`).

- **A prefetch failure now fails the run even when other networks had work.** `mustExitWithError` only fires at zero pending, so a network that was never checked was passed over silently the moment any other chain had a queued op — the reporting half of the same silent skip this PR exists to close. `prefetchNetworksWithPendingOps` returns `failedCount`; both branches exit non-zero on it, after processing the reachable networks rather than instead of.
- **A rejecting `client.close()` no longer discards an already-computed tally.** Pre-PR that cost one network; with one connection for the fleet it would have turned a teardown hiccup into "every network failed to prefetch" — and, given the change above, an exit-1.
- **`getDeploymentsFilePath` refuses a chain name that resolves outside `deployments/`**, matching `getContractAddress` and the four `proposeAll*ChainIdMappings.ts` copies, and takes the environment explicitly rather than defaulting to staging.
- **`getDeployments` keeps the underlying import error as `cause`** instead of flattening it into "not found". The `existsSync` probe stays: classifying off the loader's error taxonomy is runtime-specific, and this CLI, the tests and the propose tasks run under three different ones.
- **Nested strings in a decoded arg get the same per-network address rendering as top-level ones** — a Tron address inside a tuple showed as raw hex at the approval prompt, which the round-1 JSDoc had already (correctly) promised it would not.
- Skip resolution runs over `Promise.all` per [CONV:PARALLEL-WORK]; it reads local files only, so there is no fan-out to bound. `prefetchNetworksWithPendingOps` returns the networks its callers actually use. The queue re-read in `processNetwork` is unchanged and deliberate: a queued row is not an executable operation, and on-chain `isOperationReady` remains the authority.

New behaviour is negative-controlled as before — reverting each fix in turn fails its test and only its test (close-rejection guard: 1; traversal guard: 1; nested address: 1). The exit-code change cannot be unit-tested (the exit sits in the module that calls `runMain` at scope), so it was driven against the real functions with a genuinely corrupt `deployments/tronshasta.json` plus a network carrying a queued op, reproducing the exact previously-green case; output is in the thread.

### DIFF
--- a/script/deploy/safe/execute-pending-timelock-tx.ts
+++ b/script/deploy/safe/execute-pending-timelock-tx.ts
@@ -47,7 +47,6 @@ import { formatTimelockScheduleBatch } from './safe-decode-utils'
 import {
   classifyPrefetchResults,
   fetchPendingForNetworks,
-  type IPendingFetchResult,
 } from './timelock-prefetch'
 import {
   byOperationId,
@@ -235,9 +234,8 @@ const cmd = defineCommand({
     if (executeAll || rejectAll) {
       consola.info('🚀 Checking all networks for pending operations...')
 
-      const networksWithPending = await prefetchNetworksWithPendingOps(
-        networksToProcess
-      )
+      const { networks: networksWithPending, failedCount: prefetchFailures } =
+        await prefetchNetworksWithPendingOps(networksToProcess)
       if (networksWithPending.length === 0) return
 
       consola.info(
@@ -245,9 +243,9 @@ const cmd = defineCommand({
       )
 
       const results = await Promise.all(
-        networksWithPending.map((fetched) =>
+        networksWithPending.map((network) =>
           processNetwork(
-            fetched.network,
+            network,
             isDryRun,
             specificOperationId,
             executeAll,
@@ -302,17 +300,26 @@ const cmd = defineCommand({
           consola.warn('Failed to send batch summary notification:', error)
         }
 
-      // Exit with error code if there were failures
-      if (failedNetworks > 0 || totalOperationsFailed > 0) {
+      // Exit with error code if there were failures. A network the prefetch
+      // could not check counts: it was never looked at, and having found work
+      // elsewhere does not make that safe to pass over silently.
+      if (
+        failedNetworks > 0 ||
+        totalOperationsFailed > 0 ||
+        prefetchFailures > 0
+      ) {
+        if (prefetchFailures > 0)
+          consola.error(
+            `   ❌ Networks never checked (prefetch failed): ${prefetchFailures}`
+          )
         consola.error('\n❌ Script completed with errors')
         process.exit(1)
       }
     } else {
       consola.info('🔄 Checking all networks for pending operations...')
 
-      const networksWithPending = await prefetchNetworksWithPendingOps(
-        networksToProcess
-      )
+      const { networks: networksWithPending, failedCount: prefetchFailures } =
+        await prefetchNetworksWithPendingOps(networksToProcess)
       if (networksWithPending.length === 0) return
 
       consola.info(
@@ -322,10 +329,10 @@ const cmd = defineCommand({
       let totalFailed = 0
       let totalSucceeded = 0
 
-      for (const fetched of networksWithPending)
+      for (const network of networksWithPending)
         try {
           const result = await processNetwork(
-            fetched.network,
+            network,
             isDryRun,
             specificOperationId,
             executeAll,
@@ -342,16 +349,19 @@ const cmd = defineCommand({
               `[${result.network}] ❌ ${result.operationsFailed} operation(s) failed`
             )
         } catch (error) {
-          consola.error(
-            `Error processing network ${fetched.network.name}:`,
-            error
-          )
+          consola.error(`Error processing network ${network.name}:`, error)
           totalFailed++
         }
 
-      if (totalFailed > 0) {
+      // A network the prefetch could not check was never looked at; finding work
+      // elsewhere does not make passing over it safe, so it fails the run too.
+      if (totalFailed > 0 || prefetchFailures > 0) {
         consola.error(
-          `\n❌ Script completed with ${totalFailed} network(s) having failures`
+          `\n❌ Script completed with ${totalFailed} network(s) having failures${
+            prefetchFailures > 0
+              ? ` and ${prefetchFailures} network(s) never checked (prefetch failed)`
+              : ''
+          }`
         )
         process.exit(1)
       } else
@@ -454,19 +464,32 @@ async function fetchQueuedTimelockOps(
   }
 }
 
+/** Outcome of the fleet pre-check, as the run needs it. */
+interface IPrefetchedWork {
+  /** Networks with at least one queued op. */
+  networks: INetworksObject[string][]
+  /**
+   * Networks the prefetch could not check. Non-zero must fail the run even when
+   * other networks had work: those networks were not looked at either way.
+   */
+  failedCount: number
+}
+
 /**
  * Pre-checks the fleet, reports the outcome, and returns only the networks that
  * have queued ops.
  *
- * Exits non-zero when nothing is pending but some networks could not be checked:
- * "0 pending" is only trustworthy if every network was actually reached.
+ * Exits non-zero immediately when nothing is pending but some networks could not
+ * be checked: "0 pending" is only trustworthy if every network was actually
+ * reached. When there *is* work, the run proceeds so the reachable networks are
+ * executed, and the caller fails the run afterwards via `failedCount`.
  *
  * @param networksToProcess - Networks to pre-check.
- * @returns Networks with at least one queued op; empty when there is no work.
+ * @returns The networks to process and how many could not be checked.
  */
 async function prefetchNetworksWithPendingOps(
   networksToProcess: INetworksObject[string][]
-): Promise<IPendingFetchResult[]> {
+): Promise<IPrefetchedWork> {
   const { withPending, skipped, failed, mustExitWithError } =
     classifyPrefetchResults(await fetchPendingForNetworks(networksToProcess))
 
@@ -505,7 +528,10 @@ async function prefetchNetworksWithPendingOps(
   if (withPending.length === 0)
     consola.success('No networks with pending timelock transactions.')
 
-  return withPending
+  return {
+    networks: withPending.map((r) => r.network),
+    failedCount: failed.length,
+  }
 }
 
 async function processNetwork(
--- a/script/deploy/safe/execute-pending-timelock-tx.ts
+++ b/script/deploy/safe/execute-pending-timelock-tx.ts
@@ -229,7 +229,9 @@ const cmd = defineCommand({
     if (isDryRun)
       consola.info('Running in DRY RUN mode - no transactions will be sent')
 
-    // Process networks: scan once (one RPC per network), then process only networks with ready ops. We do not reuse prefetch RPC clients so execution always uses a fresh connection.
+    // The prefetch is queue-only over a single MongoDB connection; processNetwork
+    // then opens a fresh RPC per network so execution never reuses a connection
+    // that idled through a long interactive pause.
     if (executeAll || rejectAll) {
       consola.info('🚀 Checking all networks for pending operations...')
 
--- a/script/deploy/safe/execute-pending-timelock-tx.ts
+++ b/script/deploy/safe/execute-pending-timelock-tx.ts
@@ -45,21 +45,17 @@ import {
 } from './parked-tasks'
 import { formatTimelockScheduleBatch } from './safe-decode-utils'
 import {
+  classifyPrefetchResults,
+  fetchPendingForNetworks,
+  type IPendingFetchResult,
+} from './timelock-prefetch'
+import {
   byOperationId,
   computeOperationIdBatch,
   deserializeScheduleParams,
   getTimelockQueueCollection,
   type ITimelockQueueDoc,
 } from './timelock-queue'
-
-/** Result of pre-checking a network (queue only, no RPC). Used to decide which networks to process; processNetwork always opens a fresh RPC and re-fetches queued ops to verify on-chain readiness. */
-interface IPendingFetchResult {
-  network: INetworksObject[string]
-  /** Number of queued timelock ops for this network (on-chain ready count unknown until processNetwork runs). */
-  pendingInMongoCount: number
-  /** Set when prefetch failed; callers must not treat as "no pending" without checking. */
-  fetchError?: unknown
-}
 
 // TimelockController ABI for the functions we need
 const TIMELOCK_ABI = parseAbi([
@@ -237,44 +233,10 @@ const cmd = defineCommand({
     if (executeAll || rejectAll) {
       consola.info('🚀 Checking all networks for pending operations...')
 
-      const fetchResults = await Promise.all(
-        networksToProcess.map((network) => fetchPendingForNetwork(network))
+      const networksWithPending = await prefetchNetworksWithPendingOps(
+        networksToProcess
       )
-
-      const networksWithPending = fetchResults.filter(
-        (r) => r.pendingInMongoCount > 0
-      )
-      const networksWithFetchError = fetchResults.filter((r) => r.fetchError)
-
-      consola.info(
-        `Checked ${networksToProcess.length} network(s) (MongoDB only); ${
-          networksWithPending.length
-        } have pending timelock tx(s)${
-          networksWithPending.length > 0
-            ? `: ${networksWithPending.map((r) => r.network.name).join(', ')}`
-            : ''
-        }`
-      )
-      if (networksWithFetchError.length > 0) {
-        consola.warn(
-          `Prefetch failed for ${
-            networksWithFetchError.length
-          } network(s): ${networksWithFetchError
-            .map((r) => r.network.name)
-            .join(', ')}`
-        )
-      }
-
-      if (networksWithPending.length === 0) {
-        if (networksWithFetchError.length > 0) {
-          consola.error(
-            'No networks with pending timelock txs; some networks failed to fetch. Exiting with error to avoid silently skipping work.'
-          )
-          process.exit(1)
-        }
-        consola.success('No networks with pending timelock transactions.')
-        return
-      }
+      if (networksWithPending.length === 0) return
 
       consola.info(
         'Processing networks with pending txs (fresh RPC per network).'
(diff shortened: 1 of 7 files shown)


0xDEnYO:
@coderabbitai review

script/deploy/safe/timelock-prefetch.ts
```diff
@@ -0,0 +1,249 @@
...
+): IPrefetchOutcome {
+  const withPending = results.filter((r) => r.pendingInMongoCount > 0)
+  const failed = results.filter((r) => r.fetchError)
+  return {
+    withPending,
+    skipped: results.filter((r) => r.skipReason),
+    failed,
+    mustExitWithError: withPending.length === 0 && failed.length > 0,
+  }
+}
```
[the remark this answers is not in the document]
0xDEnYO: Good catch — fixed in 4ee8f2259.

Confirmed the mismatch: `assemblePrefetchResults` records a failure with `fetchError !== undefined`, so a falsy thrown value produced a result carrying `fetchError` that `failed` never selected — the network then read as checked-with-0-pending, `mustExitWithError` stayed `false`, and the "Checked N of M" line overstated coverage. That is precisely the silent-skip class this PR exists to remove, so it was worth closing even though every current throw site yields an `Error`.

Added a regression test (`counts a falsy thrown value as a failure, not as a checked network`) and negative-controlled it: reverting to the truthiness test fails that test and only that test.

Left `skipped: results.filter((r) => r.skipReason)` on truthiness — `TTimelockSkipReason` is a union of two non-empty string literals, so truthiness and `!== undefined` are equivalent there by type.

script/utils/deploymentHelpers.ts
```diff
@@ -19,6 +19,28 @@ type DeploymentsFileModule = Record<string, string> & {
...
+ * Resolves the on-disk path of a chain's deployments file.
+ *
+ * Callers that must tell "this chain was never deployed" apart from "the file is
+ * there but unreadable" need this: `getDeployments` reports both as not-found.
+ *
+ * @param chain - Chain the deployments file belongs to.
+ * @param environment - Production or staging.
+ * @returns Absolute path to the deployments file (which may not exist).
+ */
+export const getDeploymentsFilePath = (
```
melianessa, line 34: The newly exported `getDeploymentsFilePath` defaults to the STAGING path and carries no path-traversal guard, unlike the equivalent resolution in script/utils/utils.ts.

script/deploy/safe/safe-decode-utils.ts
```diff
@@ -450,10 +450,23 @@ function getDiamondAbiItemForSelector(selector: string): Abi[number] | null {
...
+ * @returns Display string for the argument.
+ */
+export function formatDecodedArg(arg: unknown, network?: string): string {
   if (arg === undefined || arg === null) return String(arg)
   if (typeof arg === 'bigint') return arg.toString()
-  if (typeof arg === 'object') return JSON.stringify(arg)
+  // Tuple and array args (e.g. initFrax's (chainId, eid) pairs) carry nested
+  // bigints, which plain JSON.stringify throws on — losing the whole decode and
+  // leaving the operator approving a payload they were never shown.
+  if (typeof arg === 'object')
```
melianessa, line 468: The new JSDoc promises network-formatted addresses, but the object branch the PR just rewrote ignores the `network` parameter entirely, so nested Tron addresses are still shown as raw hex at the approval prompt

script/deploy/safe/execute-pending-timelock-tx.ts
```diff
@@ -233,48 +229,16 @@ const cmd = defineCommand({
...
-
-      consola.info(
-        `Checked ${networksToProcess.length} network(s) (MongoDB only); ${
-          networksWithPending.length
-        } have pending timelock tx(s)${
-          networksWithPending.length > 0
-            ? `: ${networksWithPending.map((r) => r.network.name).join(', ')}`
-            : ''
-        }`
+      const networksWithPending = await prefetchNetworksWithPendingOps(
```
melianessa: `prefetchNetworksWithPendingOps` returns full `IPendingFetchResult` objects whose `pendingInMongoCount` every caller discards, and each pending network then re-reads the same rows from Mongo

script/deploy/safe/timelock-prefetch.ts
```diff
@@ -0,0 +1,251 @@
...
+  try {
+    deploymentData = (await getDeployments(
+      chain,
+      EnvironmentEnum.production
+    )) as { LiFiTimelockController?: string }
+  } catch (err) {
+    // getDeployments reports a corrupt or unreadable file as not-found too, and
+    // skipping one of those would hide a network that does have a timelock —
+    // the very failure this prefetch exists to stop. Only an absent file skips.
+    if (existsSync(getDeploymentsFilePath(chain, EnvironmentEnum.production)))
```
melianessa, line 126: Re-probing the filesystem with `existsSync` is a fragile way to recover a cause that `getDeployments` deliberately discards; the fix belongs in `getDeployments`

script/deploy/safe/timelock-prefetch.ts
```diff
@@ -0,0 +1,251 @@
...
+ *
+ * @param networks - Networks to pre-check.
+ * @returns One result per input network, in input order.
+ */
+export async function fetchPendingForNetworks(
+  networks: INetworksObject[string][]
+): Promise<IPendingFetchResult[]> {
+  const skipReasons = new Map<string, TTimelockSkipReason>()
+  const errorsByNetwork = new Map<string, unknown>()
+  for (const network of networks)
```
melianessa: The deployments-file resolution for all 71 networks runs in a fully sequential `for … await` loop, violating [CONV:PARALLEL-WORK] and ignoring the repo's own `mapWithConcurrency` helper

script/deploy/safe/timelock-prefetch.ts
```diff
@@ -0,0 +1,251 @@
...
+      .find(
+        {
+          network: { $in: networkNames.map((name) => name.toLowerCase()) },
+          status: 'queued',
+        },
+        { projection: { network: 1 } }
+      )
+      .toArray()
+    return tallyQueuedOpsByNetwork(rows)
+  } finally {
```
melianessa, line 89: `finally { await client.close() }` lets a close rejection escape and overwrite a successful count, and the single-connection design now amplifies that into a whole-fleet failure

script/deploy/safe/execute-pending-timelock-tx.ts
```diff
@@ -526,37 +455,57 @@ async function fetchQueuedTimelockOps(
...
+        .join(', ')}`
+    )
 
-    const rows = await fetchQueuedTimelockOps(network.name)
-    return {
-      network,
-      pendingInMongoCount: rows.length,
-    }
-  } catch (err) {
+  if (mustExitWithError) {
```
melianessa, line 522: `mustExitWithError` only fires when NOTHING is pending, so a prefetch failure is silently tolerated whenever any other network has queued ops — the exact silent-skip this PR exists to fix

script/utils/deploymentHelpers.ts
0xDEnYO: Fixed in 2cd33583f.

Added the containment check the equivalent resolution in `script/utils/utils.ts` (`getContractAddress`) and the four `proposeAll*ChainIdMappings.ts` copies already carry, so the shared helper is no longer the one path without it:

```ts
const relativePath = path.relative(base, filePath)
if (relativePath.startsWith('..') || path.isAbsolute(relativePath))
  throw new Error(`Invalid network name: ${chain}`)
```

Negative-controlled: with the check removed, `getDeploymentsFilePath('../../config/networks', production)` resolves to `/Users/…/GitHub/config/networks.json` — outside the repo entirely — and the new test fails. Framing it honestly, `chain` is typed `SupportedChain` and every current caller passes a `config/networks.json` key, so this is defence in depth rather than a live escape; it earns its place because the function is now exported and the repo already treats this as the house pattern.

On the staging default: rather than keep it, the `environment` parameter is now **required**. It was inherited from `getDeployments`, where it is at least paired with a load that fails loudly; on a bare path builder whose whole purpose is an `existsSync` probe, an omitted argument would silently answer for the wrong environment. Both call sites and the tests already passed it explicitly, so nothing changed behaviourally and `tsc` covers the rest.

script/deploy/safe/safe-decode-utils.ts
0xDEnYO: Fixed in 2cd33583f — the JSDoc was the accurate half; the object branch was the wrong half.

Nested strings now recurse through the same address branch as top-level ones:

```ts
if (typeof arg === 'object')
  return JSON.stringify(arg, (_key, value: unknown) => {
    if (typeof value === 'bigint') return value.toString()
    if (typeof value === 'string') return formatDecodedArg(value, network)
    return value
  })
```

Concretely, `formatDecodedArg({ target: '0x1231DEB6…F4EaE' }, 'tron')` rendered as raw hex before and now renders `TBdQuYCwjQzVKomfrrsgfizuSKw4q7rdUj (0x1231DEB6f5749EF6cE6943a275A1D3E7486F4EaE)` — the base58 plus the `tronHexSuffix` cross-reference, same as the top-level path.

Two regression tests, negative-controlled: dropping the string branch fails the nested-address test with exactly that raw-hex output, and the second test pins that a non-address string (`{"name":"FraxFacet"}`) is left untouched. Worth fixing here rather than narrowing the doc, for the same reason the bigint case was: this is the approval prompt, and a Tron cut payload whose nested targets show as hex is a payload the signer cannot check against `config/global.json`.

script/deploy/safe/execute-pending-timelock-tx.ts
0xDEnYO: Half fixed in 2cd33583f, half is deliberate — splitting them because they have different answers.

**The discarded return value: fixed.** `prefetchNetworksWithPendingOps` now returns what its callers use:

```ts
interface IPrefetchedWork {
  networks: INetworksObject[string][]
  failedCount: number
}
```

Both branches iterate `networks` directly (`fetched.network` → `network`), and `IPendingFetchResult` is no longer imported here. `failedCount` is the other finding on this PR (prefetch failures now fail the run), not a second discarded field.

**The re-read: intentional, and it is not the same read.** The prefetch is a MongoDB-only count — it answers "which networks have queued rows", nothing more. `processNetwork` then opens a fresh RPC per network and re-fetches, because a queued row is not an executable operation: the runner re-derives the operationId from the queue params and checks `isOperationReady` on-chain before executing anything (the property `.github/workflows/runPendingTimelockTXs.yml` relies on — a tampered queue row can only DoS, never cause unintended execution). Carrying `pendingInMongoCount` forward as if it were a work count would mean trusting the queue as the source of executability, which is exactly what that design refuses to do. The comment at the `IPendingFetchResult` declaration states this; I have left the queue re-read as-is.

script/deploy/safe/timelock-prefetch.ts
0xDEnYO: Took the root-cause half of this, kept `existsSync` — reasoning below, happy to go further if you disagree.

**Taken:** `getDeployments` no longer discards the cause. It now rethrows with `{ cause: err }`, so the import failure is preserved instead of being flattened into "not found" (verified: `(err as Error).cause !== undefined` on a corrupt file).

**Not taken — moving the classification into `getDeployments`.** Two reasons:

1. It would change a shared helper's error contract for every caller in the repo to serve one. `getDeployments` is used across `confirm-safe-tx`, the propose tasks and the health checks, all of which want "did I get deployments, yes or no"; splitting the throw into absent-vs-unreadable pushes that distinction onto callers that have no use for it.
2. The replacement would be less reliable, not more. Classifying off the cause means matching the module loader's error taxonomy (`ERR_MODULE_NOT_FOUND` and friends), which is runtime-specific — this CLI runs under `bunx tsx`, the tests under `bun test`, and the propose tasks under `bun` directly. `existsSync` asks the question we actually mean ("is the file there") and gives the same answer in all three.

The fragility this guards against is real but inverted: the risk is a *present-but-unreadable* file being classified as a skip, and `existsSync` fails safe there — it rethrows, the network is recorded as a fetch error, and (as of this round) that now fails the run. The PR's `refuses to skip a network whose deployments file exists but will not load` test covers exactly that, negative-controlled against removing the check.

With the cause preserved, a follow-up that classifies off `err.cause` is now possible without another re-probe if we decide the loader codes are stable enough across the three runtimes.

script/deploy/safe/timelock-prefetch.ts
0xDEnYO: Fixed in 2cd33583f — the loop now runs over `Promise.all`, with the per-network `try/catch` moved inside the map so one unreadable deployments file still lands in `errorsByNetwork` instead of rejecting the batch:

```ts
const resolved = await Promise.all(
  networks.map(async (network) => {
    try {
      return { network, skipReason: await resolveTimelockSkipReason(network) }
    } catch (err) { … ; return { network, err } }
  })
)
```

Result order is irrelevant here — `assemblePrefetchResults` reindexes by network name and returns in input order, which its `returns one result per input network, in input order` test pins.

On `mapWithConcurrency`: I did not reach for it, because there is nothing to bound. This pass is 71 local `import()` calls against `deployments/*.json`, memoised in `deploymentsCache` — no RPC, no DNS, no rate-limited provider. [CONV:PARALLEL-WORK] says to bound concurrency "if you hit provider rate limits", and there is no provider. Adding a pool would be ceremony over a filesystem read.

The sequential form was a leftover instinct from the bug this PR fixes — that fan-out was 71 `MongoClient`s each resolving SRV+TXT, and I carried the caution into a loop that never had the problem. Your read is right.

script/deploy/safe/timelock-prefetch.ts
0xDEnYO: Fixed in 2cd33583f.

```ts
} finally {
  await client
    .close()
    .catch((err: unknown) =>
      consola.warn('Failed to close the MongoDB connection:', err)
    )
}
```

The amplification you point at is the part that made this worth closing: pre-PR a close rejection cost one network, and the single-connection design turns the same rejection into "every network with a timelock failed to prefetch" — which, with the other fix on this PR, is now an exit-1. A teardown hiccup should not be able to do that.

To get a real negative control on it rather than a reasoned assertion, `countQueuedOpsByNetwork` takes an optional connection factory (`connect: TQueueConnector = getTimelockQueueCollection`) so the teardown path can be driven with a rejecting `close()`. Two tests: the tally survives a rejecting close, and the connection is still closed exactly once on the happy path. Reverting the `.catch` fails the first with the injected `connection reset during close`.

script/deploy/safe/execute-pending-timelock-tx.ts
0xDEnYO: Fixed in 2cd33583f. This was the strongest of the six — the PR closed the DNS half of the silent skip and left the reporting half open.

`prefetchNetworksWithPendingOps` now returns `failedCount` alongside the networks, and both branches fail the run on it:

```ts
if (failedNetworks > 0 || totalOperationsFailed > 0 || prefetchFailures > 0) { … process.exit(1) }
```

```ts
if (totalFailed > 0 || prefetchFailures > 0) { … process.exit(1) }
```

`mustExitWithError` keeps its early exit for the nothing-pending case; when there *is* work the run proceeds first, so reachable networks still get executed, and then fails. Executing what we can and reporting the gap beats aborting on both.

Proven on real data rather than asserted, since the exit itself sits in the module that calls `runMain` at scope and cannot be imported. Driving the real functions with a genuinely corrupt `deployments/tronshasta.json` and a network carrying a queued op:

```text
failedCount: 1
withPending: [ "mainnet" ]
mustExitWithError (old-only guard): false
new guard fires (failedCount > 0 with work present): true
```

That is exactly your scenario: one network never checked, work found elsewhere, run previously green.

On blast radius, since this runs on a 10-minute cron whose step fails the job on a non-zero exit and posts a Slack fallback — it is not a new alert class. The nothing-pending case already exited 1 on any prefetch failure, and that is the dominant tick. The single-connection design also means a MongoDB-side failure now fails *all* of `toCheck` at once, which lands in `withPending === 0` and was already covered. What this newly catches is the per-network deployments-read failure with work elsewhere — the case above.

gvladika (approved):

PR #2280, opened 2026-08-28
fix(healthcheck): a stalled parked claim is not removal coverage (EXSC-867)

0xDEnYO:

Fixes [EXSC-867](https://linear.app/lifi-linear/issue/EXSC-867/harden-no-stale-registered-facets-a-stalled-parked-claim-must-not)

`no-stale-registered-facets` treated the mere **existence** of an open parked task as coverage for a deprecated-but-routed facet. It could not distinguish "cleanup in flight" from "cleanup silently dead", so a stalled task kept the check green indefinitely while the facet stayed routed in production.

Found live during a fleet sweep of the parked queue: mantle `GenericSwapFacet` v2.0.0 at `0x2b7D2C78bd801Cc06DDCF91DeE2e8fAE22814f7e` (origin #2046) sat `status: proposed` with **no `safeTxHash`** — no Safe transaction was ever created for anyone to sign — for 29 days, still routing 4 selectors, while the check reported green the whole time.

**Root cause.** The drain flips `queued → proposed` (`claimForProposal`) *before* the Safe proposal exists and stamps `safeTxHash` afterwards (`linkToProposal`). A drain that dies between the two leaves a record no unattended job can move:

| Path | Behaviour on such a record |
|---|---|
| `drain-parked-tasks.ts` | claims `status: 'queued'` only → never re-claims it |
| `reconcileDecision()` | returns `'keep'` (facet on-chain, no linked proposal status) |
| `repair-orphaned-parked-tasks.ts` | counts it `unlinked`, logs "leaving for manual review", skips |
| `cancel-parked-task.ts` | refuses — `markCancelled` is `queued`-only |

So coverage had to become **liveness**, not existence:

- `queued` → always live. The next drain claims it; age alone is backlog, not breakage. Flagging old `queued` tasks would red every chain with a slow rollout (there are 16 such tasks right now from #2108).
- `proposed` **with** `safeTxHash` → live. A real proposal exists and `reconcile` resolves it once that proposal executes or reverts.
- `proposed` **without** `safeTxHash` past `STALE_PARKED_CLAIM_DAYS` → stalled. 7 days is deliberately generous: proposals are signed and executed within ~48h in practice, so a week without one is unambiguous breakage rather than slowness. The bound is not zero because a healthy drain legitimately holds that state for the seconds between claiming and linking.

`severity` also went `warning` → `error`, and both the stalled and the uncovered class now report via `ctx.logError` instead of `ctx.logWarn`. `failed = errors.length > 0` in `executeInvariant`, so previously *nothing* gated on this check; `severity: 'error'` additionally buys the transient-RPC re-verify pass. The `live` class stays an info line.

The remediation string is now split per class, because the two need opposite actions — a stalled claim must **not** be re-enqueued (the open task blocks the dedup gate) and must **not** be cancelled (that abandons a live deprecation); it needs `revertToQueued`, whose missing operator CLI is [EXSC-715](https://linear.app/lifi-linear/issue/EXSC-715/add-operator-cli-to-revert-or-cancel-a-parked-diamond-cleanup-task).

## Both halves of the hole

`no-stale-registered-facets` and `no-unexpected-facets` are the two halves of detection for this failure, as `docs/DeferredDiamondCleanupQueue.md` states: the first sees a deprecated facet **while the deploy log still names it**, the second sees it **once the deprecation PR prunes that entry** (pruning is explicitly licensed by an open `queued`/`proposed` task).

Gating only the first on liveness would have left the pruned half still downgrading a dead claim to an info line — hiding the same defect in the place the other invariant structurally cannot look, since it resolves names through the deploy log. So `no-unexpected-facets` now declines the expected-pending downgrade for a stalled claim and warns instead.

That branch cannot fire on today's fleet: the one stalled claim (mantle) is still deploy-logged, so its address never reaches the `unlogged` set. It is also warning-severity, so it cannot red a run.

## Deliberately NOT changed: `skipTestnet`

`skipTestnet: true` was the suspected reason deprecated facets survived on four testnets. It is not the cause, so flipping it would have been a cosmetic fix hiding a real gap. No testnet has a `production` target-state entry in `_targetState.json` — the entries exist but are empty objects — so `getExpectedFacetNames()` returns `undefined` and the invariant early-returns whatever the flag says. Verified by running the real `findDeprecatedLiveFacets` over the 5 active non-mainnet networks: 3 skipped for "no target-state entry", 0 findings. The comment now records this so nobody "fixes" the flag expecting a behaviour change, and the actual gap is [EXSC-868](https://linear.app/lifi-linear/issue/EXSC-868/testnet-diamonds-have-no-target-state-so-every-target-state-health).

## Falsification on real data

Same-PR tests are not evidence a new check can fire, so the shipped `splitByParkedCoverage` was run against live `facets()` reads and the live queue across **all 66 active mainnet networks** (zero unreadable):

```text
networks evaluated: 66  (STALE_PARKED_CLAIM_DAYS=7)
mantle  deprecated=1 parked=0 UNPARKED=0 STALLED=1
    stalled: GenericSwapFacet@0x2b7D2C78bd801Cc06DDCF91DeE2e8fAE22814f7e status=proposed safeTxHash=ABSENT
TOTALS: would-error-on-unparked=0, would-error-on-stalled=1
```

- fires on exactly the one known real defect, and on nothing else
- `unparked = 0` fleet-wide, so the `logError` flip reds no network that is healthy today — this change does not turn on a fleet-wide red

Negative controls are unit-tested: `queued` at 400 days, `proposed` with a `safeTxHash` at 90 days, a claim one day inside the bound, and an unreachable queue all stay green.

**Note:** mantle's queue row is being repaired in parallel, so once that lands the fleet goes green on this check. The snapshot above is from 2026-08-28T09:14Z.

### DIFF
--- a/script/deploy/healthCheckInvariants.ts
+++ b/script/deploy/healthCheckInvariants.ts
@@ -1354,6 +1354,50 @@ export function splitByParkedCoverage(
 }
 
 /**
+ * Groups open parked tasks by network and facet address for coverage lookups.
+ *
+ * Two open tasks CAN share one address: the open-status unique index is on `taskKey`, and
+ * a legacy name-keyed row does not collide with the address-keyed key `computeTaskKey`
+ * mints today (mantle still carries such a row). Collapsing them therefore has to fail
+ * **closed** — a stalled claim, once seen, is never replaced by a livelier sibling, which
+ * would otherwise mask it and re-open exactly the gap this coverage check exists to close.
+ * The result must not depend on the order the queue returned the tasks in.
+ *
+ * @param tasks - Open (`queued`/`proposed`) tasks, in any order.
+ * @returns Network (lowercased) → lowercased facet address → the task that governs coverage.
+ */
+export function collapseOpenParkedTasks(
+  tasks: readonly {
+    network: string
+    facetAddress: string
+    prUrl: string
+    status: ParkedTaskStatus
+    createdAt: Date
+    proposedAt?: Date
+    safeTxHash?: string
+  }[],
+  now: Date = new Date()
+): OpenParkedByNetwork {
+  const byNetwork: OpenParkedByNetwork = new Map()
+  for (const task of tasks) {
+    const map =
+      byNetwork.get(task.network) ?? new Map<string, IOpenParkedCoverage>()
+    const address = task.facetAddress.toLowerCase()
+    const existing = map.get(address)
+    if (!existing || !isStalledParkedClaim(existing, now))
+      map.set(address, {
+        prUrl: task.prUrl,
+        status: task.status,
+        createdAt: task.createdAt,
+        proposedAt: task.proposedAt,
+        safeTxHash: task.safeTxHash,
+      })
+    byNetwork.set(task.network, map)
+  }
+  return byNetwork
+}
+
+/**
  * Open parked tasks fleet-wide, fetched once per process and grouped by network
  * (lowercased `facetAddress` → PR URL maps). The health check evaluates dozens of networks
  * concurrently in one process, and a Mongo connect/index-check/teardown per stale
@@ -1379,21 +1423,7 @@ function fetchOpenParkedAddressesByNetwork(): Promise<
           environment: EnvironmentEnum.production,
           status: OPEN_STATUSES,
         })
-        const byNetwork: OpenParkedByNetwork = new Map()
-        for (const task of open) {
-          const map =
-            byNetwork.get(task.network) ??
-            new Map<string, IOpenParkedCoverage>()
-          map.set(task.facetAddress.toLowerCase(), {
-            prUrl: task.prUrl,
-            status: task.status,
-            createdAt: task.createdAt,
-            proposedAt: task.proposedAt,
-            safeTxHash: task.safeTxHash,
-          })
-          byNetwork.set(task.network, map)
-        }
-        return byNetwork
+        return collapseOpenParkedTasks(open)
       } finally {
         await client.close()
       }
@@ -2591,7 +2621,7 @@ export const HEALTH_CHECK_INVARIANTS: IHealthCheckInvariant[] = [
     scope: { environments: ['production'], skipTestnet: true },
     readsOnChainFacets: true,
     remediation:
-      'No task: enqueue the removal (script/deploy/safe/enqueue-parked-task.ts, with the deprecation PR URL) or run `cleanUpProdDiamond --auto --network <network>`. Stalled claim: the task is `proposed` with no Safe proposal and no unattended job can move it — it needs `revertToQueued` before a drain will re-claim it (EXSC-715). Do NOT re-enqueue (the open task blocks the dedup gate) and do NOT cancel (that abandons a live deprecation). See docs/DeferredDiamondCleanupQueue.md.',
+      'No task: enqueue the removal (script/deploy/safe/enqueue-parked-task.ts, with the deprecation PR URL) or run `cleanUpProdDiamond --auto --network <network>`. Stalled claim: the task is `proposed` with no Safe proposal and no unattended job can move it — it needs `revertToQueued` before a drain will re-claim it, and no operator CLI ships that yet (EXSC-715), so escalate to the SC on-call rather than hand-editing the queue. Do NOT re-enqueue: an address-keyed task is refused by the dedup gate, and a legacy name-keyed one is NOT — it would silently open a second task for the same address. Do NOT cancel (that abandons a live deprecation). See docs/DeferredDiamondCleanupQueue.md.',
     run: async (ctx) => {
       if (ctx.onChainFacets.length === 0) {
         consola.info(
--- a/script/deploy/healthCheckInvariants.ts
+++ b/script/deploy/healthCheckInvariants.ts
@@ -68,6 +68,28 @@ import {
 /** Severity of a failed invariant: `error` fails the run (exit 1); `warning` is reported but non-fatal. */
 export type HealthCheckSeverity = 'error' | 'warning'
 
+/**
+ * How long a claimed-but-unproposed parked task may sit before it stops counting as
+ * coverage. Signing and executing a removal proposal takes at most ~48h in practice, so a
+ * claim with no linked proposal past a week is not slow — it is a drain that died between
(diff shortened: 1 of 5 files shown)


0xDEnYO:
@coderabbitai review

0xDEnYO:
Both findings were valid and are fixed in 8b162cd7e.

**Duplicate open tasks per address (Major).** Confirmed reachable, and the precondition is live today: `computeTaskKey` is address-keyed, but the open-status unique index is on `taskKey`, so a legacy name-keyed row does not collide with it. The live queue has exactly one such legacy open row — mantle `GenericSwapFacet` — and it is the stalled one. A plain overwrite therefore let queue sort order decide whether the stalled claim survived. The collapse now lives in an extracted, unit-tested `collapseOpenParkedTasks()` and is stall-dominant, so it is order-independent.

Falsified against the live queue rather than only in tests — real 22-task queue plus mantle's real legacy row paired with a synthetic address-keyed sibling:

```text
live open tasks: 22
collapsed entries: 22 (no dupes today)   # no behaviour change for current data
mantle row survives collapse: status=proposed safeTxHash=ABSENT stalled=true
  stalled first:  kept status=proposed -> classified STALLED (fails)
  stalled second: kept status=proposed -> classified STALLED (fails)
```

**Remediation text (Minor).** Sharper than reported: the string did not just imply a CLI exists, it asserted "the open task blocks the dedup gate", which is false for a legacy name-keyed row — re-enqueueing mantle's address would *not* collide and would open a second open task. Both the remediation and the doc now state that clearing a stalled claim has no shipped operator path (EXSC-715), that re-enqueueing is not a workaround and why, and that it belongs with the SC on-call until the CLI lands.

script/deploy/healthCheckInvariants.ts
```diff
@@ -1263,29 +1284,117 @@ export function findDeprecatedLiveFacets(params: {
...
+  if (task.safeTxHash) return false
+  const claimedAt = new Date(task.proposedAt ?? task.createdAt).getTime()
+  const days = (now.getTime() - claimedAt) / 86_400_000
+  return days >= STALE_PARKED_CLAIM_DAYS
+}
+
+/** Whole-day age of a queue timestamp, for operator-facing lines (`unknown` if absent). */
+function formatDaysAgo(at: Date | undefined, now: Date = new Date()): string {
+  if (!at) return 'unknown'
+  const days = Math.floor((now.getTime() - new Date(at).getTime()) / 86_400_000)
```
melianessa: The day-in-milliseconds literal `86_400_000` is added twice with no unit comment, violating `.agents/rules/200-typescript.md` 'Inline Documentation → Constants' ('Magic numbers and timeout/delay values MUST have an inline `// <value + unit>` comment'), while the parked-task subsystem already ships `DAY_MS` and a `formatAge` for the same rows.

script/deploy/healthCheckInvariants.ts
```diff
@@ -68,6 +69,28 @@ import {
...
+ * coverage. Signing and executing a removal proposal takes at most ~48h in practice, so a
+ * claim with no linked proposal past a week is not slow — it is a drain that died between
+ * `claimForProposal` and `linkToProposal`, and no unattended job recovers it: the drain
+ * only claims `queued`, `reconcileDecision` returns `keep` without a linked proposal, and
+ * `repair-orphaned-parked-tasks.ts` skips it for manual review.
+ */
+export const STALE_PARKED_CLAIM_DAYS = 7
+
+/** The fields of an open parked task that coverage decisions read. */
+export interface IOpenParkedCoverage {
```
melianessa: `IOpenParkedCoverage` and the inline `readonly {…}[]` parameter of `collapseOpenParkedTasks` hand-copy fields of the exported `IParkedTask` instead of `Pick`, widening `facetAddress` from `Address` to `string` and leaving a 5-field object literal that silently drops any future optional field — against `.agents/rules/200-typescript.md` '## Types': 'Always check for existing types before defining new ones' / 'Reuse existing types'.

script/deploy/healthCheckInvariants.ts
0xDEnYO: Fixed in 3b5944e54. `DAY_MS` now lives in `script/deploy/shared/constants.ts` and is imported by `healthCheckInvariants.ts` (both sites), the test file (6 fixture sites), and `reconcile-parked-tasks.ts` — whose local copy is deleted, so the constant is single-sourced rather than moved. `../shared/constants` was already imported from `safe/` (`deploy-safe.ts`), so this adds no new coupling.

`formatAge` I deliberately did not reuse: it is a private helper in `list-parked-tasks.ts`, a module that calls `runMain(main)` at top level, so importing it would execute the CLI. Its contract also differs — `"3d 4h"` / `"5h"` / `"12m"` vs the whole-day `"Xd ago"` the invariant line needs.

One follow-up from the review gate, in d27f9aa2c: with the fixtures and the staleness bound both dividing by `DAY_MS`, a wrong `DAY_MS` cancels algebraically and every age assertion still passes, so there is now an `expect(DAY_MS).toBe(86_400_000)` pinning it literally.

script/deploy/healthCheckInvariants.ts
0xDEnYO: Fixed in 3b5944e54.

```ts
export type IOpenParkedCoverage = Pick<
  IParkedTask,
  'prUrl' | 'status' | 'createdAt' | 'proposedAt' | 'safeTxHash'
>
```

and the `collapseOpenParkedTasks` parameter is now `readonly (IOpenParkedCoverage & Pick<IParkedTask, 'network' | 'facetAddress'>)[]`, so `facetAddress` is back to `Address`. That tightening was load-bearing, not cosmetic: it immediately failed to compile against the test fixtures, which had been passing plain `string` addresses.

The 5-field object literal is gone too — the collapse now stores the task itself (`map.set(address, task)`), so a future optional field flows through instead of being silently dropped. Verified nothing observes the wider object: every read of the map is named property access on the five picked fields, there is no enumeration/spread/`JSON.stringify` on that path, and the map values never reach an operator or Slack line except as an already-formatted string.

0xDEnYO:
@coderabbitai review

melianessa (approved):

gvladika (approved):

PR #2286, opened 2026-08-31
chore(deploy): retire deployUpgradesToSAFE.sh and re-home its production gate (OQ2)

0xDEnYO:

Fixes [EXSC-709](https://linear.app/lifi-linear/issue/EXSC-709/s23-decide-fix-or-retire-deployupgradestosafesh) — OQ2 under [EXSC-686](https://linear.app/lifi-linear/issue/EXSC-686/multisig-signing-process-20) (Signing 2.0, WP-0.2).

> Previously stacked on #2128 (EXSC-687). That PR merged on 2026-08-31 (`67b5b2510`); this branch has
> been rebased onto `main` and now carries only the OQ2 work.

`script/deploy/deployUpgradesToSAFE.sh` proposes diamond cuts **without `--timelock`**. On
timelock-owned production diamonds that means the signers approve a proposal that then reverts — and
it was reachable as a live menu entry (`scriptMaster.sh` use case 12). (The OQ2 ruling says "three
humans sign"; per D8, verified 2026-08-31, `safeOwners[0]` is the `deployerWallet` — also the
proposer — so with threshold 3 the wasted signatures are **two human ones plus an automated key**.
The ruling's conclusion is unaffected: the proposal still reverts.) `script/tasks/diamondUpdateFacet.sh`
already does the same job correctly, proposing **with** `--timelock`. Retired outright, no tombstone.

**The part that isn't just a deletion:** the retired script was the **only caller** of
`script/deploy/github/verify-approvals.ts` — the production deploy gate. Deleting it alone would
have silently removed that control. The gate therefore moves to `diamondUpdateFacet.sh`.

## The gate condition

```bash
if [[ "$ENVIRONMENT" != "staging" ]] && ! isTestnetNetwork "$NETWORK"; then
```

Two independent reasons for the two clauses:

- **`!= staging`, not `== production`.** `getPrivateKey` matches `*staging*` as a **substring**, so
  it hands out the production key for `prod`, an empty value, or any typo that does not contain
  `staging` — those must be gated. Matching on the exact string keeps the gate at least as broad as
  the key it protects (a hypothetical `prestaging` would get the staging key yet still be gated, and
  is then rejected downstream as an unknown environment — strict, and fail-closed). This keeps
  #2128's condition verbatim rather than narrowing it. Note the comment #2128 shipped with this
  condition asserted exact-string matching in `getPrivateKey`; that was wrong, and is corrected here.
- **`! isTestnetNetwork`.** The retired script refused testnets outright
  (`"deployUpgradesToSAFE is not supported on testnet networks (no Safe)"`), so its gate never
  covered them. `diamondUpdateFacet.sh` does run on testnets, and 72 networks carry `production`
  target state — including `arbitrumsepolia` and `basesepolia`. Without this clause, deploying an
  unmerged facet to a production testnet from a feature branch would be **blocked**: the facet does
  not exist on `main`, so its closure diverges, and a pre-audit facet has no `audit/auditLog.json`
  entry. That is the standard pre-merge validation rollout. No Safe is involved on testnets, so
  exempting them costs nothing.

Gating on the existing `SHOULD_PROPOSE_TO_SAFE` predicate instead would be wrong in the other
direction: it would let `SEND_PROPOSALS_DIRECTLY_TO_DIAMOND=true` mainnet deploys skip the gate.

## Facet names, not update-script names

`CONTRACT_NAME` in `diamondUpdateFacet.sh` is the *update-script* name (`UpdateChainflipFacet`),
but the gate resolves each name to `src/Facets/<name>.sol`. Passing it verbatim would have made the
gate look for `src/Facets/UpdateChainflipFacet.sol`, which does not exist — and `fileMatchesRef`
returns `false` for a missing file, so the gate fails closed. **Every production deploy from a
non-main branch would have been blocked.** Fixed by stripping the prefix, with `UpdateCoreFacets`
special-cased to the `coreFacets` array in `config/global.json` (it cuts all 12 rather than one
facet of its own).

Verified against the tree: all 40 `script/deploy/facets/Update*.s.sol` **and** all 16
`script/deploy/zksync/Update*.zksync.s.sol` resolve through `${CONTRACT_NAME#Update}` to a real
`src/Facets/<name>.sol`, and all 12 `coreFacets` entries resolve. The only miss across all 56 is
`UpdateCoreFacets`, which is exactly the case that is special-cased. Both `UpdateCoreFacets.s.sol`
and its zksync variant read `.coreFacets` from the same `config/global.json` the gate reads, so the
gate checks precisely the set the script cuts.

## Documentation

- `docs/MultisigSigningProcess.md` referenced the retired script in four places (the §4.2
  propose entry-point list, the gate paragraph, the §5 controls table, §7 hardening backlog). The
  entry-point bullet is removed; the rest now point at `diamondUpdateFacet.sh`. The gate paragraph
  also described the **pre-#2128** semantics (single source file, audit log read from the working
  tree) and now describes what #2128 actually shipped: the transitive `src/` import closure, with
  the audit log read from `main` so a deploy cannot certify itself.
- `docs/Deploy.md:103` told operators to select `11) Propose upgrade TX to Gnosis SAFE` — already
  off by one (the entry was at 12) and now pointing at a flow that no longer exists. The bullets
  under it also described the retired script's flow ("select multiple using the spacebar", "select
  the SAFE wallet you want to use"); menu entry 1 is single-select and takes the Safe from
  `config/networks.json`. The whole "Upgrade using SAFE wallet" block is rewritten, which also
  picked up four adjacent errors it had accumulated: the path was `./scripts/scriptMaster.sh` (no
  such directory); the `yes - to LiFiDiamond` prompt was missing, and answering `no` there skips
  `diamondUpdateFacet` and therefore the gate entirely; the diamond was described as Safe-owned
  when production mainnet diamonds are owned by `LiFiTimelockController`; and the final step said
  to confirm in the Gnosis Safe app, though there is no Safe{Wallet} UI in this flow (it is
  `bun confirm-safe-tx`).

## Closing the `main` bypasses

Review of the re-homing surfaced a hole in the gate itself, inherited from #2128. It only mattered
while the gate guarded a script this PR's own description calls "effectively dead"; re-homing it onto
the primary facet-cut path made it material, so it is fixed here rather than deferred.

`verifyDeployGate` and `collectDeployGateFailures` both did:

```ts
if (input.branch === 'main') return []
```

That is a **branch-name** check with no comparison against `origin/main`. Anything in the working tree
of a checkout sitting on `main` — uncommitted edits, a half-applied patch, a local `main` that is
simply behind — reached a production Safe proposal with no comparison performed at all. Reproduced on
this branch, same tree, one appended line in `src/Facets/AcrossFacetV4.sol`:

```text
--branch some/feature  -> EXIT=1   No open PR found for branch "some/feature"
--branch main          -> EXIT=0   OK          # the bypass
```

Both short-circuits are removed, so the closure comparison always runs. The `main` case now blocks
with its own message rather than a misleading "No open PR" — no pull request can have `main` as its
head, so the open-PR exception cannot apply and GitHub is not consulted at all:

```text
--branch main, clean tree  -> EXIT=0   OK        # no GitHub call, same cost as before
--branch main, dirty tree  -> EXIT=1   Deploying from "main", but the working tree does not
                                       match it. Merge the change and pull, or reset the
                                       tree, before deploying
```

The audited-freeze exception deliberately does **not** rescue this case: on `main` there is no PR to
anchor it to, and divergence there means uncommitted or stale content, which is exactly what the gate
exists to stop.

**Two of #2128's tests encoded the hole as intended behaviour** — `allows production deploys from main
even when the working tree diverges`, and a CLI case asserting production-on-`main` exits 0 from an
empty temp dir precisely because it never touched the repo. Both are inverted. The suite goes 42 to 49
tests. Every mutation that restores a bypass is caught: re-adding the short-circuit to
`collectDeployGateFailures` fails 3 tests, to `verifyDeployGate` 2, restoring the local-`main`
fallback 1, swapping `deps.mainRef` for `HEAD` 2, and dropping the main-exemption on the GitHub
lookup 1.

### The same hole through a second door: `resolveMainRef`

Re-gating the fix above surfaced a second path to the same bypass. `resolveMainRef` fell back to
**local `main`** whenever `origin/main` could not be resolved — and local `main` is whatever the
operator last committed, so it cannot stand in for a merged ref. Reproduced in a throwaway clone with
no `origin/main`, an unreviewed facet edit committed on local `main`:

```text
before   EXIT=0   OK                                          # bypass
after    EXIT=1   Cannot resolve origin/main in this checkout. Fetch it before deploying
sanity   EXIT=0   OK    # origin/main restored, tree matches it - normal path unaffected
```

Reachable without malice too: a remote not named `origin`, or a single-branch checkout. The fallback
is removed — `origin/main` or nothing.

### Two message defects in the new `main` path

Both found by re-gating the fix commit, both in code added by it:

- The remedies were wrong. "Reset the tree" does not remove an untracked-but-imported `.sol` file
  (that needs `git clean`), and "merge and pull" is unactionable for local commits sitting on `main`,
  which have to move to a branch first. The message now says exactly that.
- A `main` block emitted a second, misleading error — `AcrossFacetV4 has changed since audited commit
  650d18eb… (11 files)` — sending the operator after an audit-log problem that cannot apply on `main`,
  and implicating 11 files they never touched. The `main` path now returns its own message plus a
  plain per-facet divergence line naming the file that actually diverged:

```text
ERROR  Deploying from "main", but the working tree does not match origin/main. Move the change
       onto a branch and open a PR, or discard it (git checkout / git clean) and pull, before deploying
ERROR  AcrossFacetV4 diverges from origin/main (src/Facets/AcrossFacetV4.sol)
```

### What this gate does and does not assert

Worth stating plainly, because the name invites a stronger reading: it enforces **main-equivalence**,
not "audited". Code whose closure matches `main` passes with no audit lookup at all — the
`audit/auditLog.json` freeze is only the *exception* that lets unmerged code through under an open PR.
Verifying that what reaches production was actually audited is the separate bytecode to audit
attestation item in §7 of `MultisigSigningProcess.md`, still backlog. §4.2 now says this explicitly so
the next reader does not over-trust the control.

### Closing the remaining gaps

**`lib/` is now compared.** Dependencies there are compiled into every facet, but their
content lives in submodules and is not in this repo's tree, so a file-by-file comparison is
impossible. The gitlink is comparable instead:
`git diff --name-only --ignore-submodules=none origin/main -- lib/` reports a submodule whose
HEAD is off its recorded commit **or** whose working tree is dirty — both verified on a
synthetic superproject. `--ignore-submodules=none` is passed explicitly so a repo-level or
user-level `ignore` setting cannot weaken it. A divergence there is not excused by an open PR
or an audit freeze, since neither says anything about a dependency. Verified to produce no
false positive on a real clone (submodules initialised and clean) or in a worktree with
submodules uninitialised.

**`origin/main` is refreshed before it is trusted.** The remote tip is read with
`git ls-remote origin main` and fetched **only when it differs**, so the common case costs one
round trip and transfers no objects. An unreachable remote **fails the gate** rather than
falling back to the local copy — comparing against a possibly-stale main is the failure mode
this closes, so quietly continuing would defeat it.

**Git reads are memoized.** One facet's closure overlaps heavily with the next one's, so the
same shared libraries were re-read once per facet. Measured on the real 12-facet
`UpdateCoreFacets` set:

```text
before   112 git calls   (108 `git show`, only 36 distinct)
after     40 git calls   ( 36 `git show`, zero redundancy)
```

**Not fixed — a deleted imported file drops out of the closure** and the gate passes. Not
exploitable: the forge compile fails immediately afterwards, and removing the import instead
would itself diverge the facet. The `continue` that exists for commented-out imports catches
this case too.

**Cost note.** The freshness check adds one `ls-remote` per gate invocation (~0.5–4s depending
on link latency). Because `diamondUpdateFacet.sh` runs once per (network, facet), a fleet
rollout would pay that per network — which is what the verdict cache below removes.

### D9, resolved — the gate stays here, and the redundancy is fixed by caching the verdict

Goran raised the placement question on the `diamondUpdateFacet.sh` call site: the verdict is
invariant across a rollout, so gating once per *(network, facet)* recomputes the identical answer
for every network. Agreed on the diagnosis. Two things came out of working through it:

- **Moving the gate to `propose-to-safe.ts` does not remove the redundancy.** That funnel is
  invoked once per (network, facet) too (`diamondUpdateFacet.sh:250` and `:265`), so the count
  goes 71 → 71. Its argv is `--to` / `--calldata` with no facet name at that layer, so gating
  there means decoding timelock-wrapped `diamondCut` calldata, reverse-mapping facet address →
  name, and adding a pass-through for every non-cut payload — plus `propose-to-safe-tron.ts` is a
  second funnel. That is a bigger change than a re-home, and it is scoped as **WP-1.4**, which
  carries WP-1.2's mandatory-ticket-link block into the same funnel. WP-1.4 removes this call
  site when it lands, so the gate is never live in two places.
- **The redundancy itself is fixed here, by caching the verdict per run** — Goran agreed to this
  split.

`script/deploy/github/deploy-gate-cache.ts` records a pass once and reuses it while the tree
stays put. Measured on this repo, same facet, cold cache then warm:

```text
cold    5185 ms   (ls-remote + closure comparison)
warm     481 ms   (three local git calls, no remote contact)
```

So a 71-network rollout goes from 71 `ls-remote` round trips to **one**, and from 71 chances for
a flaky remote to abort it fail-closed to one.

**Two properties keep the cache from weakening the control it speeds up.**

*Only a pass is ever recorded.* A failing gate aborts the rollout, so there is nothing to save on
that path, and a cached failure could outlive its cause — the PR opened to satisfy it, or the
merge that landed. A cache entry therefore cannot turn a pass into a failure.

*Anything unexpected is a miss, never a pass.* An unreadable, unparsable, expired or
non-matching entry, or a git command that fails while the key is built, all fall through to the
real check.

**What the key covers, and why it is content and not filenames.** `HEAD` plus the content of the
diff against it, plus the untracked file list, plus branch, facet set and environment. The
content half is load-bearing: `git status` porcelain output is *identical* when an
already-modified file is edited again, so a name-only fingerprint would hand the new content a
pass taken on the old one. Untracked files are keyed by name only, which is sufficient rather
than sloppy — an untracked file in a facet's closure has no counterpart on `origin/main` or at
any audited commit, so it can only ever push the verdict toward failure, whatever it holds.

`lib/` was the coverage question worth checking rather than asserting, since submodule content is
not in this repo's tree. Probed on a synthetic superproject, against what `divergedSubmodules`
actually blocks on:

```text
                                   key moves   gate would block
submodule HEAD moved off pin       yes         ["lib/dep"]
submodule tracked file edited       yes         ["lib/dep"]
submodule untracked stray (.DS_Store)  no      []            # deliberately ignored by both
```

The key moves in exactly the two cases the gate refuses and stays put in the one it deliberately
ignores, so a pass cannot be reused across a `lib/` divergence. All three are now tests, not just
a probe.

**Where the record lives.** The checkout's own git directory (`git rev-parse
--absolute-git-dir`), not the system temp directory — a world-writable location would let any
local process plant a pass for a key it can compute, and `.git` is already the trust boundary of
the checkout being deployed. Being outside the working tree also stops the cache from appearing
in the `git status` its own key is built from. The full key is re-compared on read, so a planted
or colliding entry cannot stand in for a different tree.

**Concurrency.** `proposeContractToNetworks.sh` runs `MAX_CONCURRENT_JOBS` workers per wave
(default 10), which previously all reached `resolveMainRef` together and raced each other's
`git fetch` on `refs/remotes/origin/main.lock` — fail-closed, so a lost race aborts that network.
A single-flight lock now means one worker does the network work and the rest reuse its verdict.
Verified cross-process on the real repo from a cold cache, five concurrent invocations:

```text
computed fresh: 1     reused: 4     all exit 0     locks left behind: 0
```

A lock left by a killed holder is taken over after 5 minutes, and the takeover claims it by
`rename` rather than removing it in place — with in-place removal, two waiters that both saw it
as stale can each end up believing they hold it. A lock that cannot be *created* at all
(unwritable git dir) is deliberately distinguished from one that is held, because waiting out the
full 2-minute window for a lock that will never appear would stall every invocation of the
rollout.

**What the cache trades away, stated plainly.** For up to 30 minutes the rollout is judged
against `origin/main`, and against the open-PR lookup, as they stood at its first invocation — so
`main` moving, or the anchoring PR being closed mid-rollout, does not stop the remaining
networks. Both are benign for one operator action on an unchanged tree: the code was merged, or
audited and under an open PR, when the verdict was taken. `DEPLOY_GATE_SKIP_VERDICT_CACHE=true`
forces a fresh verdict. Expiry is cheap by design — the next invocation just recomputes and
re-records, so the TTL bounds staleness without risking a stall.

**One thing the tests do not prove.** The stale-lock takeover *path* is tested, and so is the
single-flight behaviour, but the specific two-waiter interleaving the `rename` closes is not
deterministically reproducible in a test — that fix is argued from atomicity, not demonstrated.

# Evidence

`deployUpgradesToSAFE` has zero remaining references anywhere in the tree:

```text
$ git grep -n "deployUpgradesToSAFE" -- .
$ echo $?
1
```

Menu renumbers cleanly, 13 → 12, no gap:

```text
"9) Review deploy status (vs. target state)" \
"10) Create updated target state from Google Docs (STAGING or PRODUCTION)" \
"11) Update diamond log(s)" \
"12) Remove facets or periphery from diamond"
```

`bash -n` clean on both `scriptMaster.sh` and `diamondUpdateFacet.sh`.
`bun test script/deploy/github/ script/deploy/shared/propose-diamond-cut.test.ts` — 99 pass /
0 fail across the gate, cache and funnel suites; `bun test script/deploy/` — 1040 pass / 0 fail.

The cache's tests are checked by mutation rather than trusted: every property it relies on was
broken in turn and the suite caught each one.

```text
cache failures too                     -> CAUGHT   never caches a failure
drop diff content from the key         -> CAUGHT   already-modified file edited again
stop comparing the full key            -> CAUGHT   planted entry; unkeyed entry
no single-flight lock                  -> CAUGHT   concurrent callers; live holder
never expire                           -> CAUGHT   pass aged out
cache non-production too               -> CAUGHT   staging; unrecognised environment
never take over a stale lock           -> CAUGHT   lock left by a dead holder
treat an uncreatable lock as held      -> CAUGHT   unwritable cache dir
ignore the skip flag                   -> CAUGHT   skip flag; CLI remote-contact case
```

The CLI case in that last row is the one that proves the cache actually removes the remote
contact rather than merely being fast: the fixture's `origin` is repointed at a path that does
not exist between the two runs, so the second run passing means it never reached the remote — and
the same run with `DEPLOY_GATE_SKIP_VERDICT_CACHE=true` fails closed, confirming the recorded pass
was the only reason it passed.

The gate-condition test previously extracted the **first** line matching `$ENVIRONMENT` from the host
script. `diamondUpdateFacet.sh` carries an earlier unrelated `$ENVIRONMENT` condition (the
`SHOULD_PROPOSE_TO_SAFE` predicate), so the naive scan matched the wrong line and the `prod` /
empty-string cases failed. It now anchors on the `verify-approvals.ts` invocation and walks
backwards — a fix that also makes the test robust to future edits above the gate.

The condition matrix is exercised by running the real line out of the shell file, with
`isTestnetNetwork` stubbed on a marker so the test asserts that the condition *consults* it rather
than reimplementing `helperFunctions`' network classification:

| `ENVIRONMENT` | network | gate |
|---|---|---|
| `production` | mainnet | RUNS |
| `prod` | mainnet | RUNS |
| `` (empty) | mainnet | RUNS |
| `staging` | mainnet | SKIPPED |
| `production` | testnet | SKIPPED |
| `staging` | testnet | SKIPPED |

Falsification: removing `&& ! isTestnetNetwork "$NETWORK"` from the shell script fails exactly one
test (`production` / TESTNET), confirming the new row can actually fire.

Fail-closed paths confirmed by reading the code rather than assuming: `checkFailure` calls `exit 1`,
so a `jq` failure on `config/global.json` cannot yield an empty facet list; and
`collectDeployGateFailures` rejects an empty list outright (`'No facets were passed to the check'`).

CI is unaffected: `script/deploy/smokeDeploy.sh` runs `ENVIRONMENT=staging`, so the gate never fires
in `deploy-smoke-test.yml`.


0xDEnYO:
## Gate review — a real bug in the re-homing, found and fixed

The combined review-gate agent for this PR died on an API error before reporting, so I verified its highest-risk item myself. It was a genuine bug and it would have broken **every production deploy from a non-main branch.**

**What was wrong:** the re-homed gate was handed `$CONTRACT_NAME`, which in `diamondUpdateFacet.sh` is the *update-script* name (`UpdateChainflipFacet`) — that script's facet selection greps for `Update` and never strips the prefix. But `verify-approvals.ts` resolves each `--facets` entry to `src/Facets/<name>.sol` (line 123) and then reads `@custom:version` from it. So it threw rather than reaching a verdict:

```text
=== gate with the UPDATE-SCRIPT name (what I had shipped) ===
 ERROR  Could not find version for UpdateChainflipFacet

=== gate on a FEATURE branch with the correct bare facet name ===
OK
```

The retired `deployUpgradesToSAFE.sh` stripped the prefix (`sed 's/Update//g'`) before calling the gate. That step was lost in the move — a clean example of why moving a check needs its input contract re-verified, not just its call site.

**Fixed in `5876a97de`:** `${CONTRACT_NAME#Update}`, plus a case for `UpdateCoreFacets`, which has no facet source of its own — it cuts the whole `coreFacets` list, so the gate reads that list from `config/global.json`.

**Coverage falsified, not assumed.** Checked every one of the 40 real `Update*.s.sol` scripts, expanding `UpdateCoreFacets` to its 12 entries:

```text
facet names the gate can resolve: 51   unresolvable: 0
checked: 51   missing @custom:version tag: 0
```

So no selectable update script can make the gate throw instead of deciding. Note the first version of that check only asserted the `.sol` file existed; the version tag is a second requirement, and it is checked separately above.

Also confirmed the shell condition still skips staging:

```text
ENVIRONMENT=staging -> GATE_SKIPPED
```

`bash -n` clean; `bun test script/deploy/github/verify-approvals.test.ts` — 27 pass / 0 fail.

**Still open for the reviewer: D9, in the PR body.** Whether `diamondUpdateFacet.sh` is the right home at all, given it runs once per *(network, facet)* rather than once per run and has five other callers. My recommendation remains the `propose-to-safe.ts` funnel. That question is unaffected by this fix.

**Also corrected in the PR body:** it repeated the OQ2 ruling's "three humans sign a proposal that then reverts". Per D8 (verified today), `safeOwners[0]` is the `deployerWallet` and is also the proposer, so with threshold 3 that is two human signatures plus an automated key. The ruling's conclusion is unaffected — the proposal still reverts.

0xDEnYO:
@coderabbitai review

0xDEnYO:
## Review gate — escalated findings (not auto-fixed)

Gate run against `d180d4e35` after the rebase onto `main`. Auto-fixable findings were applied and
pushed; the items below are **pre-existing behaviour inherited from #2128**, not introduced here,
but re-homing the gate onto the primary facet-cut path makes them material. Each changes behaviour
on the production deploy path, so none were auto-applied. All four are now documented in
`docs/MultisigSigningProcess.md` and the PR body rather than left implicit.

### E1 — the `main` short-circuit trusts the branch *name*, never the content

`script/deploy/github/verify-approvals.ts:398`

```ts
if (input.branch === 'main') return []
```

No comparison against `origin/main` happens on this path. Uncommitted, unpushed, or stale-local-`main`
facet edits therefore reach a production Safe proposal ungated. Reproduced on this branch: the same
working tree with one appended line in `src/Facets/AcrossFacetV4.sol` exits **1 (blocked)** on a
feature branch and **0 (passed)** with `--branch main`.

Why it matters more after this PR: the exemption previously guarded `deployUpgradesToSAFE.sh`, which
this PR's own docs call "effectively dead for production diamond cuts". It now guards
`diamondUpdateFacet.sh` — reached from `deployAllContracts.sh`, `deployFacetAndAddToDiamond.sh`,
`proposeContractToNetworks.sh`, `playgroundHelpers.sh` and `helperFunctions.sh`.

Fix is one line — make the short-circuit `matchesMain` instead of the branch name; `deps.sourceClosure`
+ `fileMatchesRef` already exist and return in well under a second when everything matches. **Escalated
because it changes what a `main` deploy is permitted to do**, which is a policy decision, not a cleanup.

### E2 — Tron facet cuts are not covered by the gate

These six call `proposeDiamondCut` (`script/deploy/shared/propose-diamond-cut.ts`) directly and never
invoke `verify-approvals.ts`, yet produce real production Safe facet-cut proposals:

```text
script/deploy/tron/deploy-and-register-allbridge-facet.ts
script/deploy/tron/deploy-and-register-eco-facet.ts
script/deploy/tron/deploy-and-register-layerswap-facet.ts
script/deploy/tron/deploy-and-register-lifi-intent-escrow-facet-v2.ts
script/deploy/tron/deploy-and-register-near-intents-facet.ts
script/deploy/tron/deploy-and-register-symbiosis-facet.ts
```

Documentation-only fix applied (the §5 controls table row and the §4.2 paragraph now name this gap
explicitly). Actually gating them is a separate change.

### E3 — the gate never fetches

`resolveMainRef` uses whatever `origin/main` was last fetched locally, so "matches main" can pass
against a `main` that has since moved. Raised by the gate on #2128 and accepted there, explicitly
because adding a network call to a deploy script is an externally observable change. Unchanged by
this PR — but `diamondUpdateFacet.sh` runs once per (network, facet), so a fleet rollout now
re-evaluates the same possibly-stale ref up to 71 times instead of once.

### E4 — a block names the wrong file

`collectDeployGateFailures` surfaces `changedSinceAudit` and never `divergedFromMain`. Editing only
`src/Helpers/SwapperV2.sol` produces:

```text
ERROR AcrossFacetV4 has changed since audited commit 650d18eb... (src/Errors/GenericErrors.sol,
src/Facets/AcrossFacetV4.sol, src/Helpers/LiFiData.sol, src/Helpers/ReentrancyGuard.sol,
src/Helpers/SwapperV2.sol and 6 more)
```

11 files, led by `AcrossFacetV4.sol`, which is byte-identical to main. During an incident this points
the operator at the wrong file.

---

- **Gate body has no direct test coverage.** The new test exercises the condition line only; `${CONTRACT_NAME#Update}`, the `UpdateCoreFacets` special case, and the abort-on-failure path were verified manually but nothing defends them against regression.
- **Fleet rate-limiting.** In the audited-freeze rollout (the case the gate exists for), up to 66 mainnet workers each call `gh pr list`; secondary rate-limiting would make `countOpenPRsForBranch` throw, fail closed, and abort the rollout mid-fleet. Same root cause as the placement question already in the PR body.
- **Pre-existing, out of scope:** the interactive zkEVM facet picker in `diamondUpdateFacet.sh:90` strips only `.s.sol`, leaving `.zksync` on the name and building a `…zksync.zksync.s.sol` path that never exists. It fails closed (aborts before the gate and before any proposal), but means the interactive zkEVM update path cannot currently be exercised. Callers that pass `SCRIPT` explicitly are unaffected.

---

### What the gate verified as correct

Name derivation is 1:1 across all 56 update scripts (40 EVM + 16 zkSync); `UpdateCoreFacets` is the
only miss and is exactly the special-cased one. Both `UpdateCoreFacets` variants read `.coreFacets`
from the same `config/global.json` the gate reads, so no drift is possible. The transitive-closure
comparison is honest — editing a shared helper while leaving the facet byte-identical still blocks.
Every constructed failure mode is fail-closed: detached HEAD, unresolvable facet name, missing
`bunx`, wrong cwd, `jq` failure, unknown network. `local` declarations are split from command
substitution, so no exit-code masking. The condition-extraction test is load-bearing, not decoration
— all four mutations of the gate condition are caught (delete the block: fail; `!= staging` →
`== production`: 2 fail; drop the testnet clause: 1 fail; invert to `isTestnetNetwork`: 4 fail).
No CI workflow invokes `diamondUpdateFacet`, so nothing in CI bypasses the gate.

docs/Deploy.md
```diff
@@ -100,9 +100,8 @@ Outcomes per network: `OK` (new proposal), `SKIP` (already registered on the dia
...
 - [ ] Ensure that you have granted access to a secondary wallet to whitelist contracts and selectors
 - [ ] Make sure the facet you wish to upgrade is deployed but not added to the diamond yet
-- [ ] Run this script `./scripts/scriptMaster.sh`, select `11) Propose upgrade TX to Gnosis SAFE`
+- [ ] Run this script `./scripts/scriptMaster.sh`, select `1) Deploy one specific contract to one network`
 - [ ] Choose the network you want to run the upgrade on
-- [ ] Choose the facet(s) you want to upgrade (you can select multiple using the spacebar)
-- [ ] Hit enter and select the SAFE wallet you want to use
-- [ ] Hit enter again and wait for the script to finish
+- [ ] Choose the facet you want to upgrade
+- [ ] Wait for the script to finish. On production, non-testnet networks the diamondCut is proposed to the network's Safe (from `config/networks.json`) automatically, wrapped in a timelock `scheduleBatch`
```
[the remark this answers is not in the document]
0xDEnYO: Both points fixed in `d180d4e35`.

- Path corrected to `./script/scriptMaster.sh`.
- Added the `yes - to LiFiDiamond` / `yes - to LiFiDiamondImmutable` step. Good catch — a `no` answer routes to `deploySingleContract` only, which never reaches `diamondUpdateFacet` and therefore never runs the gate this PR re-homes, so the checklist as written skipped the one decision point that determines whether the gate applies at all.

The same pass picked up three more errors in that block: the `deployed but not added to the diamond yet` precondition (true of the retired script, false for menu 1, which deploys and cuts in one run), the diamond described as Safe-owned when production mainnet diamonds are owned by `LiFiTimelockController`, and the final step pointing at the Gnosis Safe app when there is no Safe{Wallet} UI in this flow (it is `bun confirm-safe-tx`).

0xDEnYO:
## Review gate — round 2 (`122575b8b` … `74bb86799`)

Re-gated the fix commits, since the gate's clearance covers only the commits it reviewed. It found
the same bypass through a second door, inside the fix for the first one.

### Fixed in this round

**The `main` bypass, door 2 — `resolveMainRef` fell back to local `main`.** Removing the branch-name
short-circuit (`122575b8b`) made the gate compare *content*; but the *ref it compared against* fell
back to local `main` whenever `origin/main` was unresolvable, and local `main` is whatever the
operator last committed. Reproduced in a throwaway clone with no `origin/main` and an unreviewed
facet edit committed on local `main`:

```text
before   EXIT=0   OK
after    EXIT=1   Cannot resolve origin/main in this checkout. Fetch it before deploying
sanity   EXIT=0   OK      # origin/main restored, tree matches it - normal path unaffected
```

Reachable without malice: a remote not named `origin`, or a single-branch checkout. The fallback is
gone — `origin/main` or nothing.

**Two message defects in the code `122575b8b` added.** The remedies were wrong ("reset the tree" does
not remove an untracked-but-imported `.sol`; "merge and pull" is unactionable for local commits on
`main`), and a `main` block emitted a second, misleading error pointing at an audit-log problem that
cannot apply on `main`, implicating 11 files the operator never touched. The `main` path now returns:

```text
ERROR  Deploying from "main", but the working tree does not match origin/main. Move the change
       onto a branch and open a PR, or discard it (git checkout / git clean) and pull, before deploying
ERROR  AcrossFacetV4 diverges from origin/main (src/Facets/AcrossFacetV4.sol)
```

**A false assurance in the JSDoc.** `resolveSolidityImport` claimed dependencies outside `src/` are
safe because "submodules under `lib/` are pinned by their submodule commit". The gate checks no such
thing. Corrected, and recorded as a known gap below and in §4.2.

### Escalated — still open, documented, not fixed here

1. **The closure stops at `src/`.** `lib/` content is never compared and no submodule gitlink is
   checked against main, so an edited `lib/` checkout changes deployed bytecode while the gate
   reports a match. `src/Libraries/LibAsset.sol` — in `AcrossFacetV4`'s closure — imports
   `solady/utils/SafeTransferLib.sol`.
2. **Tron facet cuts are ungated** (six `script/deploy/tron/deploy-and-register-*.ts` call
   `proposeDiamondCut` directly).
3. **The gate never fetches.** `origin/main` is now mandatory but is still whatever was last fetched.
4. **A deleted imported file drops out of the closure** and the gate passes. Not exploitable (the
   forge compile fails immediately after), but the `continue` written for commented-out imports
   catches this case too.
5. **No memoization.** A 12-facet `UpdateCoreFacets` run costs 109 `git show` calls for 37 distinct
   paths (~3.4s).

### Verification

49 tests, 0 fail. Every mutation that restores a bypass is caught: re-adding the short-circuit to
`collectDeployGateFailures` fails 3 tests, to `verifyDeployGate` 2, restoring the local-`main`
fallback 1, swapping `deps.mainRef` for `HEAD` 2, dropping the main-exemption on the GitHub lookup 1.
Independently confirmed by the gate on real repo data: the normal feature-branch path (merged code)
passes in 0.66s making **zero** GitHub calls — 14 git invocations, all `rev-parse`/`show`; unknown
environments, missing facet files, detached HEAD and stale local `main` all fail closed; filemode
changes, `.gitattributes`-normalised line endings and untracked non-imported files produce no false
positives; untracked *imported* files correctly block.

Note for reviewers: **CodeRabbit has not reviewed `122575b8b` or later** — it went rate-limited and
its check reports `pass` as a skip, not a review. Worth re-triggering before merge.

0xDEnYO:
@coderabbitai review

Context for this run: CodeRabbit's last review covered `68fed8e73` only — every commit after that hit the rate limit and the check reported `pass` as a skip rather than a review. The gate-logic changes all live in the un-reviewed range (`122575b8b`..`3a634c29b`).

docs/MultisigSigningProcess.md
```diff
@@ -112,15 +107,56 @@ All EVM funnels end in `storeTransactionInMongoDB`
...
+(staging, and any network whose `config/networks.json` type is `testnet`, which
+is how `tronshasta` stays open).
+
+Facet **removals** are deliberately outside both gates: `cleanUpProdDiamond.ts`
+and the deferred-cleanup drain (`drain-parked-tasks.ts`, which folds extra
+removal calls into whatever proposal `runPropose` is already building) propose
+real diamond cuts, but a removal installs no new bytecode, so a
+main-equivalence check has nothing to compare. Their safety comes from the
+removal-specific controls in the table below. The generic bash `sendOrPropose`
+chokepoint can likewise propose arbitrary calldata and is not gated.
```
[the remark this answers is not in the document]
0xDEnYO: Valid finding, and deliberately documented rather than introduced here — that doc line is this PR disclosing the gap, not creating it. Deferring it, with reasoning:

**Why it is real.** `sendOrPropose` (`script/helperFunctions.sh:4216`) takes an arbitrary target and calldata and proposes to the production Safe with no facet check. A hand-built `diamondCut` Add would reach the Safe ungated.

**Why not in this PR.** The fix you describe — reject facet additions on that route — needs calldata decoding in the bash chokepoint to tell an Add-cut from the traffic that legitimately flows through it: `universalCast.sh`, `diamondSyncWhitelist.sh` whitelist syncs, and `playgroundHelpers.sh` timelock admin. That is its own design question (what happens to a deliberate manual cut during an incident?), and you tagged it Heavy lift yourself. This PR retires `deployUpgradesToSAFE.sh` and re-homes its gate; bolting a decoder onto the generic chokepoint is a different change with a different blast radius.

**One correction to a worry I had while triaging this:** gating that route would *not* endanger the emergency pause. `diamondEMERGENCYPause.sh:186-188` deliberately bypasses `universalCast`/`sendOrPropose` and delegates to `script/emergency/emergencyPauseBreakGlass.sh` precisely so routing churn cannot break the pause. So the tension I assumed exists does not — which makes this more tractable as follow-up work than I first thought.

Tracking it as a follow-up under Signing 2.0 (EXSC-686) rather than expanding this PR. The gap is stated explicitly in §4.2 and in the §5 controls table so it is not silently carried.

script/tasks/diamondUpdateFacet.sh
```diff
@@ -123,6 +123,32 @@ diamondUpdateFacet() {
     SHOULD_PROPOSE_TO_SAFE=true
   fi
 
+  # getPrivateKey hands out the production key for every ENVIRONMENT that does not
+  # contain "staging", so matching on the exact string keeps the gate at least as
+  # broad as the key it protects; an unrecognised value is then rejected downstream.
+  # Testnets are exempt: deploying an unmerged facet there is how it gets validated
+  # before the audit, and no mainnet Safe is involved.
+  if [[ "$ENVIRONMENT" != "staging" ]] && ! isTestnetNetwork "$NETWORK"; then
```
gvladika, line 131: **Placement (D9): I'd move the gate to `propose-to-safe.ts` in this PR rather than defer it.**

The gate compares the working tree against `origin/main`, and that verdict is invariant across a rollout — the tree doesn't change between networks. So gating here, once per (network, facet), recomputes the identical answer for every network: a 71-network rollout runs the whole check 71 times to reach the same conclusion once would. That redundancy buys two real costs for nothing:

- 71× `ls-remote` + `gh pr list` on the production deploy path — 71 chances for a flaky remote to stall a fleet rollout (fail-closed, so it aborts mid-rollout).
- The gate now also fires from the five other `diamondUpdateFacet.sh` callers, widening the reach beyond what the retired script covered.

`propose-to-safe.ts` is the single funnel every proposal already passes through — gating there runs the check exactly once per proposal and covers future callers by construction. Re-homing the gate is already the substance of this PR, so putting it at the funnel is the same work in the right place, not a bigger change deferred to a follow-up.

0xDEnYO: Diagnosis agreed — the verdict really is invariant across a rollout. But the move doesn't remove the redundancy: `propose-to-safe.ts` is invoked per (network, facet) too. `diamondUpdateFacet.sh` calls it once per cut (`script/tasks/diamondUpdateFacet.sh:250` and `:265`), so the count goes 71 → 71, with the same remote round trips and the same 71 fail-closed chances.

The cost is also worse than the one you name, in a way neither of us called out: `proposeContractToNetworks.sh` launches `MAX_CONCURRENT_JOBS` workers per wave (default 10, `:358`). Measured here, `git ls-remote origin main` takes 4.30 / 4.41 / 4.35s — long enough that 10 workers all see a stale `origin/main` and all `git fetch` the same ref concurrently, racing on `refs/remotes/origin/main.lock`. Fail-closed, so a lost race aborts that network mid-rollout.

Two things about the funnel itself:

- Its argv is `--to` / `--calldata` — there is no facet name at that layer (`script/deploy/safe/propose-to-safe.ts:317-344`). Gating there means decoding timelock-wrapped `diamondCut` calldata, reverse-mapping facet address → name, and adding a pass-through for every non-cut payload the funnel carries (`diamondEMERGENCYPause.sh`, `diamondUpdatePeriphery.sh`, playground, the generic route at `helperFunctions.sh:4334`) — plus `propose-to-safe-tron.ts` is a second funnel. That is more than a re-home.
- The TS side is already gated in this PR, at `script/deploy/shared/propose-diamond-cut.ts:163`, which covers the six Tron `deploy-and-register-*-facet.ts` callers. So the two paths that cut a facet today are both gated, disjointly — no double run.

What I'd rather do here is fix the actual redundancy: memoize the verdict per run. The network-touching work depends on (HEAD sha, working-tree state, facet set) and never on the network, so a cache keyed on `HEAD` + a hash of `git status --porcelain` + the facet list makes the first invocation pay 4.3s and the other 70 free, and removes the concurrent-fetch race. Unreadable or stale key ⇒ cache miss, never a pass. Tradeoff worth stating: the whole rollout is then judged against `origin/main` as of run start.

One more thing in favour of your placement, stronger than the cost argument: the generic `--calldata` route can hand-build a cut and reach production through neither gate — the same hole CodeRabbit flagged as merge risk. Only a decode-at-the-funnel gate closes it. I'd do that as its own PR, scoped to the decode + reverse-map + pass-through + Tron, and justified on coverage.

Your call on whether that split is acceptable, or whether the coverage gap is worth pulling the funnel work into this PR.

0xDEnYO: You are right on the substance, and your cost analysis matches mine — the verdict is invariant across a rollout, so 71 recomputations buy nothing and add 71 chances for a flaky remote to abort a fleet deploy mid-way.

The reason it is not in this PR is a sequencing decision that was already ruled, not disagreement. Daniel ruled D9 on 2026-08-31: the gate moves into the `propose-to-safe.ts` funnel, as its own package **WP-1.4**, and *"#2286 stays exactly as it is — its `diamondUpdateFacet.sh` call site is the correct interim home, and WP-1.4 removes that call site when it lands, so the gate is never live in two places."* The reasoning was that retiring `deployUpgradesToSAFE.sh` removed the gate's only caller, so shipping the deletion without *a* host would have dropped the control entirely — and the interim host must not regress while the funnel is built.

So the disagreement is only about whether the funnel move happens here or in the next PR. Two things worth weighing, and I do not think this is mine to decide:

- **For doing it here (your position):** it is the same work in the right place, and it avoids landing a known-redundant 71× network dependency on the production deploy path even temporarily.
- **For the ruling as written:** WP-1.4 is scoped to carry *both* intent checks into that funnel — the deploy gate **and** WP-1.2's mandatory-Linear-ticket-link hard block. Doing half of it here splits one funnel change across two PRs, and WP-1.2 has not landed yet.

I have raised it with Daniel as a sequencing question on the ticket rather than picking, because changing a ruled sequencing call is not a reviewer-round decision I should make unilaterally. **If he says do it here, I will move the gate in this PR** — it is a contained change and I have the funnel work scoped already.

One thing that is not in dispute either way: this PR must not merge with the gate in two places, and it will not — whichever PR does the move removes this call site in the same change.

Separately, and unrelated to your point: this PR's only red check is `protect-critical-code`, which needs a security-manager approval because it touches security-relevant paths. That is not something I can clear.

0xDEnYO: Implemented as agreed in Slack: verdict caching lands here, the funnel move stays WP-1.4 (which removes this call site when it lands, so the gate is never live in two places).

`script/deploy/github/deploy-gate-cache.ts` records a pass once per run and reuses it while the tree stays put. Measured on this repo, same facet:

```text
cold    5185 ms   (ls-remote + closure comparison)
warm     481 ms   (three local git calls, no remote contact)
```

71 `ls-remote` round trips → 1, and 71 fail-closed chances → 1. Cross-process single-flight verified on the real repo, five concurrent invocations from a cold cache: 1 computed, 4 reused, all exit 0, no lock left behind — so the `refs/remotes/origin/main.lock` race between `MAX_CONCURRENT_JOBS` workers is gone too.

Two properties keep it from weakening the control: **only a pass is ever recorded** (a failing gate aborts the rollout anyway, and a cached failure could outlive the PR opened to satisfy it), and **anything unexpected is a miss, never a pass**.

Three things I'd point you at rather than let you find:

- The key is `HEAD` + the *content* of the diff against it, not `git status` filenames — porcelain output is identical when an already-modified file is edited again, so a name-only key would hand new content a pass taken on the old one.
- `lib/` was the coverage question worth probing rather than asserting. The key moves in exactly the two cases `divergedSubmodules` blocks on (gitlink moved, tracked file edited) and stays put for the stray-`.DS_Store` case both deliberately ignore. Now tests.
- The record lives in the checkout's git dir, not `/tmp` — a world-writable location would let any local process plant a pass for a key it can compute. Full key re-compared on read.

What it trades away, and it is in the PR body too: for up to 30 min the rollout is judged against `origin/main` **and the open-PR lookup** as they stood at its first invocation, so neither `main` moving nor the anchoring PR closing mid-rollout stops the remaining networks. `DEPLOY_GATE_SKIP_VERDICT_CACHE=true` forces a fresh verdict.

Every property above is mutation-tested (9 mutations, all caught) — same-PR tests written from the same mental model as the code are not evidence on their own. One honest gap: the stale-lock takeover path is tested, but the two-waiter interleaving the `rename` closes is not deterministically reproducible, so that fix is argued from atomicity rather than demonstrated.

0xDEnYO: Agreed on the principle — the funnel is the right home, and it is the same conclusion I reached putting the ticket-link gate in `storeTransactionInMongoDB` rather than at its eleven call sites. Two measured things change what the move buys, though, and one of them is already in this PR.

**The 71× cost is already gone.** `verifyDeployGateForRepo` runs inside `withVerdictCache` (`deploy-gate-cache.ts`), which is a **file** cache under `$GIT_DIR/lifi-deploy-gate-cache`, so it survives across the 71 separate `bunx tsx` processes. 30-minute TTL, `DEPLOY_GATE_SKIP_VERDICT_CACHE=true` to force a fresh verdict, and — the part that matters — it caches **passes only** (`deploy-gate-cache.ts:336`), so it can never turn a failing gate into a passing one. A 71-network rollout pays one real `ls-remote` and 70 cache hits.

Its docstring is your argument, near enough verbatim: *"the verdict depends on the tree and the facet set, never on the network, so a fleet rollout would otherwise pay the same `ls-remote` for every one of its networks."*

**And moving to `propose-to-safe.ts` would not reduce the invocation count anyway.** `diamondUpdateFacet` takes `NETWORK` as `$1` and calls `propose-to-safe.ts` per network (lines 250, 265), so the funnel is invoked 71 times too. The reduction available from the move is the *per-facet* multiplier, not the per-network one.

**The coverage argument is the real one, and it needs more than a move.** `propose-to-safe.ts` knows neither the environment nor the branch — no reference to either exists in the file — and the gate's `--facets` is required and resolved to `src/Facets/<name>.sol`, with an empty list failing closed by design (`verify-approvals.ts:168`). The calldata is a `diamondCut`: facet *addresses* and selectors, no names. So "covers future callers by construction" requires the funnel to decode the cut and resolve addresses to names against the deployment record — which is exactly WP-1.4/D9's scope, and what #2275 (EXSC-855) is building the decode for.

Two ways forward, and I would rather you pick than guess:

1. **Do it in WP-1.4**, where the decode lives. The funnel gate is then unconditional and genuinely covers every caller. This PR keeps the gate at `diamondUpdateFacet.sh` in the meantime, with the cache making the cost a non-issue.
2. **Do it here as three threaded flags** (`--gate-environment`, `--gate-branch`, `--gate-facets`) set by the one caller that knows them. That moves the check to the funnel and removes the per-facet multiplier, but it is **opt-in at the funnel** — no "by construction" property, and coverage identical to today. I would take this only as a stepping stone to (1), and it is worth being explicit that it does not deliver the property that motivates it.

On the wider reach: the gate is conditioned on `ENVIRONMENT != staging && !isTestnetNetwork`, so the additional `diamondUpdateFacet.sh` callers it now covers are production-mainnet facet updates. I read that as the gate reaching cases the retired script missed rather than as scope creep — but say the word if you meant specific callers that should be exempt.

Happy to implement (2) in this PR today if you prefer the funnel location now; otherwise I will carry it into WP-1.4 and link back here.

0xDEnYO: Still your call, no change since yesterday: (1) keep the gate here with the per-run cache (landed — 71 `ls-remote` → 1, cross-process single-flight verified on this repo) and move it into the funnel in WP-1.4, or (2) I thread `--gate-environment`/`--gate-branch`/`--gate-facets` through this PR now as a stopgap — removes the per-facet multiplier, but it's opt-in at the funnel, not "covers future callers by construction." Default is (1) per Daniel's D9 ruling unless you'd rather have (2) now.

Separately: fixed the lock-ownership gap CodeRabbit flagged on this same file — a holder that outlives the 5-minute stale threshold could delete a successor's lock in its own `finally`. `releaseLock` now checks an owner token before removing; added a regression test that fails on the old code and passes with the fix.

docs/MultisigSigningProcess.md
```diff
@@ -141,15 +136,80 @@ run only on the branches that actually propose; a staging or testnet-only run, a
...
+untracked files that change no bytecode — most of these submodules do not gitignore
+`.DS_Store`, so the stricter `none` would block every deploy from a Mac); a divergence
+there is not excused by an open PR or an audit freeze. The remote calls run with
+prompts disabled and a 30 s timeout, so a stalled or credential-prompting remote fails
+the gate instead of hanging the rollout. Staging is not gated, and neither are
+testnets — deploying an unmerged facet to a testnet is how it is validated before
+the audit, and no Safe is involved there.
+
+The gate runs once per *(network, facet)*, but its verdict depends only on the working
+tree and the facet set, so a fleet rollout would otherwise recompute the same answer for
```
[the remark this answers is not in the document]
0xDEnYO: Correct — the cache key also depends on branch and environment (line 172), so "depends only on the working tree and the facet set" overstated it. Fixed wording here and in the matching docstring in `script/deploy/github/deploy-gate-cache.ts` (same overstatement existed there too).

script/deploy/github/deploy-gate-cache.ts
```diff
@@ -0,0 +1,348 @@
...
+    return 'acquired'
+  } catch (error) {
+    // someone acquired it between the takeover and here; they hold it, not us
+    return isAlreadyExists(error) ? 'held' : 'unavailable'
+  }
+}
+
+const releaseLock = (lockDir: string): void => {
+  try {
+    rmdirSync(lockDir)
```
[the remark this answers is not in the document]
0xDEnYO: Split this into the two claims:

**`releaseLock` not verifying ownership — real bug, fixed.** A holder that outlives `LOCK_STALE_MS` (5 min) could wake up after a takeover had already claimed the lock for someone else and delete that successor's live lock in its own `finally`. `acquireLock` now writes an owner token (`randomUUID()`) into the lock dir on every successful claim, and `releaseLock` only removes the directory when the token still matches — a holder that lost the lock to a takeover now finds someone else's name inside and leaves it alone. Added a regression test (`does not let a stale holder tear down the lock a takeover already claimed`) that fails on the old code and passes with the fix.

**Full lease-renewal system — not doing that here.** The other half of the suggestion (wait indefinitely for a live holder, renew the lease during a long `compute`) targets a different, already-deliberate tradeoff: `LOCK_WAIT_MS` (2 min) existing at all is the point — waiting out a lock that can never be released (a crashed holder, no `LOCK_STALE_MS` reached yet) would stall every worker in the rollout for the full window instead of falling back to computing independently. A live holder finishing in its normal ~5s (measured) never gets near that fallback; reaching it means something is already abnormal, and the existing "only a pass is ever cached, anything unexpected is a miss" design bounds the blast radius of that rare case. Owner tokens fix the concrete correctness gap without taking on a lease-renewal protocol for a scenario the current design already treats as degraded.

gvladika (approved):

PR #2288, opened 2026-08-31
fix(safe): refuse to broadcast when EVM gas estimation fails (EXSC-689)

0xDEnYO:

Fixes [EXSC-689](https://linear.app/lifi-linear/issue/EXSC-689/s3-refuse-to-broadcast-when-gas-estimation-fails) — S3 / WP-0.4 of Signing 2.0.

`getGasWithFallback` caught **every** estimation error and returned a fixed limit:

```ts
} catch {
  consola.warn(`Gas estimation failed; using fallback gas limit: ${fallbackGas}`)
  return fallbackGas
}
```

Both callers then broadcast on that guess. A failed estimate usually means the transaction *would
revert* — so this burns gas for nothing and, on the Safe path, can consume a nonce.

**The failure mode is now a required parameter, not an implicit default.** That is the part that
matters: a future call site cannot inherit the unsafe behaviour by omission, because the type system
makes it declare intent.

There are **three** call sites, and they do not all want the same thing:

| Call site | Broadcasts? | Mode |
|---|---|---|
| `evm-executor.ts` → `writeContract` | yes | `refuse` |
| `evm-caller.ts` `call()` → `sendTransaction` | yes | `refuse` |
| `evm-caller.ts` `simulate()` | **no** | `fallback` |

`simulate()` is the one worth noticing. A naive "make `getGasWithFallback` throw" breaks the dry-run
path, and a simulation that throws tells the operator *less* than one that says "couldn't estimate,
assuming the fixed limit". It broadcasts nothing, so it has nothing to refuse.

`ALLOW_GAS_ESTIMATE_FALLBACK` is the deliberate override. Only an affirmative value (`true`/`1`/`yes`/
`y`/`on`) opens it, so a stray `export ALLOW_GAS_ESTIMATE_FALLBACK=` cannot silently restore the old
behaviour. Added to `.env.example` per the TS rules.

## Scope: EVM only

The two paths this fixes are `evm-executor.ts` and `evm-caller.ts`. **Tron is not covered, and it is worse than uncovered — it never estimates at all.** `tron-executor.ts` / `tron-caller.ts` delegate to `@lifi/tron-devkit`, which goes straight to `triggersmartcontract` with a hardcoded `fee_limit` and never calls the `estimateContractCallEnergy` that exists in the same package. The repo already documents the consequence in `runPendingTimelockTXs.yml:102`: a multi-call timelock batch exceeds the 50-TRX default, "aborts mid-SSTORE as OUT_OF_ENERGY and is retried forever". That is the same class of harm this PR removes for EVM. **Out of scope here — filed separately — but the title says EVM for that reason.**

## No jovay allow-list — and the evidence for that

The work package left "allow-list vs fix the RPC" open, to decide with evidence. The old module
docstring claimed jovay is a chain where `eth_estimateGas` "can revert even when `eth_call` with
unlimited gas succeeds". Probed against the public RPC in `config/networks.json`:

PR #1762 documents two layered jovay quirks. Quirk 2 is the one the fallback was built for: *"Jovay's `eth_estimateGas` binary-search simulation reverts even when `eth_call` with unlimited gas succeeds."*

A 0-wei self-send cannot exercise that — a bare value transfer has no contract execution, so there is nothing for the binary search to bisect. So the probe below is a real **contract call** against the jovay diamond (`0x026F252016A7C47CDEf1F05a3Fc9E20C92a49C37`, `owner()`):

```text
1) eth_call        -> 0x...ddb12a9303f7bdda8c7c39e58d42af3896330fed   (succeeds)
2) eth_estimateGas -> 0x87c6  (34758)                                 (succeeds)
3) mainnet, same shape -> 0x700f (28687)
```

jovay's estimator **works on contract calls today**, not just on self-sends. Higher than mainnet, consistent with the tight simulation-vs-execution delta the **multiplier** exists for — and that multiplier, from PR #1762, is untouched. What is not reproducible is the *outright failure* the fallback was built for.

**Two caveats, stated because they bound the claim:**

1. The actual failing call in #1762 was `estimateContractGas` on Safe `execTransaction` **with three valid signatures**. That cannot be constructed without real signatures, and an invalid-signature probe returns a genuine `GS020`/`GS026` revert — indistinguishable from quirk 2. So no probe here is decisive on the exact shape.
2. **This probe hit a different endpoint from the one the deploy path uses.** `config/networks.json` gives jovay `https://rpc.jovay.io` (public, keyless), but `runPendingTimelockTXs.yml:73` runs `bun fetch-rpcs` and `viemScriptHelpers.ts` reads `ETH_NODE_URI_*` from MongoDB — a likely premium, keyed URL I cannot see. That is also why the redaction fix below is critical rather than theoretical.

**Operational consequence worth knowing:** if quirk 2 still exists on the Mongo endpoint, the unattended scheduled executor now hard-fails on jovay instead of proceeding. The remedy is `ALLOW_GAS_ESTIMATE_FALLBACK=jovay` — scoped, which is why the hatch takes a network list.

So: **no allow-list.** A standing per-chain exemption rots silently — a jovay entry would already be
stale — whereas the env override is per-invocation and visible in the operator's own command. Caveat
stated plainly: this probe is a plain transfer, not a Safe `execTransaction`, so it does not prove
jovay can never fail to estimate. It does show the failure is payload/transient rather than a
property of the chain, which is what an allow-list would have encoded.

# Evidence

The refusal, on a broadcast path:

```text
Gas estimation failed on jovay for Safe execTransaction — refusing to broadcast.
  Underlying error: execution reverted: GS013
  A failed estimate usually means the transaction would revert, so broadcasting on a guessed limit burns gas for nothing and can consume a Safe nonce.
  Investigate the revert first. To broadcast anyway on a fixed limit of 500000, re-run with ALLOW_GAS_ESTIMATE_FALLBACK=true.
```

Only an affirmative value opens the hatch:

```text
ALLOW_GAS_ESTIMATE_FALLBACK="false" -> still refuses
ALLOW_GAS_ESTIMATE_FALLBACK="0"     -> still refuses
ALLOW_GAS_ESTIMATE_FALLBACK="no"    -> still refuses
ALLOW_GAS_ESTIMATE_FALLBACK=""      -> still refuses
```

Dry run unaffected: `simulate() -> 500000` (reports, no throw). Success path unchanged:
`estimate 100000 @ default 130% -> 130000`.

**The guard is proven where it counts** — `refuse-to-broadcast.test.ts` drives the real
`EvmChainExecutor` / `EvmChainCaller` with a throwing estimator and a spy on the write method:

- `EvmChainExecutor.executeTransaction` → `writeContract` called **0** times
- `EvmChainCaller.call` → `sendTransaction` called **0** times
- `EvmChainCaller.simulate` → returns `500000n`, **0** writes
- with the hatch set → `writeContract` called **1** time, deliberately

Tests: `bun test script/` — **1107 pass / 0 fail**. Baseline on `origin/main` @ `c9c70db3f`:
**1090 pass / 0 fail**. Delta is exactly the +17 new tests / +1 file; verified #2284 (the only commit
between `0fd909d20` and the base) touched no `script/` or test files, so the baseline is comparable.
No Solidity touched.


0xDEnYO:
## `/gate-review` round 1 — applied

Adversarial falsification pass. **One critical finding, which I had introduced in this PR**, plus four majors. All fixed in the fix-push.

### CRITICAL — the refusal leaked live RPC API keys

My refusal message interpolated the raw viem error. viem embeds the full request URL **including the query string** in `error.message`, and that is exactly where dRPC/Alchemy keys live:

```text
lb.drpc.org/ogrpc?...&dkey=DRPCKEY_AAAA1111  -> key present in error.message: true
eth-mainnet.g.alchemy.com/v2/ALCHEMYKEY_...  -> key present in error.message: true
```

Worse, it reaches **Slack**, which sits outside `::add-mask::` — traced `getGasWithFallback` → `chainCaller.call()` → `execute-pending-timelock-tx.ts` `notifyFailure` → `slack-notifier.ts` `extractErrorMessage` (returns `.message` verbatim) → webhook post. Local operator runs have no masking at all.

This was **newly introduced by this PR** — `origin/main`'s version never touched `error.message`. And the repo had already ruled on it: `script/utils/redactUrls.ts` exists and its docstring says verbatim *"Anything derived from an error message must therefore pass through here before it is published."* Now routed through `redactErrorReason`, with tests asserting no `https://`, no host, and no key survives:

```text
46 pass, 0 fail  (includes 2 leak cases × 6 assertions each)
```

### MAJOR — the escape hatch was global, not per-network

`execute-pending-timelock-tx.ts` runs `Promise.all` over every network with pending work. Setting `ALLOW_GAS_ESTIMATE_FALLBACK=true` because **jovay** cannot estimate would have disabled the guard for **every** network in that run — and jovay is the entire motivating case, so this was the likely first real use. It now also accepts a comma-separated network list (`ALLOW_GAS_ESTIMATE_FALLBACK=jovay`), and the refusal message suggests the scoped form rather than the global one.

### MAJOR — the dry run printed a green simulation on a failed estimate

`simulate()` correctly falls back rather than refusing (it broadcasts nothing), but the caller then printed `✅ [DRY RUN] Transaction simulation successful` for a transaction whose estimate had failed — the exact condition the real run treats as disqualifying. `--dry-run` said ✅, the real run refused. `IChainSimulateResult` now carries `estimateFailed` and both dry-run sites report **INCONCLUSIVE**.

*(Verified the safe half too: every `.simulate()` caller is inside an `if (isDryRun)` branch, so nothing ever broadcast off a fallback figure.)*

### MAJOR — the refusal was relabelled as a Safe revert

`safe-utils.ts` matched `execution reverted` in the message and rethrew as `Safe execution reverted:`. Since the refusal quotes the underlying estimation error, the operator saw *"Safe execution reverted: … refusing to broadcast"* — implying a nonce was consumed when nothing was sent. The refusal is now matched ahead of the revert case.

### MAJOR — my jovay evidence talked past the actual bug

PR #1762 documents that jovay's `eth_estimateGas` **binary-search simulation** reverts. A 0-wei self-send has no contract execution, so it cannot exercise that. Re-probed with a real contract call against the jovay diamond — `eth_estimateGas -> 34758`, succeeds — which does support the no-allow-list conclusion. Two caveats now stated in the PR body: the true failing shape (`execTransaction` with three valid signatures) is not constructible, and **my probe hit the public `rpc.jovay.io`, not the Mongo-sourced endpoint the deploy path actually uses.**

### MINOR, also fixed

- The escape-hatch test asserted only that `writeContract` was called, not the gas it received — broadcasting with `gas: undefined` would have passed.
- `.env.example` documented only `=true`; the accepted values and the network-list form are now spelled out.

### Also found, out of scope — filed as [EXSC-890](https://linear.app/lifi-linear/issue/EXSC-890/tron-safetimelock-path-broadcasts-with-no-energy-estimate-at-all)

Enumerating every on-chain send in `script/` surfaced that **Tron never estimates at all**: `tron-executor.ts` / `tron-caller.ts` go straight to `triggersmartcontract` with a hardcoded `fee_limit`, while the devkit's own `estimateContractCallEnergy` sits unused. The repo already documents the harm in `runPendingTimelockTXs.yml:102` — a multi-call batch exceeds the 50-TRX default and *"aborts mid-SSTORE as OUT_OF_ENERGY and is retried forever"*. Predates this PR; **the PR title is now scoped to "EVM"** rather than implying a repo-wide guarantee.

### Confirmed sound, no change needed

- Escape-hatch parsing across 37 values: `false`/`0`/`no`/`2`/`-1`/`01`/unicode look-alikes all fail **closed**. No value opens it accidentally.
- `refuse-to-broadcast.test.ts` fixtures were Proxy-instrumented — the executor touches exactly the three client methods provided, so no test passes for the wrong reason.

`bun test script/` — **1128 pass / 0 fail**.

0xDEnYO:
## `/gate-review` round 2 — applied (per-push rule)

Round 2 reviewed **only the round-1 fix commit**, because a fix-commit is the least-reviewed code in a PR — and round 1's critical finding had itself been introduced by an earlier commit in this same PR.

### The leak is genuinely closed — the highest-risk hypothesis was refuted

I asked whether `redactErrorReason` caps *before* redacting, which would truncate mid-URL and leave a bare `…dkey=KEY` the regex no longer matches. **It redacts first, then caps.** Proven with a message whose URL sits beyond position 180 — the URL was redacted and the *marker* got truncated, not the key:

```text
out : "xxx…(140 x's)… dkey=DRPCKEY_AAAA1111 then [redacted-ur…"
```

End-to-end through the real `getGasWithFallback` → real `safe-utils` rethrow → real `SlackNotifier` with `fetch` intercepted:

```text
Slack payload leaks the key : false
Slack payload leaks the host: false
refusal .message leaks key  : false
```

And the new leak test is a **genuine falsifier** — run against a scratch copy with only the `redactErrorReason` wrapper removed: `0 pass / 2 fail`.

### MEDIUM — my INCONCLUSIVE fix was cosmetic. Now actually fixed.

This is the same bug class round 1 flagged, and my round-1 fix only changed the log. `executeOperation` printed `⚠️ Simulation INCONCLUSIVE` and then still `return 'executed'`. Traced downstream: `operationsSucceeded++` → `operationsFailed === 0` → `success = true` → Slack posts **"✅ Network jovay processing completed successfully"** with `operationsFailed: 0`, and the exit code agrees. Only the terminal warn dissented.

Now returns a distinct `'inconclusive'` status, counted as a failure by the aggregator. The tally, the Slack summary and the exit code all dissent.

### LOW → caught a latent trap by making a field required

`IChainSimulateResult.estimateFailed` was optional, so an implementation could omit it and silently restore the green-on-failure dry run. Made required — which **immediately failed `tsc` on `tron-caller.ts`**:

```text
tron-caller.ts(103,5): error TS2741: Property 'estimateFailed' is missing
```

Not a live bug (every Tron failure path throws today rather than falling back), but exactly the trap the flag exists to prevent, and the compiler now holds the line. Fixed with an explicit `estimateFailed: false` and a comment saying why that is sound.

### HIGH, pre-existing, fixed anyway

`safe-utils.ts` interpolated a raw `errorMsg` into two throws that reach Slack via `extractErrorMessage`, which has no redaction. Verified on `origin/main` — **not introduced by this PR and not newly reachable**; the fix in fact reduces traffic to those lines. But opening the new escape hatch means broadcasting *after* an estimate already failed, which is precisely what produces the viem error that lands there. Both throws now redact.

### Also fixed

- The refusal advice contradicted itself when no network was in scope: it offered `ALLOW_GAS_ESTIMATE_FALLBACK=true` while telling the operator not to use `true`.
- The affirmative-value test table omitted `'y'`, which the guard accepts.

### Confirmed sound — attacks that failed

**Escape-hatch scoping survived every attack.** 20 cases against the real function and the real 73-network config:

```text
closed  env="true,jovay"  net="mainnet"    ← a list containing 'true' does NOT open globally
closed  env=",,,"          net="jovay"
closed  env="base"         net="basesepolia" ← exact match, no prefix bleed
closed  env="*" / "all"    net="jovay"
closed  env="jovay"        net=undefined     ← scoped value, unscoped caller
```

No network in `config/networks.json` is named `true`/`1`/`yes`/`y`/`on`, none contains a comma or whitespace.

**`resolveGas` refactor introduced no drift** — refuse messages byte-identical, multiplier arithmetic identical, throw still precedes the warn.

**Noted and deliberately NOT changed:** `redactUrls`'s regex misses credential shapes that are not `scheme://` URLs (bare `dkey=K`, `Authorization: Bearer`, a URL split across a newline). Pre-existing, and none match viem's actual `HttpRequestError` format, so the real vector is covered. Reordering redact/collapse was tested as a fix and **does not** help — recording that so nobody tries it.

`bun test script/` — **1129 pass / 0 fail**.

gvladika (approved):

gvladika (approved):

PR #2289, opened 2026-08-31
feat(deploy): repair deployment-record build provenance (EXSC-877)

0xDEnYO:

[EXSC-877](https://linear.app/lifi-linear/issue/EXSC-877/wp-71-fleet-reproduction-sweep-replace-when-touched-list) — WP-7.1.

Post-merge operational steps (backup → dry-run → `--apply` against production Mongo) are tracked separately in [EXSC-895](https://linear.app/lifi-linear/issue/EXSC-895/apply-the-deployment-record-repair-to-production-mongo-post-merge), because **merging this PR writes nothing**. EXSC-877 stays open until that runbook has been executed.

A read-only fleet bytecode-reproducibility sweep rebuilt **all 2,094 production address slots across 66 networks** from this repo and compared each against `eth_getCode`. 94.7% reproduce. The failures split into two classes, and **in neither class is the deployed code wrong — the record is**. This PR ships the repair for both, plus the backup that makes it reversible. **It writes nothing on merge**; both commands are dry-run by default.

## 1. `profile` — 14 records whose stored compiler contradicts the bytecode

These 14 store `solcVersion: 0.8.29`, but the deployed bytecode's own CBOR metadata trailer declares **`0.8.17`**. I rebuilt **each record's own commit** at `0.8.17/london` and it reproduces the deployed runtime exactly.

So the recorded commit was never wrong — only the profile. The correction rewrites `solcVersion` and `evmVersion` and **leaves `gitCommitHash` untouched**, which matters: an earlier read of this data suggested the commits were wrong too, and rewriting them would have destroyed correct provenance.

| Contract @ version | Slots | Recorded commit (kept) | `0.8.29/cancun` → |
|---|---|---|---|
| `MayanFacet@2.0.0` | 8 | `2da582742` | `0.8.17/london` |
| `SymbiosisFacet@2.0.0` | 2 (base, cronos) | `46b088402` | `0.8.17/london` |
| `Executor@2.1.0` | 2 (optimism, opbnb) | `c0cc000e8` | `0.8.17/london` |
| `ReceiverStargateV2@1.1.0` | 1 (optimism) | `c0cc000e8` | `0.8.17/london` |
| `ReceiverAcrossV4@1.0.0` | 1 (optimism) | `c0cc000e8` | `0.8.17/london` |

**Deliberately excluded: `tron/SymbiosisFacet@2.0.0`.** Its recorded solc is right, but its recorded commit `3e42b46e1` does **not** reproduce the deployed bytecode at any profile tried, while three other `contracts-tron` commits (`00b0546f3`, `a2a261f7a`, `ca21417ad`) each do. No automated rule can pick the right one, so it is left alone and needs a human call.

## 2. `attest` — 742 records with no build provenance at all

`gitCommitHash` only began being written on **2026-07-10** (#2017), so everything older has none. A bounded candidate search found a commit + profile that rebuilds to the deployed bytecode for 742 EVM slots (141 groups, 18 distinct commits, 60 networks).

This writes a **new `reproducibility` subdocument, never `gitCommitHash`**, because the two claims are not the same strength:

> A commit here reproduces the deployed bytecode; it is not proof it is the commit that was deployed, since commits frequently share a source closure.

The command refuses to touch any record that already has a real commit, so the weaker inferred evidence can never overwrite what the pipeline actually recorded.

**zkEVM is excluded** (6 slots): the sweep does not record which `zksolc` version produced the match, and zksolc pins zk codegen, so a zk attestation without it would not be reproducible.

# Safety

- **Both commands are dry-run by default**; writing needs an explicit `--apply`.
- **Both require `--backup <snapshot>`** and verify every record they are about to touch is present in it. A snapshot of all 4,442 production records has been taken (`sha256 bdf62a301e75e13c7fcb9518d8902a3e5119418b6aa91ad85ae15591ae3ffae7`).
- `profile` carries an `expect` block per record and **skips anything that has drifted** from the value the correction was derived against, so a concurrent change is never clobbered.
- Records are identified by network + contract + version **+ address**, folded for case, and an ambiguous match is skipped rather than guessed.
- `restore` deliberately only reports — replaying a whole snapshot would also undo deploys that landed after it was taken.

Dry-run output against production right now:

```text
attest   would write: 739  (3 skipped: attestation has an empty version, no safe identity)
profile  would write:  14  (0 skipped)
```

- [ ] For new facets: I have checked all points from this list

On tests: these are one-off maintenance commands over a live collection whose correctness rests on the payload data, and every row in that payload was verified by an actual rebuild-and-compare rather than by logic in this PR. Happy to add unit coverage for `lookup`/`key` folding and the drift guard if reviewers want it.

- [x] I have ensured that any new contracts have had AT A MINIMUM 1 preliminary audit conducted

No Solidity changes; nothing deployable in this PR.


0xDEnYO:
## Review-gate residuals — escalated, not auto-fixed

Ran the contracts review gate on the draft. Three findings were auto-fixed and pushed in `6e5cfd970` (snapshot/collection mismatch passing silently; the attestation caveat being denormalised into ~739 documents; `restore` renamed to `inspect` since it only ever reported). The following are **left for a human decision**.

### 1. `tron/SymbiosisFacet@2.0.0` — record defect with no safe automated repair

Its recorded commit `3e42b46e1` does not reproduce the deployed bytecode at `0.8.29/london` or `0.8.17/london`. Three other `contracts-tron` commits each reproduce it exactly at the recorded profile: `00b0546f3`, `a2a261f7a`, `ca21417ad`. They presumably share a source closure, so **which one was actually deployed is not recoverable from bytecode**. Deliberately excluded from `deploymentProfileCorrections.json` — picking one would be inventing provenance. Needs someone with context on that deploy, or it stays as a known-bad record.

### 2. `--apply` bumps `updatedAt` on ~739 records

Correct for an edit, but it is a bulk touch of a field other tooling may treat as "recently deployed". Flagging rather than changing, because suppressing it would be its own lie about the record. Worth a reviewer confirming nothing keys off `updatedAt` recency.

### 3. The `ambiguous` guard cannot fire on today's data

`lookup()` skips when network + address + contract + version matches more than one record. Checked against the live collection: **0 duplicate identity keys across all 4,442 records**, so on current data this branch is unreachable. Kept as defensive — the identity could collide after a redeploy — but it is unexercised, so treat it as unproven rather than as a working guard.

### Lower-confidence, human judgment

- **No unit tests.** These are one-off maintenance commands whose correctness lives in the payload data, and every payload row was verified by an actual rebuild-and-compare rather than by logic in this PR. The parts that *are* logic and worth covering: `key()` case-folding, `lookup()`'s none/ambiguous split, and the `expect` drift guard. Happy to add them.
- **`deploymentProfileCorrections.json` fails `prettier --check` on purpose.** Both payloads are generator output and re-emit byte-identically today (verified). contracts' lint-staged deliberately omits `*.json`, so nothing reformats them — but a future contributor running prettier by hand would break byte-identity with the generator. Per the repo rule, regenerate rather than format.

### What the gate verified rather than assumed

Every guard this PR adds was tested for whether it can actually fire, not just read:

| Guard | Proof |
|---|---|
| snapshot coverage | truncated the snapshot to 50 records → all 14 corrections correctly refused |
| drift (`expect`) | tampered one expected `solcVersion` → that record skipped, other 13 proceeded |
| collection mismatch | `--collection staging` with a production snapshot → hard refusal |
| composite-key staleness | `contractNetworkKey` / `contractVersionKey` are `name-network` / `name-version` — neither embeds `solcVersion`, so no stale key after a profile fix |
| paired artifacts | none of the 14 corrections appear in `deployments/_deployments_log_file.json`, so no flat-log divergence is introduced |

gvladika (approved):

PR #2291, opened 2026-09-01
feat(audit): gate on audited content instead of commit provenance (EXSC-688)

0xDEnYO:

Fixes [EXSC-688](https://linear.app/lifi-linear/issue/EXSC-688/wp-03-s2-audit-gate-asks-content-not-provenance) — WP-0.3 / S2.

The audit gate asked **provenance**; it now asks **content**.

The old check was *"is `auditCommitHash` in `gh pr view --json commits`?"* Git history only grows, so that question, once true, is true forever. Get an audit at commit X → log X → edit the contract at commit Y in the same PR → merges green. That is **F24**, and it is confirmed on production data below, not hypothesised.

The replacement hashes the contract's **full transitive import closure** at PR head and at each logged audit commit, and requires a match. Two consequences fall out for free:

- **The `Revert`-title exemption is deleted, not made "earned".** A clean revert restores byte-identical audited source, so it passes on content with no special case. The exemption was itself the lesser hole (a `Revert`-titled PR could carry arbitrary code).
- **No ancestor-of-`main` condition.** The originally planned version would have blocked 78 of 98 audit entries, because the repo mixes squash- and merge-commits (**F25**). Content does not care about lineage.

**Companion, and not optional: an append-only guard on `audit/auditLog.json`.** The gate compares source against a hash recorded in a file the PR author can edit. Without the guard the check is self-defeating. Adding entries stays free; changing or removing a recorded one blocks.

## Legacy entries — the D3 path is content-based too

**0 of 103** audit entries carry a `sourceClosureHash` today, so *every* real check goes down the legacy path. Per D3 that path is also content-based: the closure is recomputed **at the entry's `auditCommitHash`** and compared to head, rather than falling back to the old commit-list test. This closes F24 for all existing entries instead of forward-only, and removes this package's dependency on EXSC-706's schema field (which becomes an optimisation that saves a fetch). `git fetch origin <sha>` retrieves the audit commits unreachable from local refs; a genuine fetch failure **ERROR-blocks** after bounded retry, never falls through.

Exit codes are deliberately distinct — `1` = FAIL (a mismatch the author can fix), `2` = ERROR (the gate does not know). Both block, per T3, but one code would hide a broken gate behind what reads as an unaudited contract.

## Two fail-open defects found by this PR's own review pass, and fixed

Both were in code written earlier in this package. Both are the same shape: **a security gate reporting success when it had not actually checked anything.** Recording them because the shape is the interesting part, not the two instances.

1. **The append-only guard passed on an unreadable PR base.** It read the log at base and at head and treated a missing read as *"nothing recorded yet, so nothing can have been rewritten"* — exit 0. But a git read returns nothing both when the file is genuinely absent and when the commit could not be read. `audit/auditLog.json` has existed for the whole life of the repo, so absent-at-base on a real base commit means **the read failed**. Now: unresolvable base → ERROR (exit 2); resolved base with the file genuinely absent → pass.

2. **The gate passed when it was never told what to check.** citty drops an unrecognised flag silently, so a renamed or mistyped `--contracts-file` left the contract list empty — and an empty list is a *legitimate* pass ("no contract in this PR needs an audit"). The gate therefore printed `No contracts require an audit` and exited 0. Now a supplied-but-empty list still passes, while no source at all is an ERROR.

Both decisions were extracted as pure functions (`decideAppendOnly`, `resolveContractSource`) so each state is covered by a unit test rather than only reachable through the CLI.

## CI wiring — the "upstream step did not run" analysis

Per our CI-gate invariant, a gate must not depend on an output whose producing step can fail to execute (an unset step output coerces to `0`, so `!= '0'` guards read as "nothing to report").

This gate does **not** read a step output — it is a `run:` step that fails the job directly, so that trap does not apply. The reachable-skip analysis:

- The gate is guarded `!cancelled() && env.AUDIT_REQUIRED == 'true'`.
- `AUDIT_REQUIRED=true` is set **only** when `contracts_for_audit.txt` is non-empty, and the gate reads that same file.
- If the artifact download fails, the `Check if audit is required` step is skipped, `AUDIT_REQUIRED` keeps its job-level default `false`, the gate is skipped — **and the job is red anyway**, because the download step failed.
- Therefore: **green job + skipped gate ⇒ there were no audit-relevant contracts.** Fails closed.

`!cancelled()` rather than bare `always()` so the gate still reports alongside a failure in the step above (the author sees both problems in one run) without running after a cancellation.

The append-only guard runs **unconditionally**, not under `AUDIT_REQUIRED`: rewriting the log is exactly the move that would otherwise arrive in a PR touching no contract at all.

**Unchanged pre-existing residual, stated for the record:** if the upstream version-control job could be made to emit an empty contract list while a contract was modified, both the old and the new gate are bypassed. That surface is not altered by this PR, and the version-control check is its own required status check.

## The D14 split, now wired into the gate that runs

The previous state of this PR shipped `classifyContentVerdict` imported nowhere
but its own test, while the CI-reachable path still did the single combined-hash
comparison. The evidence below described a module CI never called. That is fixed,
and it was load-bearing rather than cosmetic: without the split the gate blocks
any contract sharing a library that moved, which on `main` is most of the audited
set.

The closure reader now returns per-file hashes alongside the combined one — a
combined hash cannot tell "this contract changed" from "a library it imports
changed", and those carry different authority. The combined hash is computed
exactly as before, so recorded `sourceClosureHash` values keep their meaning.

Both directions, run against real repo data on the real audit log:

```text
$ bunx tsx script/deploy/audit/verify-audit-gate.ts \
    --contracts "src/Periphery/ERC20Proxy.sol" --head HEAD
WARN CLOSURE-DRIFT — ERC20Proxy@1.2.0: own source is audit-relevant-identical,
  but 2 file(s) in its import closure changed since the audit
  (src/Errors/GenericErrors.sol, src/Libraries/LibBytes.sol) ...
rc=0

# same contract, one line added to its OWN source, committed:
ERROR FAIL — ERC20Proxy@1.2.0: no audit covers the source at PR head
ERROR Audit gate blocked: the source at PR head is not the source that was audited.
rc=1
```

That is the case this PR names, going from a hard `FAIL`/1 to a reported
`CLOSURE-DRIFT`/0 while own-source drift still blocks.

**Drift gets its own status check**, per D14: `audit-closure-drift`, a separate
job because a status check is per-job — folding it into `audit-verification`
would make drift indistinguishable from the blocking verdict. **Register it as
NON-REQUIRED on `main`.** It re-runs the gate with `--fail-on-drift` rather than
reading a step output, so it cannot report green because an upstream step failed
to set a variable.

**The fallback is deliberate and tested:** a caller that cannot supply per-file
detail keeps the old all-or-nothing behaviour. Absence of evidence is not
evidence the contract is unchanged — and dropping that guard makes the F24 case
itself pass, which is what the mutation run below shows.

Mutation testing on the split, since same-PR tests are not evidence they work:

| Mutation | Killed by |
|---|---|
| own-source comparison ignored | 3 tests, incl. "does not soften a real mismatch" |
| drift allowed to override a hard failure | "does not soften a real mismatch reported by another audit" |
| soft verdict without per-file detail | 3 tests, incl. **the F24 case** and the `Revert`-title case |

A moved submodule counts as drift too — its contents are not repo-owned files in
the closure, so the per-file hashes cannot see it.

## Evidence

Seven A0.5 demos, all built as **real git commits over real repo content** (subject: `ERC20Proxy@1.2.0`, audit `audit20260617`, commit `a2bb57edd`, 9-file closure), plus the falsification half of each. Harness is committed outside this repo at `spikes/wp03-audit-gate-demos/run-demos.sh` so it is re-runnable.

| # | Scenario | Verdict | Exit |
|---|---|---|---|
| 1 | Clean revert to audited source | **PASS** — byte-identical, no title exemption needed | 0 |
| 2 | Partial revert (one closure file left at `main`) | **FAIL** | 1 |
| 3 | Revert-of-a-revert | **FAIL**, then **PASS** once a fresh audit entry records head's closure | 1 → 0 |
| 4 | Same tree, PR titled `Revert "…"` | **FAIL** — title grants nothing | 1 |
| 5 | **F24**: audited at X, edited at Y in the same PR | **PASS** at X, **FAIL** at Y | 1 |
| 6a | Rewriting an existing `auditLog.json` entry | **BLOCKED**, names entry + field + both values | 1 |
| 6b | *Falsification half* — appending a new entry only | **silent**, as required | 0 |
| 7 | Unfetchable `auditCommitHash` | **ERROR**-blocks, distinct from FAIL | 2 |

Demo 7 is **synthetic by necessity** and labelled as such: all 98 SHA-bearing audit commits resolve, so the infrastructure-ERROR path is not reachable from real data. This is the DoD's stated carve-out and does not extend to any check whose failure is reachable from real repo data — demos 1–6 are all real.

Fail-closed demos for the two defects above: typo'd flag → 2, correct flag → 1, empty file → 0; unreadable base → 2, good base → 0.

`bun test script/` — **2078 pass / 0 fail**. `tsc-files --noEmit` clean, `eslint` clean, `prettier --check` clean.

## Audit remediation flow (T6)

The gate cannot tell "I fixed what the auditor asked for" from "I changed the code after the audit." Reports name scope commit **A**; remediations land at **D**.

Agreed pin model (`[CONV:AUDIT-PIN]`):

- `auditCommitHash` = **A** (what was sent / report scope)
- `finalCommitHash` = **D** (post-remediation tree the auditor signed off on)
- Fast path (CI / signer): content-hash at pin = `finalCommitHash` if set, else `auditCommitHash`. No PDF parsing; no "nothing after D" git-log check.
- Slow path (once, at filing / QA on the audit-log PR): confirm report ↔ both hashes with quoted evidence. `Fixed in #` / sidecar automation is a follow-up for `add-audit`, not the signer.

1. Auditor reviews commit A and files findings.
2. Findings are fixed (commits B, C, …) culminating in tree D the auditor signs off on.
3. `/add-audit` writes a log entry with `auditCommitHash=A` and `finalCommitHash=D` (new entry if an A-only row already exists — append-only).
4. QA on that PR checks report ↔ A/D; gate pins content equality to D.

## F24 via pragma / block-comment lines

A line is ignored only when it is wholly comment, wholly pragma, or blank. Trailing code after a terminated pragma or a closed block comment is audit-relevant on **both** the content hash and the version-bump filter. Otherwise `pragma solidity ^0.8.17; function steal() public {}` would match a clean pragma, skip `contracts_for_audit.txt`, and never reach the hash.

---

## Open question carried into review — D13, on [EXSC-883](https://linear.app/lifi-linear/issue/EXSC-883/human-decisions-needed-signing-20)

**Should dependency drift block, or be its own verdict?** Not blocking this PR — it ships as ruled (full closure, hard block, no title exemption).

Running the gate over every audited contract on `main` at `05f70748a` (59 with a resolvable audit at their current `@custom:version`) gives PASS 3 / FAIL 48 / ERROR 8. Splitting by *where* the drift sits separates two very different things:

- **Own source differs — 47 contracts.** Blocks nobody: the gate only fires on contracts a PR modifies, and modifying one forces a version bump plus a fresh audit entry, which is what gets compared. Three sampled: `WithdrawFacet@1.0.0` and `OwnershipFacet@1.0.0` differ by one SPDX line; **`ChainflipFacet@1.0.1` had `event BridgeToNonEVMChain(...)` deleted after its audit commit with no version bump** — F24 sitting on `main` today, invisible to the current check by construction.
- **Own source identical, closure drifted — the real operational cost.** `ERC20Proxy@1.2.0`'s own file is byte-identical to its audit commit, ten weeks old, and still FAILs because 33 files in its closure moved (`LibAsset`, `LibBytes`, `LibAccess`, `LibUtil`, `LibDiamondLoupe`, `OFTComposeMsgCodec`, interfaces). Normally harmless — the audit commit sits inside the PR — but a long-lived PR that merges `main` once goes red with nothing about its contract changed.

**Recommendation: split the verdict.** Own source differs → FAIL, hard block (this is F24 and must stay closed). Own source identical, closure drifted → a distinct reported-not-blocking verdict naming the drifted files. Conflating them buys no security and costs the gate the credibility it needs to survive. **Against:** a facet's behaviour genuinely can change through a library, so drift is a real signal; downgrading it accepts that in exchange for workability. Either reading is defensible, which is why it is a ruling and not a call made here.

The 8 ERROR contracts (`EcoFacet`, `FeeCollector`, `LiFiTimelockController`, `OutputValidator`, `Patcher`, `Permit2Proxy`, `ReceiverAcrossV3`, `ReceiverStargateV2`) all reach the ERROR path rather than silently passing — T3 behaving as designed.

---

## Review-gate round 1 — four fail-open defects found and closed

A three-agent review pass over this PR found four ways the gate could report success or refuse a verdict when it should not. Each was **reproduced against the real repository before being fixed**, and each fix is verified to fire on the bad input and stay silent on the good one. Two of them defeated the package outright.

### 1. Complete gate bypass with additive-only edits — CRITICAL

The append-only guard iterated only over the fields an entry **already had**, so *adding* a field to a recorded entry was zero violations. And the content check ranks a recorded `sourceClosureHash` **above** the audit-commit comparison. Combined: write that one field onto an entry an auditor already approved, and any source passes — with every existing field untouched.

Reproduced against the real 103-entry `audit/auditLog.json` with two purely additive edits:

```text
BASELINE verdict = fail
ATTACK   append-only guard violations: 0
         gate verdict = pass | blocked = false
         reason: FraxFacet@1.1.0: source closure matches the hash recorded on audit 'audit20260820'
```

All 103 entries have no `sourceClosureHash`, so every one was an open slot. `pinnedClosureHash` was the same hole. Worse, a test **asserted this as intended** (`allows adding sourceClosureHash to an entry that had none`) — the bug was pinned as a feature.

**Fixed:** an entry grows only by being *added*, never by gaining a field. After:

```text
ATTACK   append-only guard violations: 1   -> CI blocks
BENIGN   brand-new entry + new version coverage: 0 violations
```

### 2. `--audit-log` was silently ignored — MAJOR

citty resolves a **multi-word** argument's camelCase key to its **default**, so `--audit-log X` parsed into `args['audit-log']` while the code read `args.auditLog` and got the default. Isolated:

```text
HAS default  -> camel: "THE-DEFAULT" | kebab: "PASSED-A"
NO   default -> camel: "PASSED-B"    | kebab: "PASSED-B"
```

So the workflow's `--audit-log "$AUDIT_LOG_PATH"` did nothing, and a hand-run against a fixture log silently reported on the real one. Latent today only because the two paths coincide. Single-word args (`--head`, `--base`) are unaffected, which is why they worked. **Fixed** by removing the `default` and applying the fallback in code; a bogus path now throws ENOENT instead of quietly substituting.

This is the third instance in this PR of the same shape — a silently dropped flag — after the `--contracts-file` case below.

### 3. `EcoFacet` could never be verified — MAJOR

Foundry resolves a non-relative, non-remapped specifier against `libs` (`foundry.toml`: `libs = ["node_modules", "lib"]`), and `src/Facets/EcoFacet.sol:12` imports `lib/openzeppelin-contracts/.../IERC20.sol` that way. The resolver had no direct-path case, so that facet returned `closure-incomplete` → **permanent ERROR, with nothing an author could do about it.** Fail-closed, but a live facet permanently un-mergeable.

**Fixed.** Fleet sweep over all 139 `src/**.sol` at their current `@custom:version`, before and after:

```text
CLOSURE-INCOMPLETE (permanent ERROR, no way to clear):  1  ->  0
verdict tally: { fail: 74, pass: 9, no-audit: 54, error: 2 }   (unchanged)
```

### 4. Silent closure miss — a true fail-open — MAJOR

Comment stripping was two regexes with no understanding of string literals. A `/*` inside one string and a `*/` inside a later one made the comment regex span between them and **drop every import in between, with nothing recorded in `missing`** — so the hash was taken over an incomplete closure and drift in the dropped file was invisible. That is the one result this module must never produce. Reproduced:

```text
imports found : []
missing       : []
>>> SILENT MISS CONFIRMED: ./A.sol dropped from the closure with no diagnostic
```

The test that claimed this ground (`does not treat a .sol string literal in code as an import`) used a fixture containing no `import` keyword, so it could not fail against any implementation.

**Fixed** by replacing both regexes with a single scan that locates comments and string literals together. The fleet sweep's verdicts and hashes are **identical** before and after, so the rewrite is behaviour-preserving on real data.

### Also fixed

- **The append-only guard passed on an unreadable PR base**, treating a failed git read as "nothing recorded yet". A git read returns nothing both when the file is absent and when the commit cannot be read; this log has existed for the life of the repo, so absent-at-base means the read failed. Now ERROR (exit 2).
- **The gate passed when never told what to check.** An unrecognised flag is dropped silently and an empty contract list is a legitimate pass, so a renamed `--contracts-file` produced `No contracts require an audit` and exit 0. A supplied-but-empty list still passes; no source at all is now an ERROR.
- **`auditCommitHash` whitespace** made this module and the decision table disagree about whether an entry was commit-backed, surfacing as "the closure was not resolved" — an error naming neither cause nor fix. Trimmed once, at the point the entry is read.
- **An uppercase recorded hash** was accepted and then failed the lowercase `===`, reporting a content mismatch that was really a formatting slip. Normalised.
- **`git-source-reader.ts` had no tests** despite being the git seam a security-blocking gate reads history through. Covered against the real repository, because the properties under test — a squash-merged commit being readable only after a fetch, and an absent file being distinguishable from an unreadable commit — are properties of git, and a fake would assert the assumption instead of testing it.
- A comment-discipline fix (history narration in a module header).

`decideAppendOnly`, `resolveContractSource` and `classifyDuplicateKeyError`-style decisions were all extracted as **pure functions**, so every branch is unit-tested rather than only reachable through a CLI. `bun test script/` — **1234 pass / 0 fail** (baseline 1194).

## Why this PR is still a DRAFT

Independently of the above, the PR's **central safety argument is falsified on real data**, which is escalated as **D14** on [EXSC-883](https://linear.app/lifi-linear/issue/EXSC-883/human-decisions-needed-signing-20) and blocks [EXSC-688](https://linear.app/lifi-linear/issue/EXSC-688/wp-03-s2-audit-gate-asks-content-not-provenance).

The S2 ruling deletes the `Revert`-title exemption on the grounds that *"a clean revert restores byte-identical audited source and passes on content with no special case."* Run against **the actual June-2025 revert** whose blocking caused this check to be disabled in the first place (#1145 → #1186 blocked → #1187 disabled → #1893 re-enabled with the title carve-out):

```text
FAIL — LibAllowList@1.0.0: closure at audited commit eb12d93c... is 0x7d785080...,
       PR head is 0x46c3dc87... — the contract changed after it was audited

$ git diff --stat eb12d93c17cf f1a1b2220 -- src/Libraries/LibAllowList.sol
(empty — byte-identical)
$ git diff --stat eb12d93c17cf f1a1b2220 -- src/Errors/GenericErrors.sol
 src/Errors/GenericErrors.sol | 4 +++-
```

`LibAllowList.sol` **is** byte-identical to its audited source. The FAIL is entirely `GenericErrors.sol` drifting inside its import closure. So the gate as ruled would have blocked that emergency revert too — the friction #1187 existed to remove, relabelled from provenance to drift.

And it is not one case. On current `main`, **six** contracts are byte-identical to their audit and still FAIL:

```text
src/Facets/UnitFacet.sol@1.0.1              src/Periphery/GasZipPeriphery.sol@1.0.2
src/Facets/WhitelistManagerFacet.sol@1.1.0  src/Periphery/Patcher.sol@1.0.1
src/Periphery/ERC20Proxy.sol@1.2.0          src/Periphery/ReceiverAcrossV4.sol@1.0.0
```

The mechanism is a genuine semantic mismatch, not strictness: **the closure hashes raw file bytes, while this same workflow declares comment / pragma / license changes non-audit-relevant** and therefore requires no version bump and no fresh audit for them (`versionControlAndAuditCheck.yml:154-157`). One real commit — `c37da94a2`, a repo-wide SPDX header normalisation — invalidated content-equality for every contract against every prior audit. `OwnershipFacet@1.0.0`'s entire diff since its audit commit is one line:

```text
-// SPDX-License-Identifier: MIT
```

`src/Errors/` and `src/Interfaces/` are also exempt from audit yet sit inside almost every closure, so they drift freely.

So D14 asks for a ruling on two things: whether the drift-only verdict is split from the own-source verdict (my recommendation, which makes the LibAllowList revert pass while keeping F24 closed), and whether the closure hash should normalise exactly what the workflow already declares non-audit-relevant instead of hashing raw bytes. Both are now shipped: drift is non-blocking, own-source mismatch blocks, and the hash uses the shared ignore rule (whole-line comment/pragma/blank only).


script/deploy/audit/source-closure.test.ts
```diff
@@ -0,0 +1,313 @@
...
+    )
+  })
+
+  it('binds content to path — swapping two files contents changes the hash', () => {
+    const swapped = {
+      'src/Foo.sol': 'import { A } from "./A.sol";',
+      'src/A.sol': 'import { A } from "./A.sol";',
+    }
+    expect(hashOf(base)).not.toBe(hashOf(swapped))
+  })
```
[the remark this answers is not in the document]
0xDEnYO: Correct, and it is the failure mode this package has been bitten by repeatedly — a test that cannot fail for the reason it claims.

Fixed in `e13ad3b40`. The fixture now exchanges the contents of two *leaves* so the path set and the content multiset are identical in both versions, and it **asserts** that equality rather than claiming it:

```ts
expect(Object.values(before).sort()).toEqual(Object.values(exchanged).sort())
expect(hashOf(before)).not.toBe(hashOf(exchanged))
```

A hash over concatenated contents alone is now identical across the two inputs, so the test fails unless path is part of the preimage.

script/deploy/audit/audit-gate.ts
```diff
@@ -0,0 +1,255 @@
...
+}
+
+/**
+ * The audit log keys contracts by bare name, the workflow carries paths.
+ *
+ * @param path - repo-relative path of a contract.
+ * @returns the contract name the audit log uses.
+ */
+export const contractNameFromPath = (path: string): string =>
+  (path.split('/').pop() ?? path).replace(/\.sol$/, '')
```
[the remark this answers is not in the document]
0xDEnYO: Verified on real data, and the conclusion is: not exploitable today, and not something this PR can fix in isolation.

```text
$ git ls-tree -r --name-only HEAD -- src/ | grep "\.sol$" | xargs -n1 basename | sort | uniq -d
(blank — zero duplicate basenames across all 139 src contracts)

$ audit/auditLog.json -> auditedContracts keys: 109, keys containing a slash: 0
```

The bare-name lookup is not something the gate chose — **`audit/auditLog.json` itself is keyed by bare contract name**, all 109 of them, with no path component anywhere. So `contractNameFromPath` is matching the schema it has to read; keying by path would find no entry at all.

Worth stating precisely what the residual is, since it is narrower than it looks: the content comparison uses the **real path**, not the name. So a future basename collision could only produce a *false PASS* if the two same-named contracts had **byte-identical transitive import closures** — at which point they are the same source. The realistic failure is the opposite direction, a false FAIL when the wrong entry is selected, which fails closed.

Fixing it properly means adding a path (or a path-qualified key) to the audit-log schema, which is InfoSec-owned and belongs with EXSC-706's schema work rather than here — and until then the append-only guard shipping in this PR means the key set cannot be quietly rewritten either. Leaving the code as-is deliberately; flagging it on the ticket so the schema change carries it.

0xDEnYO:
## Review gate — escalated, not fixed

Gate run on `df4c17f01` (the two D14 commits plus the main merge). No auto-fixes applied; the one substantive finding is a governance call, not a typo.

**E1 — the D14 split is not wired into the gate that actually runs.**

`classifyContentVerdict` in `script/deploy/audit/closure-drift.ts` is imported nowhere except its own `closure-drift.test.ts`. The CI-reachable path — `verify-audit-gate.ts` → `verifyAuditContent` in `audit-gate.ts` — still does the single combined-closure-hash comparison, unchanged by `2bbed6371` and `009fa9582`.

So the evidence table in the PR body, which says the six byte-identical contracts now come back `closure-drift` / `blocks=false`, describes a module the shipping gate never calls. Against real repo data:

```text
$ bunx tsx script/deploy/audit/verify-audit-gate.ts --contracts "src/Periphery/ERC20Proxy.sol" --head HEAD
FAIL — ERC20Proxy@1.2.0: no audit covers the source at PR head
Audit gate blocked: the source at PR head is not the source that was audited.
rc=1
```

Still a hard `fail` / exit 1. There is no integration test tying `closure-drift.ts` to `audit-gate.test.ts` or `verify-audit-content.test.ts`, so `bun test` cannot see the disconnect — the suite is 2069 pass / 0 fail with the module orphaned.

This does not contradict the PR's own "Why this PR is still a DRAFT" section; it confirms, on the post-fix commits, that the stated blocker is still fully present. Wiring it in is the D14 ruling deferred to EXSC-883, so the gate escalates rather than deciding it.

**What the gate verified as solid** (real-data, not same-PR tests):

| Check | Evidence |
| --- | --- |
| Append-only guard | Round-1 additive-field bypass re-run against the real 103-entry `auditLog.json` → caught (`field-added`). A novel same-count audit-id-swap → caught (`coverage-removed`). `baseResolved` ERROR-blocks ahead of the "nothing recorded" fallback |
| Draft-skip bypass | `audit-verification` is gated `draft==false`, but `on.pull_request.types` includes `ready_for_review` and `synchronize`, so the ready-flip and any push re-run it against final HEAD — no tamper window |
| Closure completeness | Swept all 139 real `src/**.sol`: 0 `closure-incomplete`. EcoFacet resolves cleanly (13 files, 0 missing) |
| SPDX/comment normalization | Closure hash for `OwnershipFacet.sol` immediately before/after the real `c37da94a2` normalization commit — hashes match, so this half genuinely works in the shipped path |
| Removed provenance check | Grepped repo-wide including `docs/`: commit-list membership and the `Revert`-title exemption are fully gone, no stale references |

**Also worth knowing before this flips to ready:** `CodeRabbit` reports `pass` but the detail is *"Review rate limited"*, and its last real review was 2026-09-01T10:14Z — before `2bbed6371`, `009fa9582` and `df4c17f01`. Those three commits have had zero CodeRabbit review, and `audit-verification` shows `skipping` while the PR is a draft, so the local runs above are the only verification this commit state has had.

Escalated items need a decision before this PR is ready.

0xDEnYO:
**Ready.** The escalation from the previous gate round (E1 — the D14 split not wired into the CI-reachable path) is closed in `e2e5fd6a4`, and the drift status check in `7fe300a38`.

Two things a reviewer needs to act on, neither of which I can do:

1. **`audit-closure-drift` must be registered NON-REQUIRED on `main` branch protection.** If it is left unregistered the drift is invisible; if it is registered as *required* it blocks merges D14 explicitly says should not be blocked. This is the one manual step this PR depends on.
2. **`protect-critical-code` will stay red** until an InfoSec Manager or the CTO approves — this PR edits `.github/workflows/`, which is the point of that gate.

One design decision I made rather than escalating, flagged so it can be overruled cheaply: when a contract has several audits and one finds own-source drift while another finds only closure drift, the hard failure wins. A single audit proving the contract itself changed is not softened by another proving only its libraries moved.

0xDEnYO:
**Caveat on this PR's own CI, so the green is not read as more than it is.**

`audit-closure-drift` passes here, but vacuously — this PR changes no Solidity, so the contract list is empty and the job takes the nothing-to-verify path:

```text
[success] No contracts require an audit — nothing to verify.
```

`audit-verification` is green for the same reason. So neither job has exercised a real comparison in CI on this PR; the evidence that the split works is the local run against real repo data in the description (`CLOSURE-DRIFT`/rc=0 for library drift, `FAIL`/rc=1 once ERC20Proxy's own source changes) plus the mutation table.

The first PR that actually modifies an audited contract is where these two jobs get their first real CI exercise. Worth knowing when deciding how much this PR's green is worth.

`protect-critical-code` has not run yet either — it triggers on review submission, and this PR edits `.github/workflows/`, so expect it to go red and need the InfoSec approval.

0xDEnYO:
**Correction to my earlier comment: item 1 was wrong. `audit-closure-drift` needs no registration and no action.**

Required checks on `main` are an explicit allow-list in the `main protection` ruleset, and a new job that is not in it is non-required by default:

```text
validate-json, audit-verification, solc-floor-build-required, protect-critical-code,
validate-scripts-required, unit-tests-required, version-control, protect_audit_labels,
check-new-network-health, smoke-test-required, ts-tests-required
```

`audit-closure-drift` is absent and nothing in this PR adds it, so it reports without blocking the moment this merges — which is exactly the D14 behaviour. There is no "register it as non-required" step; that is the default state.

It also would not be a click if we ever did want it required: the ruleset is defined in `lifinance/lifi-infrastructure` (`modules/github-repositories/rulesets.tf`, per-repo `inputs/github/lifinance/repositories/repositories.yaml`), so making a check required is a PR there.

The other item stands: `protect-critical-code` is in the required list and this PR edits `.github/workflows/`, so it needs an InfoSec Manager or CTO approval before merge.

script/deploy/audit/verify-audit-content.ts
```diff
@@ -0,0 +1,308 @@
...
+  const drifts: {
+    reason: string
+    auditId: string
+    driftingDependencies: string[]
+  }[] = []
+
+  for (const entry of entries) {
+    const { kind } = classifyAuditEntry(entry)
+
+    if (kind === 'recorded') {
```
gvladika, line 181: **The D14 split does not apply on the `recorded` path.**

An entry with a `sourceClosureHash` classifies as `recorded`, and a mismatch pushes to `failures` and `continue`s at line 190. It never reaches `compareClosures`. The own-source-vs-drift distinction only exists on the `commit` path.

Same inputs, same drift (own file identical, one library moved):

```text
A  commit-only entry    -> closure-drift   (does not block)
B  + sourceClosureHash  -> fail            (blocks)
```

Invisible today because 0 of 103 entries carry the field. The moment EXSC-706 starts writing it, every contract sharing a drifted library hard-blocks again — and the regression will look like the gate breaking for no reason, since nothing in the log or the message points at the new field as the cause.

Two related notes:

- `audit-gate.ts:247` resolves `closureAtAuditCommit` for these entries — a network fetch plus a full closure walk — and the result is never read.
- `computeClosureDetail` stamps `version: 2` into the envelope, so any hash recorded against today's `main` implementation is permanently mismatched. Worth stating on EXSC-706 explicitly.

Suggested fix: run `compareClosures` on the `recorded` path too, using `closureAtAuditCommit` when the entry has a resolvable commit. A recorded hash should be a fast path, not a different policy.

.github/workflows/versionControlAndAuditCheck.yml
```diff
@@ -344,6 +346,22 @@ jobs:
...
+        uses: oven-sh/setup-bun@0c5077e51419868618aeaa5fe8019c62421857d6 # v2
+
+      - name: Install dev dependencies
+        run: bun install
+
+      # Runs unconditionally, not under AUDIT_REQUIRED: rewriting the log is precisely
+      # the move that would otherwise arrive in a PR touching no contract at all.
+      - name: Verify the audit log is append-only
+        env:
+          BASE_SHA: ${{ github.event.pull_request.base.sha }} # via env, never inline, matching the PR_TITLE handling below
```
gvladika, line 359: **The append-only baseline should be `HEAD^1`, not `base.sha`.**

Line 686 of this file states the model correctly: on a `pull_request` event HEAD is the merge commit. So `HEAD^1` *is* the base tip that HEAD merged. `base.sha` is not guaranteed to be that commit.

The gap: entry `E` lands on `main` after the PR's `base.sha`. The author merges `main` into the branch, deletes `E`, pushes. Merge-base is now `main`, so the merge commit carries the deletion. `before` (at `base.sha`) never had `E`, `after` doesn't have `E` — **zero violations**, and an entry leaves recorded history with the guard reporting clean.

```diff
-            --base "$BASE_SHA" \
+            --base HEAD^1 \
```

That also removes the `BASE_SHA` env plumbing, and is consistent with the model the file already documents.

script/deploy/audit/verify-audit-content.ts
```diff
@@ -0,0 +1,308 @@
...
+      verdict: 'closure-drift',
+      reason: best.reason,
+      matchedAuditId: best.auditId,
+      driftingDependencies: best.driftingDependencies,
+    }
+  }
+
+  return {
+    verdict: 'fail',
+    reason: `${subject}: no audit covers the source at PR head:\n  ${failures.join(
```
gvladika, line 304: **What happens to fixes made in response to audit findings?**

This is the case every SC engineer will hit, and I can't find it addressed anywhere in the PR.

Per `503-sc-feature-pr-lifecycle.md`, stage 8 is "audit, then resolve the findings" and stage 9 is "file the audit". `/add-audit` extracts `auditCommitHash` **from the PDF** (`add-audit.md:15`), and the PDF names the commit the auditor reviewed — before remediation.

So the normal flow is: auditor reviews X → you fix findings at Y → the log records X → the gate compares X against head and hard-FAILs, because the contract's own source changed. Which is correct: the gate genuinely cannot distinguish "I fixed what the auditor asked for" from "I changed the code after the audit". But it does mean the recorded hash must now be the *post-remediation* commit, and the append-only guard means it cannot be corrected after the fact.

Two asks:

1. Confirm the intended flow — does the auditor need to sign off on the final commit, and does the report have to name it? Then put it in the PR description and the audit docs.
2. Make this message say it. Right now it reports that the source does not match and stops. Something like *"if this follows audit remediation, record the post-remediation commit in a new audit entry — the report may be reused"* turns a dead end into an instruction.

Without this, the first person to hit it after an audit will reasonably conclude CI is broken.

script/deploy/audit/verify-audit-gate.ts
```diff
@@ -0,0 +1,203 @@
...
+
+  return paths.map((path) => {
+    const source = reader.readFile(path)
+    const version = source ? extractContractVersion(source) : undefined
+
+    // Refused here rather than allowed through: with no version the gate cannot
+    // look up coverage, and the resulting "no audit found" would read as an
+    // unaudited contract instead of a malformed one.
+    if (!version)
+      throw new Error(
```
gvladika, line 48: **This throw exits 1, not 2, breaking the exit-code contract in the module header.**

```text
$ bunx tsx script/deploy/audit/verify-audit-gate.ts --contracts-file cl.txt --head HEAD
ERROR  could not read a @custom:version from src/Facets/DoesNotExist.sol at HEAD
EXIT=1
```

`runMain` catches the throw and exits 1. This is a gate-doesn't-know condition arriving as FAIL — the exact collapse the header argues against. Same shape at `verify-audit-log-append-only.ts:42`, whose own comment insists malformed must be distinguished from absent.

Both should use the `consola.error` + `process.exit(EXIT_ERROR)` shape the other input failures already use.

script/deploy/audit/verify-audit-gate.ts
```diff
@@ -0,0 +1,203 @@
...
+    }
+
+    // Reported, not blocked. Every contract's own source matched its audit; what
+    // moved is code they import, which is a different claim and carries
+    // different authority. The drift check reports this on its own status.
+    if (report.verdict === 'closure-drift') {
+      const drifted = report.results.filter(
+        (result) => result.verdict === 'closure-drift'
+      )
+      if (args.failOnDrift) {
```
gvladika, line 182: **A red `audit-closure-drift` does not mean drift.**

`--fail-on-drift` only affects the `closure-drift` branch. `fail` and `error` still exit 1/2 at line 199, so whenever `audit-verification` is red the drift job is red too — for a different reason, on a check whose whole purpose is to isolate drift as a distinct signal. It also repeats every git fetch and closure walk.

In `--fail-on-drift` mode, suggest exiting 0 on `fail`/`error` (the required check already owns those) and non-zero only on drift.

.github/workflows/versionControlAndAuditCheck.yml
```diff
@@ -344,6 +346,22 @@ jobs:
           fetch-depth: 0
           persist-credentials: false
 
+      - name: Set up Bun
+        uses: oven-sh/setup-bun@0c5077e51419868618aeaa5fe8019c62421857d6 # v2
+
+      - name: Install dev dependencies
+        run: bun install
```
gvladika, line 353: `bun install` → `bun install --frozen-lockfile`, here and at line 737. Repo convention in 6 of 10 workflows, and the right default for a security gate: one that can resolve a different dependency tree than the lockfile pins is a weaker gate.

script/deploy/audit/source-closure.ts
```diff
@@ -0,0 +1,381 @@
...
+
+  if (path === undefined) return undefined
+
+  const submoduleDir = SUBMODULE_DIR_RE.exec(path)?.[1]
+  if (submoduleDir) return { path, external: true, submoduleDir }
+
+  // Not readable from git at any tree-ish, so it cannot be hashed or pointed at.
+  // Reported as external so the closure walk records it rather than silently
+  // dropping it; the caller still sees it as a dependency it could not pin.
+  if (path.startsWith('node_modules/'))
```
gvladika, line 244: **This is the EcoFacet trap, one remapping away from being live.**

A `node_modules/` path becomes `{external: true, submoduleDir: 'node_modules'}`, then `readSubmodulePointer('node_modules')` runs `git ls-tree` on a gitignored directory, gets `undefined`, and lands in `missing` → `closure-incomplete` → permanent ERROR with nothing the author can do. Exactly the failure you fixed for EcoFacet.

`remappings.txt:1` has `@uniswap/=node_modules/@uniswap/`. No `src/` contract imports it today, so this is latent, not live.

Either fail at resolve time with a message naming the cause, or pin these by their `package.json` version. Either way a test on the `@uniswap/` case would keep it from regressing.

gvladika (changes requested):
Reviewed the gate end to end and verified the risky claims against the real repo rather than reading them.

**Verified and dismissed** — no action needed:
- Anonymous `git fetch origin <sha>` retrieves orphaned audit commits. Tested `d62200244…` into a credential-free clone; `persist-credentials: false` is not a problem.
- Zero duplicate contract basenames in `src/`, so the name-keyed audit lookup is safe today.
- `bun test script/deploy/audit/` — 176 pass / 0 fail.

**Three things I think should block merge:**

1. The D14 drift split does not apply on the `recorded` path, so it disappears the moment EXSC-706 populates `sourceClosureHash`. Reproduced.
2. The append-only guard baselines on `base.sha` rather than the merge parent, which leaves a way to delete a recorded entry silently.
3. Nothing in the PR or the failure message covers what happens to fixes made in response to audit findings — the case every SC engineer will hit.

Four smaller items inline. The architecture is right and the own-source / drift separation is the correct call; these are about the edges.

script/deploy/audit/verify-audit-content.ts
0xDEnYO: Agreed — the gate cannot tell “I fixed what the auditor asked for” from “I changed the code after the audit,” so the recorded hash has to be the post-remediation commit.

Intended flow, now in the PR body, `/add-audit`, `501-audits`, and lifecycle stage 9:

1. Auditor reviews X and files findings.
2. Findings are fixed at Y.
3. Auditor signs off on Y (same report, addendum, or confirmation). The report may be reused.
4. `/add-audit` writes a **new** log entry whose `auditCommitHash` is Y. The append-only guard rejects in-place edits of the X row.

The FAIL path already named that way out (`REMEDIATION_HINT`). This commit also closes the related F24 hole: a line is ignored only when it is wholly comment / wholly pragma / blank, so `pragma …; function steal()` no longer matches a clean pragma and skip both gates.

.github/workflows/versionControlAndAuditCheck.yml
```diff
@@ -154,12 +156,37 @@ jobs:
...
+                  REST="${DIFF_LINE:1}"
+                  REST="${REST#"${REST%%[![:space:]]*}"}"
+                  [[ -z "$REST" ]] && continue
+                  [[ "$REST" == //* ]] && continue
+                  if [[ "$REST" == '/*'* ]]; then
+                    if [[ "$REST" != *'*/'* ]]; then
+                      continue
+                    fi
+                    AFTER_BLOCK="${REST#*'*/'}"
+                    [[ -z "${AFTER_BLOCK#"${AFTER_BLOCK%%[![:space:]]*}"}" ]] && continue
```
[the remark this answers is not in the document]
0xDEnYO: Accepted as known limitation on this PR — not an F24 bypass (trailing code on pragma/comment openers is already handled). LI.FI-owned facets use `///` almost exclusively; multi-line `/*` bodies are not house style. Follow-up: forbid `/*`/`/**` under owned globs via lint/CI rather than teaching the audit hasher a block-comment lexer. Tracking ticket incoming.

script/deploy/audit/audit-relevant-source.ts
```diff
@@ -0,0 +1,48 @@
...
+/**
+ * Drops the lines the version-bump check would ignore.
+ *
+ * @param source - Solidity source, as read from disk or git.
+ * @returns The remaining lines, newline-joined, with no trailing newline.
+ */
+export const normaliseAuditRelevantSource = (source: string): string =>
+  source
+    .split('\n')
+    .filter((line) => !isAuditNonRelevantLine(line))
```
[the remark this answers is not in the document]
0xDEnYO: Accepted as known limitation on this PR — not an F24 bypass (trailing code on pragma/comment openers is already handled). LI.FI-owned facets use `///` almost exclusively; multi-line `/*` bodies are not house style. Follow-up: forbid `/*`/`/**` under owned globs via lint/CI rather than teaching the audit hasher a block-comment lexer. Tracking ticket incoming.

0xDEnYO:
Addressed the QA Low from `f871bd63`:

2. Removed the leftover auditor-handle `gh api` block after the per-audit loop. The inverted check inside the loop already fail-closes; the outer copy only re-validated the last `$AUDITOR_GIT_HANDLE`.

1. The High is the existing `protect-critical-code` InfoSec-Manager / CTO human gate on `.github/` — not a code fix. Still waiting on that approval.

Fix: `5d5ea0721`

script/deploy/audit/verify-audit-content.ts
```diff
@@ -0,0 +1,348 @@
...
+        const classified = classifyContentVerdict(subject, recordedComparison)
+        if (classified.verdict === 'closure-drift') {
+          drifts.push({
+            reason: `${classified.reason} (audit '${entry.auditId}', recorded hash ${entry.sourceClosureHash})`,
+            auditId: entry.auditId,
+            driftingDependencies: recordedComparison.driftingDependencies,
+          })
+          continue
+        }
+      }
```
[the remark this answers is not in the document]
0xDEnYO: Fixed: when a recorded `sourceClosureHash` disagrees with head but the recomputed per-file closures match (e.g. hasher envelope bump), we now return pass — same as the commit-only path. Regression test added.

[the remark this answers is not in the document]
0xDEnYO: Follow-up: root cause was `closureMatches` comparing `combined` hashes. It now uses per-file (+ submodule) equality, and both the recorded and commit paths return `pass` when that classifier says so.

0xDEnYO:
P5 is fixed in `abe88cbc2`: `compareClosures` now derives `closureMatches` from per-file (+ submodule) equality, and the recorded path returns `pass` when that recompute matches even if `sourceClosureHash` is a stale hasher envelope. Regression test covers the envelope-mismatch case.

P1 remains the InfoSec-Manager / CTO `protect-critical-code` gate.

gvladika (approved):

melianessa (approved):

PR #1904, opened 2026-06-10
feat: Slack notification when proposals are created (count + sign reminder)

0xDEnYO:
# Which Linear task belongs to this PR?

[EXSC-433](https://linear.app/lifi-linear/issue/EXSC-433/slack-notification-when-proposals-are-created-count-sign-reminder) — Slack notification when proposals are created (count + sign reminder)

# Why did I implement it this way?

All proposal-creation entry points in `multiNetworkExecution.sh` (`iterateAllNetworksGrouped`, `executeAllNetworksForContract`, `executeNetworksByEvmVersion`) funnel through `executeNetworksByGroup`, so the new `notifyProposalsCreatedToSlack` is called exactly once there — right after the final execution summary and before `cleanupProgressTracking` removes the progress file. This guarantees **one message per run** (no duplicates) with the count of successfully created proposals and the target chains, plus a sign reminder.

The function:

- no-ops unless the run's `actionType` is `proposal` (read from the progress tracking file), so deployment/verification runs are unaffected
- no-ops if zero proposals were created
- reuses the existing Slack notifier path: `sendMessageToSlackSmartContractsChannel` from `script/helperFunctions.sh`, which posts via `SLACK_WEBHOOK_SC_GENERAL` from `.env` (no new credentials, no hardcoded webhook URLs)
- never fails the run if the Slack post fails

The legacy `iterateAllNetworksOriginal` path intentionally does not get the notification — no current entry point uses it for proposals.

## Verification

- `bash -n script/multiNetworkExecution.sh` — passes
- `bash -c 'set -euo pipefail; source script/multiNetworkExecution.sh'` — sources clean in strict mode
- Dry-ran `notifyProposalsCreatedToSlack` in isolation with a mocked Slack sender against 6 cases: proposal run with successes (1 message, correct count/chains/reminder), non-proposal run (no message), proposal run with zero successes (no message), missing progress file (no-op), invalid JSON (no-op), unset `PROGRESS_TRACKING_FILE` under `set -u` (no crash). All passed.

## Note on /pr-ready

`/pr-ready` (local CodeRabbit) could not be completed: the CLI returned "Review limit reached" (org usage quota exhausted), so this PR is opened as a **draft**. A self-review pass (mechanical greps + semantic read + executable dry-run) was performed instead; cloud CodeRabbit will still review on CI. Please re-run `/pr-ready` once quota is available before flipping to Ready for Review.

# Checklist before requesting a review

- [x] I have performed a self-review of my code
[...]

PR #2299, opened 2026-09-01
feat(safe): render the signing ask from the proposal document

0xDEnYO:

Ref [EXSC-696](https://linear.app/lifi-linear/issue/EXSC-696/s10-slack-card-rendered-from-the-proposal-document) — WP-1.3 (S10).

Ref EXSC-696 — deliberately `Ref` and not `Fixes`: the ticket also asks for **thread updates on status transitions**, which cannot be built with an incoming webhook — see "What is not in this PR".

The signing ask was assembled from the runner's own progress file:

```text
2x proposals created: AcrossFacetV4 across 2 networks — please sign and schedule 🙏
```

That file knows a contract name and a network count. It does not know why the change is being made, who made it, from which commit, or how to review it — so the card asked people to sign and told them nothing to sign on.

The card is now rendered from the proposal documents:

```text
1x proposal created on mainnet — please sign and schedule 🙏

*Reason:* add AcrossFacetV4 to the whitelist
*Proposed by:* Alice Example &lt;alice@example.com&gt;
*Commit:* 1234567890ab
*PR:* https://github.com/lifinance/contracts/pull/2125

*To review:*
• `bun confirm-safe-tx --network mainnet`  (0xabababab)
```

and, on a wider run with less to go on:

```text
12x proposals created across 12 networks — please sign and schedule 🙏

*Reason:* _none given_
*Proposed by:* ci-bot (bot)
*Working tree:* dirty — src/Facets/Foo.sol
*Proposer-side check (advisory):* codehash matched on 12/12 networks

*To review:*
• `bun confirm-safe-tx --network chain0`  (0x00000000)
• …and 11 more — same command per network
```

## Decisions worth a reviewer's attention

**Every proposer-controlled string is escaped *and* flattened onto one line.** Two different risks, and only the first is obvious: `<!channel>` in a reason would notify a whole channel from a field nobody reviews, and a **newline lets a value forge the card's own structure** — a reason containing `\n*Reason:* something else` renders as a second, entirely plausible Reason line. These values are sanitized when stored, but a card that trusts the document is one hand-edited row away from lying to every signer who reads it.

**An absent reason is stated, not omitted.** `_none given_` tells a signer the intent was never captured. A card that simply skips the line is indistinguishable from one that failed to render it.

**A proposer-side check is labelled advisory**, per the ticket. Authoritative status lives in the signer-side attestation, and a card stating a result without that word invites signing on the card's word.

**The expected-hash table is a parameter, not a placeholder line.** The ticket asks the design to reserve the slot for WP-7.2's DM service. Reserving it as an optional argument keeps the slot without shipping a "not yet available" line — filler sections train operators to skip sections.

**Truncation at 8,000 characters, and the middle is what goes.** Not Slack's 3,000 — that is the per-`blocks[].text` limit, and this card posts as a top-level `text`, whose limit is 40,000. 8,000 is a readability bound, not a platform one. Slack drops overflow silently and the review commands sit at the *bottom*, so the cut is taken out of the body: the headline and the commands both survive, and the card says it was cut.

**The advisory-check line and the hash table have no production caller yet.** `render-proposal-card.ts` supplies neither `checkSummary` nor `expectedHashes`, so the `*Proposer-side check (advisory):*` line, the `*Expected code hashes:*` block, and the `Proposer-side check` entry in the divergence warning cannot render on a real run today — they are the reserved WP-7.2 slot, exercised by tests only. The second example card above shows the shape, not live output.

**Rendering is split from posting.** `send-slack-webhook-message.ts` stays the only thing that talks to Slack; `render-proposal-card.ts` writes text and `proposal-card.ts` is pure. The bash caller **falls back to the old summary line** if the render fails for any reason — a worse message is acceptable, a missing signing ask is not, and `notifyProposalsCreatedToSlack`'s contract is that it never fails the run.

# Evidence

`bun test script/` — **1606 pass / 0 fail** (1538 baseline; 68 new), identical with and without ambient env config. `tsc-files --noEmit`, `eslint`, `bash -n` and `markdownlint-cli2` all clean.

**14 mutations, all die.** Two initially **survived**, and both are worth naming because neither was a weak guard:

| Mutation | Result |
|---|---|
| stop escaping Slack markup | 2 fail |
| escape `<` but not `&` | 1 fail |
| omit the reason line when absent | 1 fail |
| drop the "advisory" label | 1 fail |
| render the full hash instead of a short one | 1 fail |
| render the hash header with no rows | 1 fail |
| emit one command per network, unbounded | 1 fail |
| render an empty card instead of refusing | 1 fail |
| hide the dirty working tree | 1 fail |
| drop the bot marker | 1 fail |
| stop flattening newlines | 1 fail |
| drop the truncation | 1 fail |

- One survived on a **bad assertion of mine** — I checked for `'expected hash'` where the card says `'expected code hashes'`.
- One survived because **two guards caught the same input**: an explicit `length === 0` check plus an `if (!first)` that existed only for the type checker. I removed the redundant one rather than adding a fixture, so the remaining guard is the one that is actually load-bearing.

The newline-flattening and truncation guards came from an **enumeration pass, not from mutation** — listing what a document value can lie about (markup, structure, length) rather than mutating the guards I had already written. Mutation testing cannot surface a check that was never there, which is how a missing URL-scheme check got through review round 1 on #2298 despite 19 passing mutations.

# What is not in this PR

**Thread updates on status transitions.** The ticket asks for them and they are not buildable as things stand: Slack **incoming webhooks return no message timestamp**, so there is nothing to thread a reply to, and the repo has no bot token — `SlackNotifier` is webhook-only and `.env.example` has only `WEBHOOK_*` entries. Threading needs a Slack app with `chat.postMessage`, which is a credential to provision rather than code to write. Raised as **D18 on EXSC-883**; the card itself is useful without it, so I shipped that half rather than blocking on it.

**`ticketUrl` is not rendered.** The field arrives with WP-1.2 ([#2298](https://github.com/lifinance/contracts/pull/2298), still a draft). Built off `origin/main` deliberately rather than stacking — a squash-merge breaks a stack, and stacked PRs get no automatic CodeRabbit review. Adding the row is a two-line follow-up once #2298 lands.

## Reversible choices

- **Four networks** is the threshold for naming versus counting, and for one command versus a summarising line.
- **The newest pending proposal per network** is the selector. Nothing in a document identifies the run that created it, so a network carrying an older unsigned proposal would contribute that one instead; the run's own network list bounds the query. Networks with no pending proposal are named in a warning rather than silently dropped.
- **The first proposal's provenance heads the card, and divergence is named rather than hidden.** A run proposes the same change everywhere, so reason/proposer/PR are identical in practice; when they are not, the card warns and says only the first is shown. The two fields with a safety consequence are unioned across every row instead of read from the first — a dirty working tree or a non-human actor on any network is never suppressed by another network's clean row, and a value in the dirty-tree field that cannot be read as a path list reports as unreadable rather than as clean.

---

# Accepted residuals, decided rather than dropped

**A bare URL in a free-text field is not neutralised.** Only `prUrl` is guarded, and a reviewer was right that leaving this unmentioned across three rounds read as dropped. The decision, so it is on the record:

The distinction is whether **the card asserts the value is a link**. `*PR:* <url>` is the card saying "this is the PR" — an unvalidated value there borrows the card's authority, which is why it is scheme-checked and withheld when it fails. `*Reason:*` is proposer-supplied prose, and a signer reads it as prose. Slack will auto-link a URL inside it, but neutralising URLs in prose breaks a legitimate reason that cites one, and the alternative — rendering prose the card has silently edited — is worse than a visibly clickable string in a field labelled as someone's words.

`reason` is capped at 200 characters, sanitised through `sanitizeProvenanceText`, and stripped of the mrkdwn markers that could disguise it as a label. What remains is a proposer able to put a link in their own stated reason, which they could equally do in the Linear ticket the card links to.

**Two smaller ones**, both low reachability and shared with the signing prompt:

- `_none given_` is a forgeable sentinel. Forging the absence of one's own reason gains an attacker nothing.
- `expectedHashes` and `checkSummary` have no production caller today — `render-proposal-card.ts` maps six fields and neither is among them. They are the reserved WP-7.2 slot. This bounds every truncation finding in this PR to hardening: a real 41-network card measures a few hundred characters against an 8,000 cap.

# What the working-tree line now says

Four states, matching `provenance-display.ts`'s model for the signing prompt, each naming its own networks:

```text
clean everywhere      (no line)
capture failed on one  *Working tree:* capture incomplete on arbitrum
dirty + unreadable     *Working tree:* dirty on mainnet — src/A.sol; not captured on arbitrum
dirty + incomplete     *Working tree:* dirty on mainnet — src/A.sol; capture incomplete on arbitrum
```

The reason that matters: a failed `git status` writes an **empty path list plus `captureErrors`**, and the card mapped neither field — so such a row fell into no bucket and rendered **no working-tree line at all**. The two likeliest probe failures are the 5-second timeout and the 8 MB buffer cap, both of which correlate with a *very* dirty tree. The card read clean exactly when the tree was dirtiest.

### DIFF
--- a/docs/MultisigSigningProcess.md
+++ b/docs/MultisigSigningProcess.md
@@ -128,9 +128,14 @@ live `getMinDelay()` with `config/timelockController.json` fallback, timestamp
 salt), resolves the nonce (`getNextNonce`), **signs immediately** (EIP-712),
 computes the `safeTxHash` via the Safe's on-chain `getTransactionHash`, and
 stores. The document (`ISafeTxDocument`) carries the raw Safe tx fields, the
-proposer's wallet address, an `intentHash` dedup key, and a
-`pending → submitted → executed / reverted` status — but **no description, PR
-link (except drain `parkedTaskRefs`), git commit, or human identity**.
+proposer's wallet address, an `intentHash` dedup key, a
+`pending → submitted → executed / reverted` status, the origin-PR
+`parkedTaskRefs` on a drained facet removal, and a `provenance` block
+(`IProposalProvenance`) holding the rationale, proposer identity and actor, git
+commit and branch, scoped working-tree dirtiness, the branch's PR link and any
+capture errors. **Rows stored before that block existed carry none of it**,
+which is why both the signing prompt and the Slack card state each absence
+explicitly instead of dropping the line.
 `notifyProposalsCreatedToSlack` (`script/multiNetworkExecution.sh`) posts the
 signing-ask card rendered by `script/deploy/safe/render-proposal-card.ts` —
 reason, proposer, PR, commit, working-tree state and a per-network
--- a/docs/MultisigSigningProcess.md
+++ b/docs/MultisigSigningProcess.md
@@ -131,9 +131,11 @@ stores. The document (`ISafeTxDocument`) carries the raw Safe tx fields, the
 proposer's wallet address, an `intentHash` dedup key, and a
 `pending → submitted → executed / reverted` status — but **no description, PR
 link (except drain `parkedTaskRefs`), git commit, or human identity**.
-`notifyProposalsCreatedToSlack` (`script/multiNetworkExecution.sh`) posts
-count + contract + network to `#dev-sc-multisig-proposals`; the rich
-signing-ask thread is a manual step per `.agents/commands/multisig-rollout.md`.
+`notifyProposalsCreatedToSlack` (`script/multiNetworkExecution.sh`) posts the
+signing-ask card rendered by `script/deploy/safe/render-proposal-card.ts` —
+reason, proposer, PR, commit and a per-network `bun confirm-safe-tx` command — to
+`#dev-sc-multisig-proposals`, falling back to a count + contract + network line
+only if the render fails.
 
 ### 4.3 Confirm / sign
 
--- a/docs/MultisigSigningProcess.md
+++ b/docs/MultisigSigningProcess.md
@@ -133,9 +133,9 @@ proposer's wallet address, an `intentHash` dedup key, and a
 link (except drain `parkedTaskRefs`), git commit, or human identity**.
 `notifyProposalsCreatedToSlack` (`script/multiNetworkExecution.sh`) posts the
 signing-ask card rendered by `script/deploy/safe/render-proposal-card.ts` —
-reason, proposer, PR, commit and a per-network `bun confirm-safe-tx` command — to
-`#dev-sc-multisig-proposals`, falling back to a count + contract + network line
-only if the render fails.
+reason, proposer, PR, commit, working-tree state and a per-network
+`bun confirm-safe-tx` command — to `#dev-sc-multisig-proposals`, falling back to
+a count + contract + network line only if the render fails.
 
 ### 4.3 Confirm / sign
 
--- a/script/deploy/safe/proposal-card.test.ts
+++ b/script/deploy/safe/proposal-card.test.ts
@@ -5,7 +5,11 @@ import {
   // eslint-disable-next-line import/no-unresolved
 } from 'bun:test'
 
-import { renderProposalCard, type ICardProposal } from './proposal-card'
+import {
+  SLACK_TEXT_LIMIT,
+  renderProposalCard,
+  type ICardProposal,
+} from './proposal-card'
 
 const proposal = (overrides: Partial<ICardProposal> = {}): ICardProposal => ({
   network: 'mainnet',
@@ -134,17 +138,88 @@ describe('renderProposalCard', () => {
   })
 
   it('truncates, and stays under the cap, when the card still overflows', () => {
-    const card = renderProposalCard([
-      proposal({ dirtyTreeScoped: [`src/${'a'.repeat(4000)}.sol`] }),
-    ])
+    // Every single field is capped now, so no one of them can overflow the card
+    // on its own. The hash table's ROW COUNT is not capped, and each `&` in a
+    // value becomes five characters once escaped — so this is what an overflow
+    // actually looks like, rather than one enormous string.
+    const card = renderProposalCard([proposal({ reason: '&'.repeat(200) })], {
+      expectedHashes: Array.from({ length: 200 }, (_, i) => ({
+        network: `chain${i}`,
+        expected: `0x${'ab'.repeat(32)}`,
+        actual: `0x${'cd'.repeat(32)}`,
+      })),
+    })
 
-    expect(card.length).toBeLessThanOrEqual(2900)
+    expect(card.length).toBeLessThanOrEqual(SLACK_TEXT_LIMIT)
     expect(card).toContain('card truncated')
     // An error-path string never runs in a happy-path test, which is how a
     // command that was never a `bun` alias shipped as the recovery instruction.
     expect(card).toContain(
       'bunx tsx script/deploy/safe/list-pending-proposals.ts'
     )
+  })
+
+  it('keeps the review commands on an overflowing card', () => {
+    // They are the only actionable part and they sit at the bottom, so a
+    // tail-truncated card lost exactly the thing it exists to deliver. Measured
+    // before the fix: every one of 5226 overflowing configurations dropped the
+    // whole review section.
+    const card = renderProposalCard(
+      ['mainnet', 'arbitrum'].map((network) => ({
+        network,
+        safeTxHash: `0x${'ab'.repeat(32)}`,
+      })),
+      {
+        expectedHashes: Array.from({ length: 200 }, (_, i) => ({
+          network: `chain${i}`,
+          expected: `0x${'ab'.repeat(32)}`,
+        })),
+      }
+    )
+
+    expect(card).toContain('card truncated')
+    expect(card).toContain('*To review:*')
+    expect(card).toContain('bun confirm-safe-tx --network mainnet')
+    expect(card).toContain('bun confirm-safe-tx --network arbitrum')
+    expect(card.length).toBeLessThanOrEqual(SLACK_TEXT_LIMIT)
+  })
+
+  it('keeps underscores and tildes, which real paths and emails contain', () => {
+    // Stripping these mangled real data for no security gain: they only
+    // italicise or strike text, so they cannot forge a label or break out of a
+    // code span. `alice_smith@…` and `alicesmith@…` rendered identically on a
+    // card whose job is saying who proposed, and 115 tracked paths in this repo
+    // carry an underscore.
+    const card = renderProposalCard([
+      proposal({
+        proposerHandle: 'A B <alice_smith@example.com>',
+        dirtyTreeScoped: ['.github/pull_request_template.md'],
+        reason: 'bump ~1.2.0',
+      }),
+    ])
+
+    expect(card).toContain('alice_smith@example.com')
+    expect(card).toContain('.github/pull_request_template.md')
+    expect(card).toContain('~1.2.0')
+  })
+
+  it('flags a divergent commit or PR, not only a divergent reason', () => {
+    // gitCommit is what a signer re-derives calldata against and prUrl is the
+    // rationale they read, so these diverging matters more than the prose doing
+    // so. The first version of this flagged only the prose.
+    const card = renderProposalCard([
+      proposal(),
+      proposal({
+        network: 'arbitrum',
+        gitCommit: 'ffffffffffff',
+        prUrl: 'https://github.com/lifinance/contracts/pull/999',
+        proposerHandle: 'Bob Other <bob@example.com>',
+      }),
+    ])
+
+    expect(card).toContain('gitCommit')
+    expect(card).toContain('PR')
+    expect(card).toContain('proposerHandle')
   })
 
   it('never cuts a surrogate pair or an entity in half', () => {
@@ -336,5 +411,49 @@ describe('renderProposalCard — it cannot understate the ask', () => {
     ])
 
     expect(card.toLowerCase()).toContain('differ')
+  })
+})
+
+describe('renderProposalCard — the PR link is a link, or it is not shown', () => {
+  it.each([
+    ['javascript:', 'javascript:alert(1)'],
+    ['data:', 'data:text/html,<script>x</script>'],
+    ['file:', 'file:///etc/passwd'],
+    ['plain http', 'http://evil.example/pull/1'],
+    ['a bare word', 'nope'],
+    ['a protocol-relative URL', '//evil.example/pull/1'],
+  ])('does not present a %s value as a link', (_label, prUrl) => {
+    // The capture path already accepts only https, so a value that is not one
+    // reached the document by hand. Presenting it as a PR lends it the card's
+    // credibility, and Slack auto-links a bare URL — the same gap as the ticket
+    // link's missing scheme check on #2298, on the field next to it.
+    const card = renderProposalCard([proposal({ prUrl })])
+
+    expect(card).toContain('not a link, ignored')
+    expect(card).not.toMatch(/\*PR:\* \S+$/m)
+  })
+
+  it('keeps a real https PR link', () => {
+    expect(
+      renderProposalCard([
+        proposal({ prUrl: 'https://github.com/lifinance/contracts/pull/2125' }),
+      ])
+    ).toContain('*PR:* https://github.com/lifinance/contracts/pull/2125')
+  })
+
+  it('says a link was rejected rather than staying silent about it', () => {
+    // Silently omitting looks identical to "no PR was recorded", which is a
+    // different fact and a less alarming one.
+    expect(
+      renderProposalCard([proposal({ prUrl: 'javascript:alert(1)' })])
+    ).toContain('not a link')
+  })
+
+  it('does not treat an https prefix inside a longer scheme as https', () => {
+    expect(
+      renderProposalCard([
+        proposal({ prUrl: 'nothttps://github.com/x/pull/1' }),
+      ])
+    ).toContain('not a link')
   })
 })
--- a/script/deploy/safe/proposal-card.test.ts
+++ b/script/deploy/safe/proposal-card.test.ts
@@ -131,7 +131,7 @@ describe('renderProposalCard', () => {
       }),
     ])
 
-    expect(card.length).toBeLessThanOrEqual(2900)
+    expect(card.length).toBeLessThanOrEqual(SLACK_TEXT_LIMIT)
     expect(card).toContain('bun confirm-safe-tx --network mainnet')
     expect(card).toContain('0xabababab')
     expect(card).toContain('*Proposed by:*')
@@ -160,10 +160,8 @@ describe('renderProposalCard', () => {
   })
 
   it('keeps the review commands on an overflowing card', () => {
-    // They are the only actionable part and they sit at the bottom, so a
-    // tail-truncated card lost exactly the thing it exists to deliver. Measured
-    // before the fix: every one of 5226 overflowing configurations dropped the
-    // whole review section.
+    // They are the only actionable part and they sit at the bottom, so a tail
+    // cut would drop exactly the thing the card exists to deliver.
     const card = renderProposalCard(
       ['mainnet', 'arbitrum'].map((network) => ({
         network,
@@ -185,11 +183,11 @@ describe('renderProposalCard', () => {
   })
 
   it('keeps underscores and tildes, which real paths and emails contain', () => {
-    // Stripping these mangled real data for no security gain: they only
+    // Stripping these mangles real data for no security gain: they only
     // italicise or strike text, so they cannot forge a label or break out of a
-    // code span. `alice_smith@…` and `alicesmith@…` rendered identically on a
-    // card whose job is saying who proposed, and 115 tracked paths in this repo
-    // carry an underscore.
+    // code span, while `alice_smith@…` and `alicesmith@…` render identically on
+    // a card whose job is saying who proposed, and many tracked paths in this
+    // repo carry an underscore.
     const card = renderProposalCard([
       proposal({
         proposerHandle: 'A B <alice_smith@example.com>',
@@ -206,7 +204,7 @@ describe('renderProposalCard', () => {
   it('flags a divergent commit or PR, not only a divergent reason', () => {
     // gitCommit is what a signer re-derives calldata against and prUrl is the
     // rationale they read, so these diverging matters more than the prose doing
-    // so. The first version of this flagged only the prose.
+    // so.
     const card = renderProposalCard([
       proposal(),
       proposal({
@@ -217,23 +215,37 @@ describe('renderProposalCard', () => {
       }),
     ])
 
-    expect(card).toContain('gitCommit')
-    expect(card).toContain('PR')
-    expect(card).toContain('proposerHandle')
+    // The warning names the card's own labels, so asserted on the whole
+    // sentence — `PR` and `Commit` both occur elsewhere on the card.
+    const warning = card.split('\n').find((l) => l.startsWith('⚠ These differ'))
+
+    expect(warning).toContain('Commit')
+    expect(warning).toContain('PR')
+    expect(warning).toContain('Proposed by')
   })
 
   it('never cuts a surrogate pair or an entity in half', () => {
     for (const filler of ['\u{1F64F}', '&', 'a'])
       for (let pad = 1; pad <= 6; pad++) {
-        const card = renderProposalCard([
-          proposal({
-            network: 'n'.repeat(pad),
-            dirtyTreeScoped: [filler.repeat(4000)],
-          }),
-        ])
-        const body = card.slice(0, card.indexOf('\n… card truncated'))
+        // The overflow has to come from the row COUNT: every field is capped, so
+        // no single value can reach the card's cap on its own. `pad` walks the
+        // cut one character at a time so it lands mid-pair and mid-entity.
+        const card = renderProposalCard(
+          [proposal({ network: 'n'.repeat(pad) })],
+          {
+            expectedHashes: Array.from({ length: 40 }, () => ({
+              network: filler.repeat(200),
+              expected: filler.repeat(200),
+            })),
+          }
+        )
+        const cut = card.indexOf('\n… card truncated')
+        const body = card.slice(0, cut)
 
-        expect(card.length).toBeLessThanOrEqual(2900)
+        // Without this the assertions below run against a whole, uncut card and
+        // prove nothing about the cut.
+        expect(cut).toBeGreaterThan(0)
+        expect(card.length).toBeLessThanOrEqual(SLACK_TEXT_LIMIT)
         // A lone high surrogate renders as U+FFFD; a bare `&am` as literal junk.
         expect(/[\uD800-\uDBFF]$/.test(body)).toBe(false)
         expect(/&[a-z]{0,3}$/i.test(body)).toBe(false)
@@ -251,10 +263,10 @@ describe('renderProposalCard', () => {
   })
 
   it('strips control and bidi characters from every rendered field', () => {
-    // EXSC-693: collapsing `\s+` does not touch Cc/Cf, so ANSI escapes and the
-    // bidi overrides behind Trojan Source reached the card unfiltered. Asserted
-    // per field, because the length-capped fields and the uncapped ones reach
-    // Slack by different escape paths.
+    // Collapsing `\s+` does not touch Cc/Cf, so ANSI escapes and the bidi
+    // overrides behind Trojan Source need their own removal. Asserted per field,
+    // because the length-capped fields and the uncapped ones reach Slack by
+    // different escape paths.
     const payload = '\u001b[2K\u202ereversed\u200bhidden\u0007'
     const banned = ['\u001b', '\u202e', '\u200b', '\u0007']
     const fields: (keyof ICardProposal)[] = [
@@ -393,7 +405,6 @@ describe('renderProposalCard — it cannot understate the ask', () => {
   })
 
   it('names the contract in the headline, which the message it replaces did', () => {
-    // Reviewer feedback on PR #1904 asked to keep the count plus the contract.
     // A name absent from every other fixture field, and asserted on the headline
     // rather than the whole card: the default fixture's reason mentions
     // AcrossFacetV4, so `toContain` on the card passed with no headline change.
@@ -425,8 +436,7 @@ describe('renderProposalCard — the PR link is a link, or it is not shown', () 
   ])('does not present a %s value as a link', (_label, prUrl) => {
     // The capture path already accepts only https, so a value that is not one
     // reached the document by hand. Presenting it as a PR lends it the card's
-    // credibility, and Slack auto-links a bare URL — the same gap as the ticket
-    // link's missing scheme check on #2298, on the field next to it.
+    // credibility, and Slack auto-links a bare URL.
     const card = renderProposalCard([proposal({ prUrl })])
 
     expect(card).toContain('not a link, ignored')
--- a/script/deploy/safe/proposal-card.test.ts
+++ b/script/deploy/safe/proposal-card.test.ts
@@ -256,3 +256,85 @@ describe('renderProposalCard', () => {
     expect(card).toMatch(/reason[^\n]*none given/i)
   })
 })
+
+describe('renderProposalCard — values cannot forge card structure', () => {
+  it('strips markdown markers that entity-escaping does not neutralise', () => {
+    // A backtick closes the code span the review command sits in, and the text
+    // after it renders as card prose. `*` and `_` forge a bold label inline, so
+    // a reason can fake the advisory line. Slack mrkdwn has no escape sequence
+    // for these, so the only options are strip or let them through.
+    const card = renderProposalCard([
+      proposal({
+        network: 'main`net`*URGENT: already reviewed, just sign*',
+        reason: '*Proposer-side check (advisory):* all clear',
+      }),
+    ])
+
+    expect(card).not.toContain('`URGENT')
+    expect(card).not.toContain('*URGENT')
+    // Exactly one code span opens and closes on the command line.
+    const commandLine = card
+      .split('\n')
+      .find((l) => l.includes('confirm-safe-tx'))
+    expect((commandLine?.match(/`/g) ?? []).length).toBe(2)
+    // And the forged advisory label is inert. Checked as "no stray marker
+    // survives on the reason line", not as "no line starts with the label" —
+    // the forgery renders mid-line, so the line-start form passed regardless.
+    const reasonLine = card.split('\n').find((l) => l.startsWith('*Reason:*'))
+    expect(reasonLine).toBeDefined()
+    expect((reasonLine?.match(/\*/g) ?? []).length).toBe(2)
+  })
+
+  it('neutralises a control sequence in the hash, which is not a hex string', () => {
+    // `<!channel>` is exactly the short-hash width, so it survived the slice
+    // unescaped and notified the channel from the one field the renderer
+    // trusted.
+    const card = renderProposalCard([proposal({ safeTxHash: '<!channel>' })])
+
+    expect(card).not.toContain('<!channel>')
+  })
+})
+
+describe('renderProposalCard — it cannot understate the ask', () => {
+  it('says so when fewer proposals were found than the run created', () => {
+    // A card reading "38x proposals" after 41 networks succeeded leaves three
+    // proposals that nobody is told to sign. Undercounting the signing ask is
+    // the one error here with a direct safety consequence.
+    const card = renderProposalCard(
+      [proposal(), proposal({ network: 'arbitrum' })],
+      {
+        expectedCount: 5,
+      }
+    )
+
+    expect(card).toContain('5')
+    expect(card.toLowerCase()).toMatch(/only 2|3 (missing|not found)/)
+  })
+
+  it('stays quiet when the counts agree', () => {
+    const card = renderProposalCard([proposal()], { expectedCount: 1 })
+
+    expect(card.toLowerCase()).not.toContain('missing')
+  })
+
+  it('names the contract in the headline, which the message it replaces did', () => {
+    // Reviewer feedback on PR #1904 asked to keep the count plus the contract.
+    // A name absent from every other fixture field, and asserted on the headline
+    // rather than the whole card: the default fixture's reason mentions
+    // AcrossFacetV4, so `toContain` on the card passed with no headline change.
+    const headline = renderProposalCard([proposal()], {
+      contract: 'WhitelistManagerFacet',
+    }).split('\n')[0]
+
+    expect(headline).toContain('WhitelistManagerFacet')
+  })
+
+  it('flags divergent reasons rather than printing the first as if it covered all', () => {
+    const card = renderProposalCard([
+      proposal({ reason: 'reason A' }),
+      proposal({ network: 'arbitrum', reason: 'reason B' }),
+    ])
+
+    expect(card.toLowerCase()).toContain('differ')
+  })
+})
--- a/script/deploy/safe/proposal-card.test.ts
+++ b/script/deploy/safe/proposal-card.test.ts
@@ -19,6 +19,10 @@ const proposal = (overrides: Partial<ICardProposal> = {}): ICardProposal => ({
   reason: 'add AcrossFacetV4 to the whitelist',
   prUrl: 'https://github.com/lifinance/contracts/pull/2125',
   gitCommit: '1234567890abcdef',
+  // A measured empty array, which is what capture actually writes for a clean
+  // tree. Omitting it made the default fixture look like a pre-provenance row,
+  // for which the card now correctly says "not captured".
+  dirtyTreeScoped: [],
   ...overrides,
 })
 
@@ -585,7 +589,7 @@ describe('renderProposalCard — an unreadable dirty-tree field is not clean', (
       const line = card.split('\n').find((l) => l.startsWith('*Working tree:*'))
       expect(line).toBeDefined()
       expect(line).toContain('arbitrum')
-      expect(line).toMatch(/could not be read/)
+      expect(line).toMatch(/not captured/)
     }
   )
 
@@ -658,5 +662,60 @@ describe('renderProposalCard — the unions cannot become the overflow', () => {
 
     expect(small).toBeGreaterThan(0)
     expect(large - small).toBeLessThan(10)
+  })
+})
+
+describe('renderProposalCard — the dirty-path list cannot understate itself', () => {
+  it('says so when capture truncated the path list', () => {
+    // Capture sets dirtyTreeTruncated and the signing prompt renders "20+
+    // dirty". The card had no such field, so exactly MAX_DIRTY_PATHS paths took
+    // the "<= limit" branch and a 500-file dirty tree rendered identically to a
+    // 20-file one — the same understating-a-warning class as reading the field
+    // from row zero.
+    const card = renderProposalCard([
+      proposal({
+        dirtyTreeScoped: Array.from({ length: 20 }, (_, i) => `src/F${i}.sol`),
+        dirtyTreeTruncated: true,
+      }),
+    ])
+
+    const line = card.split('\n').find((l) => l.startsWith('*Working tree:*'))
+    expect(line).toMatch(/more/)
+  })
+
+  it('stays exact when capture was not truncated', () => {
+    const card = renderProposalCard([
+      proposal({
+        dirtyTreeScoped: Array.from({ length: 20 }, (_, i) => `src/F${i}.sol`),
+      }),
+    ])
+
+    const line = card.split('\n').find((l) => l.startsWith('*Working tree:*'))
+    expect(line).not.toMatch(/more/)
+  })
+})
+
+describe('renderProposalCard — an uncaptured working tree is not a clean one', () => {
+  it('reports an absent dirtyTreeScoped rather than rendering nothing', () => {
+    // `provenance-display.ts` already holds this rule for the signing prompt:
+    // only a measured empty array may read as clean, and a non-array — absent
+    // and null included — is 'unreadable'. Emitting no line for those left the
+    // card and the prompt disagreeing about one field.
+    const card = renderProposalCard([
+      proposal({ dirtyTreeScoped: undefined }),
+      proposal({ network: 'arbitrum', dirtyTreeScoped: undefined }),
+    ])
+
+    expect(card).toContain('*Working tree:*')
+    expect(card).toMatch(/not captured/)
+  })
+
+  it('is silent only for a measured empty array', () => {
+    const card = renderProposalCard([
+      proposal({ dirtyTreeScoped: [] }),
+      proposal({ network: 'arbitrum', dirtyTreeScoped: [] }),
+    ])
+
+    expect(card).not.toContain('*Working tree:*')
   })
 })
--- a/script/deploy/safe/proposal-card.test.ts
+++ b/script/deploy/safe/proposal-card.test.ts
@@ -439,7 +439,7 @@ describe('renderProposalCard — the PR link is a link, or it is not shown', () 
     // credibility, and Slack auto-links a bare URL.
     const card = renderProposalCard([proposal({ prUrl })])
 
-    expect(card).toContain('not a link, ignored')
+    expect(card).toContain('not a link, withheld')
     expect(card).not.toMatch(/\*PR:\* \S+$/m)
   })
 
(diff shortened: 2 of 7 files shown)


0xDEnYO:
## `/gate-review` round 1 — escalated findings

Auto-fixed in 4a204cf4d (details in the commit). The items below were **not** changed — each
needs a decision. All reproduced by executing the real module.

### E1 — `escape()` does not neutralise backticks; a value breaks out of the card's own code span

`escape` handles `& < >`, which is Slack's documented escape set for mentions, so every
`<!channel>` / `<@U…>` / `<!subteam^…>` form is genuinely neutralised. Backticks are not, and
the review-command line wraps a proposer-controlled value **inside a code span**:

```
network = 'mainnet` *URGENT: already reviewed, just sign* `'
→ • `bun confirm-safe-tx --network mainnet` *URGENT: already reviewed, just sign* ``
```

Slack renders that bold text outside the span. A single backtick is enough to corrupt it: a
`reason` containing one leaves an odd backtick count, so the span boundaries shift and the
actual command loses its formatting. Three backticks open a code fence that swallows the rest
of the card.

Not auto-fixed because Slack offers no escape for backticks — strip them, substitute U+02CB,
or reject the value are all defensible, and it changes rendered output.

### E2 — `*` and `_` let a value forge a label on the line it already occupies

Newline-flattening stops a value creating a *new* line, which is what the enumeration pass
covered. It does not stop a value forging a label inline:

```
reason = 'routine *Proposer-side check (advisory):* codehash matched 41/41'
→ *Reason:* routine *Proposer-side check (advisory):* codehash matched 41/41
```

Both labels render bold, so the card appears to attest a check that never ran — the exact
thing the `checkSummary` docstring says the card must not do. Separately, `reason = '_none
given_'` renders identically to the absent-reason sentinel. Same fix decision as E1.

### E3 — `prUrl` is rendered under a `*PR:*` label with no validation

`javascript:alert(1)`, `http://evil.example/pull/1`, `https://evil.example/…/pull/2299` and
`not a url at all` all render as-is. The capture path already gates this
(`git-provenance.ts:456`, `clean.startsWith('https://')`), so the guard exists one layer up
and is absent here. Escalated because the tightening level is a judgment call (`https://` vs
`https://github.com/` vs the lifinance org) and it changes what a legacy row renders — and
because PR #2298 is deciding the same question for the same field.

### E4 — the card can silently understate how many proposals need signing

`render-proposal-card.ts` derives the headline count from the rows it found, not from the
networks the run actually succeeded on. A network whose row does not match (any reason) is
dropped, `consola.warn`s to the runner's terminal, and the card posts as though complete: 41
successful networks with 38 matching rows reads `38x proposals created`. The old message used
the runner's own `PROPOSAL_COUNT`, which was ground truth. Proposals that exist and need
signing are omitted from the ask with no in-card trace.

Fix shape is a real choice: pass the expected count from bash and assert it, render a "N
networks had no pending row" line, or fail the render into the fallback.

### E5 — reason / proposer / commit / PR are read from `proposals[0]` and applied to all networks

Only `network` and `safeTxHash` vary per row; everything else comes from the first proposal.
The selection is "newest pending row per network, independently" — and the code comment
concedes nothing in the document identifies the run. So network A's row can come from an
earlier run than network B's, and the card prints one run's reason as if it covered both.
Verified: differing reasons across rows render only the first.

### E6 — the contract name is gone, and this is live reviewer feedback

The old line named it (`1x proposal created: AcrossFacetV4 on mainnet`). The card never
renders it — `PROPOSAL_CONTRACT` is now read only for the fallback. When no reason was given
the card identifies nothing being changed at all. On PR #1904, the immediately-preceding
change to this same message, the reviewer asked to "keep the top message short (just the count
+ contract)". That is unaddressed. The contract is available from the progress file and
decodable from `safeTx.data` via `getTargetName`.

### E7 — 2900 is the wrong limit for this payload, and the exit-code contract is lost

Two separate items, both needing a call:

- `SLACK_TEXT_LIMIT = 2900` is copied from `slack-notifier.ts`, whose own comment attributes
  it to the **per-block** 3000-char `invalid_blocks` limit. The card is posted as top-level
  `text` with no blocks, and `SlackNotifier` only truncates the `blocks[].text` it builds
  itself — verified, so there is no double truncation, but the cap is ~13x tighter than the
  payload needs. I corrected the false comment and left the number alone.
- `render-proposal-card.ts` opens a raw `MongoClient` (deliberately, so index creation cannot
  fire), which costs the actionable lifi-connect tunnel message and the directory's exit-code
  contract (`list-pending-proposals.ts`: 2 = recoverable misconfig, 1 = real error). The bash
  treats every non-zero identically, so a down tunnel and a genuine bug both degrade to the
  count-only line indistinguishably. Fixing it properly needs a matching bash change.

### E8 — `render-proposal-card.ts` has no test file

The 100%-coverage rule for new TS helpers is unmet, and the untested part is the part most
likely to be wrong: newest-per-network selection and output ordering. Not authored here — I
did not want to add untested test code inside a review pass. Siblings (`parked-tasks.test.ts`,
`reconcile.test.ts`) stub the Mongo `Collection` directly.

I did execute the query against a throwaway MongoDB 7. Results: `status: 'pending'` matches
`SafeTxStatus` and correctly excludes `executed`; rows missing `timestamp` sort **last** under
`{timestamp: -1}`, so they are only picked when they are the network's only pending row (the
desired behaviour); `network` lowercasing agrees with `storeTransactionInMongoDB`, which is
the only insert path, Tron included. Two latent mis-orderings are unreachable on the
production path but would bite a hand-edited row: a `timestamp` stored as a string sorts below
every `Date` (BSON type order), and a mixed-case stored `network` never matches and is silently
dropped into E4.

---

- `actor` renders verbatim for anything but `human`, unconstrained by `ProvenanceActor`, so a
  row can put `(VERIFIED BY SECURITY)` in the parenthetical.
- `checkSummary` has no production caller: nothing in the stored schema carries it and
  `render-proposal-card.ts` never sets it, so that branch is unreachable outside the test.
  `expectedHashes` is the same, though it is documented forward scaffolding.
- The card drops `captureErrors` and renders the working-tree line only when dirty, so a row
  whose dirty-tree capture failed shows no working-tree line — reading as clean.
  `provenance-display.ts` calls that impression the one it must never give.
- The bash `NETWORK_LIST` join is correct — verified under `set -euo pipefail` with empty, one
  and many elements, and with glob/backtick/space characters; the jq already filters
  space-bearing names, and `PROPOSAL_COUNT` is `${#SUCCESSFUL_NETWORKS[@]}` so the non-empty
  assumption is tautological. The write mode is safe: `writeFileSync` and the fallback's `>`
  both truncate, so no partial-write path exists. The function still returns 0 on every path.
  I dropped a pointless `2>&1` that merged the renderer's stderr into the run's stdout.
  `shellcheck` reports only pre-existing SC2155/SC2207 in this function.

### Verified this round

`bun test script/` 1558 pass / 0 fail (was 1551; the card file goes 13 → 20 tests).
`bunx eslint` and `bunx tsc-files --noEmit` clean on all changed files. `bash -n` clean.
Ten mutations of the guards added here, all killed — including one that caught a flaw in my
own first fix (capping after escaping split an entity, so capping now precedes escaping).

Not verified: behaviour against the live Slack API. Every mrkdwn claim above is from Slack's
documented escaping rules plus the rendered text, not from a posted message.

0xDEnYO:
## `/gate-review` round 3 — escalated findings

Auto-fixed in `f8cfe93e0` (details in the commit). CodeRabbit is rate-limited on this PR (`CodeRabbit` reports `pass` with "Review rate limited"), so this gate is the only review that has run.

The items below were **not** changed — each needs a decision.

### E1 — A card can say "clean tree, human, all green" while another network was dirty, bot-proposed and failing (`proposal-card.ts:245-300`)

`dirtyTreeScoped`, `actor` and `checkSummary` are rendered from `proposals[0]` only and are **not** in the divergence set. Executed, two proposals, first clean:

```
input:  [{network:'mainnet', dirtyTreeScoped:[], actor:'human', checkSummary:'all green'},
         {network:'arbitrum', dirtyTreeScoped:['src/Facets/Evil.sol'], actor:'bot', checkSummary:'FAILED: 3 mismatches'}]

card:   *Proposed by:* alice
        *Proposer-side check (advisory):* all green
        (no *Working tree:* line, no ⚠ divergence line, no mention of Evil.sol)
```

`dirtyTreeScoped` and `actor` are both populated by the production caller (`render-proposal-card.ts:122,125`), so this is reachable on any multi-network run where the tree went dirty or an actor differed mid-run. The divergence mechanism exists precisely to stop first-row values reading as covering every network, and the three fields with the most direct safety consequence are the ones outside it. There is no comment or test recording the omission as deliberate.

Not auto-fixed: adding them changes the card for real runs, and there is more than one reasonable design (add to the divergence list / render per-network / render `dirtyTreeScoped` as an explicit "clean" so absence stops meaning clean). `dirtyTreeScoped` also needs a comparison that survives arrays — `String()` collapses `['a','b']` and `['a,b']` to the same value.

Related: `provenance-display.ts:69-78` already codes the inverse invariant for the sibling CLI — a failed dirty-tree probe renders yellow `UNKNOWN`, never green `clean` (EXSC-693, `80c3c1bf6`). This card renders no line at all when `dirtyTreeScoped` is absent or non-array, so a failed capture reads as clean. That is a regression against a decision the repo already made.

### E2 — `renderLink` guards the one field already validated, and the label it prints is false

Two separate problems, same helper.

**The threat model is inverted.** The docblock's rationale — "Slack auto-links a bare URL, so a signer gets a clickable attacker-chosen destination" — applies verbatim to `reason`, `proposerHandle`, `checkSummary` and `dirtyTreeScoped`, none of which are guarded. Executed:

```
*Reason:* see https://evil.example for approval
*Proposed by:* bot https://evil.example
*Working tree:* dirty — https://evil.example
*Proposer-side check (advisory):* ok https://evil.example
```

Meanwhile `prUrl` — the only guarded field — is the one whose sole production writer (`git-provenance.ts`, `openPrUrl`) already returns `undefined` unless the value starts with `https://`.

**The label states something untrue.** `startsWith('https://')` is case-sensitive, and the rejected value is left in the message verbatim:

```
"HTTPS://evil.example/pull/1" -> *PR:* HTTPS://evil.example/pull/1 — not a link, ignored
"http://evil.example/pull/1"  -> *PR:* http://evil.example/pull/1 — not a link, ignored
```

Slack auto-links both (documented behaviour — I could not execute this half, no webhook was posted). If so, the card asserts the value was ignored next to a live clickable link, which is worse than omitting the line: a signer who reads "ignored" stops evaluating the URL. The tests do not catch it — `expect(card).not.toMatch(/\*PR:\* \S+$/m)` passes trivially because the trailing label means the line no longer ends in a bare token, so the assertion proves the label is present, not that the link is neutralised.

The check is also scheme-only, with no destination validation: `https://`, `https:///`, `https://good.example@evil.example/pull/1` all pass and render as links.

Options: lower-case the scheme test and validate the host (this is a GitHub PR URL); or wrap a rejected value in a code span, which Slack does not auto-link, and which cannot be broken out of because `stripMarkdownMarkers` removes every backtick from values; or drop the value entirely. Escalated because it is signer-facing and there is no single obvious fix.

**One question worth settling cheaply:** post one card with `prUrl: 'HTTPS://evil.example'` to a scratch channel. If Slack renders it as a link, E2 is the most serious item in the PR.

### E3 — The divergence warning fires on values that render byte-identically

7 of 7 constructed cases flagged, all rendering the same text:

```
zero-width space only     flagged  (rendered identical)
control char only         flagged  (rendered identical)
bidi override only        flagged  (rendered identical)
stripped backtick only    flagged  (rendered identical)
stripped asterisk only    flagged  (rendered identical)
trailing whitespace only  flagged  (rendered identical)
tab vs space              flagged  (rendered identical)
differ past the 200 cap   flagged  (rendered identical)  <- correct, keep
```

`trailing whitespace` is the realistic one: `'deploy facet '` vs `'deploy facet'` prints "⚠ These differ … **Check each network individually**", sending a signer to hand-verify up to 41 networks over a trailing space. That is how a safety warning gets learned-ignored.

The last row is the case the raw comparison was introduced for and must keep firing. Comparing `sanitizeProvenanceText(v)` **uncapped** keeps it while removing the other six. Not auto-fixed because it changes when the warning appears.

Same comparison, opposite failure: `String(p[field] ?? '')` reports two genuinely different values as agreeing. `{a:1}` vs `{b:2}` both stringify to `[object Object]` → no warning, and the card displays `*Reason:* [object Object]`. `ICardProposal` types every field `unknown` precisely because a hand-edited or half-migrated row can hold a non-string, so this is the case the interface was designed for. `JSON.stringify` would fix it.

### E4 — `checkSummary` and `expectedHashes` cannot render in production

`render-proposal-card.ts:117-126` maps only `network, safeTxHash, proposerHandle, actor, reason, prUrl, gitCommit, dirtyTreeScoped`, and no caller passes `expectedHashes`. So `*Proposer-side check (advisory):*` and the whole `*Expected code hashes:*` table are unreachable, and the `MAX_FIELD_CHARS` docblock cites `checkSummary` as a field that can consume the budget — a case that cannot occur.

This also bounds every truncation finding in this PR: with real network names and no hash table, a 41-network card measures **633 characters** against the 8000 cap. Truncation is unreachable through the production caller, and the only fixture that overflows is the hash table no caller supplies. Worth deciding whether to wire both up or drop them, rather than leaving rationale pointing at dead code.

### E5 — The shortfall warning is inside the truncated range; the review commands are not

Round 3 moved the review section out of the truncatable range on the grounds that it is "the only actionable part". The ⚠ shortfall line — "N proposals missing from this card", described in the code as the one error with a direct safety consequence — stayed in. With a hostile ~1000-char contract and four 200-character network names the cut lands inside the headline and the shortfall warning is dropped entirely.

Unreachable through the real caller (see E4), so this is a design question, not a live bug: if the review commands earn unconditional placement, the missing-proposal warning arguably earns it more.

### E6 — `render-proposal-card.ts` has no test, and the round-3 shortfall logic is the untested part

The pure renderer has 38 tests. The `withoutPending` / `settled` / `unaccounted` split — the round-3 change with the most behavioural weight — has none. Repo precedent does allow untested CLI entry points (`list-pending-proposals.ts`, `propose-to-safe.ts`, `confirm-safe-tx.ts` are all untested), so this is a recommendation rather than a rule violation. I verified the logic by hand against a throwaway `mongo:7` instead; a `render-proposal-card.test.ts` stubbing `Collection` (the `parked-tasks.test.ts` pattern) would make that repeatable.

### E7 — `_none given_` is now forgeable

`_` is no longer stripped, so a `reason` whose sanitized text is literally `_none given_` renders identically to the absent-reason sentinel. Cosmetic — it downgrades information rather than escalating trust — but a sentinel of `(none given)` would remove the collision at no cost.

---

### On the round-3 claims I was asked to check

Confirmed correct, with executed evidence:

- **The shortfall redesign.** Against a real MongoDB: `submitted`, `executed` and `reverted` rows all read as settled and stay off the card; a network with no row in any status is counted and produces the ⚠ line with the right number. The reasoning about the runner's cumulative success list is sound — `SUCCESSFUL_NETWORKS` is every network ever marked `success` in the progress file.
- **`settled` / `unaccounted` cannot both miss a network** for scalar `network` values; they partition `withoutPending` exactly.
- **The card never exceeds `SLACK_TEXT_LIMIT`.** ~3500 configurations swept (all-`&`/`<`/`>`, astral, lone surrogates, 200-char fields, 5000-row hash tables, 400 networks): 0 over limit, maximum exactly 8000. The negative-slice case you asked about is **not reachable** at today's constants — worst-case `reviewText` measures 4369 against a 3536-char budget. It is reachable at `NAMED_NETWORK_LIMIT` 8 or `MAX_FIELD_CHARS` 380, which is why it is now clamped.
- **`SLACK_TEXT_LIMIT = 8000` is correct.** `send-slack-webhook-message.ts` posts `{ text }` — top-level, not `blocks` — so 3000 does not apply and 40,000 does. 2900 was copied from `slack-notifier.ts`, whose own comment scopes it to block text.
- **Narrowing the strip set to `` [`*] `` lets nothing exploitable through.** The load-bearing fact is that `sanitizeProvenanceText` collapses `\s+` to a single space, so a value can never contain a newline and therefore can never start a card line. With `*` still stripped, every injection lands mid-line after `*Reason:* ` and cannot produce a bold label. `<`/`>`/`&` escaping kills `<!channel>`, `<!here>`, `<@U123>`, `<url|text>` and `>` blockquote. What `_`/`~` buy an attacker is italic and strikethrough mid-sentence — cosmetic. **The trade was right**; data fidelity was worth more. The one residual, `_none given_` (E7), is a sentinel collision rather than a forgery.
- **`actor` and `checkSummary` were not consciously excluded** from the divergence set — no comment, no test, nothing in the four commits. See E1.

Corrected: **the backtick-balance guard was never in the production code.** `65e404867` added only a per-line test assertion (`proposal-card.test.ts`, backtick count on the command line `=== 2`), which is still there. Nothing was removed.

But the reasoning behind leaving it out was **wrong**, and the measurement was incomplete. The review section is no longer in the truncated range, but the **shortfall line carries its own backtick pair and is** — so a cut can land between them. 17 of 3000 swept configurations produced an odd total backtick count, shifting every downstream pairing and rendering the review commands as prose. Your 5226-configuration sweep evidently did not include entity-expanding network names alongside a long contract, which is what moves the cut onto that pair. Reachability through the real caller is nil (it needs ~160-character network names), so the defect is minor — but the reasoning would have misled whoever next moves `NAMED_NETWORK_LIMIT`. Now guarded; the sweep reports 0.

Also: the PR description still says "Truncation at Slack's 2900 characters". Worth updating before this goes to a human.

0xDEnYO:
## `/gate-review` round 4 — escalated findings

Auto-fixed in `44dd721c3`: four assertions that passed against the code they were written
to test. Proof: the round-3 assertions run verbatim against `f8cfe93e0` gave **6 pass / 2
fail** where all eight had to fail; the rewritten nine give **0 pass / 9 fail**.

CodeRabbit still cannot review this PR (org credit limit), so this is the only review
before a human.

The items below were **not** changed — each needs a decision.

---

### 1. The card cannot tell a clean working tree from an unmeasurable one — production-reachable

`captureErrors` is stored on the document (`safe-utils.ts:1510`) and rendered by the
signer-side CLI as `⚠ incomplete (N): …` (`provenance-display.ts:178`). `IRow`
(`render-proposal-card.ts:38`) does not read it and `ICardProposal` has no such field, so a
row whose git probe failed — `dirtyTreeScoped: []` plus a recorded error — renders **no
working-tree line at all**, i.e. exactly as a measured-clean tree.

`provenance-display.ts:69-78` refuses to paint that green on purpose:

> An empty list is what a failed probe, a missing/malformed field, and a clean tree all
> produce. Only a measured empty array with no capture errors may be painted green.

`c4108eb40`'s message claims the card now holds the invariant that CLI holds. It holds the
one-network-vs-all half; this half is still open, and it needs no hand-edited row — a git
spawn failure in a container is enough.

Same shape, hand-edit only: a row whose `dirtyTreeScoped` is a **truthy non-array** is
silently excluded by the `Array.isArray` filter and also reads as clean
(`proposal-card.ts:275`). The legacy-row test covers `'not-an-array'` but only asserts the
card renders.

### 2. The dirty list claims to be complete when capture already truncated it — production-reachable

Capture caps at `MAX_DIRTY_PATHS` (20) and sets `dirtyTreeTruncated`
(`git-provenance.ts:392-395`). The card never receives the flag, and computes its own
overflow from what it got: `paths.length > MAX_DIRTY_PATHS` is `20 > 20` → false.

Executed: a row carrying 20 paths renders 20 paths and **no "…and N more"**. A network with
50 dirty files therefore shows 20 and asserts nothing was dropped. `formatWorkingTree`
renders `⚠ 20+ dirty: … , …` for the same input.

Two lines to thread through `IRow` → `ICardProposal`, but it is a new field and a behaviour
change, so it is your call whether it rides here or follows.

### 3. Both new unions are unbounded per line — hand-edit only, but it can bury the dirty warning

`MAX_FIELD_CHARS` bounds each field to 200 chars, and the docstring at
`proposal-card.ts:28-32` states why: *"a single oversized … can consume the whole budget and
push the review command … off the bottom."* Both unions are now O(rows × 200), so that
per-line bound no longer holds, and the docstring overstates what the cap buys.

Executed, 41 rows, 40 with distinct 200-char `actor` values:

| | |
|---|---|
| `*Proposed by:*` line | 7,677 chars |
| card | truncated at 8,000 |
| review commands | **survive** |
| `*Working tree:*` (one network genuinely dirty, `src/Facets/CRITICAL.sol`) | **gone** |

So the answer to "can either union push the review commands off the card" is **no** — they
are appended after truncation. But the actor union can push the *dirty-tree warning* off,
which is the failure class `c4108eb40` exists to close, reached by a different route. Same
for the `(on …)` list: 40 dirty rows with 300-char network names give a 7,556-char
working-tree line.

Bounded in production: `ProvenanceActor` is `'human' | 'bot' | 'ci' | 'UNKNOWN'`
(`git-provenance.ts:111`) and `safe-utils.ts:1483-1486` re-clamps any other value to
`UNKNOWN`, so the actor union holds at most three short strings; network names come from
config. A production-shaped 41-network card measures **370 chars against the 8,000 cap**.
Hardening, not a live hole — but it is the one bound this commit removed.

### 4. `(on …)` names every network where the headline counts them — internal inconsistency

`describeNetworks` (`proposal-card.ts:169`) switches to `"N networks"` past
`NAMED_NETWORK_LIMIT` and the headline uses it. The new working-tree qualifier names all of
them: a 41-network run with 40 dirty renders a 388-char list under a headline that reads
"41x proposals created across 41 networks". Reusing the helper fixes both this and the bound
in item 3.

Confirmed working as intended: `MAX_DIRTY_PATHS` still bounds the path list under the union
(41 distinct paths → 20 shown + "…and 21 more", 426-char line); the qualifier **is** omitted
when every row is dirty; an `actor` that sanitizes to empty is dropped from the union.

### 5. PR body drift

- *"Truncation at Slack's 2900 characters"* — the code has used `SLACK_TEXT_LIMIT = 8000`
  since `f096f97ac`, and its docstring explains why 2900 is the wrong number here.
- *"1551 pass / 0 fail (1538 baseline; 13 new)"* — now 1583.
- Under Reversible choices, *"The first proposal's provenance heads the card … if they ever
  diverge, the card shows the first"* — no longer true of `dirtyTreeScoped` or `actor`.

---

- `_none given_` is forgeable: `escape` deliberately keeps `_`, so a reason of literally
  `_none given_` renders identically to an absent one. Verified. Cheapest close is a sentinel
  without `_`; the exposure is a proposer making the card understate their own intent, which
  is why I did not touch it.
- `checkSummary` and `expectedHashes` are supplied by no production caller
  (`render-proposal-card.ts:121-130` maps six fields; neither is one), so the new
  `COMPARED_FIELDS` entry and the whole hash-table branch are exercised only by tests. Fine
  as forward API — worth one line in the PR body so a reviewer does not read the divergence
  warning as live.
- Non-string members of `dirtyTreeScoped` render as `1, [object Object], x` — coercion is
  documented as deliberate, but `[object Object]` on a signer's card is a poor tell.
- A dirty row whose only path sanitizes to empty renders `*Working tree:* dirty (on mainnet) — `
  with a dangling em dash. Pre-existing.
- Paths are deduped **after** escaping, so `src/F*.sol` and `src/F.sol` collapse to one and
  the "…and N more" count understates by the difference.
- A capture failure renders `*Proposed by:* UNKNOWN (UNKNOWN)`.
- A valid https URL over 200 chars is truncated and still presented as a link. Not reachable
  from a GitHub PR URL.

---

**Escalated items need your decision before this PR is ready.** Items 1 and 2 are the two
that need no hand-edited row.

0xDEnYO:
## `/gate-review` round 5 — escalated findings

Auto-fixed in `5b1102143` (two card comments that described the previous behaviour) plus
the PR body (truncation limit was documented as 2,900 where the code is 8,000; test count
was 1551 where it is 1589; the reserved WP-7.2 slot is now labelled as having no production
caller). CodeRabbit is still rate-limited on this PR, so this gate is the only review
standing between `8e542954a` and a human.

**Round-4 technique re-run independently.** The six test cases added in `8e542954a` were
run verbatim against `8e542954a^`. Five fail there; the sixth
(`stays clean when the field is absent or an empty array`) passes on both, which is correct
— it is an over-fire guard on the new branch, not an assertion of new behaviour. No error
found in the author's own audit.

Two items below were **not** changed. Both were reproduced by executing the real module.

### 1. The card never reads `dirtyTreeTruncated`, so a very dirty tree renders as exactly 20 files

`captureGitProvenance` records at most `MAX_DIRTY_PATHS` (20) paths and sets
`dirtyTreeTruncated: true` when it dropped some (`git-provenance.ts:594-595`).
`buildProvenance` persists the flag (`safe-utils.ts:1501`) and the signing prompt renders it
as `20+ dirty` (`provenance-display.ts:155`, `formatWorkingTree` line 65).

`ICardProposal` has no such field (`proposal-card.ts:43-58`) and `render-proposal-card.ts:121-130`
does not map it. With one dirty row the card therefore receives exactly 20 paths, and
`summarise(paths, 20)` takes the `items.length <= limit` branch — so **no overflow marker is
emitted at all**:

```text
*Working tree:* dirty — <20 paths>
```

A tree with 500 dirty files and a tree with exactly 20 render identically. This is the same
understating-a-warning class the PR has already closed twice by other routes (round 3's
per-network union, round 5's shape check), and it is the one remaining route by which the
working-tree line can read as less alarming than the truth. It predates `8e542954a` — it has
been there since `1ed285540` — but it is live.

Not auto-fixed: it changes the text a signer reads. The fix is mechanical —
`dirtyTreeTruncated?: unknown` on `ICardProposal`, map it in `render-proposal-card.ts`, and
force the `…and N more` branch (or append `+`) when any reported row carries it.

### 2. Absent provenance reads as a clean tree on the card, where the signing prompt refuses to

`provenance-display.ts:130-139` treats a provenance block whose `dirtyTreeScoped` is not an
array — **`undefined` and `null` included** — as `'unreadable'`, and paints it yellow
`UNKNOWN (unreadable)`. The comment at `provenance-display.ts:69-73` states the rule
explicitly ("only a measured empty array with no capture errors may be painted green"), and
there is a test for exactly this: *"does not throw when dirtyTreeScoped is missing, and does
not call it clean"* (`provenance-display.test.ts:208`).

`proposal-card.ts:286-288` now treats `undefined` and `null` as absent, emitting no
working-tree line — which reads as clean. And `render-proposal-card.ts:129`
(`row.provenance?.dirtyTreeScoped`) collapses two different facts into the same `undefined`:
*no provenance block at all* and *a block whose dirty-tree field is missing*. The prompt
distinguishes them (`formatProvenanceLines` returns "not recorded (proposal predates
provenance capture)" for the first); the card has no equivalent line for either.

On the direct question — `null` vs `undefined` — the repo has already ruled, in the sibling
renderer, that neither is clean. Treating them as absent is defensible for a card that is
advisory while the prompt is authoritative, but it should be a stated decision rather than a
divergence, because the card is what people read in Slack and the prompt is what they read
after they have already decided to sign.

More than one valid fix (thread a provenance-present flag through, or render "provenance not
recorded" whenever every provenance field is absent), so not auto-fixed.

### Lower-confidence — human judgment, not acted on

- A `usable` row whose paths all sanitize away contributes no `; some paths could not be
  read` note when another row has real paths: `['']` + `['src/Real.sol']` renders
  `dirty — src/Real.sol`, with no hint that one row's paths were unrenderable. `shapeNote`
  keys off `unreadable` only. Reachable only via a hand-edited row.
- `reported.map((p) => escape(p.network))` is not deduped, unlike `nonHuman`, so two rows
  with the same network render `(on dup, dup)`. Unreachable in production —
  `render-proposal-card.ts:75-82` dedupes and emits one row per network.
- `summarise(items, limit)` with `limit <= 0` would emit a leading `", …and N more"`. Both
  call sites pass `4` and `20`; the JSDoc does not state the precondition.
- An item containing `, ` is indistinguishable from two items (`(on a, b)` for one network
  literally named `a, b`). Unreachable: network names arrive from a comma-split.
- `_none given_` is forgeable, since `_` is deliberately not stripped. Forging *absence*
  gains an attacker nothing.
- The advisory `checkSummary` line is not qualified with its source network. Consistent with
  `Reason`/`Commit`/`PR`, all row-zero with the shared divergence warning — and unreachable
  in production anyway (see the PR body).
- The working-tree line is now the card's largest single consumer: `MAX_DIRTY_PATHS` (20)
  × `MAX_FIELD_CHARS` (200) measured **4,078 characters** of the 8,000 budget in a
  40-row worst case, with `*Proposed by:*` at 1,038. Bounded, and the review commands and
  truncation notice both survived with no unpaired backtick — but the two constants multiply.
  Pre-existing, not introduced by `8e542954a`.

### Verdict

One substantive item (#1) and one design decision (#2). Everything else standing from
rounds 3–5 is cosmetic or unreachable in production.

0xDEnYO:
## `/gate-review` round 6 — residual findings

Gate on `7a04ab927`. Auto-fixed and pushed as `b31420eca` (tests + comments only, no
renderer logic). Everything below is **escalated — not changed** and needs a decision
before this goes ready.

`bun test script/` 1593 pass / 0 fail with ambient config and with `--env-file=/dev/null`.
`tsc-files --noEmit`, `eslint`, `bash -n`, `markdownlint-cli2` all clean. `git ls-files |
grep typechain` empty.

### The round's headline: the state word was unasserted (fixed)

Mutating `proposal-card.ts` `const state = … : 'not captured'` to `'clean'` left the suite
at **55 pass / 0 fail**. Both `/not captured/` assertions were satisfied by the *detail*
fragment `'paths not captured'`, so two describe blocks named "is not clean" and "is not a
clean one" passed against a card reading `*Working tree:* clean — paths not captured`.
The check the commit exists to install could not fire. The same mutation now fails 5.

Of the five new assertions, exactly one (`says so when capture truncated the path list`)
died on a real behavioural regression; three were parameterisations of one wording-only
regex. The "5 fail against `7a04ab927^`" count is 5 *test cases* over 3 distinct
assertions. Post-fix: 6 fail against `7a04ab927^`, and all three mutations die
(state word → 5 fail, drop the truncated branch → 1, suppress the uncaptured line → 5).

### E1 — critical, carryover from round 4: a failed `git status` probe reads as clean

A row whose porcelain probe failed carries `dirtyTreeScoped: []` **plus** `captureErrors`,
and the card renders **no `*Working tree:*` line at all**.

- `git-provenance.ts:385` returns `{ paths: [], truncated: false }` on `porcelain === null`;
  `runCommand` (`:283-289`, not `quiet`) has already pushed the failure into `ctx.errors`,
  so `captureGitProvenance` (`:594-600`) writes both fields.
- `safe-utils.ts:1510` persists `captureErrors`.
- `render-proposal-card.ts:36-45,121-131` — `IRow.provenance` has no `captureErrors` member
  and the mapper never reads it; `ICardProposal` has no such field. It is structurally
  invisible to the card.
- `proposal-card.ts:291` — `[]` is an array, so the row is in neither `dirty` nor
  `uncaptured`, and the `if` is false. Verified: `[{ network: 'mainnet',
  dirtyTreeScoped: [] }]` renders no line.
- The signing prompt for the identical row: `Working tree: UNKNOWN (capture incomplete)`
  plus `Capture: ⚠ incomplete (1): …`.

The two likeliest probe failures are the 5 s `GIT_TIMEOUT_MS` and the 8 MB
`MAX_BUFFER_BYTES` (`git-provenance.ts:79,82`) — both correlate with a *very* dirty tree.
The card says clean exactly when the tree is dirtiest.

Escalated rather than fixed: it adds a field to the Mongo read surface and a signer-facing
state word. It is also the item that decides whether "the two renderers of one field no
longer disagree" is true — on the shape axis it now is; on the capture-errors axis, the
axis `provenance-display.ts:69-72` was written for, it is not.

### E2 — major: a mixed dirty + uncaptured card drops the `(on …)` qualifier entirely

`proposal-card.ts:302-310` — `reported = [...dirty, ...uncaptured]`, and `where` is
suppressed when `reported.length === proposals.length`. Verified renderings:

| rows | rendered |
|---|---|
| dirty(mainnet `src/A.sol`) + uncaptured(arbitrum) | `dirty — src/A.sol; some rows captured no path list` |
| the same two + clean(optimism) | `dirty (on mainnet, arbitrum) — src/A.sol; …` |
| dirty(mainnet) + 4 clean | `dirty (on mainnet) — src/A.sol` ✅ |

Row 1 names no network, so one path reads as the complete inventory across both — and
arbitrum's tree was never measured. Row 2 names arbitrum as dirty while showing only
mainnet's path.

Production-reachable, not hypothetical: `render-proposal-card.ts:95-113` takes the newest
*pending* row per network across all history, and its own comment says "A network can carry
an older unsigned proposal". Mixed-vintage provenance shapes in one card are the designed
case. This is `c4108eb40`'s class — a warning about one network read as a claim about all —
arriving from the other direction, and no test in `7a04ab927` uses a mixed set.

### E3 — major: the truncated branch discards the known overflow count

`proposal-card.ts:312-318`. Verified, 40 unique paths across two rows:

- one row flagged `dirtyTreeTruncated` → `…src/F19.sol, and more that capture stopped counting`
- the same 40 paths, no flag → `…src/F19.sol, …and 20 more`

The strictly worse situation renders the vaguer, smaller-sounding warning.
`provenance-display.ts:63-66` keeps both facts (`⚠ 40+ dirty`).

Bounding, for the record: the branch **is** bounded (20 × `MAX_FIELD_CHARS` ≈ 4 000, the
same ceiling `summarise` has), and both branches show `min(paths.length, 20)` paths, so they
cannot disagree on the count shown — only on how the remainder is described.

### E4 — major, judgment: the state word was downgraded for rows carrying evidence of dirt

`dirtyTreeScoped: 'src/A.sol'` rendered `dirty — paths could not be read` before
`7a04ab927` and renders `not captured — paths not captured` now. "not captured" reads as a
tooling gap; "dirty" reads as a code-integrity problem. Defensible — it matches
`provenance-display.ts`'s yellow `UNKNOWN (unreadable)` — but it is the one place the
alignment was reached by moving the *card down* to the prompt's severity rather than the
prompt up, inside a commit whose subject is not understating a warning. Your call.

### E5 — minor, low reachability, shared with the prompt

- `proposal-card.ts:301` — `dirty.some((p) => p.dirtyTreeTruncated === true)`; a row in
  `uncaptured` cannot contribute its flag. `proposals.some` would. Verified: dirty(20 paths,
  no flag) + uncaptured(flag) renders the exact list with no "more".
- `dirtyTreeScoped: []` + `dirtyTreeTruncated: true` renders no line, because `truncated` is
  computed inside the `if`. Verified. `provenance-display.ts` paints the same row green
  `clean`, so it is a shared hole, and its only routes are the `sanitizeOverride` test seam
  (`safe-utils.ts:1498-1503`, "Production code never sets it") or a hand-edited row.
- `proposal-card.ts:319-322` — the `; some rows captured no path list` note is suppressed
  when `paths.length === 0`, so `[' ', '  ']` + an uncaptured row renders
  `dirty — paths not captured` and the uncaptured network vanishes from the line.

### Lower confidence — human judgment, not acted on

- **Round 3 E2, never resolved.** Only `prUrl` is guarded against Slack auto-linking
  (`renderLink`). A bare `https://…` in `reason`, `proposerHandle`, `checkSummary` or a
  dirty path renders unguarded and Slack links it on the screen a signer reads before
  approving. Not re-raised in rounds 4–6, so it reads as dropped rather than deferred.
- `render-proposal-card.ts` has **no test file** — 196 lines of Mongo read plus the
  `dirtyTreeTruncated` mapping this commit added. A typo fails `tsc`; an omission is caught
  by nothing.
- `proposal-card.test.ts` has no module header; `.agents/rules/200-typescript.md` requires
  one for every new TS file. Sibling test files also lack one.
- `docs/MultisigSigningProcess.md:133-136` lists the card's fields and omits the
  `*Working tree:*` line — the field this round is about.
- `dirtyTreeTruncated` as a non-boolean truthy (`'true'`, `1`) is dropped by `=== true`.
  Consistent with `provenance-display.ts:155`, and only ever written as `true` by
  `git-provenance.ts:595` / `safe-utils.ts:1501`. Sensible as is; noted only because the
  card's stated contract is to be total over any document shape.

### Confirmed clean

No stale references to the removed wording or identifiers (`paths could not be read`,
`some paths could not be read`, `shapeNote`, `dirtyRows`, `usable`) anywhere in the repo.
The truncation/cut-budget tests are byte-for-byte unaffected by the fixture change, because
`dirtyTreeScoped: []` restores the pre-fix absent-line rendering. CodeRabbit cannot review
this PR (org credit limit), so this gate is the only review before a human.

0xDEnYO:
## `/gate-review` round 7 — verdict: ready to flip

Gate on `1834ddc65`. Auto-fixed and pushed as `9f3e0721c` + `5ed8a6d38`. **All of rounds
3–6 is closed.** Nothing below blocks: what remains is one doc item outside this PR's scope,
two hand-edit-only holes shared verbatim with already-merged code, one over-warning that is
correct by the model this PR deliberately aligns to, and one typing dedupe.

`bun test script/` **1606 pass / 0 fail**, identical with ambient config and with
`bun --env-file=/dev/null`. `tsc-files --noEmit`, `eslint`, `bash -n
script/multiNetworkExecution.sh`, `markdownlint-cli2` all clean. `git ls-files | grep
typechain` empty.

### Pre-fix audit — both commits, my numbers

**`c9a76a866` → 5 fail against `c9a76a866^`**, not 4. Your four new assertions all fail, and
the fifth is the pre-existing `reports an absent dirtyTreeScoped rather than rendering
nothing`, whose expectation this commit tightened from `'not captured — paths not captured'`
to the exact `'not captured'`. The sixth new case (`stays silent for a measured empty array
with no capture errors`) correctly passes on both — it is an over-fire guard, not an
assertion of new behaviour. Your count is right on its own terms.

**`1834ddc65` → 9 of 10 independent mutants killed.** I did not reuse your six. Each was
applied to `proposal-selection.ts` and run against the 8 tests:

| mutant | result |
|---|---|
| drop the first-wins guard, so the last row per network wins | 1 fail |
| iterate query order instead of caller order | 2 fail |
| `settled` = every network without a pending row | 1 fail |
| `unaccounted` predicate inverted | 3 fail |
| `captureErrors` dropped from the mapping | 1 fail |
| `dirtyTreeTruncated` dropped from the mapping | 1 fail |
| `dirtyTreeScoped` dropped from the mapping | 1 fail |
| `withoutPending` predicate inverted | 3 fail |
| the `row !== undefined` filter removed | 3 fail |
| `String(row.network)` → `row.network as string` in the map key | **survived** |

The survivor is an equivalent mutant, not a gap: `IRow.network` is `string` and the Mongo
`$in` matched it, so the coercion is a no-op in production. Not worth a fixture.

### B — can any working-tree combination read as less alarming than the truth?

**No, on any shape `captureGitProvenance` can produce.** 20 combinations executed against
the real module; the four-state line is correct in every one:

```text
all four states at once   dirty on mainnet — src/A.sol; capture incomplete on arbitrum; not captured on base
dirty + errors, one row   dirty — src/A.sol
captureErrors: [{…},null,7]  capture incomplete
captureErrors: ['', '  ']    capture incomplete
dirtyTreeScoped: ' M src/A.sol'   not captured
dirtyTreeScoped: null            not captured
paths that all sanitize away      dirty — paths unusable
41 rows over all four states      dirty on net0, net4, net8, net12, …and 7 more — …; capture incomplete on net1, … ; not captured on net2, …
```

`on()`'s suppression condition is right, and it has to be "covers every **row**" rather than
"every reported row" — the latter is round 6's E2 exactly: one dirty plus one clean would
suppress the qualifier and the single path would read as the whole card. Verified per state:
with any clean sibling present the network is always named; suppression fires only when a
state is universal, and since the three buckets are disjoint at most one can be. The 41-row
card is 535 characters, and 3,344 with every field at its 200-char ceiling — the 8,000 cap is
not in play.

Round 6's **E5(c) is fixed** (`dirty on mainnet — paths unusable; not captured on arbitrum`
— the uncaptured network no longer vanishes). **E3 is fixed**: 40 paths with the flag now
render `…and 20 more, and capture stopped counting before the end`.

### C — the two `withoutPending` computations cannot disagree

They are the same predicate. The reader asks `!rows.some(r => String(r.network) === n)`;
`selectProposals` asks `!newest.has(n)`, and `newest`'s key set is exactly
`{String(row.network)}` over every pending row, unfiltered. Executed over 20,000 random
cases — network values drawn from strings, mixed case, `123`, `null`, `undefined`, an object
with a `toString`, and an array — **0 mismatches**, no duplicate coverage, and
`proposals ∪ settled ∪ unaccounted` partitioned `networks` exactly every time.

Worth recording why a disagreement could not have been dangerous either: the reader's set
only bounds the `distinct` query. Too large → the query asks about extra networks, harmless.
Too small → a network never gets queried, `known.has` is false, and it lands in
`unaccounted`, which fires the shortfall warning. Both directions fail loud.

---

## Escalated — not changed

### 1. §9 still lists shipped work under "Nothing below exists in the repo today"

`docs/MultisigSigningProcess.md:291` — "**Provenance on proposals** — attach human identity,
git commit/branch, and a PR link/description to each proposal, shown at signing" sits under a
heading that asserts nothing in the list exists. It shipped in EXSC-693
(`provenance-display.ts`), and this PR is the second consumer of it. Not auto-fixed because
deciding which of the five bullets are now shipped (integrity asserts, executability
simulation, bytecode↔audit attestation) is a doc pass of its own, not this PR's.

### 2. A non-array `captureErrors` reads as clean — hand-edit only, shared with the prompt

`proposal-card.ts:301-307` requires `Array.isArray(p.captureErrors)`, so a row with
`dirtyTreeScoped: []` and `captureErrors: 'git status timed out'` matches **no bucket** and
renders **no working-tree line at all**. Executed; same for an object or a number.

This is E1's shape, and it is genuinely low-severity: `captureGitProvenance` only ever writes
`errors.map(sanitizeProvenanceText)`, i.e. `string[]`, so the only routes are a hand-edited
row or a future schema change. `provenance-display.ts` has the identical hole —
`toSanitizedList` returns `[]` for a non-array, and the row is then painted green `clean`.

Escalated rather than fixed because the fix has to land in **both** renderers in the same
change or they diverge again on the axis this PR spent three rounds aligning. The shape that
closes the class rather than the instance: define `clean` positively and put the residue in a
bucket, so no row can fall through to silence.

### 3. A dirty row carrying `captureErrors` reports only "dirty"

The signing prompt renders both a dirty line and `Capture: ⚠ incomplete (N): …`
(`provenance-display.ts:178`); the card renders only `dirty — src/A.sol`. Not under-alarming
on any capture-produced shape — a non-empty path list means `status --porcelain` succeeded,
so the list is complete and the errors are about other probes, whose sentinels (`*Commit:*
UNKNOWN`) the card does show. Recording it because the card is the one surface that never
says "capture was partial" when the tree is also dirty.

### 4. A measured-clean tree reads "capture incomplete" when an unrelated probe failed

`captureErrors` is **one shared collector** for `rev-parse --show-toplevel`, `rev-parse
HEAD`, `rev-parse --abbrev-ref HEAD`, `status --porcelain` and `branch --remotes --contains`
(`git-provenance.ts:283-289`). Only the fourth says anything about the working tree.
Executed — a row whose tree was measured clean but whose remote-containment probe failed:

```text
*Commit:* UNKNOWN
*Working tree:* capture incomplete
```

Fail-loud, and it matches `provenance-display.ts` exactly, which is the alignment this PR
wanted — so it is not a divergence and not a bug. Raising it only because the reachable
version (a shallow CI checkout, where `branch --remotes --contains <GITHUB_SHA>` can fail on
every run) would put "capture incomplete" on the safety field of every card, and alarm
fatigue there is the one outcome that would undo six rounds of work. Closing it properly
means attributing errors per probe in `git-provenance.ts`, which is its own change.

### 5. `IRow` duplicates `IStoredProposal`

`render-proposal-card.ts:34-49` restates `proposal-selection.ts:14-30` field for field, plus
`status`. No field is currently lost — all of them flow at runtime — but this is the exact
place three fields have been dropped already, and the duplication puts the drift risk back.

Verified as a clean replacement (`tsc-files --noEmit` exit 0):

```ts
interface IRow extends IStoredProposal {
  status?: SafeTxStatus
}
```

Escalated because it retypes the Mongo filter's `network` from `string` to `unknown`, which
weakens checking on the query, so there is more than one defensible shape (the alternative is
narrowing `IStoredProposal.network` to `string`).

### Closed as unreachable

- Round 6 E5(a) — `dirty.some(p => p.dirtyTreeTruncated === true)` cannot see a flag on a
  non-dirty row. `scopedDirtyTree` sets `truncated` only when `unique.length > 20`, and then
  records exactly 20 paths, so a non-dirty row never carries a real flag. Both mixed cases
  executed: the other row's state is named either way, so nothing is hidden.
- Round 6 E5(b) — `dirtyTreeScoped: []` with `dirtyTreeTruncated: true` renders no line.
  Same proof; hand-edit only, and `provenance-display.ts` paints the same row green.
- Round 3 E2 (bare URL in free text) and round 6 E4 (the state word for an unreadable field)
  are now **decided in writing** in the PR body rather than open.

### Auto-fixed

- `9f3e0721c` — module headers on `proposal-card.test.ts` and `proposal-selection.test.ts`.
  `.agents/rules/200-typescript.md:107` requires one on every new TS file, and the nearest
  siblings (`provenance-display.test.ts`, `safe-utils.test.ts`) both carry one, so round 6's
  "sibling test files also lack one" was wrong. Also dropped the call-site comment above
  `notifyProposalsCreatedToSlack`, which still described the count-only summary that is now
  only the fallback, and added the working-tree state to the card's field list in
  `docs/MultisigSigningProcess.md`.
- `5ed8a6d38` — §4.2 said the document carries "no description, PR link, git commit, or
  human identity" one sentence before saying the card renders exactly those from it.
  `ISafeTxDocument.provenance` has held them since EXSC-693. This closes the **unactioned
  CodeRabbit inline comment from 2026-09-01T18:36Z**, which landed three minutes before round
  6's push and shipped through two commits.
- PR body: evidence line was 1589 pass / 51 new; it is 1606 / 68.

CodeRabbit cannot review this PR (org credit limit), so this gate is the only review before a
human.

gvladika (approved):

PR #2300, opened 2026-09-01
fix(safe): validate --accountIndex on the remaining Ledger-signing scripts (EXSC-874 D11)

0xDEnYO:

EXSC-874 (D11 follow-up)

**Stacked on #2297** — `parseAccountIndex`, `resolveSafeSigningOptions` and `cli-flags.ts` only exist on that branch. Base is `fix/exsc-874-ledger-on-safe-script-helpers`; retarget to `main` once #2297 merges.

#2297 fixed `--accountIndex` in `cleanUpProdDiamond.ts` and `propose-to-safe.ts`. The same `args.accountIndex ? Number(args.accountIndex) : 0` was still in the other three scripts that offer the flag. Verified against real citty 0.1.6 with each script's own arg schema:

| typed | citty yields | old code signed from |
|---|---|---|
| `--accountIndex "$UNSET_VAR"` | `''` | account 0 |
| bare `--accountIndex` / `--accountIndex=` | `''` | account 0 |
| bare `--account-index` / `--account-index=` | boolean `true` | account 1 |
| `--accountIndex 3.7` | `'3.7'` | account 3 (BIP32 truncates) |
| `--accountIndex -1` | `'-1'` | account 2147483647 (wraps) |
| `--accountIndex abc` | `'abc'` | `m/44'/60'/0/0` — the `NaN` segment is dropped entirely |

Every one derives a different, valid-looking address with no error at any layer, which is why the refusal has to happen at the flag.

**What each script got**

- All three: `readValueFlag` + `parseAccountIndex`, the pair the two already-fixed scripts use. The raw string is forwarded unconverted so the guard sees what the operator typed — `Number('')` is 0, so converting first destroys the evidence.
- `confirm-safe-tx.ts` and `add-safe-owners-and-threshold.ts` additionally refuse an `--accountIndex` the Ledger path would ignore. `getLedgerAccount` reads it only on the Ledger Live branch ([`ledger.ts:54-60`](https://github.com/lifinance/contracts/blob/main/script/deploy/safe/ledger.ts#L54-L60)), so `--accountIndex 3` without `--ledgerLive` used to sign from account 0 silently. Same guard, same wording as `resolveSafeSigningOptions`.
- Those two also read `--ledgerLive` via `readBooleanFlag`. citty passes the kebab spelling's value through raw, so `--ledger-live=false` arrived as the truthy string `'false'` — an operator turning Ledger Live off got it on. The new `accountIndex` guard depends on `ledgerLive` being right, so it could not be left as-is.

**Not routed through `resolveSafeSigningOptions`**, despite it being the natural home: it resolves the signing key as well as the flags. `confirm-safe-tx.ts` picks its key through an interactive prompt, and `add-safe-owners-and-threshold.ts` reads it through `getPrivateKey`, which strips a leading `0x` — `SafeClient.init` hands the raw value to `new TronWalletClient(privateKey)`, which passes it to TronWeb unprefixed. Swapping in the helper's plain env read would have broken Tron signing on a path this PR has no reason to touch.

`--derivationPath` carried the same defect in those two scripts and is fixed alongside: read from the parsed args, an empty value reached `splitPath('')`, which returns an empty BIP32 path and derives an address rather than erroring.

**Left alone in `ledger-flex-calibrate.ts`**: its `--ledgerLive` defaults to `true`, so `--accountIndex` is honoured and needs no pairing guard, and it has no `--derivationPath`. Its `--ledger-live=false` has the same truthy-string flaw, but `readBooleanFlag` cannot express "absent means true", so the fix is a decision rather than a typo — escalated in the review-gate comment below, along with three related items. Note it does sign on the device (throwaway calibration payloads, nothing broadcast), so this is not purely cosmetic.

**Behaviour change for operators**: `--accountIndex <n>` without `--ledgerLive` now errors on the two Safe scripts instead of silently using account 0. Anyone in that habit was already signing from the wrong account.

**Verification** — every case below run against the real scripts, not just the units:

```text
ledger-flex-calibrate --accountIndex=      → accountIndex must be a non-negative integer, got ''
ledger-flex-calibrate --account-index      → --accountIndex needs a value, e.g. --accountIndex <value>.
add-safe-owners      --accountIndex=3.7    → accountIndex must be a non-negative integer, got '3.7'
add-safe-owners      --accountIndex=3      → 'accountIndex' only selects an account on the Ledger Live path.
confirm-safe-tx --accountIndex=3 --ledger-live=false → same refusal (was: truthy 'false')
confirm-safe-tx --accountIndex=3 --ledgerLive        → Using derivation path: m/44'/60'/3'/0/0
```

`ledger-flags.test.ts` covers `parseAccountIndex` directly, the `readValueFlag`+`parseAccountIndex` composition over real argv, and source-level invariants over every script that declares the flag. The script list is **discovered**, not listed, so a new script with the same shape is caught.

The review gate found the first version of those source assertions was near-decorative — they matched `Number(args.accountIndex)` literally, so naming the raw value and converting it a line later reinstated the entire bug with the suite green. They now match any numeric conversion of an account-index-named identifier. Every assertion was falsified against a real regression before being kept:

```text
critical-rename    -> 2 fail     bracket-access -> 3 fail
optional-chain     -> 3 fail     destructured   -> 2 fail
parseInt           -> 3 fail     pre-PR sources -> 6 fail
```


0xDEnYO:
## Review gate — escalated, not fixed

Auto-fixed in 7b7d249e1: the test's source-scan blind spot, `--derivationPath` empty-value refusal in the two Safe scripts, and an over-broad docs row. These four were escalated instead.

**E1 — `ledger-flex-calibrate.ts` ignores `--ledger-live=false` and `--no-ledger-live`.** Its `ledgerLive` is `type: 'boolean', default: true`, so citty applies the default to the camelCase key even when the kebab spelling was typed, and the kebab value comes through as the truthy string `'false'`. `--ledger-live=false --accountIndex 3` therefore takes the Ledger Live path and signs from account 3 while the operator believes they turned it off. `readBooleanFlag` cannot express "absent means true", so the fix is a decision, not a typo: add a default-aware reader to `cli-flags.ts`, or drop the citty default. The camel spellings (`--ledgerLive=false`, `--no-ledgerLive`) work correctly today.

**E2 — the same script has no `--accountIndex` without `--ledgerLive` guard.** Adding it depends on E1: the guard is only sound once `ledgerLive` is trustworthy.

**E3 — `--ledger` itself is still read from the parsed args** in `confirm-safe-tx.ts` and `add-safe-owners-and-threshold.ts` (`args.ledger ?? true`), so `--ledger=no` means "use Ledger". Same `default: true` problem as E1, and fail-safe in direction — it errs toward the hardware wallet — but inconsistent with the rationale this PR states.

**E4 — the argv-reading block is now duplicated five times**, and the `'accountIndex' only selects an account on the Ledger Live path` message three times (`safe-utils.ts` plus the two inline copies). This is #2297's escalated item E9, grown rather than shrunk. A shared `resolveLedgerFlags()` would fold all five, but it is a scope decision — the two inline scripts deliberately do not route through `resolveSafeSigningOptions`, because it also resolves the signing key and would drop `getPrivateKey`'s `0x` stripping that `TronWalletClient` depends on.

- Flag validation is not fail-fast: `add-safe-owners-and-threshold.ts --check --accountIndex=abc` runs the whole read-only audit before reaching the parser, and `ledger-flex-calibrate.ts --list --accountIndex=abc` prints the table. Neither signs, so it is cosmetic.
- When `--derivationPath` and `--accountIndex` are given together, the refusal says "Add --ledgerLive", which would then hit the mutual-exclusion throw. Only "drop --accountIndex" is actionable there.

0xDEnYO:
Heads-up: this now conflicts with its own base, and part of it is duplicated. My fault, not yours — I changed `confirm-safe-tx.ts` in the parent branch (#2297) *after* this branched, so we fixed the same thing twice.

**What overlaps.** #2297 now routes `confirm-safe-tx.ts` through `readValueFlag` + `parseAccountIndex`, and additionally fixes a desync this PR does not touch: the compat guard and the operator log were still reading `args.ledgerLive` from citty, which parses `--ledger-live=false` to the truthy string `'false'` — so the console announced a Ledger Live derivation while the client derived from the default path. That is `8ca9fd742`.

**What is genuinely yours and worth keeping:**

- `add-safe-owners-and-threshold.ts` (+56) — #2297 does not touch this script at all.
- `ledger-flex-calibrate.ts` (+12) — same.

**Suggested disposition** (a human call, so I have not touched this branch): rebase onto the current `fix/exsc-874-ledger-on-safe-script-helpers` and drop the `confirm-safe-tx.ts` hunks, leaving the two scripts above. #2297 is ready, CI green, and posted for review, so the base is stable now.

One note on `ledger-flags.test.ts`, offered rather than asserted: it validates by reading source text and asserting on its content. That shape passes against the obvious rewrite of the same bug — a script that moves its validation, or spells the call differently, keeps the test green while reintroducing the defect. #2297 took the other route: `readBooleanFlag`/`readValueFlag` are pure functions over an `argv` array, so the behaviour is asserted directly rather than the source's shape, and the call sites are covered by spawning the real command. Reusing those two readers here would cover `add-safe-owners-and-threshold.ts` and `ledger-flex-calibrate.ts` with behavioural tests instead.

Also worth knowing before you re-run anything: **Bun auto-loads `.env` into the test process, and `DOTENV_CONFIG_PATH` does not stop it.** Two of my placement tests passed locally and failed in CI for exactly that reason. Worth checking with the file hidden.

0xDEnYO:
Follow-up on the review-gate HIGH item: the script-discovery regex in `ledger-flags.test.ts` matched `type: 'string'`, which silently skipped a declaration spelling the type differently, quoting it differently, or carrying a `}` before it — a `${...}` in the description is enough, and one already appears at `ledger-flex-calibrate.ts:197`. A skipped script is a pass, not a failure. Now matches any `accountIndex: {` declaration (f2dce0bc4).

Falsified: patching `confirm-safe-tx.ts` to `type: 'number'` or to a template-literal description, with the bug reinstated, went 3 fail under the new regex and would have been silently skipped by the old one. The remaining known gap — an arg declared through a shared spread — is noted in the code; the shared object itself becomes the file to assert on.

The four escalated items above are unchanged.

0xDEnYO:
Closing as superseded by #2297.

`main` now carries `script/deploy/safe/cli-flags.ts` and `script/deploy/safe/ledger-guards.ts` from #2297, and main's `cli-flags.ts` is a **superset** of the one here (it adds `IBooleanFlagOptions` on top of `readBooleanFlag` / `readValueFlag`). This branch's copy is the older, narrower ancestor, so a merge produces an add/add conflict on the file this PR exists to introduce.

Merging `main` in costs 9 conflicts — `cli-flags.ts`, `cli-flags.test.ts`, `confirm-safe-tx.ts`, `propose-to-safe.ts`, `safe-utils.ts`, `safe-utils.test.ts`, `safeScriptHelpers.ts`, `cleanUpProdDiamond.ts`, `MultisigSigningProcess.md` — every one of them a re-decision against code that already merged. That is more work than re-deriving the residue, and riskier to review.

**Residue worth keeping** (tracked on EXSC-874, still In Progress): `--accountIndex` validation is still missing on `add-safe-owners-and-threshold.ts`, `ledger-flex-calibrate.ts`, `cleanUpProdDiamond.ts` and `script/safe/safeScriptHelpers.ts`. Re-cut against main's `cli-flags.ts` it is a small, self-contained PR.

Branch kept, nothing deleted.

PR #2302, opened 2026-09-02
fix(tron): record the constructor arguments a contract was deployed with (EXSC-907)

0xDEnYO:

Part of [EXSC-907](https://linear.app/lifi-linear/issue/EXSC-907/wp-22-record-hardening) (ref EXSC-907) — the `constructorArgs` leg of WP-2.2. Deliberately does not close the ticket; the other four items are listed at the bottom with why each is separate.

## The defect

Seven Tron periphery contracts are deployed with real constructor arguments and recorded with the literal `'0x'`:

| Contract | Deployed with | Recorded as |
|---|---|---|
| `ERC20Proxy` | 2 addresses | `'0x'` |
| `Executor` | 2 addresses | `'0x'` |
| `FeeCollector` | 1 address | `'0x'` |
| `FeeForwarder` | 1 address | `'0x'` |
| `TokenWrapper` | 3 addresses | `'0x'` |
| `OutputValidator` | 1 address | `'0x'` |
| `LiFiTimelockController` | `uint256, address[], address[], address, address, address` | `'0x'` |

A record is what a verifier rebuilds creation code from, so each of these describes a deployment that did not happen. Constructor arguments are also what set immutables, so nothing downstream can check those either. This is the `constructorArgs "0x"` item in WP-2.2's scope, and it is live on real deployments rather than latent.

The eighth site did encode, but from types **guessed off JavaScript runtime values** — a string starting with `0x` or `T` became `address`, any other string `string`, a number `uint256`, anything else `bytes`. Every current call site passes hex addresses, so today the guess is right; a `uint256` gas value read from config as a string would encode as a dynamic string at a different offset. And on failure it caught the error and concatenated raw hex and UTF-8 bytes — not ABI encoding at all — then wrote that to the record as if it were.

## The fix

Types come from the contract's own ABI (`constructorInputTypes`), tuples expanded to `(address,uint128)` form. Every failure throws: an unreadable ABI, an arity mismatch, an encoder error, a non-hex result. A deploy that stops is recoverable; a record holding the wrong arguments is not, because nothing downstream can tell.

`recordTronDeployment` is now the only path from a Tron deployment to the log. That matters more than fixing seven call sites: the sites could pass a raw string, which is how one literal spread to seven places. They now pass the artifact and the values, and the funnel decides what gets recorded. `logDeployment` is no longer imported by the periphery script, so the old shape is gone rather than merely unused.

`assertRecordedArgsMatchAbi` then refuses either direction of disagreement — a record claiming none for a contract that takes some, and a record carrying arguments a contract cannot take (which means it is for a different contract).

## Evidence

**The guard against real ABIs.** All seven fire; genuine no-arg facets stay silent; and the same seven go silent once real arguments are recorded:

```text
MUST FIRE — real ABIs, recorded as '0x' by deploy-and-register-periphery.ts:
  ERC20Proxy               FIRES   ERC20Proxy takes 2 constructor arguments (address, address) but the record says it has none…
  Executor                 FIRES   Executor takes 2 constructor arguments (address, address)…
  FeeCollector             FIRES   FeeCollector takes 1 constructor arguments (address)…
  FeeForwarder             FIRES   FeeForwarder takes 1 constructor arguments (address)…
  TokenWrapper             FIRES   TokenWrapper takes 3 constructor arguments (address, address, address)…
  OutputValidator          FIRES   OutputValidator takes 1 constructor arguments (address)…
  LiFiTimelockController   FIRES   LiFiTimelockController takes 6 constructor arguments (uint256, address[], address[], address, address, address)…

MUST STAY SILENT — contracts that genuinely take no arguments:
  AccessManagerFacet       silent  (0 declared: none)
  PeripheryRegistryFacet   silent  (0 declared: none)

MUST STAY SILENT — the same contracts once real encoded args are recorded:
  ERC20Proxy               silent  (2 declared: address, address)
  Executor                 silent  (2 declared: address, address)
  FeeCollector             silent  (1 declared: address)
```

**Placement.** The periphery script no longer imports `logDeployment`, and `recordTronDeployment` takes an artifact and values, so there is no parameter a raw string could be passed through and a site left behind would be a TS2304. Correcting an earlier version of this description: I wrote that `tsc` named all eight sites when the encoder signature changed. It did not. Only one of the eight called the encoder; the other seven passed the literal to `logDeployment`, whose signature this PR does not touch. They were found by reading the file. The compiler guarantees completeness from here, not discovery.

**30 unit tests, 15 mutations, all dying.** Including: accepting a no-args record for a contract with arguments, dropping the mirror check, returning `[]` instead of throwing on an unreadable ABI, allowing an arity mismatch, accepting a non-hex encoder result, not expanding tuples, and treating only `''` as an empty record.

`bun test script/` — 1568 pass, 0 fail, run with `.env` moved aside so the suite sees what CI sees.

## What the review round changed

The gate found the ordering wrong, and it inverted this PR's own argument. `recordTronDeployment` throws on an ABI or arity problem, and it ran **before** `saveContractAddress` — so a throw left a contract deployed on chain with its address never written to `deployments/<network>.json`. `checkExistingDeployment` reads only that file, with no on-chain fallback, so the next run deploys a duplicate. "A failed deploy is recoverable" is not true of a deploy whose address is lost. All eight sites and `deployContractWithLogging` now save the address first and record second.

Three smaller things from the same round: the record check tested emptiness and nothing else, so `not-hex`, `0` and a truncated `0x11…` were all accepted for a contract taking two addresses (it now requires hex and at least one 32-byte word per declared argument); the `tronUtils` encoder had no callers left once the recorder called the pure helper directly and it asserted nothing, so it was a funnel bypass waiting for a caller, now deleted; and three comments recounted the defect, which belongs in this description rather than in the code.

**One behavioural change worth naming:** encoding now uses the target network's TronWeb codec rather than always mainnet's. ABI encoding is network-independent and both Tron networks share the address prefix, so no output differs — but it is a change.

**Adjacent, pre-existing, not fixed here:** `deploy-core-facets.ts` passes one argument to `LiFiDiamond` (which declares two) and none to `LiFiIntentEscrowFacetV2` (which declares one). Both already fail before any TRX is spent — the devkit checks arity at energy-estimation time — so this PR neither fixes nor worsens them. Worth its own ticket.

## Round two of review: the check moved before the broadcast

Reordering save-before-record removed the duplicate-deployment risk but left the real problem — a bad ABI or arity still failed *after* TRX was spent, leaving a contract on chain that could not be recorded. Everything the check needs is pure (the artifact's ABI and the values), so it belongs before the deploy, where failing costs nothing.

`assertTronDeploymentRecordable` now runs immediately before `deployer.deployContract` at all eight periphery sites and in `deployContractWithLogging`. Ordering verified at every site rather than asserted:

```text
404-440:   PREFLIGHT -> DEPLOY -> SAVE -> RECORD  ok
515-549:   PREFLIGHT -> DEPLOY -> SAVE -> RECORD  ok
639-675:   PREFLIGHT -> DEPLOY -> SAVE -> RECORD  ok
746-782:   PREFLIGHT -> DEPLOY -> SAVE -> RECORD  ok
918-954:   PREFLIGHT -> DEPLOY -> SAVE -> RECORD  ok
1021-1057: PREFLIGHT -> DEPLOY -> SAVE -> RECORD  ok
1161-1197: PREFLIGHT -> DEPLOY -> SAVE -> RECORD  ok
1304-1332: PREFLIGHT -> DEPLOY -> SAVE -> RECORD  ok
```

What the preflight adds beyond the devkit: `TronContractDeployer` already checks arity at energy-estimation time, so an arity mismatch was stopped pre-broadcast. It does not cover an unreadable ABI, a non-hex encoder result, or the record-vs-ABI mirror check — those are what this closes. The save-before-record ordering stays as the backstop.

Also this round: `isHex` accepted odd-length payloads, so `0x0` passed both as an encoder result and as a record. Whole bytes are now required, with a whole-byte control case so a stricter version rejecting everything could not pass instead.

## Not in this PR

The other four WP-2.2 items, each separable:

- **`repo` field on `IDeploymentRecord`** and **the verifier's repo check** — worth landing together, since the field exists to be checked. The check also needs a decision first: WP-2.2's scope says "commit reachable from the declared repo's default branch", but F18/F25 and the D3 ruling establish that ancestry is meaningless under squash-merge (78 of 98 audit entries fail an ancestor test), which is why WP-0.3 was re-scoped to content equality. Presence-in-the-declared-repo is the check that survives that; I am not going to quietly implement an ancestry assertion the project already ruled against.
- **Refuse a dirty or unpushed tree** — `git-provenance.ts` already has `getScopedDirtyTree` (with F22's exclusions) and `isCommitOnRemote`, so this is a policy layer, not new plumbing. One catch to handle first: `isCommitOnRemote` reads local refs only, so a stale checkout reports `false` for a commit that is in fact pushed — fine for a recorded field, not fine as a refusal.
- **Store the verified codehash post-deploy** — needs the primitives in [#2301](https://github.com/lifinance/contracts/pull/2301), which is in review.

- [x] I have updated any required documentation — none: this corrects what a deployment record holds, with no user-visible behaviour change beyond a deploy now stopping rather than recording something false

### DIFF
--- a/script/deploy/tron/constructor-args.ts
+++ b/script/deploy/tron/constructor-args.ts
@@ -69,6 +69,7 @@ const canonicalType = (input: unknown, contractName: string): string => {
  * @param contractName - Named in every message; the caller always knows it.
  * @returns Canonical type strings in declaration order, empty when the contract
  * declares no constructor.
+ * @throws When the ABI, a constructor input, or a tuple components list cannot be read.
  */
 export const constructorInputTypes = (
   abi: unknown,
@@ -120,6 +121,19 @@ export const assertRecordedArgsMatchAbi = (
       } constructor arguments (${types.join(
         ', '
       )}) but the record says it has none. A verifier rebuilding creation code from this record would compute a different deployment.`
+    )
+
+  if (types.length > 0 && !isHex(recorded))
+    throw new Error(
+      `${contractName}'s recorded constructor arguments are not hex, so they are not what the constructor received.`
+    )
+
+  // Each static parameter occupies one 32-byte word, and a dynamic one occupies
+  // at least an offset word, so anything shorter is truncated.
+  const words = (recorded.trim().replace(/^0x/iu, '').length / 64) | 0
+  if (types.length > 0 && words < types.length)
+    throw new Error(
+      `${contractName} takes ${types.length} constructor arguments but the record holds only ${words} words of them.`
     )
 
   if (types.length === 0 && !empty)
--- a/script/deploy/tron/constructor-args.ts
+++ b/script/deploy/tron/constructor-args.ts
@@ -1,0 +1,182 @@
+/**
+ * Constructor arguments for a Tron deployment record.
+ *
+ * Import this wherever a deployment is recorded: the recorded arguments are what
+ * a verifier rebuilds creation code from, so a record claiming a contract took
+ * none when it took two describes a deployment that never happened.
+ */
+
+/** What a record holds when a contract genuinely takes no arguments. */
+const NO_ARGS = '0x'
+
+/** Both spellings appear in existing records; `''` predates the sentinel. */
+const isEmptyRecord = (recorded: string): boolean =>
+  recorded.trim() === '' || /^0x$/i.test(recorded.trim())
+
+const isHex = (value: string): boolean => /^0x[0-9a-f]*$/i.test(value.trim())
+
+interface IAbiParameter {
+  type: string
+  components?: readonly unknown[]
+}
+
+const asParameter = (input: unknown, contractName: string): IAbiParameter => {
+  if (typeof input !== 'object' || input === null)
+    throw new Error(
+      `${contractName}: a constructor input is not an object, so its type cannot be read`
+    )
+  const { type, components } = input as { type?: unknown; components?: unknown }
+  if (typeof type !== 'string' || type === '')
+    throw new Error(
+      `${contractName}: a constructor input has no type, so it cannot be encoded`
+    )
+  return {
+    type,
+    ...(Array.isArray(components) ? { components } : {}),
+  }
+}
+
+/**
+ * Renders one ABI parameter as its canonical type string.
+ *
+ * Tuples have to be expanded into `(a,b)` because an encoder given the literal
+ * word `tuple` has no idea what it is encoding.
+ */
+const canonicalType = (input: unknown, contractName: string): string => {
+  const { type, components } = asParameter(input, contractName)
+  if (!type.startsWith('tuple')) return type
+
+  if (!components)
+    throw new Error(
+      `${contractName}: a tuple constructor input has no components, so its type cannot be built`
+    )
+
+  const inner = components
+    .map((component) => canonicalType(component, contractName))
+    .join(',')
+  // Anything after `tuple` is array suffixes, e.g. `tuple[2][]`.
+  return `(${inner})${type.slice('tuple'.length)}`
+}
+
+/**
+ * Reads a constructor's parameter types out of a contract's ABI.
+ *
+ * Throws rather than returning an empty list when the ABI cannot be read: an
+ * empty list means "this contract takes no arguments", and letting a malformed
+ * artifact say that is how a record ends up claiming a deployment had none.
+ *
+ * @param abi - The `abi` array from a Forge artifact.
+ * @param contractName - Named in every message; the caller always knows it.
+ * @returns Canonical type strings in declaration order, empty when the contract
+ * declares no constructor.
+ */
+export const constructorInputTypes = (
+  abi: unknown,
+  contractName = 'contract'
+): string[] => {
+  if (!Array.isArray(abi))
+    throw new Error(
+      `${contractName}: the artifact holds no readable ABI, so its constructor cannot be read`
+    )
+
+  const constructor = abi.find(
+    (entry) =>
+      typeof entry === 'object' &&
+      entry !== null &&
+      (entry as { type?: unknown }).type === 'constructor'
+  )
+  if (!constructor) return []
+
+  const { inputs } = constructor as { inputs?: unknown }
+  if (inputs === undefined) return []
+  if (!Array.isArray(inputs))
+    throw new Error(
+      `${contractName}: the constructor's inputs are not a list, so its types cannot be read`
+    )
+
+  return inputs.map((input) => canonicalType(input, contractName))
+}
+
+/**
+ * Checks a record's constructor arguments against what the contract can take.
+ *
+ * @param contractName - Contract the record is for.
+ * @param recorded - The string about to be written to the record.
+ * @param types - Declared constructor types, from {@link constructorInputTypes}.
+ * @throws When the record and the ABI disagree about whether there are
+ * arguments, in either direction.
+ */
+export const assertRecordedArgsMatchAbi = (
+  contractName: string,
+  recorded: string,
+  types: readonly string[]
+): void => {
+  const empty = isEmptyRecord(recorded)
+
+  if (types.length > 0 && empty)
+    throw new Error(
+      `${contractName} takes ${
+        types.length
+      } constructor arguments (${types.join(
+        ', '
+      )}) but the record says it has none. A verifier rebuilding creation code from this record would compute a different deployment.`
+    )
+
+  if (types.length === 0 && !empty)
+    throw new Error(
+      `${contractName} takes no constructor arguments, but the record carries ${
+        recorded.trim().length
+      } characters of them. Either the record is for a different contract or the arguments are wrong.`
+    )
+}
+
+/** Encodes ABI parameters. Tron's encoder accepts base58 addresses; viem's does not. */
+export type AbiParamEncoder = (
+  types: string[],
+  values: readonly unknown[]
+) => string
+
+/**
+ * Encodes constructor arguments using the contract's declared types.
+ *
+ * Every failure throws. Substituting a best-effort encoding writes bytes that
+ * are not what the constructor received, and nothing downstream can tell the
+ * difference — a failed deploy is recoverable, a wrong record is not.
+ *
+ * @param encoder - ABI encoder to call, e.g. TronWeb's.
+ * @param args - Values passed to the constructor, in declaration order.
+ * @param types - Declared types, from {@link constructorInputTypes}.
+ * @param contractName - Named in every message.
+ * @returns Hex-encoded arguments, or `0x` when the contract takes none.
+ */
+export const encodeConstructorArgs = (
+  encoder: AbiParamEncoder,
+  args: readonly unknown[],
+  types: readonly string[],
+  contractName: string
+): string => {
+  if (args.length !== types.length)
+    throw new Error(
+      `${contractName} expects ${types.length} constructor arguments, got ${args.length}`
+    )
+
+  if (types.length === 0) return NO_ARGS
+
+  let encoded: string
+  try {
+    encoded = encoder([...types], args)
+  } catch (error) {
+    throw new Error(
+      `${contractName}: encoding constructor arguments failed (${
+        error instanceof Error ? error.message : String(error)
+      })`
+    )
+  }
+
+  if (typeof encoded !== 'string' || !isHex(encoded))
+    throw new Error(
+      `${contractName}: the ABI encoder did not return hex, so there is nothing safe to record`
+    )
+
+  return encoded
+}
--- a/script/deploy/tron/constructor-args.ts
+++ b/script/deploy/tron/constructor-args.ts
@@ -13,7 +13,12 @@ const NO_ARGS = '0x'
 const isEmptyRecord = (recorded: string): boolean =>
   recorded.trim() === '' || /^0x$/i.test(recorded.trim())
 
-const isHex = (value: string): boolean => /^0x[0-9a-f]*$/i.test(value.trim())
+const isHex = (value: string): boolean => {
+  const normalized = value.trim()
+  // Whole bytes only: '0x0' is a nibble, and half a byte of a record is not a
+  // value the constructor could have received.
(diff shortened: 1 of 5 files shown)


script/deploy/tron/tronUtils.ts
```diff
@@ -194,20 +200,15 @@ export async function deployContractWithLogging(
...
-        result.contractAddress,
+        address: result.contractAddress,
         version,
-        constructorArgsHex,
-        false
-      )
+        artifact,
+        constructorArgs,
+        verified: false,
+      })
```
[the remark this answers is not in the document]
0xDEnYO: Taken, and it is the better fix — `e3f09ced7`.

`assertTronDeploymentRecordable` now runs immediately before `deployer.deployContract` at all eight periphery sites and in `deployContractWithLogging`. Everything it checks is pure (the artifact ABI and the values), so failing there costs nothing.

Verified ordering at every site rather than asserting it:

```text
404-440:   PREFLIGHT -> DEPLOY -> SAVE -> RECORD  ok
515-549:   PREFLIGHT -> DEPLOY -> SAVE -> RECORD  ok
639-675:   PREFLIGHT -> DEPLOY -> SAVE -> RECORD  ok
746-782:   PREFLIGHT -> DEPLOY -> SAVE -> RECORD  ok
918-954:   PREFLIGHT -> DEPLOY -> SAVE -> RECORD  ok
1021-1057: PREFLIGHT -> DEPLOY -> SAVE -> RECORD  ok
1161-1197: PREFLIGHT -> DEPLOY -> SAVE -> RECORD  ok
1304-1332: PREFLIGHT -> DEPLOY -> SAVE -> RECORD  ok
```

For completeness on the address half: an earlier commit in this PR (`0d3970f68`) already moved `saveContractAddress` ahead of the record, so even if the recorder somehow throws, the deployed address is on disk and `checkExistingDeployment` will find it. The preflight makes that path unreachable for the three causes you listed rather than merely survivable.

Worth noting what the preflight adds beyond the devkit: `TronContractDeployer` already checks arity at energy-estimation time, so an arity mismatch was stopped pre-broadcast. It does not cover an unreadable ABI, a non-hex encoder result, or the record-vs-ABI mirror check — those are the ones this closes.

script/deploy/tron/constructor-args.ts
```diff
@@ -0,0 +1,182 @@
...
+    encoded = encoder([...types], args)
+  } catch (error) {
+    throw new Error(
+      `${contractName}: encoding constructor arguments failed (${
+        error instanceof Error ? error.message : String(error)
+      })`
+    )
+  }
+
+  if (typeof encoded !== 'string' || !isHex(encoded))
```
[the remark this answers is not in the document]
0xDEnYO: Fixed in `e3f09ced7`. `isHex` now requires whole bytes, so `0x0`, `0x123` and a 63-byte-and-a-nibble payload are all refused as encoder results and as records.

Added a whole-byte control case alongside the three refusals — without it, an `isHex` that rejected everything would satisfy the new assertions just as well.

gvladika (commented):
Three notes on whether the guard layering is all load-bearing. The core fix (ABI-typed encoding + single funnel + save-before-record) is right and I'd approve it — these are about redundancy in the guards around it, not correctness. Tests pass and tsc is clean.

script/deploy/tron/constructor-args.ts
```diff
@@ -0,0 +1,201 @@
...
+/**
+ * Checks a record's constructor arguments against what the contract can take.
+ *
+ * @param contractName - Contract the record is for.
+ * @param recorded - The string about to be written to the record.
+ * @param types - Declared constructor types, from {@link constructorInputTypes}.
+ * @throws When the record and the ABI disagree about whether there are
+ * arguments, in either direction.
+ */
+export const assertRecordedArgsMatchAbi = (
```
gvladika, line 123: `assertRecordedArgsMatchAbi` is only called from `recordTronDeployment` and `assertTronDeploymentRecordable`, and both feed it the string `encodeConstructorArgs` just produced from the same `types`. So in every path that calls it, it's validating the encoder's own fresh output — which the encoder already guarantees is hex, correct arity, and non-empty-iff-types-exist. It can't fire there.

Where it *would* earn its keep is auditing records the encoder didn't produce — e.g. the seven `'0x'` records still in `deployments/*.json` until those contracts are re-deployed. But it isn't wired as a standalone record auditor anywhere in this PR. As-is it's a well-built function guarding the encoder against itself, carrying test + mutation coverage for a check that can't trigger in production.

Suggestion: either wire it over the existing deployment records (higher-value — it'd catch the bad records that are live right now), or note explicitly that it's pure defense-in-depth for a future auditor.

script/deploy/tron/tronUtils.ts
```diff
@@ -229,49 +239,88 @@ export async function deployContractWithLogging(
...
+ * that decides what the `constructorArgs` field holds, so a call site cannot
+ * hand the log a string of its own.
+ *
+ * @param params.artifact - The Forge artifact the contract was deployed from.
+ * @param params.constructorArgs - Exactly the values passed to the constructor.
+ * @throws When the arguments cannot be encoded from the ABI. Callers must save
+ * the deployed address BEFORE calling this: the contract is already on chain by
+ * then, and losing its address costs a duplicate deployment.
+ */
+export async function recordTronDeployment(params: {
```
gvladika, line 295: `recordTronDeployment` re-runs `constructorInputTypes` → `encodeWithTypes` → `assertRecordedArgsMatchAbi`, which `assertTronDeploymentRecordable` (line 266) already ran moments earlier at the same call site. So every deploy encodes twice.

It's cheap and pure, and "each function self-contained" is a legit reason to keep the funnel independent. But worth naming that the preflight's *unique* marginal value over the save-before-record backstop is narrow: the failure modes it alone catches pre-broadcast (unreadable ABI, encoder throw) are deterministic artifact problems that surface on the first contract before any TRX is at risk. Two mechanisms for one property — fine to keep for money-spending tooling, just flagging it's redundant rather than each catching a distinct case.

script/deploy/tron/constructor-args.ts
```diff
@@ -0,0 +1,201 @@
...
+  }
+}
+
+/**
+ * Renders one ABI parameter as its canonical type string.
+ *
+ * Tuples have to be expanded into `(a,b)` because an encoder given the literal
+ * word `tuple` has no idea what it is encoding.
+ */
+const canonicalType = (input: unknown, contractName: string): string => {
```
gvladika, line 50: Tuple expansion here handles a case none of the eight affected contracts have — no Tron periphery constructor takes a tuple. Without it, a tuple would make the encoder throw (safe) rather than mis-encode, so this adds capability, not safety.

Not arguing to remove it (correct ABI handling is defensible), just noting it's generality beyond the current call sites per the repo's "no speculative features" line.

script/deploy/tron/constructor-args.ts
0xDEnYO: Confirmed by testing: fed `encodeConstructorArgs`'s real output back into `assertRecordedArgsMatchAbi` across address/uint256, bool, string, bytes, address[], empty string/bytes, and a tuple — zero failures. The only way I could trip the word-count check was handing it a deliberately broken (truncated-output) encoder. So on both current call sites it's guarding against the underlying ABI encoder misbehaving, not a record/ABI mismatch this code could itself produce — matches your read exactly.

Went with your second option: added a note on `assertRecordedArgsMatchAbi`'s docstring making this explicit, rather than wiring a standalone auditor in this PR — that needs read access to the historical Mongo deploy log (out of scope here, and worth doing properly). Flagged it as a follow-up task so it isn't lost, since you're right that it'd catch real bad records — ERC20Proxy alone was unconditionally recorded as `'0x'` before this PR despite taking two constructor args.

Doc commit is ready locally; holding the push because merging in current `origin/main` pulls in an unrelated, already-known TS failure (`script/deploy/safe/proposal-card.ts` importing `MAX_PROPOSAL_REASON_LENGTH` from the wrong file) that trips the pre-push validator — someone's already fixing it on a separate branch. Will push as soon as that clears.

script/deploy/tron/tronUtils.ts
0xDEnYO: Confirmed — `recordTronDeployment` re-runs `constructorInputTypes` → `encodeWithTypes` → `assertRecordedArgsMatchAbi` independently of the `assertTronDeploymentRecordable` preflight at the same call site, so it does encode twice per deploy. Agreed it's worth keeping as-is: same story as the thread above — the only thing the second pass could still catch beyond the first is the encoder itself behaving differently between two calls a few lines apart, which is cheap insurance for money-spending tooling. No change made.

script/deploy/tron/constructor-args.ts
0xDEnYO: Agreed none of the eight current sites pass a tuple. Kept the expansion because `constructorInputTypes`/`canonicalType` is the general ABI-type reader for any Tron periphery or facet constructor going forward, not scoped to this PR's eight contracts — dropping tuple support wouldn't shrink the deployment-recording feature, it'd just make the ABI reader wrong the day a Tron contract does take a struct (encoder throw instead of silent mis-encode, per your own note, but still a gap). Reads to me as correctly implementing "read the declared types" rather than speculative scope creep, but open to your call if you'd still cut it.

script/deploy/tron/constructor-args.ts
0xDEnYO: Update: the docstring clarification is pushed now — `034d6472f` on this branch (the push was blocked earlier by an unrelated pre-existing TS failure on main, since fixed in #2312 and cherry-picked here as `81e7414a8`).

To restate what landed vs. what's deferred: `assertRecordedArgsMatchAbi` now documents that both current call sites feed it a fresh encode, so only the word-count check can fire there (and only against a misbehaving encoder) — that's the code-visible part of your finding. The higher-value half — wiring it as a standalone auditor over the historical Mongo deploy records — is still open and still needs your judgment on scope/placement; not doing that in this PR. Leaving this open for that.

gvladika (approved):

PR #1303, opened 2025-07-25
Tron deployments

ezynda3:
# Which Jira task belongs to this PR?

# Why did I implement it this way?

# Checklist before requesting a review

- [x] I have performed a self-review of my code
- [x] This pull request is as small as possible and only tackles one problem
- [ ] I have added tests that cover the functionality / test the bug
- [ ] For new facets: I have checked all points from this list: https://www.notion.so/lifi/New-Facet-Contract-Checklist-157f0ff14ac78095a2b8f999d655622e
- [ ] I have updated any required documentation

# Checklist for reviewer (DO NOT DEPLOY and contracts BEFORE CHECKING THIS!!!)

- [ ] I have checked that any arbitrary calls to external contracts are validated and or restricted
- [ ] I have checked that any privileged calls (i.e. storage modifications) are validated and or restricted
- [ ] I have ensured that any new contracts have had AT A MINIMUM 1 preliminary audit conducted on <date> by <company/auditor>

PR #2305, opened 2026-09-02
fix(tron): pass the constructor arguments the core facets declare (EXSC-911)

0xDEnYO:

Fixes [EXSC-911](https://linear.app/lifi-linear/issue/EXSC-911/sc-tron-core-facet-deploy-script-passes-wrong-constructor-arity-to)

Reuses `constructorInputTypes` / `assertTronDeploymentRecordable` from #2302 (EXSC-907), which has merged — this PR now targets `main` directly and its diff is the three Tron files below.

`script/deploy/tron/deploy-core-facets.ts` passed the wrong number of constructor arguments to two contracts. Nothing on chain is affected — both paths are unreachable today, and arity is checked at energy-estimation time, so neither could ever have spent TRX.

**`LiFiDiamond` got one argument against a two-argument constructor.** Not drift: `constructor(address _contractOwner, address _diamondCutFacet)` has been in place since the repo's initial commit (`36f87e399`, 2022-03-04), and the one-argument call is in the very first version of this script (`97ef15adc`, 2025-08-26, #1303). That same PR recorded the Tron diamond at `TU3ymit…`, so the diamond reached chain another way and every run since has taken the `Skipping LiFiDiamond (already deployed)` branch above it.

**The owner value was not encodable either.** It was built as `0x` + `tronWeb.address.toHex(...)` — a 21-byte `0x41…` string. TronWeb's `encodeParams(['address'], …)` rejects it outright (`invalid address`), so even a two-argument call would have failed. `tronAddressToHex` returns the 20-byte form the encoder accepts, which is what every other branch already passes.

**`getConstructorArgs()` returned `[]` for `LiFiIntentEscrowFacetV2`**, which declares `constructor(address _inputSettler)`. This one is real drift, but in config rather than in the contract: the facet entered `config/global.json` → `coreFacets` on 2026-07-28 (#2120) and 2026-08-17 (#2190), long after this switch was last touched. `getCoreFacets()` just reads that array, so adding a name to it silently extends the deploy loop while the switch's fall-through is `return []`. The facet is live on Tron (2026-08-20, #2220) via its own script, which passes the settler correctly.

Fixing the two call sites alone would leave the mechanism intact, so the arguments are now resolved for the whole facet set and checked against each compiled ABI **before the first deployment**. A facet added to `coreFacets` without a branch fails loudly, naming the facet and what its constructor declares, with nothing spent — rather than being deployed with no arguments. The check is scoped to facets that will actually be deployed, so facets skipped as already-deployed do not require an artifact.

The argument builder moved to `script/deploy/tron/helpers/coreFacetConstructorArgs.ts` because `deploy-core-facets.ts` calls `runMain` at module scope and cannot be imported from a test — the same reason `propose-calls-tron.ts` keeps its logic `runMain`-free.

Verified against real ABIs (`FOUNDRY_PROFILE=solc_floor forge build`), not the Solidity source:

| Case | Result |
|---|---|
| `LiFiDiamond` + `[owner]` | throws `expects 2 constructor arguments, got 1` |
| `LiFiIntentEscrowFacetV2` + `[]` | throws `expects 1 constructor arguments, got 0` |
| `LiFiDiamond` + `[owner, diamondCutFacet]` (20-byte) | encodes |
| `0x41…` 21-byte owner | `invalid address` from the encoder |

Tests: 14 new, 100% line and function coverage on the new helper. Full TS suite 2006 pass / 0 fail. `bunx eslint` and `bunx tsc-files --noEmit` clean on all three files. The suite stays self-contained — the artifact loader is injectable, so no `out/` is needed in CI.

## The diamond was outside the check it motivated

`LiFiDiamond` is deployed **after** all eleven facets and is not in `coreFacets`,
so the pre-flight did not cover it — its arity was still checked inside
`deployContractWithLogging`, which is where the expensive failure happens. On
`origin/main` that is exactly what a staging run does: eleven facets deploy, then
`LiFiDiamond expects 2 constructor arguments, got 1`.

Its real arguments do not exist that early (the DiamondCutFacet address is
runtime), so `assertDiamondConstructorShape` probes the ABI with placeholders to
prove the arity and encodability. Against the real compiled artifact:

```text
1 arg  -> LiFiDiamond expects 2 constructor arguments, got 1 The diamond is
          deployed with [owner, diamondCutFacet] in deploy-core-facets.ts.
2 args -> accepted
```

The probe is skipped when the diamond is already deployed, matching the facets'
own skip rule — otherwise an ABI change would fail a run that never deploys it.

## Behaviour change on staging, stated

`deploy-core-facets.ts` selects `tronshasta` for every non-production
environment, and `config/lifiintentescrow.json` has no `tronshasta` key. The run
therefore stops at argument resolution with nothing deployed.

This is the same failure as before, moved earlier: on `origin/main` the identical
run deploys ten facets, then fails on `LiFiIntentEscrowFacetV2`, then fails on the
diamond. So the delta is 12 broadcasts to 0 — the point of the ticket — but it is
a visible change for anyone running staging, and it stays blocked until a
`tronshasta` entry is added. Not fixed here: adding that config value is a
separate decision about what the settler address should be.

## Also from the review gate

- The module docstring invited callers to pass any core facet set. `GasZipFacet`
  is in `coreFacets`, has a constructor, and has no branch here — it is only kept
  out by the `exclude` at the call site. The docstring now says so.
- The wrapper appended "They are built by `getConstructorArgs`" to every failure,
  including an unreadable-ABI failure the args have nothing to do with. Reworded
  to a pointer rather than a cause.
- `loaderFor({})` returns `{ abi: undefined }`, which the real loader never
  produces — it rejects outright on a missing `out/`. That left the real
  stale-artifact condition untested; there is now a case whose loader rejects.

## Second gate round, on the fix commit

The hoisted `checkExistingDeployment` was the risky part of the fix, so it was
checked rather than assumed: it only reads `deployments/<network>.json` by
contract name and prompts, the facet loop only writes facet-named keys, and
`LiFiDiamond` is not in `coreFacets` — so the hoisted read cannot go stale. Proven
by reading the same key before the loop and after all eleven writes (identical),
and the skip branch was enumerated over all 16 combinations of
`exists x address in {null, '', '0', addr} x shouldRedeploy` with 0 disagreements
against the previous inline condition.

One real finding, fixed in `e33e3ef7a`: the probe counted a literal `2` while the
deploy built its own `[ownerHex, diamondCutFacetHex]`, with nothing tying them
together. Drift in one direction fails safe, but the other re-opens exactly this
ticket's bug — the probe would pass a shape the deploy does not use, and the
mismatch would surface after every facet was paid for. Both now derive from
`DIAMOND_CONSTRUCTOR_PARAMS`; adding a parameter is a compile error at the deploy
site:

```text
deploy-core-facets.ts(315,13): error TS2741: Property 'thirdParam' is missing in
type '{ owner: string; diamondCutFacet: string; }' but required in type
'DiamondConstructorArgs'.
```

Also corrected: the probe's docstring claimed it proves "the arity and the ABI".
It proves the arity and that address-shaped placeholders encode — a parameter
changing to another word-sized type still passes, and an array or tuple parameter
would fail for a reason that is not arity. Said so explicitly.


0xDEnYO:
@coderabbitai review

0xDEnYO:
@coderabbitai review

0xDEnYO:
Reviewed the one nitpick from the CodeRabbit run (body-only, no thread — `coreFacetConstructorArgs.ts:23-25`, "use an interface for the artifact result").

Rejecting: it's self-flagged Trivial/Low-value, and the inline `Promise<{ abi?: unknown }>` shape matches the existing convention for this kind of injectable loader type in the same directory tree (e.g. `DrainOpener` in `script/deploy/safe/drain-parked-tasks.ts:393`). Not changing it.

0xDEnYO:
## Review gate — escalated findings

Auto-fixed and pushed in `169b053a2`: artifact loading hoisted out of the `try` (loader failures were being reported as constructor-argument failures, pointing at the wrong file), and two comments removed that narrated the defect's history rather than a non-obvious *why*.

The rest needs an author decision — they are behavioural choices, not mechanical repairs.

### 1. Regression: a dry run of a fresh network now fails

`deploy-core-facets.ts:281-290` sends `facetAddresses['DiamondCutFacet'].address` through `tronAddressToHex`. When `DiamondCutFacet` is deployed in the same run, that value is the devkit's simulated placeholder `TXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX` (`TronContractDeployer.js:262-267`), which fails base58 checksum decoding:

```text
$ node -e "const {TronWeb}=require('tronweb'); new TronWeb({fullHost:'…'}).address.toHex('T'+'X'.repeat(33))"
THREW: Invalid address provided
```

The surrounding `try/catch` records that as `status: 'failed'`, so `hasFailures` is true and the script exits 1. `--dry-run` on a first-time network — the case this script exists for — reports a hard failure. The previous code built the diamond's only argument from `networkInfo.address`, a real address, so it never converted a simulated one.

Fail-safe rather than unsafe, but it breaks the rehearsal path, and there is no existing pattern in `deploy-and-register-periphery.ts` for a same-run simulated address to copy — hence escalating rather than picking a dry-run semantic.

### 2. The new refusal hard-blocks every `tronshasta` run, and a test pins it there

`coreFacetConstructorArgs.ts:78-84`. `config/lifiintentescrow.json` has only a `tron` key, `LiFiIntentEscrowFacetV2` is in `coreFacets`, and non-production runs target `tronshasta`:

```text
STAGING THREW: lifiEscrowInputSettler not found for tronshasta in config/lifiintentescrow.json
```

Previously the run deployed 10 facets plus the diamond and failed only on the escrow facet; now it aborts before the first broadcast. That is the intended fail-fast, but nobody finds out until a staging bring-up. Compounding it, `coreFacetConstructorArgs.test.ts:106-114` asserts that failure by name — so adding the `tronshasta` config entry breaks the unit test. Either add the entry here, or retarget the test at a synthetic network key.

### 3. Coverage gap: `LiFiDiamond` sits outside the pre-flight it motivated

`deploy-core-facets.ts:168-175` probes only `coreFacets`. The diamond's arity is checked inside `deployContractWithLogging` (line 297+), i.e. after 11 facets are paid for. Wrong diamond arity is one of the two bugs this PR exists to fix, so for that contract the mechanism is still intact: add a third constructor parameter and the script keeps passing two, discovering it at the end. The *values* genuinely cannot be resolved early (one is a facet address), but the arity can — `constructorInputTypes(await loadForgeArtifact('LiFiDiamond'))` asserted `=== 2` in the pre-flight block closes it.

### Lower-confidence — your judgment

- `deploy-core-facets.ts:209-211` — the `No constructor arguments resolved` guard cannot fire: the filter predicate at line 171 is character-identical to the loop's skip predicate at line 189 over the same index-aligned arrays. Keep it for TS narrowing of `Map.get`, but it is not a check.
- `coreFacetConstructorArgs.test.ts:168-178` — the "no readable ABI" case relies on `loaderFor({})` returning `{abi: undefined}`; the real `loadForgeArtifact` throws instead and never returns an ABI-less object. The branch is real, but the test proves the injection seam, not the deploy path.
- All four `resolveCoreFacetConstructorArgs` tests inject a loader, so the exact regression this PR prevents — a constructor-bearing facet added to `coreFacets` — is caught only when a human runs the deploy script. A test asserting every name in `getCoreFacets()` is either handled or constructor-less would make it CI-enforced without needing `out/`.
- `encodeParams` rewrites the caller's array in place (base58 → lowercased `0x` hex), so the map does not hold what `getConstructorArgs` returned. Benign and idempotent, but it falsifies the "everything it checks is pure" docstring on `assertTronDeploymentRecordable`, and this is the first call site that keeps the array and deploys from it later.
- New directory `script/deploy/tron/helpers/` — every other script sub-area uses `utils/` (`script/utils`, `script/deploy/facets/utils`, `script/deploy/zksync/utils`, `script/troncast/utils`), and the closest sibling `constructor-args.ts` is flat in `tron/`.

### Adjacent, out of scope — worth a ticket

- `tronAddressToHex` passes a `0x` EVM address straight through, so a `config/lifiintentescrow.json` entry holding the EVM settler — which is what that file's top-level default holds — clears both the guard and the encoder, and would deploy the facet pointing at an address with no code on Tron. Nothing checks value *correctness*, only encodability.
- `register-facets-to-diamond.ts:66-78` hardcodes a 9-facet list omitting `LiFiIntentEscrowFacetV2` and `DiamondCutFacet` — the same config-drift mechanism this PR fixes, one script over.
- History note: the diamond branch has been broken two ways since #1303 (wrong arity *and* a malformed 21-byte `0x41` owner), so it could never have succeeded. The `LiFiDiamond` address recorded in `deployments/tron.json` therefore did not come from this script. Doesn't block this PR — the skip-if-exists branch means it never re-runs — but the provenance is an open question for WP-2.2.

gvladika (approved):

PR #2308, opened 2026-09-02
feat(immutables): per-immutable expectation registry + gate (EXSC-869)

0xDEnYO:

Part of [EXSC-869](https://linear.app/lifi-linear/issue/EXSC-869/wp-23-per-immutable-expectation-registry) (ref EXSC-869) — WP-2.3 **part (i)**: the registry mechanism, its CI gate, and the InfoSec protection D20 asks for. Part (ii), the authoring pass, is not here.

**This PR will be red on `protect-critical-code` until an Information Security Manager or the CTO approves it** — it edits `.github/`, which is exactly the gate it is adding paths to. That is the designed cost, not a problem to route around.

## What the registry is, and where it lives

Each entry names the immutable it governs and how its expected value is established:

| `source` | Means | Requires |
|---|---|---|
| `config` | the value comes from a config path | `configData`: a label under the same contract in `deployRequirements.json` |
| `derived` | computed at deploy time | `rule`: how |
| `unchecked` | deliberately not verified | `reason` |
| `unverifiable` | not getter-coverable (F5: no compiler-generated getter, and no explicit accessor) | `reason` |

`authorityBearing: true` flags the ones the drift report must call out. Per D20 that is a marker, not a new approver tier.

**It is its own file, and that is forced by the protection rather than by taste.** Path-based protection only means anything if the path holds only the protected thing. Folding the registry into `deployRequirements.json` would put every routine deploy-requirement edit — a new contract's config paths — behind an InfoSec approval nobody asked for. Verified against the workflow's own matcher:

```text
PROTECTED  script/deploy/resources/immutableRegistry.json
PROTECTED  script/deploy/immutables/registry-schema.ts
PROTECTED  .github/workflows/x.yml
free       script/deploy/resources/deployRequirements.json
free       src/Facets/AcrossFacet.sol
```

The gate's own code is protected too: editing the checker so it always passes is equivalent to editing the expectation.

## Why the link between the two files is explicit

There is no derivable link, and this is the finding that shaped the whole design. A `configData` key is a **free-form label** — `helperFunctions.sh:4903` reads `configData | keys[]` only to iterate, then follows `configFileName`/`keyInConfigFile`, so nothing constrains its spelling. `AcrossFacet` is the worked case:

```text
constructor(IAcrossSpokePool _spokePool, address _wrappedNative)
    stores:   address private immutable wrappedNative
    registry: "_wrappedNativeAddress"
```

Three spellings, one binding. Matching them by name across the repo produces **50 orphaned entries on a correct repo** — noise, not findings. So the entry names the immutable explicitly, and `registry-coverage.ts` exists only to *suggest* the 54 likely links for a human to confirm during part (ii). Its header says so.

## Errors and warnings are different things

An immutable nobody has documented yet is a **warning**: the authoring pass is still running, and WP-2.3's own ordering constraint puts the hard fail after it. A registry that asserts something false is an **error in either mode** — a link to a `configData` key that does not exist, an entry for an immutable the contract no longer declares (a rename leaves an expectation nothing checks), an exemption with no reason, or an unrecognised `source`. That last one matters because a typo in `source` would otherwise make an immutable silently exempt.

## Evidence

Every error path firing on the real fleet, and the benign direction staying silent:

```text
79 immutables, no entries      -> 0 errors, 79 warnings, exit 0
AcrossFacet authored (2)       -> 0 errors, 77 warnings, exit 0
--strict with the gap          -> exit 1
link to a nonexistent label    -> 1 error, exit 1 even in warn-only
entry for a renamed immutable  -> 1 error  ("...does not declare that immutable")
exemption with no reason       -> 1 error  ("...cannot be reviewed")
```

Also measured: **79 immutables in 42 files** (the ticket says 111, which came from a grep whose `^\s*//` filter could never match past `grep -n`'s `file:line:` prefix), and 104 `configData` entries across 61 contracts. And writing the enumeration found `address payable public immutable POLYMER_FEE_RECEIVER` in `PolymerCCTPFacet.sol`, which a single-token type pattern reads straight past — the worst failure available to a registry gate, since the immutable is never *asked for* rather than showing as unanswered. The sweep now reports zero declarations it cannot read.

`jq '.AcrossFacet.configData | keys[]'` still returns both labels, so the deploy path reads `deployRequirements.json` exactly as before. 109 tests here, 100% lines on all three modules; the repo's `run-ts-tests` job is green on this branch.

## Not in this PR

Part (ii): the authoring pass. It is **not** "add 29 entries" — it is authoring all 79, with the suggester's 54 name matches doing the mechanical half. The `--strict` flip belongs at the end of that, per the ticket's ordering constraint. The three-state drift report (wrong-at-deploy / superseded-since / current) also waits, since it needs deployed values — that is `immutableBindings.ts` from [#2213](https://github.com/lifinance/contracts/pull/2213), still open.

- [x] I have updated any required documentation — the registry's vocabulary is documented in `registry-schema.ts`; `docs/MultisigSigningProcess.md` gains nothing until the gate blocks


script/deploy/immutables/immutable-declarations.ts
```diff
@@ -0,0 +1,74 @@
...
+ * @param file - Repo-relative path, carried onto each result.
+ * @returns One entry per declaration, in source order.
+ */
+export const parseImmutableDeclarations = (
+  source: string,
+  file: string
+): IImmutableDeclaration[] => {
+  const found: IImmutableDeclaration[] = []
+
+  source.split('\n').forEach((text, index) => {
```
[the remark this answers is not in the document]
0xDEnYO: Taken, and this was the most valuable of the four — `482e024c7`.

You are right that strict mode could miss it, and the reason is worse than a missed match: a shape the parser cannot read produces **nothing**, so the immutable is never asked for and never appears as unanswered. My own docstring claimed "a shape it cannot read shows up as a missing entry" — that was false, and I have corrected it in the file.

Rather than chase syntaxes, the gate now asks what the enumerator failed to read. `findUnreadableImmutableLines` reports any line mentioning `immutable` that did not parse (excluding `//`, block-comment bodies and natspec tags), and those are errors in either mode. Demonstrated by wrapping a real declaration in `AcrossFacet.sol`:

```text
wrapped across lines : 78 immutables found (was 79)  ERROR src/Facets/AcrossFacet.sol:25 mentions immutable but could not be read  exit 1
restored             : exit 0
```

Note the count dropping to 78 — that is exactly the invisibility you identified. Six tests cover it, including the wrapped form, and one pins the `address payable` form as *readable* so a future narrowing of the type pattern becomes an error rather than silence.

I did not switch to token-aware parsing. It is the better long-term answer, but the blind-spot report closes the class today for any shape rather than the shapes I anticipate, and it does so without a Solidity parser dependency in a CI gate.

script/deploy/immutables/registry-schema.ts
```diff
@@ -0,0 +1,163 @@
...
+        continue
+      }
+
+      if (source === 'config') {
+        const label = entry.configData
+        if (!nonEmptyString(label)) {
+          errors.push(`${where} is config-sourced but names no configData key.`)
+          continue
+        }
+        if (!((label as string) in (contractRequirements.configData ?? {})))
```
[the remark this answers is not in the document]
0xDEnYO: Fixed in `482e024c7` — `Object.prototype.hasOwnProperty.call`. Three tests, one each for `toString`, `constructor` and `hasOwnProperty`, all failing against the previous commit.

Worth saying: this is the second time today you have caught me on this exact bug — the CBOR map decoder in #2301 had `key in entries` with the same prototype problem. Recorded as a pattern to check rather than a one-off.

script/deploy/immutables/registry-schema.ts
```diff
@@ -0,0 +1,163 @@
...
+      const where = `${contract}.${name}`
+
+      if (!declaredHere.has(name)) {
+        errors.push(
+          `${where} has a registry entry but ${contract} does not declare that immutable. A rename leaves an expectation nothing checks.`
+        )
+        continue
+      }
+
+      if (entry.authorityBearing === true) authorityBearing.push(where)
```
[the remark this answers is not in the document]
0xDEnYO: Fixed in `482e024c7`. A present `authorityBearing` must now be a boolean, and a non-boolean is an error rather than a silent omission.

The consequence was the bad part: `"true"` left an authority-bearing immutable unflagged in the drift report, which is the one place it has to appear. Two tests — the quoted value errors, and an honestly `false` marker still passes without being listed.

.github/workflows/protectSecurityRelevantCode.yml
```diff
@@ -52,8 +52,10 @@ jobs:
               echo "Warning: File $FILE not found"
               continue
             fi
-            ##### check for github actions, pre-commit checker, and emergency-pause break-glass paths
-            if echo "$FILE" | grep -iE '^\.github/|^\.husky/pre-commit|^script/emergency/'; then
+            ##### check for github actions, pre-commit checker, emergency-pause break-glass paths,
+            ##### and the per-immutable expectation registry plus the gate that reads it — editing
+            ##### either is equivalent to editing the expectation itself (EXSC-869, D20)
+            if echo "$FILE" | grep -iE '^\.github/|^\.husky/pre-commit|^script/emergency/|^script/deploy/resources/immutableRegistry\.json$|^script/deploy/immutables/'; then
```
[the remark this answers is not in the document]
0xDEnYO: Taken — `482e024c7`. This is a real bypass and I reproduced both halves on a scratch repo before changing anything:

```text
DELETING script/deploy/resources/immutableRegistry.json
  current : SKIPPED (not found)          -> no approval required
  fixed   : PROTECTED

RENAMING it to script/deploy/resources/harmless.json
  current : free (only the new name is listed)  -> no approval required
  fixed   : free harmless.json  +  PROTECTED immutableRegistry.json
```

Both changes applied: `--no-renames` on the diff so a rename reports both paths, and the `[[ ! -f "$FILE" ]]` guard removed. The matcher is pure string matching on paths, so a path that no longer exists in the working tree is safe to test — the guard was buying nothing and costing the deletion case.

Flagging the scope honestly: this is **pre-existing** and not something this PR introduced. I fixed it here rather than filing it separately because the protection this PR adds is defeated by it — adding `immutableRegistry.json` to a matcher that skips deletions would have been protection in name only. If you would rather it landed as its own PR against `.github/`, say so and I will split it; it is a self-contained two-line change.

One thing I did not change: the `if [[ -z $FILES ]]` abort above it still treats an empty diff as a workflow bug and exits 1. That is orthogonal and looked deliberate.

0xDEnYO:
## Review-gate residual — items escalated for a human decision

The gate's auto-fixes landed in 7d8100a. These four it did not touch, because each is a design call rather than a defect.

**1. The `src/**` paths-filter gap is wider than the comment claimed, and there is precedent against shipping it.** A src-only PR runs neither the warn-only check nor the unreadable-declaration check — and the latter is a hard failure in both modes, so the gap is not confined to warn-only material. The job is also skipped on drafts, and the workflow has no `push` trigger, so it never runs on `main` either. The commit corrects the comment to say all of this.

The precedent: on #1901, the PR that created `validateScripts.yml`, @gvladika raised the structurally identical finding — a sweep that only fires on manifest changes, so the footgun it targets can be introduced without detection. That was not accepted as a documented known-gap; it was fixed before merge (`6106e88e`). The counter-argument here is real (adding `src/**` also runs the forge build and TypeChain generation on every Solidity PR), but it is a trade-off for a reviewer to accept, not one to assume. Options: add `src/**` and eat the build cost, split the registry check into its own cheap workflow with its own filter, or accept the gap explicitly as a `--strict` prerequisite.

**2. The registry's one cross-file check can be satisfied from an unprotected file.** An entry's `configData` link is validated against `deployRequirements.json`, which `protectSecurityRelevantCode.yml` does not cover. A non-InfoSec edit therefore cannot *author* an expectation — verified: planting a full `immutables` block in `deployRequirements.json` yields `0 carry a registry entry` — but it can *satisfy* the link check the registry exists to keep honest, by adding the label the entry points at. Whether that matters depends on what the link is meant to prove.

**3. `contractOf()` keys everything by file basename, in two modules independently.** Nothing enforces the invariant it assumes. Verified clean today by AST walk (0 duplicate basenames, 0 immutable-bearing contracts whose name ≠ basename), but a file renamed without its `contract` declaration, or one file declaring two immutable-bearing contracts, would silently mis-key entries against `deployRequirements.json`. Fix could be a shared helper plus a uniqueness check, or reading the `contract` declaration instead of the filename — more than one valid answer.

**4. A wrapped declaration yields a wrong `type` and drops `visibility`.** `IExecutor\n    public immutable EXECUTOR;` parses as `{type: "public", name: "EXECUTOR"}` with no visibility. `name` is the only field this validator reads, so there is no gate impact today — but #2213's binding verifier reads `getter`-adjacent shape information, and this is the one input where the enumerator produces a confidently wrong record instead of reporting an unreadable line.

Also worth knowing for triage: `CodeRabbit` reported `pass` with *"Review rate limited"*, so its last real review covered only the first two commits — the commit that wires the gate into CI has had no bot review. Re-requested above. The four round-1 threads are fixed and confirmed, but CodeRabbit could not self-resolve them, so they still read as open.

0xDEnYO:
@coderabbitai review

script/deploy/immutables/immutable-declarations.ts
```diff
@@ -0,0 +1,243 @@
...
+ * `transient` is deliberately not excluded. Solc rejects it as a data location
+ * on an immutable, so it can never be the modifier this exclusion guards
+ * against, while `uint256 public immutable transient` compiles. Excluding it
+ * would only make a legal declaration unreadable, which fails the gate in both
+ * modes.
+ */
+const RESERVED_AFTER_IMMUTABLE =
+  'public|private|internal|override|constant|immutable'
+
+const DECLARATION_RE = new RegExp(
```
gvladika, line 140: **Biggest simplification available here: enumerate immutables from the compiler instead of parsing source.** `forge build` already emits every immutable per contract (AST + `immutableReferences`), and the `validateScripts` job this runs in already compiles (Foundry install + `typechain:incremental` a couple of steps above). Reading the artifact would delete this parser *and* the entire `findUnreadableImmutableLines` hard-fail class — that failure mode only exists because a regex can miss shapes like `address payable public immutable`; the compiler can't miss them. ~695 lines (this file + its test) collapse, same set produced. The tradeoff is coupling the gate to build output — but this job already produces it.

script/deploy/immutables/registry-coverage.ts
```diff
@@ -0,0 +1,113 @@
...
+
+/**
+ * Proposes links in both directions.
+ *
+ * @param declarations - Every immutable found in `src/`.
+ * @param entries - Contract name to its `configData` keys.
+ * @returns Likely matches, unmatched declarations, and unmatched entries. None
+ * of the three is a verdict; they are the input to the authoring pass.
+ */
+export const assessRegistryCoverage = (
```
gvladika, line 72: This suggester (~266 lines incl. test) exists because the config-key↔immutable link is treated as hand-author-only. But it's largely *derivable*: immutable ← constructor-param is in the AST, and constructor-param ← config-path already lives in the deploy script that constructs the contract. Deriving `expected` from the same config path the deployer reads makes the registry and the deploy wiring unable to drift by construction, and shrinks the manifest to the genuinely underivable residue. Worth asking whether this module is needed at all versus deriving the links.

script/deploy/immutables/registry-schema.ts
```diff
@@ -0,0 +1,225 @@
...
+ * `config` names a `configData` label rather than repeating its path, so the
+ * deploy requirement and the expectation cannot drift apart. The link is
+ * explicit because nothing derives it: a label, its constructor parameter and
+ * the immutable it lands in are three independent spellings.
+ */
+export type ImmutableSource =
+  | 'config'
+  | 'derived'
+  | 'unchecked'
+  | 'unverifiable'
```
gvladika, line 30: `unverifiable` (no getter / "written to a mapping") is mostly an artifact of reading via getters. An immutable is by definition spliced into the deployed bytecode — `immutableReferences` gives the byte offsets, so the value is readable from `eth_getCode` with or without a getter. Reading actuals from bytecode collapses this category to near-empty and removes a whole `source` kind and its branches. (Minor: an immutable is never in a mapping — that would be storage, not an immutable.)

script/deploy/immutables/immutable-declarations.ts
```diff
@@ -0,0 +1,243 @@
...
+  line: number
+  /** The declared type, e.g. `address`, `uint256`, `IGasZip`. */
+  type: string
+  /** As written; undefined when the declaration omits it. */
+  visibility?: 'public' | 'private' | 'internal'
+  name: string
+}
+
+/**
+ * Replaces every comment and string literal with spaces, keeping all other
```
gvladika, line 29: Style (non-blocking): docstrings like this narrate the debate and alternatives considered ("the two cheap ways of approximating this both hid a real immutable…"), which `000-global-standards` says belongs in the PR description rather than the source. A fair amount of this PR's perceived heft is prose like this wrapped around ~700 lines of logic. Pure trim, no behavior change — worth tightening to the non-obvious *why* only.

gvladika (commented):
Recap of the biggest simplification opportunities (non-blocking). The core concept is sound — a source-enumerated ledger that forces every immutable into "checked-and-how" or "exempt-and-why". I'd keep the enforcement inversion, the validator's specific checks (`hasOwnProperty` vs `in`, the boolean check, rejecting unknown `source`), and the InfoSec protection + bash hardening. These four are about *how much* is hand-built vs. derived from facts the compiler and the deploy path already own.

script/deploy/immutables/immutable-declarations.ts
0xDEnYO: You're right, and I checked it against a real build rather than just reasoning about it. `forge build --ast` already gives exactly the enumeration this file hand-parses:

```
$ forge build src/Facets/AcrossFacet.sol --ast
$ jq -c '.ast.nodes[] | select(.nodeType=="ContractDefinition") | .nodes[] | select(.mutability=="immutable") | {id,name,type:.typeDescriptions.typeString,visibility}' out/AcrossFacet.sol/AcrossFacet.json
{"id":534,"name":"spokePool","type":"contract IAcrossSpokePool","visibility":"private"}
{"id":537,"name":"wrappedNative","type":"address","visibility":"private"}
```

name/type/visibility/mutability, straight from the compiler — and the same artifact's `deployedBytecode.immutableReferences` (present on a default build, no `--ast` needed) is keyed by those same AST ids. The AST can't miss a declaration the way a regex can, so this doesn't just handle more shapes than `DECLARATION_RE` — it deletes the entire `findUnreadableImmutableLines` failure class, not patches it. I confirmed that class is still live today: wrapping the real `wrappedNative` declaration in `AcrossFacet.sol` across three lines (a purely cosmetic reformat, valid Solidity) turns the gate red in both modes:

```
ERROR src/Facets/AcrossFacet.sol:29 mentions immutable but could not be read as a declaration: immutable wrappedNative;
```

An AST read would never produce that false block — the exact thing five of this PR's six fix commits exist to chase down one shape at a time.

Two costs before I'd do the swap: (1) `validateScripts.yml` already runs `forge build src` for TypeChain, but not with `--ast` — adding it needs adding, and I didn't confirm whether changing the output selection forces forge to invalidate the existing cache and do a full recompile (didn't want to burn a full-repo build twice in a review pass to find out). (2) it trades a self-contained ~700-line TS module for a dependency on `out/` staying in sync with `src/`, which this repo has hit as a real failure mode elsewhere (stale `out/` bytecode causing salt collisions, zkEVM build-profile drift across `out/`).

I don't think I should swap the mechanism out mid-drain on a PR that's explicitly scoped as "part (i), mechanism only" (ticket ordering puts the authoring pass and the `--strict` flip after this) — that's an architecture call, not a bug fix, and the file is InfoSec-protected specifically so a change like this gets that review. Leaving this open rather than resolving it: worth doing, but as a deliberate follow-up (before the `--strict` flip would be the natural point) rather than something I unilaterally rewrite in this pass.

script/deploy/immutables/registry-coverage.ts
0xDEnYO: Checked this against the real repo rather than taking the docstring's word for it, since it's making the opposite claim from yours and I wanted to know which is actually true.

The "config-path already lives in the deploy script" link isn't a direct name-based mapping the way that makes it sound. `deployRequirements.json`'s `configData` keys are free-form labels — `helperFunctions.sh:4903` reads them with `jq -r '.[$CONTRACT].configData | select(type=="object") | keys[]'` purely to iterate, then follows `configFileName`/`keyInConfigFile` off each one; nothing ties the label's spelling to the constructor parameter. Reproduced on the worked example:

```
$ grep -n 'wrappedNative\|constructor' src/Facets/AcrossFacet.sol
27: address private immutable wrappedNative;
47: constructor(IAcrossSpokePool _spokePool, address _wrappedNative)
49: wrappedNative = _wrappedNative;

$ jq '.AcrossFacet.configData' script/deploy/resources/deployRequirements.json
{ "_spokePool": {...}, "_wrappedNativeAddress": {...} }
```

Constructor param `_wrappedNative`, immutable `wrappedNative`, configData key `_wrappedNativeAddress` — three spellings, one binding, confirmed against actual source rather than the module's own prose about itself.

Where I think you're onto something real: `configData`'s key order in `deployRequirements.json` does appear to track constructor-parameter order here (`_spokePool` first = ctor param 1, `_wrappedNativeAddress` second = ctor param 2) — so a *positional* AST-derived link, not a name-matched one, might actually hold fleet-wide. I haven't verified that across all 61 contracts with `configData` entries, and it's a real research task (walk every contract's AST constructor params against its `configData` object's key order), not something I did as part of replying to a comment. It would also be silently fragile the moment a config-irrelevant constructor param sits between two config-bound ones, or someone reorders the JSON without realizing order now carries meaning.

So: I don't think "delete the module" holds as-is — the free-form-label problem is real and verified — but "tighten or replace it with positional AST derivation, if the ordering assumption holds" is a legitimate question nobody's answered yet. Leaving this open; it belongs to WP-2.3 part (ii) (the authoring pass this module exists to bootstrap), not this PR's mechanism-only scope, and it's not wired into any CI gate today (only reachable from its own test file), so nothing here is currently load-bearing.

script/deploy/immutables/registry-schema.ts
0xDEnYO: Right on the mechanism, and right that "in a mapping" is wrong — an immutable can't be one, that's storage. That phrasing is in the PR description's summary table, not in the code (grepped `registry-schema.ts` for "mapping" — no hit), so there's nothing in the diff to correct, but I won't repeat that framing going forward.

On the substance: `immutableReferences` (byte offsets into `deployedBytecode`, present on a default `forge build`, no `--ast` needed) does make a getter unnecessary to read an immutable's actual value on a standard EVM chain, and I'd agree that collapses most of what `unverifiable` is for. One thing worth flagging before anyone builds that reader, since it's easy to miss and this repo has already been bitten by it: it only holds for the EVM chains. This fleet also deploys to zkEVM (zksync/zksolc) networks, and this repo already treats those as bytecode-incompatible for exactly this kind of sweep — `reproducibilityAttestations.json` explicitly excludes zkEVM "because the sweep does not record which zksolc version produced the match, and zksolc pins zk codegen." zksolc doesn't splice immutables into deployed bytecode at all; it writes them into `ImmutableSimulator` at a fixed address, indexed by ordinal. `eth_getCode` + `immutableReferences` offsets against a zkEVM deployment reads the wrong thing entirely, not just a missing thing — so "read every immutable from bytecode, no getter needed" is correct for most of the fleet but needs an explicit zkEVM branch (reading `ImmutableSimulator` instead), not a blanket replacement for `unverifiable`.

None of this touches what's actually in #2308 — I checked the full diff (`git diff origin/main...HEAD --stat`) and there is no bytecode or `eth_getCode` read anywhere in it; this PR is pure Solidity-source parsing plus registry cross-checks. `unverifiable` also has zero live entries right now (the registry has one seed contract, `AcrossFacet`, both sourced as `config`), so this is a schema question for whoever builds the drift reader in part (ii)/#2213, not a defect in what's shipping here. Leaving open for that — flagging the zkEVM caveat so it's on the table when someone picks it up rather than discovered after the fact.

script/deploy/immutables/immutable-declarations.ts
0xDEnYO: Agreed, and it's a fair hit against `000-global-standards` — trimmed both docstrings you'd flag hardest in this file. The top-of-file comment no longer names the specific `address payable public immutable` shape that motivated it, and `maskCommentsAndStrings`'s docstring no longer narrates the two rejected masking approaches ("pairing quotes with a regex", "skipping any line that opens a comment") — both now just state the invariant a maintainer needs (why masking has to happen at all) without the history of what was tried first. Left the technical asides that explain a non-obvious *why* in the regex itself (the `$`-in-identifier / word-boundary issue, why `transient` isn't excluded) since those are decisions a future editor of the pattern actually needs, not narrated debate.

109 tests still pass (`bun test script/deploy/immutables/`), no behavior change. Pushed as part of this round; resolving.

script/deploy/immutables/immutable-declarations.ts
0xDEnYO: You are right that the compiler knows this and a regex does not, and I would rather not ship a parser if the compiler can answer it. I measured what is actually available before answering, and there is a gap worth knowing about.

`immutableReferences` **is** in the default artifact, but it is keyed by AST node id and carries no names:

```text
out/EmergencyPauseFacet.sol/EmergencyPauseFacet.json
  deployedBytecode.immutableReferences =
    {"16224": [{"start":204,"length":32}, ...], "16231": [...]}
```

So on its own it gives the count and the byte offsets — enough to read values out of `eth_getCode`, not enough to say which immutable is which. Mapping node id to name needs the AST, and this repo does not emit one:

```text
$ grep -n "extra_output" foundry.toml     -> no match
$ python -c "...out/build-info/<latest>.json"
  top keys: ["id", "language", "source_id_to_path"]
  sources in output: 0
```

Foundry's current build-info is a stub, so the AST would have to be turned on repo-wide (`extra_output = ["ast"]`, or `--ast` on the build) — a change to every build's artifact size and time, and a new dependency for this check on the `validateScripts` job actually having compiled.

So I read your point as correct but as a two-step: the parser is the wrong long-term answer, and the right one is not free. Rather than quietly rewrite it under a review round, I would like your call on which you prefer:

1. Enable AST output and derive declarations from it, dropping the parser and the whole `findUnreadableImmutableLines` hard-fail class with it.
2. Keep the source scan for names, but take actual values from `immutableReferences` + `eth_getCode`, which needs no AST and, as you note on `registry-schema.ts`, collapses the `unverifiable` category to near-empty.
3. Land as-is and do either as a follow-up.

My preference is (2) now and (1) as a follow-up: (2) removes the category you flagged and the getter dependency without touching the build config, while (1) is the larger change and wants its own PR since it affects every compile.

Also taking your correction on the wording: an immutable is spliced into bytecode, never in a mapping — that phrasing in `registry-schema.ts` is wrong regardless of which option we take, and I will fix it either way.

script/deploy/immutables/registry-coverage.ts
0xDEnYO: Agreed that this is largely derivable, and the direction is right: immutable ← constructor-param is in the AST, and constructor-param ← config-path is already in the deploy script.

The blocker is the same one as on `immutable-declarations.ts`: this repo does not emit an AST today (`extra_output` is unset and Foundry's build-info is a stub with no `sources`), so the first half of that chain is not readable yet. The second half is also less uniform than it looks — the EVM deploy scripts reach config through several shapes, so "the same config path the deployer reads" is a per-script convention rather than one lookup.

I would rather not half-derive it: a suggester that derives some links and hand-authors the rest is harder to reason about than one that is honestly hand-authored. Tied to the same call I asked for on the other thread — if we enable the AST, this becomes derivable properly and the suggester shrinks to the underivable residue, which is the outcome you are describing and the one I would want.

script/deploy/immutables/registry-schema.ts
0xDEnYO: Correct on both counts, and the second is a plain error on my side: an immutable is spliced into the deployed bytecode, so it is never "written to a mapping" — that would be storage. I will fix that wording regardless of what we decide about the rest.

On the substance: `immutableReferences` gives the byte offsets, so the value is readable from `eth_getCode` with or without a getter, and that does collapse `unverifiable` to near-empty and removes a `source` kind and its branches. Verified the data is there in the default artifact (offsets present, names absent — see the other thread). This is the strongest of your three points because it needs no build-config change, and it is option (2) in the call I asked for on `immutable-declarations.ts`.

0xDEnYO:
Both changes-requested items are addressed in `950212ff4`.

**1. `src/**`-only PRs no longer bypass the gate.** Took your option (b): the check reads Solidity source and two JSON files — no forge build, no TypeChain — so putting `src/**` on the heavy job would have bought a full compile on every Solidity PR for nothing. It is now its own job gated on `src/**/*.sol` plus the registry files, costing a `bun install`.

Doing that surfaced a second problem worth naming: adding the job to the aggregator's `needs` is not enough. `validate-scripts-required` only read `validate-scripts.result`, so a **red** registry job would have been reported green — the exact failure the aggregator exists to prevent. It now loops over both, and I checked all 16 result combinations:

```text
validate=success   registry=failure   -> FAIL (verify-immutable-registry=failure)
validate=success   registry=cancelled -> FAIL (verify-immutable-registry=cancelled)
validate=skipped   registry=failure   -> FAIL (verify-immutable-registry=failure)
validate=success   registry=skipped   -> OK
... 16/16 as intended
```

**2. Entry-point tests.** `mergeRequirements` and the exit routing (now `decideExit`) are exported and unit-tested; `main()` is behind `import.meta.main` so the module can be imported. The CLI cases spawn the real script in a throwaway git repo, because neither the file set (`git ls-files`) nor the exit code is observable from the pure functions — same pattern as `verify-approvals.test.ts`.

15 new tests covering every case you listed. They are not self-confirming — each mutation kills both a unit test and a CLI test:

| Mutation | Killed by |
|---|---|
| `--strict` ignored | `decideExit > exits 1 on warnings alone with --strict`, `CLI > exits 1 for the same input with --strict` |
| `mergeRequirements` drops registry-only contracts | 2 unit tests + `CLI > exits 1 on a malformed registry` |
| unreadable declarations stop failing | `decideExit > exits 1 on an unreadable declaration`, `CLI > exits 1 on a declaration it cannot read` |

The `git ls-files` case puts the identical contract at `src/`, `test/` and `script/` and asserts only the `src/` one is named — a leaking glob fails it.

Full suite 2026 pass / 0 fail.

Note the `--strict` flip itself is still EXSC-869 part (ii); this closes the prerequisite you identified, it does not flip it.

Separately, I have replied on your three architectural threads about reading the compiler instead of parsing source — short version: `immutableReferences` is in the artifact but carries no names, and this repo emits no AST, so it needs a build-config change. Options and a recommendation are on the `immutable-declarations.ts` thread; that one is your call.

script/deploy/immutables/immutable-declarations.ts
gvladika: Suggesting the AST swap gets done in this PR rather than as a follow-up. Two things move me off the "land as-is" option I was leaning toward.

**The blocker doesn't hold.** `--ast` is a build-invocation flag, and `extra_output` can sit in a dedicated profile — already this repo's pattern (`[profile.zksync]` redirects `out` and `cache_path`; `[profile.solc_floor]` exists for one CI check). Verified rather than assumed:

```text
FOUNDRY_OUT=…/out-imm FOUNDRY_CACHE_PATH=…/cache-imm forge build src --ast
  7.65s, 195 artifacts with an AST
  default out/AcrossFacet.json md5: identical before and after
```

Nothing lands on anyone else's build. The real cost is that the gate job needs a cached `setup-foundry` plus ~8s — not a repo-wide change.

**Your `immutableReferences` finding is right, and it argues for (1) over (2).** I reproduced your `EmergencyPauseFacet` numbers exactly — same offsets, no names, no `.ast`. But the ids are a snapshot, not an identifier:

```text
your comment            : 16224 / 16231
isolated --ast build    : 16383 / 16390
repo default out/       : 82539 / 82546
```

Node ids are assigned per compilation unit, so they shift with the source set. The name↔offset join is only well-defined **inside one build's output** — which `--ast` gives for free. Without it, option (2) can only pair name to offsets by ordinal (sorted ids vs declaration order), which is the same unasserted positional assumption you argued against on the `registry-coverage.ts` thread. It holds today (no `src/` contract inherits immutables — I checked, zero cases), but nothing asserts it.

**Why here rather than a follow-up:** `script/deploy/immutables/**` is covered by the matcher this PR adds, and `protect-critical-code` already needs an InfoSec approval on this PR. Doing it here costs zero extra approval; deferring costs a second InfoSec cycle on the same protected directory. On top of that the registry has one seed entry so nothing has to be re-validated, `--strict` is blocked on this anyway, and it's a net deletion (239 + 452 lines, plus the whole `findUnreadableImmutableLines` class).

For the record your parser is correct today — I diffed it against the AST over all of `src/`: 80 vs 80, identical sets, 0 unreadable. This isn't a defect, it's deleting a failure class at the cheapest moment to do it.

If you agree, it scopes small: swap the body of `immutable-declarations.ts` behind the same export, drop `findUnreadableImmutableLines` and its tests, add `setup-foundry` + the profile build to the new job. `registry-schema.ts` and the entry point don't move.

Your call in the end, but I think the case for doing it now is stronger than deferring it.

script/deploy/immutables/registry-coverage.ts
gvladika: The "repo emits no AST today" half of this doesn't hold — see the other thread (`--ast` is a per-invocation flag, a dedicated profile isolates it, default `out/` verified byte-identical). So immutable ← constructor-param becomes readable the moment that swap lands.

Your second reason stands on its own though: constructor-param ← config-path really is a per-script convention rather than one lookup, and a half-derived suggester is worse than an honestly hand-authored one. Fine to carry to part (ii).

Good instinct on the positional hypothesis — if the `configData` key order really does track ctor-param order fleet-wide, it wants an assertion in the gate, not an assumption in a suggester.

Separate and smaller: this module is imported by nothing but its own test (113 + 153 lines). Either land it with the authoring pass that consumes it, or keep it with a one-line header note saying it's a deliberate part-(ii) bootstrap, not wired into any gate. The second is fine by me — I just don't want the next reader to have to work out whether it's dead.

script/deploy/immutables/registry-schema.ts
gvladika: Agreed, and the zkEVM catch is a good one I didn't have — zksolc writing immutables into `ImmutableSimulator` by ordinal means a naive `eth_getCode` + offsets reader returns the wrong bytes there, not missing ones. Worth carrying into whatever picks this up rather than leaving it only in this thread.

You're right there's no defect here: `unverifiable` has zero live entries and nothing in this diff reads bytecode. Keep the category — with your own caveat, "collapses to near-empty" holds for most of the fleet but not all of it.

One thing still outstanding: you said you'd fix the "written to a mapping or dynamic array" wording regardless. You were right it isn't in `registry-schema.ts` (grepped, no hit) — but it's still live in the **PR description's** summary table, which is what a reader hits first.

.github/workflows/validateScripts.yml
```diff
@@ -67,6 +72,17 @@ jobs:
...
+            # adding a facet has to be able to run the gate without also
+            # touching script tooling.
+            immutables:
+              - 'src/**/*.sol'
+              - 'script/deploy/immutables/**/*.ts'
+              - 'script/deploy/resources/immutableRegistry.json'
+              - 'script/deploy/resources/deployRequirements.json'
+              - 'package.json'
+              - 'bun.lock'
+              - '.github/workflows/validateScripts.yml'
```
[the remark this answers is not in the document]
0xDEnYO: Fixed, and widened past the two paths you named — they were the weaker half.

`foundry.toml` pins `solc_version = '0.8.29'` with `auto_detect_solc = false`, and its `skip`/`src`/`libs` decide which files get compiled at all. It was in neither filter. Demonstrated on a real build: adding `skip = ['src/Facets/**']` to `[profile.default]` — a `foundry.toml`-only change matching nothing in either filter — takes the enumeration from 79 immutables to 24 and hard-fails the gate with 57 unenumerated sources, on a PR where the job never runs. `validate-scripts-required` treats the skipped job as pass, so that reports green.

`immutables` now carries `.foundry-version`, `.github/actions/setup-foundry/**`, `foundry.toml`, `remappings.txt`, `lib/**` and `.gitmodules` — the list `setup-foundry` already keys its build cache on, minus the scopes this job does not build. Verified each path matches the filter and that unrelated files still skip.

script/deploy/immutables/immutable-ast.ts
```diff
@@ -0,0 +1,209 @@
...
+        if (!member.name || !member.src) continue
+
+        let starts = offsetsByFile.get(file)
+        if (!starts) {
+          starts = lineOffsets(file)
+          offsetsByFile.set(file, starts)
+        }
+
+        const byteOffset = Number(member.src.split(':')[0])
+        byPosition.set(`${file}:${member.src}`, {
```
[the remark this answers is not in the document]
0xDEnYO: Fixed as suggested: `IImmutableDeclaration` now carries the enclosing `ContractDefinition` name, and both `validateImmutableRegistry` and `assessRegistryCoverage` key on it — the two `contractOf` basename helpers are gone.

Proven on a real `forge build src --ast`, not only in the unit suite. Appending a second contract to `src/Facets/AcrossFacet.sol`:

```
contract=AcrossFacet   name=spokePool      line=23
contract=AcrossFacet   name=wrappedNative  line=27
contract=AcrossHelper  name=spokePool      line=145
```

Against the real `immutableRegistry.json`: basename keying reports `errors: [] warnings: []` — the helper's undocumented immutable is absorbed by AcrossFacet's entry. Contract-name keying warns `AcrossHelper.spokePool has no registry entry`. The basename-collision half was confirmed the same way: a second file declaring `ScratchDupFacet` compiles into the same `out-immutables/AcrossFacet.sol/` directory, where the old keying produced a false "has a registry entry but does not declare that immutable" error.

Regression cases for both layouts are in `registry-schema.test.ts` and `registry-coverage.test.ts`, plus multi-contract attribution in `immutable-ast.test.ts`. Each was checked to fail with the production change reverted.

On today's tree the change is a no-op: 79 declarations, zero where the AST contract name differs from the file basename, zero files with more than one immutable-declaring contract, zero basename collisions under `src/`. It only changes what happens for the two layouts it fixes.

script/deploy/immutables/verify-immutable-registry.test.ts
```diff
@@ -0,0 +1,284 @@
...
+            },
+          ],
+        },
+      ],
+    },
+  })
+
+/** Builds a throwaway repo the CLI can run against, and returns its path. */
+function makeRepo(files: Record<string, string>): string {
+  const root = mkdtempSync(join(tmpdir(), 'immutable-registry-'))
```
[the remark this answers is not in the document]
0xDEnYO: Fixed: the roots are tracked and removed with `rmSync(..., { recursive: true, force: true })` in an `afterEach`, matching the pattern `immutable-ast.test.ts` already uses. The leak was real — `${TMPDIR}` had 126 stale `immutable-registry-*` directories from previous runs; the count is now flat across a full run.

0xDEnYO:
### Review-gate residual — two items left for a decision

Both sit in the `verify-immutable-registry` job this PR adds, and neither is fixed here because each has more than one valid answer.

**1. `cache-key-suffix: immutable-ast` on that job is a no-op** (`.github/workflows/validateScripts.yml:170-182`). `setup-foundry` caches `cache` and `out`, while the job builds with `FOUNDRY_OUT=out-immutables` / `FOUNDRY_CACHE_PATH=cache-immutables` — so it stores and restores directories the job never touches and recompiles all of `src/` from scratch every run. The step comment describes collision-avoidance; what is actually happening is total cache absence. Either give the action the gate's directories to cache, or drop the input and say plainly that this job is uncached. (Worth noting the absence is load-bearing in the good direction today: a cached `out-immutables` would carry stale artifacts of deleted contracts into the enumeration.)

**2. The workflow has no `push` trigger** (lines 25-30), so `main` is never re-verified. Widening the `immutables` filter closes the pre-merge hole, but if drift ever does land, nothing surfaces it until an unrelated PR happens to touch a matching path. Adding a `push` trigger is out of this PR's scope; flagging it as the remaining gap.

gvladika (dismissed):

script/deploy/immutables/immutable-declarations.ts
0xDEnYO: Taken in full. The parser and the whole `findUnreadableImmutableLines` class are gone — the enumeration is now `forge build src --ast`, and `findSourcesWithoutAst` replaces the unreadable-line hard-fail with a question the compiler can answer definitely: a source file it emitted no AST for fails the gate, instead of an unparseable shape contributing nothing.

Measured both enumerations against each other before swapping, on the whole tree: 79 immutables each, no set difference in either direction, no line-number drift. So the swap is behaviour-preserving here and only removes the class of shapes a regex could miss.

You were also right that the `--ast` blocker didn't hold.

script/deploy/immutables/registry-coverage.ts
0xDEnYO: Agreed, and the blocker you dismantled is now gone for real — the AST is available in this job, so immutable ← constructor-param is derivable today.

Not doing it in this PR, and tracked as [EXSC-935](https://linear.app/lifi-linear/issue/EXSC-935/derive-immutable-expectations-instead-of-hand-authoring-them-and-read) instead. Reason is scope rather than disagreement: this PR has been through two rounds and the AST swap already replaced its core, and the derivation is not the small half of the work — see below.

What's left after the AST gives you the first half: `configData` keys are free-form labels, not constructor parameters. `AcrossFacet` files its requirement under `_wrappedNativeAddress` for a parameter called `_wrappedNative`, and nothing constrains that spelling — `helperFunctions.sh` only ever reads `configData | keys[]` and then follows `configFileName`/`keyInConfigFile`. So param → config-path has to be read out of the deploy script that constructs the contract, which is the part worth doing properly. Getting that right is what makes the registry and the deploy wiring unable to drift by construction, which is your actual point and the reason the ticket exists rather than a "won't do".

The suggester stays as-is for now, still explicitly a bootstrap for the authoring pass and not a gate.

script/deploy/immutables/registry-schema.ts
0xDEnYO: Agreed on the mechanism, and the "written to a mapping" wording was simply wrong — that's storage, an immutable can't be one. It was in the PR description's summary table, not the code, and it's fixed there.

Keeping `unverifiable` in this PR, tracked as [EXSC-935](https://linear.app/lifi-linear/issue/EXSC-935/derive-immutable-expectations-instead-of-hand-authoring-them-and-read). Collapsing it means building a bytecode-actuals reader (`immutableReferences` offsets + `eth_getCode`), which is a bigger change than removing a `source` kind, and it has to carry your own zkEVM caveat: zksolc doesn't splice immutables into the bytecode at all — it writes them into `ImmutableSimulator` at `0x…8005` indexed by ordinal × 32, so a naive offsets reader returns the wrong bytes there rather than missing ones. That exception is on the ticket so it doesn't get rediscovered later.

melianessa (approved):

script/deploy/immutables/getter-exemptions.json
```diff
@@ -0,0 +1,41 @@
+{
+  "AcrossFacetPackedV4.SPOKEPOOL": "config-backed and annotatable; annotation deferred until a fleet dry run confirms it matches on chain",
```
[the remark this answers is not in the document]
0xDEnYO: Correct on substance, and it applies to all 34 `config-backed and annotatable` rows, not just this one — while a getter sits in this file the gate proves the getter is known but does not compare the deployed value against `config/`.

Deferring rather than fixing here: annotating before a fleet dry run risks turning the healthcheck red on a live chain, which is exactly what the reason string in each row records. Tracked with the full annotation pass (plus the `601-healthcheck-invariants.md` "may only shrink" wording nit from this review) in EXSC-936, related to WP-2.4 and the WP-2.3 replay ticket.

gvladika (approved):

PR #2309, opened 2026-09-02
feat(healthcheck): whitelist invariant reads the timelock queue — queued syncs report as expected-pending (EXSC-918)

0xDEnYO:

Fixes [EXSC-918](https://linear.app/lifi-linear/issue/EXSC-918/intent-aware-whitelist-invariant-a-queued-whitelist-sync-reports-as)

Follow-up to EXSC-847 (#2270), which made `facets-registered` and `periphery-registered` intent-aware. The whitelist check was the one invariant of that family still reporting a rollout in flight as drift.

`config/whitelist.json` is merged before the Safe proposal syncing it executes, so in between, `checkWhitelistIntegrity` errored on every pair the queued operation was about to add — on both of its steps (`Source of Truth FAILED` from `isContractSelectorWhitelisted`, and `Pair Array is missing N pairs from config` from `getAllContractSelectorPairs`). Live today: `robinhood` red on `0xCC89feed…8599 / 0xf8989325` while its sync operation sat `queued`; same shape on `arc` (09-01) and `injective` (08-27).

Two independent reasons #2270's machinery didn't already cover it, both addressed here:

- **The decoder ignored whitelist calls.** `extractRegistrations` handled only `diamondCut` and `registerPeripheryContract` — a whitelist batch was explicitly documented as contributing nothing. It now also decodes `batchSetContractSelectorWhitelist` — the call `diamondSyncWhitelist.sh` and `proposePeripheryWithWhitelist.ts` actually emit — plus the singular `setContractSelectorWhitelist` defensively, both only with `_whitelisted` **true**; `false` un-whitelists and is ignored, the counterpart of a `Remove` cut. A length-mismatched batch reverts on-chain (`InvalidConfig`), so it decodes to nothing rather than to the pairs it could have paired up.
- **The downgrade was wired only into the two registration invariants.** A new pure `splitByPendingWhitelist` splits missing pairs by whether a queued operation whitelists **exactly** that contract *and* selector on **that** network's diamond. Matching the address alone would report a sync that is never coming — the same reasoning that makes the periphery caller match the registry name.

Design points worth a reviewer's attention:

- **Records now carry an explicit `kind`** (`facet-cut` / `periphery` / `whitelist`). `facets-registered` previously inferred "this is a cut" from `peripheryName === undefined`, which a whitelist record would have satisfied — a latent false-green on an error gate. Each consumer now matches its own shape, and there's a regression test that a queued whitelist entry does **not** downgrade a missing facet.
- **One lazy queue read, shared by both steps.** Resolved only when a pair is actually missing, so a synced network never touches MongoDB (tested).
- **An unreachable queue keeps every error** and adds a coverage warning, matching `facets-registered`: the pair comparison stands on an independent on-chain signal, and a Mongo blip turning a genuinely unsynced whitelist green is far worse than a false alert during a rollout.
- **Additions only.** A stale pair (on-chain, no longer in config) still errors even while its removal is queued — the removal side reads parked tasks, which carry no whitelist payloads. Called out in `[CONV:HEALTHCHECK-INTENT]`.
- The two "is synced" success lines now name the pending count instead of claiming the diamond already holds what only a queued operation will put there.

**Verified against the live queue (904 rows), not just fixtures.** The single `queued` row is robinhood's sync — carrying exactly the red pair and targeting robinhood's own diamond (`0xB477751B…14Af3`) — so this turns today's alert into expected-pending. Every inner-call selector present across all 904 rows was resolved: `batchSetContractSelectorWhitelist` appears 266 times (408 whitelist records, of which 91 are `_whitelisted=false` and correctly decode to nothing), and none of the undecoded selectors writes whitelist storage. The legacy writers that would have been a genuine coverage gap (`batchAddDex`, `batchSetFunctionApprovalBySignature`, `setContractWhitelist`) appear zero times and have no emitter left in the repo. 74 distinct networks have carried a whitelist row historically, so the behaviour change is fleet-wide, not robinhood-specific.

Mutation-tested each new guard rather than trusting the suite: dropping `kind === 'whitelist'`, dropping `kind === 'periphery'`, widening the selector comparison, or removing the "skip the queue when nothing is missing" short-circuit each fail exactly one dedicated test. The first two initially passed — the fixtures standing in for "a record of another kind" did not carry the field the surviving predicate compares, so the comparison did the rejecting and the discriminator was never exercised. Fixed in the second commit.

### DIFF
--- a/.agents/rules/601-healthcheck-invariants.md
+++ b/.agents/rules/601-healthcheck-invariants.md
@@ -95,7 +95,8 @@ invariant:
 - A rollout proposed **without** `--timelock` writes no row at all, and so is never downgraded.
 - Only whitelist **additions** are covered. A pair the diamond holds and config no longer
   declares (`Pair Array has N stale pairs`) reports as a hard error even while its removal
-  is queued — the removal side reads parked tasks, which carry no whitelist payloads.
+  is queued: that branch consults no intent source at all, and the parked-task queue the
+  other removal gate reads carries no whitelist payloads to consult.
 - **Tron** rolls out through `contracts-tron` and has no EVM queue row; it is skipped by branch.
 - A row is honoured only while it is plausibly still waiting. A never-scheduled or
   directly-cancelled operation is reported and skipped by the execution runner *without* a
@@ -111,10 +112,10 @@ Two boundaries are not negotiable:
   compensating write. Deploy logs stay a pure function of the loupe
   ([docs/DeploymentLogs.md](../../docs/DeploymentLogs.md)).
 - **An unreachable queue must never suppress a finding.** What decides the degradation is
-  what the check is *for*, not its severity — all three are error-severity.
+  what the check is *for*, not its severity — all four are error-severity.
   `no-stale-registered-facets` exists _only_ to police queue coverage, so without the queue
   every finding it could make is noise: it skips and reports the reduced coverage.
-  `facets-registered` and `periphery-registered` stand on an independent on-chain signal, so
-  they keep every error and add a warning naming the degraded coverage — a MongoDB blip
-  turning genuinely missing registrations green is far worse than a false alert during a
-  rollout.
+  `facets-registered`, `periphery-registered` and `whitelist-integrity` stand on an
+  independent on-chain signal, so they keep every error and add a warning naming the
+  degraded coverage — a MongoDB blip turning genuinely missing registrations green is far
+  worse than a false alert during a rollout.
--- a/.agents/rules/601-healthcheck-invariants.md
+++ b/.agents/rules/601-healthcheck-invariants.md
@@ -61,10 +61,10 @@ JSDoc on exports, `bunx eslint` + `bunx tsc-files --noEmit`).
 ## Intent-aware invariants, chain-only generators ([CONV:HEALTHCHECK-INTENT])
 
 Several invariants compare on-chain reality against a _desired_ state that a merged PR
-already records — `_targetState.json` for `facets-registered` and `periphery-registered`, a
-deleted source file for `no-stale-registered-facets`. Between that merge and the multisig
-operation acting on it the two legitimately disagree, and the remediation is "wait", not
-"fix".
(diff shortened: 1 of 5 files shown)


0xDEnYO:
Review-gate pass (local trial) — nothing escalated, no critical findings. Recording what was **not** fixed, so a reviewer does not have to re-derive it:

- **`STALE_QUEUE_GRACE_MS` headroom is thinnest on exactly this consumer.** Across 901 executed queue rows the create→execute spread is p50 3.3 h / max 70.7 h, and that fleet-wide max **is** a whitelist row (facet-cut rows top out at 20.2 h). The window is delay + 72 h ≈ 75 h, so ~4.3 h of margin; 0 rows have ever exceeded it. Left alone — widening it would weaken the never-landed-cut gate the bound exists for, and the failure direction is over-alerting.
- **`blocked` rows report as findings.** `listPendingRegistrationsByNetwork` reads `status: 'queued'` only, so a whitelist sync sitting `blocked` still errors even though it becomes executable once an operator clears the cause. Nothing bounds how long that takes, so honouring it would be unbounded masking. Pre-existing from #2270; this PR adds a third consumer to it, and the docstring now names the case instead of implying the status list is exhaustive.
- **The "that network's diamond" filter has no whitelist-path test.** It lives in `resolvePendingRegistrations` (`record.target === diamond`), is locked by `facets-registered`'s own wrong-target test, and is verified on real data here — but `splitByPendingWhitelist` does not re-check it. Documented as a caller-owned precondition rather than duplicated.
- **A coverage warning is lost if step 2 throws after step 1 recorded an unreachable queue.** Cosmetic: every error is still reported, and the throw itself is logged.

Also worth knowing, since it changes when this code touches MongoDB: the queue read now fires on *any* missing whitelist pair, which is more common fleet-wide than a missing facet. During a Mongo outage the "queue unreachable" warning will therefore appear on materially more networks than #2270's did. Errors are preserved in that state.

0xDEnYO:
@coderabbitai review

0xDEnYO:
@coderabbitai full review

gvladika (approved):

PR #2310, opened 2026-09-02
feat(healthcheck): enforce the getter annotation for new immutable bindings (EXSC-919)

0xDEnYO:

Fixes [EXSC-919](https://linear.app/lifi-linear/issue/EXSC-919/mechanically-enforce-the-getter-annotation-for-new-immutable-bindings)

Follow-up to EXSC-784 / #2213 (merged). Deferred backlog it produced: EXSC-917.

Coverage of the `immutable-bindings-match-config` invariant added in #2213 is **opt-in** — it only checks constructor args carrying a `getter` annotation in `deployRequirements.json`. `immutableBindings.test.ts` already proves existing annotations are sound, but nothing asserted that a *new* immutably-bound counterparty gets annotated at all, so the bug class #2213 closed reopens silently for any contract added without one.

> **Stacked on #2308** (`feat/exsc-869-immutable-registry`) — review that one first. The diff here is the getter gate only.

The `verify-immutable-registry` CI job now fails when a public immutable address getter in the deployed source trees (`src/Facets`, `src/Periphery`, `src/Security`) is neither annotated nor recorded in `script/deploy/immutables/getter-exemptions.json` with a reason.

**Why it reads the compiler AST.** Revised on review (@gvladika, and he was right). The earlier version parsed Solidity source, because the only job running `bun test:ts` is deliberately Foundry-free. The swap moves the gate out of that suite entirely and into #2308's `verify-immutable-registry` job, which already compiles `src/` with `--ast` for the registry gate — so this rides on the same build rather than carrying a second reader. What that buys is exactness: `visibility` and the declared type are the compiler's answers, so a `private` immutable (no generated getter) and an `enum`/`uint256` (cannot hold a counterparty) need no heuristic. It retired the parser, its comment-stripping, and the enum/user-defined-type scan of all of `src/`.

Measured before changing anything, on the current tree: **the two enumerations agree exactly — 79 immutables each, no set difference in either direction, no line-number mismatches.** So the recorded exemptions still describe the code, and the swap is behaviour-preserving here.

**Why the two gates share a job.** They enumerate the same immutables to ask two questions — "does the registry account for it?" and "is its getter checked or exempt?". One AST build, one job, two checks; a second job would compile `src/` again to reach the same list.

**Why exemptions moved from a TypeScript constant to JSON.** A compiled-in constant is a claim about one repo. Running the gate anywhere else — including its own CLI tests, which build a synthetic repo — reported all 39 entries as stale. As data it is read from the repo under test. `checkDeployRequirements()` iterates `.configData | keys[]`, so a sibling key there still breaks deploys, which is why this is its own file rather than a nested key; it sits under `script/deploy/immutables/`, where the InfoSec protection already covers it.

**Falsification.** Both directions were made to fire against the real repo, not just against fixtures: deleting one real exemption (`ChainflipFacet.CHAINFLIP_VAULT`) makes the gate name that getter and exit 1; inventing an exemption for a getter no contract declares makes it name that instead. Restoring the file returns exit 0.

**Known limitation, stated plainly.** The gate enforces *"if you expose a getter, annotate it"*; it cannot force a contract to expose one. 25 counterparties are bound to `private`/`internal` immutables (`AcrossFacet.spokePool`, `SquidFacet.squidRouter`, `SymbiosisFacet.*`) and stay unverifiable on chain. `101-solidity-contracts.md` now requires `public immutable` for new external protocol addresses, so this shrinks going forward.

**Documentation**, so the annotation is part of facet authoring rather than health-check lore: `601-healthcheck-invariants.md` (gate mechanics), `101-solidity-contracts.md` (authoring requirement — fires on any `src/**/*.sol` edit), `099-finish.md` (finish checklist), `docs/AddingANewBridge.md` (new-bridge walkthrough), `templates/facet.template.hbs` (at the scaffolded immutable). The first two were updated for the AST swap; `ts-unit-tests.yml` goes back to its original triggers, since the gate no longer lives in that suite.


script/deploy/shared/immutableGetterCoverage.ts
```diff
@@ -0,0 +1,229 @@
...
+ * Matches a `<type> public immutable <NAME>;` declaration.
+ *
+ * @remarks Solidity accepts the two specifiers in either order, and the whole declaration may be
+ *   wrapped across lines, so neither is assumed — a form this pattern could not read would be a
+ *   getter the gate never sees, which is the one failure mode it must not have. `address payable`
+ *   is spelled out as the one two-word type in use; `public immutable` appears only in state
+ *   variable declarations, so no line anchor is needed to avoid matching inside a function body.
+ */
+const PUBLIC_IMMUTABLE_DECLARATION =
+  /(address\s+payable|[A-Za-z_][A-Za-z0-9_]*)\s+(?:public\s+immutable|immutable\s+public)\s+([A-Za-z_][A-Za-z0-9_]*)\s*;/g
```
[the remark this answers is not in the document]
0xDEnYO: Fixed in eeae811d1 — the pattern now accepts an optional inline initializer, and a test covers it.

One correction on the evidence: there are **no** inline-initialized public immutables in the gated trees today. The `rg` pattern above allows `[\s\S]{0,200}?` between the declaration and the `=`, so it matched across unrelated following statements rather than an initializer on the declaration itself; `grep -rnE "(public[ \t]+immutable|immutable[ \t]+public)[ \t]+[A-Za-z_][A-Za-z0-9_]*[ \t]*="` over `src/Facets src/Periphery src/Security` returns 0.

The concern still stands as robustness, which is why it is fixed: this gate exists so an unannotated binding cannot pass silently, and a declaration form the parser cannot read is precisely a getter it never sees. Detection still resolves the same 49 getters, so the widening introduced no false positives.

.agents/rules/101-solidity-contracts.md
```diff
@@ -15,4 +15,5 @@ paths:
...
 - When adding a new deployable contract that has constructor arguments, add/update its entry in `script/deploy/resources/deployRequirements.json` so deploy scripts validate the presence of required constructor args.
+- Constructor args that bind an **external protocol address** (spoke pool, vault, router, portal) must be stored in a `public immutable` and annotated with `getter` in that same `deployRequirements.json` entry, so `immutable-bindings-match-config` compares the bound address against config on every chain. A `private`/`internal` immutable exposes nothing to read and leaves the binding permanently unverifiable — immutably, since only a redeploy can change it. `immutableGetterCoverage.test.ts` fails on a public immutable address getter that is neither annotated nor recorded as exempt; see `601-healthcheck-invariants`.
```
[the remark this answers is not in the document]
0xDEnYO: Fixed in eeae811d1 — you are right that the wording overstated it. A `private`/`internal` immutable generates no getter, but a contract can still expose the value through an explicit public accessor, so "exposes nothing to read" and "permanently unverifiable" were wrong.

Reworded to say that such an immutable generates no getter and therefore cannot be read on chain unless the contract adds an explicit accessor. Also added the coverage nuance your comment implies: the gate detects compiler-generated getters, so a hand-written accessor over a non-public immutable is invisible to it and has to be annotated deliberately.

script/deploy/shared/immutableGetterCoverage.ts
```diff
@@ -0,0 +1,230 @@
...
+ *
+ * @remarks Solidity accepts the two specifiers in either order, allows an inline initializer in
+ *   place of constructor assignment, and lets the declaration wrap across lines, so none of those
+ *   is assumed — a form this pattern could not read would be a getter the gate never sees, which
+ *   is the one failure mode it must not have. `address payable` is spelled out as the one
+ *   two-word type in use; `public immutable` appears only in state variable declarations, so no
+ *   line anchor is needed to avoid matching inside a function body.
+ */
+const PUBLIC_IMMUTABLE_DECLARATION =
+  /(address\s+payable|[A-Za-z_][A-Za-z0-9_]*)\s+(?:public\s+immutable|immutable\s+public)\s+([A-Za-z_][A-Za-z0-9_]*)\s*(?:=[^;]*)?;/g
```
gvladika, line 42: **Why regex-parse source instead of reading the compiler's AST?**

The merged `immutable-bindings-match-config` check reads live chain state, so it never had to parse Solidity. This gate introduces source-parsing only to police annotations — but the compiler already knows precisely which state variables are `immutable` (the AST node carries `mutability: "immutable"`; the ABI alone can't tell immutable from constant from mutable). Reading build-info AST would drop this parser and the "a Solidity style the regex can't read is a getter we never see" failure mode the calibration tests guard against.

The stated reason is that `ts-unit-tests.yml` is Foundry-free, so `out/` doesn't exist and an artifact gate would skip. That's sound — but it's a choice, not a constraint: the same check could live in `forge-unit-tests.yml` (which compiles) and read the AST. Could you note in the PR why it stays in the TS job rather than the compiled one? "Decoupled from a build" is a fine answer; it just shouldn't be implicit.

**Concrete cost of the regex — comment false-positive.** The pattern matches raw text, so a NatSpec/commented line trips it:

```
parsePublicImmutableGetters('    /// e.g. address public immutable ROUTER;')
  -> [{ getter: 'ROUTER', solidityType: 'address' }]
```

No such comment exists in the scanned trees today, so the gate is green and the "re-derives all 10 annotations" calibration holds — but a future author writing an example in NatSpec gets a spurious failure demanding they annotate a getter that doesn't exist. This is the mirror of the inheritance blind spot already called out (same AST-vs-regex root). Same category covers user-defined value types and enums, which `NON_ADDRESS_VALUE_TYPE` doesn't exclude. If the check stays regex-based, stripping comments before matching closes it.

Not a blocker — the ratchet design is the right shape. This is about where the check lives and the parser's edges.

0xDEnYO: You are right that the compiler knows this and a regex does not, and I would rather not ship a parser if the compiler can answer it. I measured what is actually available before answering, and there is a gap worth knowing about.

`immutableReferences` **is** in the default artifact, but it is keyed by AST node id and carries no names:

```text
out/EmergencyPauseFacet.sol/EmergencyPauseFacet.json
  deployedBytecode.immutableReferences =
    {"16224": [{"start":204,"length":32}, ...], "16231": [...]}
```

So on its own it gives the count and the byte offsets — enough to read values out of `eth_getCode`, not enough to say which immutable is which. Mapping node id to name needs the AST, and this repo does not emit one:

```text
$ grep -n "extra_output" foundry.toml     -> no match
$ python -c "...out/build-info/<latest>.json"
  top keys: ["id", "language", "source_id_to_path"]
  sources in output: 0
```

Foundry's current build-info is a stub, so the AST would have to be turned on repo-wide (`extra_output = ["ast"]`, or `--ast` on the build) — a change to every build's artifact size and time, and a new dependency for this check on the `validateScripts` job actually having compiled.

So I read your point as correct but as a two-step: the parser is the wrong long-term answer, and the right one is not free. Rather than quietly rewrite it under a review round, I would like your call on which you prefer:

1. Enable AST output and derive declarations from it, dropping the parser and the whole `findUnreadableImmutableLines` hard-fail class with it.
2. Keep the source scan for names, but take actual values from `immutableReferences` + `eth_getCode`, which needs no AST and, as you note on `registry-schema.ts`, collapses the `unverifiable` category to near-empty.
3. Land as-is and do either as a follow-up.

My preference is (2) now and (1) as a follow-up: (2) removes the category you flagged and the getter dependency without touching the build config, while (1) is the larger change and wants its own PR since it affects every compile.

Also taking your correction on the wording: an immutable is spliced into bytecode, never in a mapping — that phrasing in `registry-schema.ts` is wrong regardless of which option we take, and I will fix it either way.

0xDEnYO:
@coderabbitai review

script/deploy/immutables/getter-exemptions.json
```diff
@@ -0,0 +1,41 @@
+{
```
gvladika, line 1: **This file is not a CI trigger, so a PR that only edits it skips the gate.**

The `immutables` filter in `validateScripts.yml` matches TypeScript under this directory, plus two JSON files under `script/deploy/resources/`:

```yaml
immutables:
  - 'src/**/*.sol'
  - 'script/deploy/immutables/**/*.ts'          # .ts only
  - 'script/deploy/resources/immutableRegistry.json'
  - 'script/deploy/resources/deployRequirements.json'
```

Nothing matches `script/deploy/immutables/getter-exemptions.json`, and `validate-scripts-required` reports green when `verify-immutable-registry` is skipped.

That leaves the exemption list as the one input to this gate that can change without the gate running — and editing it is the documented flow for both directions of the ratchet: recording a new exemption, and dropping one that has since been annotated.

**Failure scenario.** A PR whose only change is adding `"FooFacet.BAR": ""` to this file touches no `.sol` and no `.ts`, so `detect-changes` sets `immutables=false`, the job is skipped, the aggregator is green, and it merges. Both errors that should have fired never run — I confirmed both fire when the job does run:

```
$ jq '. + {"FooFacet.BAR": ""}' getter-exemptions.json > … && bunx tsx verify-immutable-registry.ts --out-dir out-immutables
ERROR  FooFacet.BAR is exempted but no contract declares it. …
ERROR  FooFacet.BAR is exempted with no reason given.
exit=1
```

Same for the shrink direction: dropping `GasZipFacet.GAS_ZIP_ROUTER` alone exits 1 with "nothing checks", but only if something else in the PR triggers the job.

**Fix** — one filter entry, alongside the existing `.ts` line:

```yaml
- 'script/deploy/immutables/**/*.json'
```

`validateScripts.yml` belongs to #2308, but this PR introduces the file the filter misses, so the entry seems to belong here rather than in the base.

Worth noting the InfoSec side is already correct: `protectSecurityRelevantCode.yml` matches on the `^script/deploy/immutables/` prefix, so it does cover this file — the docstring's claim on `EXEMPTIONS_PATH` checks out. It is only the paths-filter that is extension-scoped.

0xDEnYO: Confirmed and fixed — `4e11a5bf4` + `f55d600c7`.

One deviation from the fix you prescribed, flagging it rather than leaving it silent: instead of adding a `.json` entry alongside the `.ts` one, I **replaced** the `.ts` entry with the directory.

```yaml
immutables:
  - 'src/**/*.sol'
  - 'script/deploy/immutables/**'
  ...
  - 'foundry.toml'
  - 'remappings.txt'
  - '.foundry-version'
  - '.github/actions/setup-foundry/**'
```

Reasoning: the defect isn't that JSON was missing, it's that the filter was extension-scoped over a directory where every file is a gate input — a `.json` entry fixes the instance and leaves the next non-`.ts`, non-`.json` input to be found the same way. `protectSecurityRelevantCode.yml` already gates on the `^script/deploy/immutables/` prefix, so this also makes the two agree, and `ts-unit-tests.yml` sets the in-repo precedent with a plain `script/**`.

Applying your own argument one step further turned up a second instance of it, which is the `foundry.toml`/`remappings.txt`/`.foundry-version`/`setup-foundry` block above: the gate's real enumeration source is `forge build src --ast`, so those inputs can change while `immutables=false` skips the job — the sibling `scripts` filter already listed two of them. Milder than the exemptions hole, since `findUnreadableImmutableLines`' successor `findSourcesWithoutAst` reds the *next* triggering PR, but the offending PR itself still merges unchecked.

Verified rather than assumed, under the matcher paths-filter v3 actually uses (picomatch, `dot: true`):

```text
script/deploy/immutables/**   -> 11/11 files, incl. getter-exemptions.json  (old pattern: 10/11)
regression check              -> none lost; every former .ts trigger still matches
every other entry             -> resolves to >=1 real tracked file
control: docs/Deploy.md       -> false
```

Your exit-1 assumption holds in all three directions, and none of them is warn-only: `coverageErrors` folds into `errors` at `verify-immutable-registry.ts:181`, and `decideExit` fails on `errors > 0` independently of `--strict`. A malformed-JSON edit to the exemption list is fail-closed too — `readGetterExemptions` returns `{}`, so all 39 exempt getters report unaccounted rather than silently passing.

gvladika (approved):

PR #2312, opened 2026-09-02
fix(safe): import MAX_PROPOSAL_REASON_LENGTH from where it lives (EXSC-921)

0xDEnYO:

Fixes EXSC-921.

A parallel session opened the byte-identical fix as #2313 with that ticket seven minutes after
this one; #2313 is closed in favour of this PR, which was already posted for review and green.
Also labelled `trivial` (single-line import fix).

`script/deploy/safe/proposal-card.ts` imports `MAX_PROPOSAL_REASON_LENGTH` from `./safe-utils`,
but that constant lives in `./proposal-intent`. `bun test script/` fails on `main` with:

```text
SyntaxError: Export named 'MAX_PROPOSAL_REASON_LENGTH' not found in module
  .../script/deploy/safe/safe-utils.ts
```

**How it got in.** Two PRs that were each green on their own crossed:

| PR | merged (UTC) | what it did |
| --- | --- | --- |
| #2298 | 11:03:36 | moved `MAX_PROPOSAL_REASON_LENGTH` out of `safe-utils.ts` into `proposal-intent.ts` |
| #2299 | 11:06:59 | added `proposal-card.ts`, which imports it from `safe-utils.ts` |

Three minutes apart, and neither PR's CI ever saw the other's change, so both merged green and
the combination is what breaks. Nothing was wrong with either diff in isolation.

**Why this direction.** The constant's home is `proposal-intent.ts`, and the sibling consumer
`provenance-display.ts` already imports it from there — so pointing `proposal-card.ts` at the
same place matches what's already the convention. Re-exporting it from `safe-utils.ts` would
also go green, but it would leave the constant with two import paths and re-couple
`safe-utils.ts` to something #2298 deliberately moved out.

Evidence, on a clean detached checkout of `origin/main` at `0bc30a141`:

| | `proposal-card.test.ts` | `bun test script/` |
| --- | --- | --- |
| `origin/main` as-is | 0 pass / **1 fail** | **1 fail** |
| with this change | passes | **1902 pass / 0 fail** |

`Run TypeScript Unit Tests` on `main` at `0bc30a141` is a `failure`, and every open PR inherits
that red — including #2286, whose `run-ts-tests` / `ts-tests-required` red is this and not its
own.

No new test: `proposal-card.test.ts` already covers this and already fails on `main` — it is the
check that caught it. The gap is that no job runs the merged combination of two green PRs before
they land, which is a CI-ordering issue rather than a missing unit test.

### DIFF
--- a/script/deploy/safe/proposal-card.ts
+++ b/script/deploy/safe/proposal-card.ts
@@ -10,7 +10,7 @@ import {
   sanitizeProvenanceText,
 } from '../shared/git-provenance'
 
-import { MAX_PROPOSAL_REASON_LENGTH } from './safe-utils'
+import { MAX_PROPOSAL_REASON_LENGTH } from './proposal-intent'
 
 /** How many networks are named individually before the card switches to a count. */
 const NAMED_NETWORK_LIMIT = 4


0xDEnYO:
Closing — the fix reached `main` by another route and this PR now contributes nothing.

#2286 merged as `fe6a44372` and carried the same one-line import change (it had picked it up to get its own push past `.husky/pre-push`). `git log -S` confirms that commit is what introduced `import { MAX_PROPOSAL_REASON_LENGTH } from './proposal-intent'` on `main`, and `proposal-card.ts` is now byte-identical between `origin/main` (`9a0a7aaf1`) and this branch — so merging this would be a no-op on the only file it touches.

`main` is green again, which was the whole point. EXSC-921 is satisfied by `fe6a44372`.

For the record, since it is a reusable lesson rather than a one-off: the underlying break was #2298 moving `MAX_PROPOSAL_REASON_LENGTH` out of `safe-utils.ts` while #2299 added a consumer importing it from the old home — merged three minutes apart, each green against a `main` that lacked the other's change. CI never tests the merged result, so nothing could have caught it pre-merge. It also blocked pushes on every branch, because `.husky/pre-push` type-checks merged-in files and so failed citing a file the pusher never touched — which is why five separate branches ended up carrying an identical copy of this fix.

PR #2313, opened 2026-09-02
fix(safe): import MAX_PROPOSAL_REASON_LENGTH from proposal-intent (EXSC-921)

0xDEnYO:

[EXSC-921](https://linear.app/lifi-linear/issue/EXSC-921/proposal-cardts-imports-max-proposal-reason-length-from-the-wrong)

`proposal-card.ts` imported `MAX_PROPOSAL_REASON_LENGTH` from `./safe-utils`, which does not export it — `safe-utils.ts` only *imports* other members from `./proposal-intent`. The constant is defined and exported in `proposal-intent.ts`, which is where the sibling modules (`provenance-display.ts`, `provenance-display.test.ts`) already import it from. Re-pointing the one import matches them and needs no re-export.

Direction matters here, not just the name: `proposal-intent.ts` documents itself as the dependency-free leaf ("what lets `safe-utils.ts` import it without a cycle"). Importing from the leaf keeps `proposal-card.ts` matching its own "pure: rows in, text out" docstring — re-exporting the constant from `safe-utils` instead would have pulled that module's Safe/Mongo/network graph into the renderer.

Effect on `main` today:

```text
script/deploy/safe/proposal-card.ts(13,10): error TS2305: Module '"./safe-utils"' has no exported member 'MAX_PROPOSAL_REASON_LENGTH'.
```

Anything that merges `main` and type-checks this file inherits the failure — found while merging `main` into #2293. Introduced in ac3dd16a4 (#2299).

Verification on this branch:

- `bunx tsc-files --noEmit script/deploy/safe/proposal-card.ts` → clean (fails on `main` at 0bc30a141)
- `bunx tsx script/utils/validateScripts.ts --base origin/main` → passes
- `bun test script/deploy/safe/` → 1086 pass, 0 fail
- `bunx eslint script/deploy/safe/proposal-card.ts` → clean

Two things found while verifying, deliberately **not** in this PR:

1. `validate-scripts` reported `pass` on #2299 even though `proposal-card.ts` was in that PR's changed-file set — running `tsc-files --noEmit` over exactly that five-file set locally exits 2 with the error above. Why the CI type check didn't fail is worth its own look; tracked on EXSC-921.
2. `script/deploy/safe/test-ledger.ts` treats `getLedgerAccount()`'s result as the account itself, but it returns `{ account, transport }` (4x TS2339, and the `signMessage` guard makes the script exit 1 unconditionally). Separate defect, separate PR.

Also noted and left alone: the `MAX_FIELD_CHARS` docstring says `reason` has its "own tighter cap", but `MAX_PROPOSAL_REASON_LENGTH` is also 200. Pre-existing and unrelated — kept out so this stays a one-line, cleanly cherry-pickable fix.

### DIFF
--- a/script/deploy/safe/proposal-card.ts
+++ b/script/deploy/safe/proposal-card.ts
@@ -10,7 +10,7 @@ import {
   sanitizeProvenanceText,
 } from '../shared/git-provenance'
 
-import { MAX_PROPOSAL_REASON_LENGTH } from './safe-utils'
+import { MAX_PROPOSAL_REASON_LENGTH } from './proposal-intent'
 
 /** How many networks are named individually before the card switches to a count. */
 const NAMED_NETWORK_LIMIT = 4


0xDEnYO:
## `/gate-review` — nothing auto-fixed, nothing escalated

Gate on `7869ab76d`. The parallel-agent fan-out the gate prescribes was not used — this session forbids spawning agents unless explicitly asked — so the passes were run inline. Diff is one line, so the scope was cheap to cover directly.

**Verdict: no finding cleared 80, nothing blocking.**

Confirmed on real data rather than by reading:

- The failure is real on `main` — `tsc-files --noEmit` over the exact five-file changed set of #2299 exits **2** with the TS2305; over the same set on this branch it exits **0**.
- `bunx tsx script/utils/validateScripts.ts --base origin/main` picks up the file and passes.
- `bun test script/deploy/safe/` 1086 pass / 0 fail; `proposal-card.test.ts` 62 pass. Runtime value resolves to `200`.
- No storage / selector / access-control / event / amount-math surface touched — no always-escalate change-class in scope.

Direction check (collaborator pass): `proposal-intent.ts` is the documented dependency-free leaf, so importing from it keeps `proposal-card.ts` consistent with its own "pure: rows in, text out" docstring. Re-exporting from `safe-utils` would have been the wrong fix.

### Lower-confidence — human judgment, not changed

1. `proposal-card.ts:24-32` — the `MAX_FIELD_CHARS` docstring says `reason`, `gitCommit` and `safeTxHash` have "their own tighter caps". True for the latter two (12 / 10), but `MAX_PROPOSAL_REASON_LENGTH` is also 200, so `reason`'s cap is equal, not tighter. Pre-existing; left alone to keep this a one-line cherry-pickable fix.
2. `validate-scripts` reported `pass` on #2299 with `proposal-card.ts` in its changed-file set, which locally fails. Worth understanding before trusting that check — note that CodeRabbit was rate/credit-limited across all seven gate rounds on #2299, so this gate was the only review that ran there. Tracked on EXSC-921.
3. `script/deploy/safe/test-ledger.ts` — treats `getLedgerAccount()`'s result as the account, but it returns `{ account, transport }`. 4x TS2339, and the `signMessage` guard makes the script `exit 1` unconditionally. Separate defect, separate PR.

0xDEnYO:
## Answering my own open question above — the CI gap is confirmed

Item 2 in the gate comment said it was unclear why `validate-scripts` didn't fail on #2299. It's now proven, and this PR demonstrated it live.

`validate-scripts` is gated on `github.event.pull_request.draft == false`, and `validate-scripts-required` treats `skipped` as success so it can serve as a required check. The workflow comment reads that skip as "no relevant files changed" — but the same `if:` also skips on draft, and the aggregator can't tell the two apart. So on a draft, the required check is green with nothing type-checked.

On this very PR:

```text
run 33635261934 (draft):  validate-scripts = skipped   validate-scripts-required = success
run 33635434504 (ready):  validate-scripts = success   validate-scripts-required = success
```

And in between, `gh pr checks 2313` showed `validate-scripts-required pass SUCCESS` — served by the draft run — while the ready run's real job was still `IN_PROGRESS`. The stale green outlives `gh pr ready`, so there's a window where branch protection is satisfied by a run that checked nothing.

That's how a file that doesn't compile reached `main`. This PR itself is properly validated: run 33635434504's `validate-scripts` is `success`, not `skipped`.

Fixing the gate is out of scope here (it touches `.github/workflows/**` and applies to the unit-test, smoke-test and solc-floor aggregators too) and is queued as separate work. Worth knowing before trusting a `*-required` green: check the inner job's conclusion for the current head SHA, not the aggregator's name.

0xDEnYO:
Closing as a duplicate of #2312 — byte-identical one-line fix, opened seven minutes apart by parallel sessions that each found the broken import independently.

Keeping #2312 because it is already posted to `#dev-sc-review`, its CI has settled green, and its body carries the root-cause write-up (#2298 moved the constant, #2299 imported it from the old home, three minutes apart, each green against a main lacking the other's change). I have moved `Fixes EXSC-921` onto #2312 so the ticket still closes.

Nothing wrong with this PR — it just would not do anything #2312 isn't already doing, and two identical open PRs cost a reviewer twice.

PR #2316, opened 2026-09-03
fix(deploy): attest the earliest reproducing commit, not an arbitrary one (EXSC-877)

0xDEnYO:

[EXSC-877](https://linear.app/lifi-linear/issue/EXSC-877/wp-71-fleet-reproduction-sweep-replace-when-touched-list). Follow-up to #2289, which is already merged and applied to production.

#2289 backfilled 739 deployment records with a `reproducibility.attestedCommit`. Those attestations are **true but imprecise**, and this corrects them in place.

## The defect

Any commit that leaves a contract's source closure untouched reproduces the same bytecode. The sweep's recovery pass took the *first* candidate that matched, which was frequently a distant descendant of the change that actually produced the code — one attestation pointed **143 days** past its own deploy.

## The fix, and the rule that is NOT being applied

Each attestation is re-selected to the **earliest known commit whose build reproduces the deployed runtime** — the point at which that exact source closure was introduced. **686 of 739 move.**

The tempting rule, "the commit must predate the deploy", is wrong here and is deliberately not used: **LI.FI deploys first and merges the PR afterwards**, so the commit carrying the deployed source legitimately lands days later. Applying that constraint left 219 of 742 with no candidate at all. What matters is the size of the gap, not its sign, so the payload now records `daysAfterDeploy` per row:

| Gap | Count | Reading |
|---|---|---|
| Commit predates the deploy | 522 | Code landed earlier, deployed to a chain later |
| 0–7 days after | 151 | Normal deploy → PR → merge |
| 8–30 days after | 47 | Slow merge, still plausible |
| >30 days after | 19 | Closure unchanged; pick is weak, wants a deeper search |

The 19 are the honest residual: no candidate close enough to the deploy was ever built for them.

## Why not constrain by the audit commit

This was considered and rejected on evidence. Of 69 distinct contract@versions in the payload, **34 have no usable audit anchor** — 16 have no audit entry, 1 has a non-SHA hash, and 17 audit commits do not resolve even after fetch-by-SHA. Of the remainder, at the *same* `@custom:version` the source at the recorded `auditCommitHash` differs in code from the deployed source in **25 cases**, so the audit commit is frequently not the deployed content either. Constraining selection to it would be unenforceable for half the fleet and unsound for much of the rest.

(Those 25 divergences are a separate finding, relevant to WP-0.3's content-equality gate rather than to this PR. At least one — `AcrossFacet@2.0.0`, `pragma solidity ^0.8.17` vs `0.8.17` — is bytecode-neutral, so the count is an upper bound. Distinguishing them needs a rebuild-and-compare per audit commit, which is that gate's job.)

## `--refresh`

Because #2289 is already applied, `attest` would skip all 739 as already-attested. `--refresh` rewrites an existing attestation **only** when the payload names a different commit, still only under `--apply`. Without the flag the previous behaviour is unchanged.

Dry-run against production on this branch:

```text
would write (new):   0
would refresh:     686
already correct:    53
skipped:             0
```

- [x] I have ensured that any new contracts have had AT A MINIMUM 1 preliminary audit conducted

No Solidity changes; nothing deployable.


0xDEnYO:
## Applied to production, then verified against the database

`attest --refresh --apply` was run: **686 refreshed, 53 already correct.** I then read production back and checked it against the payload rather than trusting the writer's own summary — which caught something the summary could not.

**First verification: 710/739 matched.** Commits were correct everywhere (0 differ), but **29 records had a stored build profile differing from the payload** — all `GenericSwapFacetV3@1.0.0` at commit `548f79ddd`, differing only in `evmVersion` (`london` stored vs `cancun` in the payload).

### Two defects, both fixed in `32b279d3e`

**1. The refresh guard compared only the commit.** A correction that keeps the commit and changes only the profile was silently skipped. It now compares the profile too. Those 29 were skipped for exactly this reason.

**2. The profile is not recoverable, and I should not have been re-picking it.** Several build profiles reproduce the same bytecode — `evmVersion` frequently changes nothing for a given contract — and there is no evidence to choose between them: the metadata trailer records `solc` but **never** `evmVersion`, and **714 of these 739 records carry no `solcVersion`/`evmVersion` of their own**. I first tried preferring `foundry.toml`'s canonical pairings (`0.8.29↔cancun`, `0.8.17↔london`); that made it worse, flipping 166 rows that had previously agreed. It was an invented rule dressed as evidence.

So the payload now **adopts the applied profile wherever the commit agrees**, and the caveat states the ambiguity explicitly. The commit is the field that carries information, and it is correct on all 739.

### Final state

```text
commit AND profile match payload: 739
commit differs:                     0
profile differs:                    0
attestation on a record with a real gitCommitHash: 0
```

A further `--apply` is a **no-op** (`would refresh: 0`, all 739 already attested), so this PR is now a documentation-and-guard change rather than a pending write. The gap distribution as stored: 522 predate the deploy, 151 within a week, 47 within a month, 19 beyond.

PR #2317, opened 2026-09-03
fix(deploy): treat EXIT_ON_ERROR="false" as a soft return in the missing-deploy-script guard (EXSC-922)

0xDEnYO:

Fixes [EXSC-922](https://linear.app/lifi-linear/issue/EXSC-922/deploysinglecontract-kills-the-whole-shell-when-a-deploy-script-is)

The "check if deploy script exists" guard in `deploySingleContract` still used the stale form:

```bash
if [[ -z "$EXIT_ON_ERROR" ]]; then
```

Every other guard in that function already uses `[[ -z "$EXIT_ON_ERROR" || "$EXIT_ON_ERROR" == "false" ]]`, established by EXSC-805 (9c3e00c3e) and EXSC-808 (3fe628d60) in #2234. This one was missed.

Six callers pass the literal string `"false"` as the 5th positional arg expecting a soft return:

- `script/scriptMaster.sh:187,242`
- `script/helperFunctions.sh:4097,4101`
- `script/deploy/deployCoreFacets.sh:51`
- `script/deploy/deployFacetAndAddToDiamond.sh:105`

`"false"` is non-empty, so `-z` is false and the guard took the `exit 1` branch — terminating the whole shell process rather than returning to the caller. Concretely: scriptMaster use case 2 ("deploy one contract to all networks") loops over networks, so a missing deploy script on the *first* network killed the entire interactive session instead of warning and continuing.

## Scope

Three sites, but only one is a live bug. The other two carry the same stale pattern with no caller that currently passes `"false"`, so they do not misbehave today:

| Site | Status |
|---|---|
| `deploySingleContract.sh:133` | **live bug** — six callers pass `"false"` |
| `helperFunctions.sh:2782` (`removeFacetFromDiamond`) | latent — function is currently unreferenced |
| `diamondEMERGENCYPause.sh:113` | latent — its only caller (`scriptMaster.sh:318`) leaves the flag unset |

I fixed all three so `grep -rn 'z "$EXIT_ON_ERROR" \]\]' script/` now returns nothing. Leaving the latent two would mean the next reader rediscovers and re-triages the same pattern.

That grep covers this one stale form, not all `EXIT_ON_ERROR` handling in the repo: `script/tasks/diamondUpdatePeriphery.sh:153` uses a third, structurally different convention (`== "true"` → `exit`, everything else soft-continues). It is logically compatible with the two here and out of scope for this PR.

## On the test

The new cases assert on the distinction that actually matters — *return* vs *exit* — rather than just the return code. The call runs inside a command substitution, so an `exit` tears down only that subshell and swallows the trailing `echo`; a `return` lets it through. That makes the two outcomes directly observable.

Falsification check: with the `deploySingleContract.sh` fix reverted, the `"false"` case fails with `killed-the-calling-shell` while the other two stay green — i.e. the test fails for the right reason, and only for that reason.

Found while reviewing #2315, which fixed this same pattern in a guard it introduced. Kept out of that PR as unrelated churn.


0xDEnYO:
@coderabbitai Not applying the explicit-return-type nitpick: no repo rule or lint rule requires it (`.eslintrc.cjs` has neither `explicit-function-return-type` nor `explicit-module-boundary-types`), and all 13 `it()` callbacks in this file — the 10 that predate this PR included — omit it. Annotating only the three new ones would make them the outlier in the file they live in.

script/deploy/deployRegistrationPropagation.test.ts
```diff
@@ -202,3 +202,31 @@ describe('deployToNetworkWorker result file', () => {
...
+
+describe('deploySingleContract missing deploy script', () => {
+  // Runs the call inside a command substitution so the two outcomes are distinguishable: a
+  // `return` lets the trailing echo run, an `exit` tears the subshell down and swallows it.
+  const harness = (exitOnError: string) => `
+    source "$REPO_ROOT/script/deploy/deploySingleContract.sh"
+    ${STUB_PRELUDE}
+    bunx() { :; }
+    isZkEvmNetwork() { return 1; }
+    checkIfFileExists() { return 1; }
```
gvladika, line 214: After the rebase, this test no longer exercises the guard the PR fixes.

The harness stubs `source` to a no-op, so the `source .../assertTreeRecordable.sh` inside `deploySingleContract` never runs and `assertTreeRecordableOrFail` is undefined at call time. The rebase pulled in the earlier `EXIT_ON_ERROR` guard from #2315 (now `deploySingleContract.sh:55`), so `if ! assertTreeRecordableOrFail ...` hits command-not-found, enters that block, and returns/exits at line 55 — before reaching the missing-script guard at line 146 that this PR changes.

Net effect: all three cases are satisfied by the line-55 guard (which already had the fixed form), so the test passes whether or not the line-146 fix is present. I confirmed this against the branch: with the `deploySingleContract.sh` change reverted, the `"false"` case still returns `rc=1` and stays green. The falsification check in the description held on the pre-rebase branch but no longer does.

One-line fix — add to the stubs in this block so execution reaches the guard under test:

```ts
checkIfFileExists() { return 1; }
assertTreeRecordableOrFail() { return 0; }  // reach the missing-script guard instead of short-circuiting at line 55
```

After adding it, please re-confirm the falsification: revert the `deploySingleContract.sh` change and check the `"false"` case fails.

0xDEnYO: Confirmed and fixed in `faaf445d9` — you were right that the rebase moved the short-circuit to the line-55 recordability guard, so all three cases passed with the line-146 fix reverted.

Added `assertTreeRecordableOrFail() { return 0; }` to the stubs. Re-ran the falsification you asked for:

- fix present: 13 pass / 0 fail
- `deploySingleContract.sh` line-146 change reverted: 12 pass / **1 fail** — `returns instead of killing the shell when EXIT_ON_ERROR is "false"`

So the test now fails without the fix, which it did not before.

One extra change while in there: `CONTRACT_REMINDERS` pointed at the real `contractSpecificReminders.sh`, which the pre-commit secret checker flags (it matches a `.env` key of the same name) and which would also couple this test to that file's contents. It now points at an empty `mktemp` file.

gvladika (approved):

PR #2319, opened 2026-09-03
chore(deploy): remove the dead deployment-logger module (EXSC-907)

0xDEnYO:

Ref [EXSC-907 · WP-2.2 Record hardening](https://linear.app/lifi-linear/issue/EXSC-907/wp-22-record-hardening) — deliberately `Ref` and not `Fixes`: scope items on the ticket remain open after this, so it must not auto-close.

`script/deploy/shared/deployment-logger.ts` has never run. It is a MongoDB-first
logging scaffold from the 2025 mongo-scripts era, and on `main` its only importer
has always been its own test file. Deleting it rather than annotating it, because
it is not a credible future replacement for the live path — see below.

**The live write path, for the record.** Both entrypoints shell out to the CLI:

- `script/helperFunctions.sh:113` → `bunx tsx script/deploy/update-deployment-logs.ts add`
- `script/utils/utils.ts:328` (`logDeployment`, used by `script/deploy/tron/tronUtils.ts:316`) → the same CLI

The record itself is built at `script/deploy/update-deployment-logs.ts:732`. Note that
the live `logDeployment` and the deleted module's `logDeployment` are two unrelated
functions that share a name — which is part of how this module reads as live.

**Why this is worth a PR and not just tidiness.** `getCurrentGitCommitHash` — the
`'UNKNOWN'`-sentinel hole this work package is about — has three call sites on
`main`: `deployment-logger.ts:125`, `deployment-logger.ts:210`, and
`update-deployment-logs.ts:750`. Only the third one executes. Reading the symbol's
call sites therefore gives the wrong picture of the deploy-record write path, and it
has done so repeatedly: #2315's description got it wrong, and the mistake goes back
to #2017, the merged PR that introduced the hash, whose body states it is "injected
by `DeploymentLogger.log()`/`logBatch()` and the `update-deployment-logs.ts add` CLI
path" — only the CLI half of that has ever been true. After this PR the symbol has
exactly one call site, which is the live one.

The module is also drifting away from the record schema in a way that would be a bug
the moment anything called it. Both `log()` and `logBatch()` compose their own Mongo
`$set`/`$setOnInsert` inline, so they bypass the `provenanceUpdate()` helper that
#2318 (open) introduces, and neither calls that PR's `getCurrentRepo()`. They would
write records with `repo` absent — which #2318 defines as meaning "written before the
field existed", so a fresh record would misread as legacy data. That is only harmless
while the module stays dead.

**How I confirmed it is unreferenced** (all against freshly-fetched `origin/main`):

- Module path and class name: no hits outside the module and its test.
- Every one of its five exports — `ILogOptions`, `ILoggerConfig`, `DeploymentLogger`,
  `logDeployment`, `logDeploymentBatch`: no external hits.
- No barrel file in `script/deploy/shared/`, no `tsconfig` path aliases, and no dynamic
  `await import()` / `require()` of it anywhere — the only variable-path dynamic imports
  under `script/` build `deployments/*.json` paths.
- No reference in any `*.sh`, `*.yml`, `*.json`, `*.md`, `*.sol` or `*.toml`, and no
  `package.json` script. `script/utils/validateScripts.ts` walks `script/**` generically,
  with no hardcoded file set. The module's top level is pure declarations, so nothing
  depended on importing it for a side effect.
- History: the only importer that has ever existed outside its own test is two commits
  on the `contracts-tron` branch `tron-sync-deployment-log` (`db2f071ad`, `fdc98882e`,
  2026-05-26) — a Tron dual-write scaffold that was never merged and never had a PR
  opened. `tron/main` carries the module equally dead.

Nothing is orphaned by the removal: `DatabaseConnectionManager`, `mongoEq`,
`createDefaultCache`, `DeploymentCache`, `IDeploymentRecord` and `IConfig` all keep live
users elsewhere.

**What the deleted test suite actually covered, and the one piece I kept.** The suite
stubbed `DatabaseConnectionManager.getInstance` throughout (28 `spyOn` calls, no
`mock.module`), so no real Mongo — but it was not mocked all the way down, and two of
its assertions were the only place in the repo asserting that `mongoEq` produces
`{ $eq: value }`. That property is load-bearing beyond this module:
`.agents/references/aikido-false-positive-catalog.md:32` cites `mongoEq` by name as the
sanitizer that justifies dismissing Aikido NoSQL-injection findings across the
deployment-log scripts. Deleting the suite would have left nothing holding it true, so
this PR adds `script/deploy/shared/mongo-log-utils.test.ts` covering `mongoEq` and
`deploymentRecordEqFilter` — including that an injected operator object gets nested
rather than passed through, and that the filter drops keys outside its whitelist.
I verified these assertions can fail: with `mongoEq` mutated to a pass-through, all 8
of them do.

**One gap I am not closing here.** `script/deploy/shared/deployment-cache.ts` had no
test file of its own and was incidentally executed by the deleted suite (one test
constructed a real `createDefaultCache` and ran a real `invalidate`, with no assertions
about either). After this PR it has zero executed coverage despite four live consumers.
That is a pre-existing hole rather than a regression this PR introduces, and covering a
600-line cache module is its own change — flagging it rather than expanding scope.

**Tests.** `bun test script/` → 1961 pass / 0 fail across 70 files before adding the new
suite (the 31 removed tests all belonged to the deleted module), and the new suite is 8
pass / 0 fail. `bunx eslint` and `bunx tsc-files --noEmit` clean on the added file.
`bunx tsc --noEmit` repo-wide reports 88 errors, all pre-existing on `main` (missing
`typechain`/`out` artifacts plus `script/deploy/safe/test-ledger.ts`); none mention
`deployment-logger`, which is the mechanical proof that no stale import survives. No
Solidity touched, so no `forge test`.


0xDEnYO:
@coderabbitai review

0xDEnYO:
@coderabbitai review

gvladika (approved):

PR #2325, opened 2026-09-07
feat(deploy): verify .foundry-version on the deploy path (EXSC-932)

0xDEnYO:

[EXSC-932](https://linear.app/lifi-linear/issue/EXSC-932/check-foundry-version-on-the-deploy-path-not-only-in-ci-and-pre-commit)

`script/utils/verify-foundry-version.sh` already existed and was already enforced in exactly two
places — the CI `setup-foundry` action and the husky pre-commit hook. No deploy script called it, so
a developer whose local `forge` had drifted from `.foundry-version` passed every existing guard and
still produced bytecode a verifier cannot reproduce from the recorded commit.

**No second checker.** `assertFoundryVersionOrFail` (`script/deploy/shared/assertFoundryVersion.sh`)
delegates to the same script, so all three enforcement points reach one verdict. It mirrors
`assertTreeRecordable.sh` from #2315 in shape and placement.

**One shared chokepoint, plus one earlier call where a build precedes it.**

- `executeAndParse` (`helperFunctions.sh`) is the seam every deploy-path `forge` invocation passes
  through: all 8 `forge script` call sites in the repo are `COMMAND` arguments to it, spread across
  exactly the six entry points the ticket names. Gating there covers all six and cannot be missed by
  the next script added — the same reasoning D16 used for the Ledger connect seam and D9 for the
  proposal funnel.
- `deploySingleContract` gets an additional call at function entry, because its zk path runs
  `./foundry-zksync/forge build` and derives the CREATE2 salt through `ensureStandardArtifactForSalt`
  (a standard `forge build`) *before* its first `executeAndParse`. Measured: gate at line 51, first
  direct `forge` invocation at line 288. Without the entry call the seam would fire after a build had
  already run.

**`diamondUpdateFacet.sh` is not touched** — #2324 is open against it. It reaches the gate through
`executeAndParse`; its `forge script` strings are assigned to `COMMAND` and handed to the seam, so it
has no `forge` invocation at command position. A test asserts all three of those facts, so the
coverage is not an argument.

**No environment exemption.** Unlike `assertTreeRecordableOrFail`, this function takes no
`ENVIRONMENT` argument at all, so there is nothing to exempt. A drifted compiler produces
irreproducible bytecode on a testnet and in staging too, and staging deploys are the rehearsal for
production. `deploy-smoke-test.yml` is the only workflow that runs these scripts and it uses
`setup-foundry`, which installs the pin and runs the same checker, so it stays green.

**Fails closed.** Only an explicit exit 0 from the checker permits the deploy: a missing checker, an
unrunnable one, an absent `forge`, and an unparseable `forge --version` all refuse. The seam is also
reached through `helperFunctions.sh`, so if that source line ever fails the function is undefined and
`executeAndParse` refuses rather than continuing.

**No memoisation, deliberately.** A "already verified" flag would be a shell global, therefore
settable from the environment (and `.env` is sourced by these scripts), i.e. a bypass. The check
costs one `forge --version` per invocation and is silent on success (`--quiet`).

**The result globals are reset on refusal.** `checkExecutorAndReceiver.sh` and
`acceptOwnershipTransferPeriphery.sh` call `executeAndParse` without checking its status and read the
verdict out of `RAW_RETURN_DATA` / `RETURN_CODE` via `handleForgeScriptError`, where a previous
call's success payload would have read as a completed forge run.

## Falsification, on real repo data

Real checker, real `.foundry-version` (1.7.1), real entry-point scripts; only the `forge` binary is
substituted, because the drifted case cannot be produced any other way.

| Case | Result |
|---|---|
| real matching `forge` | checker `✓ foundry 1.7.1`, exit 0 |
| `forge` reporting 9.9.9 | `✗ foundry version mismatch / expected: 1.7.1 / actual: 9.9.9` + refusal, rc 1 |
| no `forge` on PATH | `✗ forge not on PATH` + refusal, rc 1 |
| checker absent from the tree | `foundry version checker not found at …` + refusal, rc 1 |
| checker restored, same redirected tree | rc 0 — the counterpart, so the row above is not "refuses any redirected tree" |
| `executeAndParse` + drifted `forge` | command **did not run**; `RETURN_CODE=1`, `RAW_RETURN_DATA` cleared |
| `executeAndParse` + matching `forge` | command **ran**, rc 0 |
| seeded globals, no call | seed survives — so the cleared globals above are the refusal's doing |
| real `deploySingleContract`, drifted | `Cannot confirm the local foundry matches …/.foundry-version … Nothing has been broadcast.` |
| real `deploySingleContract`, matching | foundry message absent; refusal comes from the *next* guard (`assertTreeRecordableOrFail`) instead |
| `ENVIRONMENT` = `staging` / `testnet` / empty | still refuses on drift, still passes on a match |

The last two rows are what makes the gate's placement observable rather than argued: with a matching
`forge` the deploy gets past this gate and is stopped by the guard behind it.

41 new tests in `script/deploy/shared/assert-foundry-version.test.ts`. `bun test script/`:
**2432 pass / 0 fail** (baseline `origin/main` 2391/0). The suite also passes with `forge` removed
from `PATH`, which is the condition `ts-unit-tests.yml` runs in — it installs no Foundry.

`deployRegistrationPropagation.test.ts` needed one added stub: its "missing deploy script" case
stubs `source`, so the new guard was undefined there and refused, and the case passed for the wrong
reason without ever reaching the guard it tests.

## Not in this PR

- `docs/MultisigSigningProcess.md` is owned by open PR #2324, so the user-visible refusal is
  documented in `docs/Setup.md` §3 instead (which previously claimed only the hook and CI enforce the
  pin). The signing-process doc is owed once #2324 merges.
- The group warm-up `forge build` in `updateFoundryTomlForGroup` (`deployGroupingHelpers.sh`) is not
  gated. It builds the tree; it deploys nothing, and the per-network `deploySingleContract` that
  follows refuses at its entry gate. Gating it would also change the return contract of a function
  two tolerant callers rely on.

## How the placement claim is proven — and why it was rewritten four times

The interesting history of this PR is in the tests, not the gate.

The first version asserted placement by **classifying shell text** — find the gate's line
number, find the first `forge` invocation's line number, compare. `/gate-review` broke that
four times, each time with a shape that occurs in this repo:

| Round | Evasion that kept the suite green with the gate deleted |
|---|---|
| 1 | `gateIndex` matched a **comment** that names `executeAndParse`; and the ordering check was skipped for the four entry points with no command-position forge |
| 2 | a line inside a **multi-line double-quoted string** counted as a gate call |
| 3 | a **heredoc body** naming the gate counted as a gate call; a **single-quoted** multi-line string likewise |
| 4 | a forge started after `;`, `&&`, `else`, `exec`, bare `&`, or `VAR=$(cmd x) forge` was invisible; and the string tracker mis-classified **63% of `helperFunctions.sh`** as string interior |

The fifth version stops classifying shell text. **The classifier is deleted** (261 lines) and
four of the six entry points are now **run for real** against a `forge` stub that appends every
argv to a log. Each case asserts the gate refused, that at least one forge ran (so the run
actually reached the gate rather than returning early), and that every argv was exactly
`--version` — the checker's own probe and nothing else. The question "did a forge start before
the gate?" is now answered by bash, not by a regex.

Two entry points are not driven: `acceptOwnershipTransferPeriphery` and
`deployAndStoreCREATE3Factory` both stop on a missing private key before their forge call, so a
case for either would assert on a run that never reached the gate. Supplying a key to reach them
is deliberately not done — it would put a real key in `cast`'s argv on any machine that has one.
Both reach the gate through `executeAndParse`, whose refusal is itself executably proven.

A sixth round then found two defects in the fifth. The `deployAndStoreCREATE3Factory` case
**failed outright with no `.env`** — the exact CI trap an earlier commit claimed to have closed,
repeated on the case added to close it. And all four function stubs in the harness were **inert**:
four of these scripts re-source `helperFunctions.sh` inside their own body, redefining every stub
before it could take effect, so three comments asserting the stubs had an effect were false. Both
are fixed; the whole suite now runs green in a tree with no `.env` and no `forge`, which is the
check that would have caught either one.

### Mutation results on the final revision

| Mutation | Killed by |
|---|---|
| gate deleted from `deploySingleContract`, comment left | 6 tests, incl. the executable `deploySingleContract` case |
| gate deleted from `executeAndParse` | 8 tests, incl. four executable cases |
| `checkExecutorAndReceiver` rerouted to `executeAndCapture` | its executable case |
| `assertFoundryVersionOrFail` always returns 0 | 13 tests |
| heredoc naming the gate + rerouted call (killed round 4's parser) | its executable case |
| `if [ -d out ]; then :; else forge build; fi` above the gate | 2 tests |

### One finding was in the tests' own side effects

Driving `diamondUpdateFacet` **with the gate mutated away** let it run on to
`saveDiamondFacets`, which rewrote a tracked deployment log. Deployment records are read-only
for this project, so the harness now stubs the log writers and asserts
`git status --porcelain deployments config` is empty after every run. Verified in both
directions: the assertion fires when the gate is removed, and is empty on the normal path.

## Residual scope, stated rather than implied

- **zkEVM bytecode is not covered.** On zk networks the compiler that produces the deployed
  bytecode is `./foundry-zksync/forge`, pinned separately in `foundry.toml [external.zksync]`.
  `.foundry-version` says nothing about it. The gate still earns its place on the zk path,
  because the CREATE2 salt comes from `ensureStandardArtifactForSalt` → a plain `forge build`.
- **The TRON deploy path is not covered.** `script/deploy/tron/*.ts` reads Forge artifacts out of
  `out/` and broadcasts through tronweb; neither this gate nor the sibling recordability gate is
  on that path. `docs/Setup.md` says "the shell deploy scripts" for that reason. Worth its own
  package.
- **The verdict is about the local pin, not the recorded commit.** The checker compares the local
  `forge` to the working-tree `.foundry-version`. In `deploySingleContract` the next guard
  (`assertTreeRecordableOrFail`) closes that; on the `executeAndParse`-only entry points there is
  no such pairing, so the claim is "the local forge matches the local pin", nothing stronger.
- **Retry noise.** On the `executeAndParse`-only entry points the gate sits inside the existing
  retry loop, so a drifted `forge` is refused once per attempt rather than once. Nothing is
  broadcast and the globals are cleared each time; hoisting a second call into each of those
  scripts was rejected as scope creep against D9's "one evaluation per route".
- **`./foundry-zksync/forge` is not intercepted** by the argv log. The stub is found through
  PATH; the zk fork is invoked by relative path. So a pre-gate forge started that way — e.g.
  `ZKFORGE=./foundry-zksync/forge; "$ZKFORGE" build` — is a **known bypass of the test, not of the
  gate**: the gate still refuses, but the suite cannot observe such a build. The argv evidence
  covers the pinned `forge`, which is the binary `.foundry-version` governs.
- The `deploySingleContract` executable case runs a non-zk network, so the zk branch that
  motivated its extra entry-point gate is not itself exercised; the ordering there rests on the
  whole-line `gateCallLineIndex` assertion, which a heredoc could spoof (it fails in the safe
  direction on reformatting). The executable cases carry the claim; those two ordering tests are
  a cheap regression net, not evidence.
- Two more ungated direct `forge build` call sites exist beyond the two the code comment names:
  `deployContractToNetworks.sh:398` and `proposeContractToNetworks.sh:453`. Both build the tree;
  neither deploys. Worth its own package alongside `scriptMaster.sh` and
  `deployGroupingHelpers.sh`.


0xDEnYO:
Dispositions for the three CodeRabbit comments on this PR, all actioned:

1. **`docs/Setup.md` — manual `foundryup` rerun.** Applied. The paragraph now says CI and the shell deploy scripts refuse on mismatch, that the pre-commit hook only warns (`.husky/pre-commit:512` is explicitly "warn only, do not block commit" — the previous "refuse to run" claim was wrong for the hook), and that a pin bump needs `foundryup --install "$(cat .foundry-version)"` locally or deploys will refuse.

2. **`assert-foundry-version.test.ts` — detect an executable Forge command.** The concern was right, the suggested fix was not: `commandPositionOf` returns the quoted string for the four entry points whose forge invocation is an `executeAndParse` argument, so `/^forge\s+(script|build)/` matches **0 of 6** entry-point files (measured). That whole line-scanning approach has since been deleted — it was evadable by comments, multi-line strings, heredocs, and `;`/`&&`/`else`/`exec`. Placement is now proven by running four entry points against an argv-recording `forge` stub.

3. **`touchedRecords` — compare against a baseline.** Applied as suggested; it is now `recordsChanged: recordStatus() !== recordsBefore`. This mattered in practice — concurrent agents transiently dirty `deployments/`, which the absolute check would have blamed on the deploy script.

gvladika (approved):

gvladika (approved):

PR #2327, opened 2026-09-07
fix(tron): pre-flight the fee limit on the three direct-EOA send paths (EXSC-920)

0xDEnYO:

Fixes [EXSC-920](https://linear.app/lifi-linear/issue/EXSC-920/three-tron-direct-eoa-send-paths-broadcast-with-no-fee-limit) — the direct-EOA follow-up to EXSC-890 / #2311, which closed the same defect on the Tron Safe and timelock paths.

Tron sends were broadcasting without comparing the energy estimate against the fee limit they run under. `troncast send` had no estimate at all under a 1000 TRX default; the batched facet registration computed the estimate, printed it, and then never used it as a gate under 5000 TRX; the direct `transferOwnership` had no estimate under 10 TRX. A fee limit that cannot pay for a call does not fail cleanly — it runs until the limit is spent and aborts part-way with the energy still charged. `troncast send` is one of the paths `script/emergency/emergencyPauseBreakGlass.sh` reaches, so that shape lands mid-incident.

## Hard refusal, not print-and-confirm

The ticket left open whether `troncast send` should refuse at all, on the grounds that it is an interactive operator tool. It is not only that: it is reached from `script/universalCast.sh:153,238` and `script/emergency/emergencyPauseBreakGlass.sh:244`, both non-interactive shell callers. A confirmation prompt there would hang an incident response at the worst moment, and break-glass is exactly where a part-applied pause hurts most. So every path hard-refuses, and the escape hatch is the `ALLOW_GAS_ESTIMATE_FALLBACK=<network>` switch that already exists and is already documented for the Safe paths — one switch to learn, one to audit.

## Transport failure vs unaffordable — kept fail-closed

The pre-flight does not split "estimate says unaffordable," "simulation reverts," and "TronGrid 429 / HTTP RPC died after retries." All three are "no usable estimate" and refuse the broadcast.

The first two should refuse: an unaffordable call aborts mid-execution with energy charged; a revert-looking estimate usually is a revert, and Tron charges for those too.

The third — a 429 after `estimateTronEnergy`'s retries — also blocks `emergencyPauseBreakGlass`, because that script reaches `troncast send`. That is accepted, not an oversight. Broadcasting a pause with no energy figure reintroduces the half-applied pause this PR exists to prevent. The EVM side of the same script can fail open on a gas-price read because the failure mode there is overpay, not a part-applied multi-call.

The override is the existing hatch: `ALLOW_GAS_ESTIMATE_FALLBACK=tron`. Scope it to the network. Do not add a second, transport-only bypass.

`script/emergency/README.md` is left unchanged on purpose: that path is InfoSec-gated (`protect-critical-code`). The hatch is documented in `.env.example` and `.agents/commands/interact-tron.md`.

## One seam instead of three guards

`sendGuardedTronContractCall()` (new, `script/deploy/tron/tron-guarded-send.ts`) takes the broadcast as a callback and runs the existing `assertTronBroadcastAffordable` first. Each call site therefore holds no bare `.send()`, so "guard wired in on the wrong side of the send" is not expressible at a call site — which is the mistake this repo has shipped twice with a green suite. The refusal policy stays single; only the estimate transport differs (raw calldata via `estimateTronEnergy`, decoded arguments via the new `estimateTronEnergyBySelector`, the diamondCut via the existing `estimateDiamondCutEnergy`).

Two strengthenings to the shared module, both needed by the new callers:

- **A non-positive estimate is now treated as no estimate.** Only `estimateTronEnergy` refused a zero; the devkit's `estimateContractCallEnergy`, which the diamondCut path uses, does not. Priced, zero clears any fee limit, so the guard would have been a no-op on exactly the calls it exists to stop.
- **`raiseFeeLimitHint` is caller-supplied.** These paths are capped by `--feeLimit` and by constants, not by `TRON_SAFE_EXEC_FEE_LIMIT_SUN`, and a refusal naming the wrong control sends the operator to a setting that changes nothing.

Pricing on the diamondCut paths uses `tronEnergyCostInSun`, not the devkit's `getCurrentPrices`, for the reason `tron-energy-estimate.ts` already documents: `getCurrentPrices` substitutes a real-looking constant when the read fails and yields `0` for an empty price string, so neither answer can be told apart from a correct one.

## The ticket's third path was the wrong function

The ticket names `script/deploy/tron/tronUtils.ts:425-450`. `registerFacetToDiamond`, which contains it, **has no callers anywhere in the repo** — `grep -rn registerFacetToDiamond` over everything but `node_modules` returns only its own definition. The batched diamondCut the deploy runbook actually invokes (`.agents/commands/deploy-contract-tron.md:70`, `docs/TronFork.md:507`) is in `register-facets-to-diamond.ts`, and it is the same shape: estimate computed, cost printed, then a bare `.send()` at a hardcoded 5000 TRX. Its own comment recorded a *failed* transaction as the reason that limit was raised — the exact failure mode this PR exists to prevent.

So both are guarded. The live one reuses the estimate it already computes, so no second round trip. `registerFacetToDiamond` is guarded too but stays dead; deleting it (≈160 lines) is a separate change and not folded in here.

## Reversible choices

- **The two shared Tron energy modules moved** from `script/deploy/safe/executors/` to `script/deploy/tron/`. After this PR most of their callers sit outside the Safe path, and `.agents/rules/202-tron-scripts.md`'s globs now actually cover them. Four import lines updated; `gas-with-fallback.ts` deliberately stayed put, since it is the shared EVM+Tron escape-hatch policy, not a Tron module. No reference to the old paths remains anywhere in `script/`, `.agents/`, `.github/`, `docs/` or `*.sh`.
- **`TRANSFER_OWNERSHIP_FEE_LIMIT_SUN` raised from 10 to 20 TRX.** At mainnet's 210 SUN/energy, 10 TRX buys 47,619 energy, and the estimate's 1.2 safety margin then refuses above 39,682 raw. That send demonstrably fit under 47,619 before, so the margin would have eaten the whole headroom a working send relied on, and the refusal's only remedy is a code change mid-deploy. 20 TRX keeps the guard and gives the margin somewhere to live.
- **Native TRX transfers are exempt.** They run no VM code, so the fee limit caps nothing they can exceed. Asserted as a test, not just a comment.
- **The pre-flight sits after each path's dry-run early return**, so dry-run behaviour is unchanged and a refusal cannot swallow it. The existing checks at each site were written down in order and then the placement was proven by running the paths, not by reading them (see the mutation table).
- **Each guarded site got its own small module** (`send-transfer-ownership.ts`, `send-guarded-facet-registration.ts`, and `sendGuardedDiamondCut` in `tronUtils.ts`) so the real send is reachable from a test without loading the enclosing script's artifact, deployment-file and key handling — which is also what makes those tests runnable in CI at all.
- **Every suite is named after the module it imports** — the diamondCut cases live in `tronUtils.test.ts`, and the two extracted sends have suites of their own.

## Eight defects found while verifying, all fixed here

1. **`--fee-limit` and `--dry-run` were silently ignored.** citty resolves the spelling a caller did *not* type to the argument's `default`, so with `default: '1000'` on `feeLimit` the value in `--fee-limit 5000` never reached the code — which made the new guard compare against a cap the send does not run under, and made its own "re-run with a higher `--feeLimit`" hint a no-op. `--dry-run` was worse: it left `args.dryRun` at `false`, so a run the operator asked to simulate broadcast. Both spellings are the documented interface (`.agents/commands/interact-tron.md`, `script/troncast/README.md`). Dropping the citty defaults, with the fallbacks applied in the body, makes both spellings carry the passed value. A valueless `--fee-limit`, which citty passes through as `true`, is now rejected rather than capping the send at 1 TRX via `Number(true)`.
2. **The raw-calldata estimate refused every hex address.** It posts `visible: true`, so it needs base58, while `isValidAddress` also accepts `0x…` and `41…` — forms the broadcast normalises. `troncast send 41… --calldata …` worked on `main` and would have been refused.
3. **The call value was rounded before the guard that checks for rounding.** The function-signature path did `BigInt(Math.round(Number(callValue)))` and handed the result to `estimateTronEnergyBySelector`, whose `MAX_SAFE_INTEGER` check could therefore never fire. It now converts from the SUN string directly, and a malformed `--value` refuses with a message instead of a `BigInt` RangeError.
4. **`--dry-run` broadcast for real on two deploy scripts.** The same citty defect, live in `transfer-ownership-to-timelock.ts` and `register-facets-to-diamond.ts`: `bun script/deploy/tron/transfer-ownership-to-timelock.ts --step 1 --dry-run` transferred diamond ownership. `--delay-seconds` and `--no-propose` were dropped the same way. Both now declare those flags without a default and read them through `flagIsOn`, new in `cli-flags.ts` — the module that already exists for this class of ambiguity, so the coercion has one definition rather than three. Its tests run real citty over both argument shapes, including the one that carries a default, because a test that builds `args` by hand cannot see this class at all. Roughly a dozen other `script/deploy/tron/*.ts` CLIs share the pattern; that sweep is a follow-up, not this PR.
5. **`parseValue` produced non-integral SUN.** `4.1tron` stringified as `4099999.9999999995` — 2128 of ~14,000 decimal-TRX inputs do — and `1000000000000000tron` as `1e+21`. TronWeb's own integer validator refused both at the broadcast, so neither was a working path. Rounding fixed the artefact but lost sub-SUN amounts: `0.0000004tron` is 0.4 SUN and became `0`, which would have broadcast a zero-value call for a nonzero `--value` (caught by CodeRabbit). The TRX form now shifts the decimal string digit-wise instead of multiplying, which is exact for every amount and uses no float at all; an amount finer than one SUN, or a malformed or negative one, is refused rather than reaching the node. That also means the call-value estimate's `MAX_SAFE_INTEGER` guard can finally fire, since the value no longer loses precision on the way to it.
6. **A valueless `--feeLimit` silently applied the default.** citty resolves it to `''` for the camel spelling and `true` for the kebab one; the first fell through to the 1000 TRX default the operator was trying to override. Both are refused now.
7. **The new tests could not run in CI at all.** Both deploy-side suites imported the CLI script they cover, whose chain reaches `demoScriptHelpers.ts` and from there `typechain` — generated, and not built by the ts-tests job. They passed locally only because a stray `typechain` symlink sat in the worktree. `run-ts-tests` reported `Cannot find module '../../../typechain'` the moment the PR left draft. The two guarded sends now live in `send-transfer-ownership.ts` and `send-guarded-facet-registration.ts`, which import only the pre-flight and their constants, and each suite sits beside the module it covers.
8. **Documentation described a failure mode that no longer happens.** `.agents/commands/interact-tron.md` told operators that an energy problem shows up *after* broadcast and to retry with a bigger `--fee-limit`; it is now a pre-broadcast refusal that names the figure. The owner-mismatch prompt in `transfer-ownership-to-timelock.ts` promised a broadcast that the pre-flight refuses first.

## Two more sends the guard refused that used to work

Both found after the fee-limit margin fix, both fixed here.

9. **The estimate built a non-canonical selector.** The function-signature path joined the input types straight out of `parseFunctionSignature`, which canonicalises nothing and returns whatever was typed. The broadcast resolves `contract[name]` against the fetched ABI, and TronWeb formats that selector through ethers — so `troncast send <addr> "transfer(address,uint)" …` estimated `0x6cb927d8` while the send used `0xa9059cbb`. The simulation reverted and the new guard hard-refused a call that worked on `main`; the same held for `byte`, `int`, and a tuple written as a struct name. The selector and the parameter types now come off the ABI entry `contract[name]` is bound to (`resolveBroadcastCall`, `script/troncast/utils/abi.ts`), so estimate and broadcast read one source of truth — a canonicalisation table would be a second one, and #2327's own WP-1.4 already shipped the bug where a guard re-parsed its input instead of consuming the caller's. A name the ABI does not declare, a differing arity, and an overload the bare name cannot reach are refused before the estimate, naming the candidates, rather than guessed at. Its suite builds real TronWeb `Contract` wrappers, because the property under test is agreement with TronWeb's own selector formatting; a hand-rolled fake would assert a second copy of the canonicalisation.
10. **HTTPS-only refused a local dev node, and only on one branch of the command.** `troncast send --calldata … --rpcUrl http://localhost:8090` was refused by the moved `estimateTronEnergy`, while the function-signature path on the same `--rpcUrl` was unaffected — the two branches of one invocation disagreed. The restriction exists because an on-path attacker who answers plaintext HTTP picks `energy_used` and so picks whether the pre-flight lets a send past the fee limit; loopback has no on-path attacker. Plaintext HTTP to `localhost`, `127.0.0.1` and `[::1]` is now allowed, every other host still requires HTTPS, and the refusal names the rule it applied and the escape hatch. Matched on the parsed hostname, not a prefix, so `http://localhost.an-attacker.example` is still refused — asserted as a test.

## One merge commit, and why

`origin/main` is merged in rather than left behind, because `protect-critical-code` computes the PR's file list with a **two-dot** `git diff --name-only origin/main HEAD`. That reports the symmetric difference, so a branch based on an older `main` inherits every `.github/` file `main` has changed since — here `syncLedgerClearSigning.yml` from #2326, which this PR never touched — and the gate then demanded an Information Security Manager approval for it. `git diff --name-only origin/main...HEAD` on the same commits matches nothing. Merging clears it; the workflow's two-dot diff is a separate fix, and one that will itself need security approval since it edits `.github/`.

## Evidence

`bun test` over every file this PR touches, plus all of `script/deploy/tron/`, `script/troncast/` and `script/deploy/safe/executors/`: **389 pass / 0 fail**, checked both with and without a `typechain` directory present — that one environment difference between a local worktree and the ts-tests runner is exactly what hid a real failure for a while. Before the `origin/main` merge, the whole of `bun test script/` was **2445 / 0** against **2391 / 0** on `origin/main` at `46d991a33` (`origin/main` has since moved to `46d18f4ef`, a workflow-rename commit touching no `script/` file). `tsc-files --noEmit`, `eslint` (0 errors), `prettier --check` and `markdownlint-cli2` are clean on every changed file. CI is the arbiter for the whole suite: a handful of tests in `assert-tree-recordable`, `deploy-gate-cache`, `propose-diamond-cut` and `verify-immutable-registry` build real git repos and submodules in temp directories and fail non-deterministically in a local worktree — a different set on each run, none in a file this PR touches, and `run-ts-tests` is green on the runner.

Every path has a negative test asserting the broadcast spy was never reached *and* a positive test asserting the same spy fires on an affordable estimate — the guard-refuses-everything failure mode #2311's own gate found does not pass here. Both directions, printed:

```text
PATH  troncast send --calldata     POSITIVE   spy: BROADCAST REACHED    hits: ["wallet/triggersmartcontract","BROADCAST REACHED"]
PATH  troncast send "pause()"      POSITIVE   spy: BROADCAST REACHED    hits: ["BROADCAST REACHED"]
PATH  sendGuardedFacetRegistration POSITIVE   spy: BROADCAST REACHED    sends: 1
PATH  sendGuardedDiamondCut        POSITIVE   spy: BROADCAST REACHED    sends: ["BROADCAST REACHED"]
PATH  sendTransferOwnership        POSITIVE   spy: BROADCAST REACHED    sends: ["BROADCAST REACHED"]

PATH  troncast send --calldata     NEGATIVE   spy: broadcast NOT reached
  Energy estimation failed on tron for raw calldata call to TAuE…CU2 — refusing to broadcast.
    Underlying error: Tron simulation failed: {"result":false,"message":"REVERT"}
    Investigate the revert first. To broadcast anyway, re-run with ALLOW_GAS_ESTIMATE_FALLBACK=tron — …

PATH  troncast send "pause()"      NEGATIVE   spy: broadcast NOT reached
  Energy estimation failed on tron for pause() on TAuE…CU2 — refusing to broadcast.

PATH  troncast send --calldata     NEGATIVE (1 TRX limit)   spy: broadcast NOT reached
  Estimated energy exceeds the fee limit on tron for raw calldata call to TAuE…CU2 — refusing to broadcast.
    Estimated 24000 energy, costing 2400000 SUN, against a fee limit of 1000000 SUN.
    Re-run with --fee-limit 3 (TRX) or higher. …

PATH  sendGuardedDiamondCut        NEGATIVE   spy: broadcast NOT reached
  Energy estimation failed on tron for diamondCut registering OwnershipFacet — refusing to broadcast.
  Estimated energy exceeds the fee limit … Estimated 60000000 energy, costing 6000000000 SUN, against a
  fee limit of 5000000000 SUN. Register fewer selectors at a time, or raise the devkit's
  DEFAULT_FEE_LIMIT_TRX to at least 6000 TRX. …

PATH  sendTransferOwnership        NEGATIVE   spy: broadcast NOT reached
  Estimated energy exceeds the fee limit on tron for transferOwnership(TVQY…VfR) on TAuE…CU2 — refusing.
    Estimated 240000 energy, costing 24000000 SUN, against a fee limit of 20000000 SUN.
    Raise TRANSFER_OWNERSHIP_FEE_LIMIT_SUN in script/deploy/tron/constants.ts to at least 24000000. …
```

Nine mutations, each killed. The placement one matters most: this repo has twice shipped a guard wired in on the wrong side of the thing it guards, with a fully green suite.

| Mutation | Tests killed |
| -- | -- |
| the broadcast callback runs *before* `assertTronBroadcastAffordable` | 18 across all four site suites |
| `assertTronBroadcastAffordable` never refuses | 14 across the site suites |
| the pre-flight removed from `sendGuardedFacetRegistration` | 5 of 6 in its suite |
| citty `default: '1000'` restored on `feeLimit` | 2 |
| citty `default: false` restored on `dryRun` | 1 |
| the raw estimate posts `args.address` again | 1 |
| the call value converted via `Number`/`Math.round` again | 1 |
| the pricing refusal skips `redactErrorReason` | 1 |
| the non-positive-estimate rule deleted | 2 |
| `flagIsOn` narrowed to a strict `=== true \|\| === 'true'` test | 2 |
| `parseValue` back to an unrounded TRX→SUN multiplication | 1 |
| a valueless `--feeLimit` falls through to the default again | 1 |
| `parseValue` back to `Math.round` on the TRX multiplication | 8 |

Every `not.toContain` / `toEqual([])` in the new suites has the same marker asserted **present** in a positive case on the same fixtures — `'broadcast'`, `'estimate-calldata'`, `'estimate-selector'`, a non-empty `sends`, `'TRON_SAFE_EXEC_FEE_LIMIT_SUN'`, and `'failed for [redacted-url]'` for the credential-redaction case — so none of them can pass by the marker simply never appearing.

No test reaches a real Tron RPC: every estimate is a stubbed `fetch` or a stubbed `transactionBuilder`, and the only host that appears is `https://tron.invalid`.

## Known, not fixed here

- **`registerFacetToDiamond` in `tronUtils.ts` is dead** (grep above). Guarded, but a ≈160-line unreferenced function; worth deleting in its own PR.
- **The transport hardening covers one of the three estimators.** The HTTPS-or-loopback rule and `redirect: 'error'` live in `estimateTronEnergy`, the estimator they were written for. `sendGuardedDiamondCut` prices through the devkit's `estimateContractCallEnergy`, and `estimateTronEnergyBySelector` goes through tronweb's `HttpProvider`; neither carries either check. So the threat model this PR states — an on-path attacker forging `energy_used` to steer a send past the fee limit — is mitigated on one of the three paths that decide a broadcast. Closing it means changing or wrapping devkit behaviour, which is its own package and its own PR.
- **The pre-flight's `estimateFailed` arm is unreachable from every new call site.** `script/deploy/tron/tron-guarded-send.ts` hardcodes `estimateFailed: false` in the `IChainSimulateResult` it synthesises, so the arm that exists to stop a fallback figure being read as a real estimate can never fire for a guarded send. It is safe **today** only because the devkit estimator throws rather than substituting a fallback, and the throw is what the guard actually catches. It goes live the moment an estimator starts reporting a fallback instead of throwing, and nothing here would notice.
- **`TRON_SAFE_EXEC_FEE_LIMIT_SUN` is spelled out in two sibling constants** (`TRON_FEE_LIMIT_SUN_ENV` in the estimator, `TRON_FEE_LIMIT_ENV` in the pre-flight) plus the devkit's own export. Both pre-date this PR; they only became siblings when the modules moved. Deduping means editing files this PR otherwise only moves.
- **One more Tron send of the same class stays unguarded**: `deploy-and-register-periphery.ts:112` (`.send({feeLimit: 10_000_000})`, no estimate). `.agents/rules/202-tron-scripts.md` now states the convention that covers it, and that an unguarded send elsewhere in the tree is not precedent. `deploy-safe-tron.ts:312,564` are fine — they derive the fee limit *from* the estimate, so they are affordable by construction, and the rule now says so rather than leaving the two techniques looking like a pair.
- **The dropped-flag sweep.** Roughly a dozen other `script/deploy/tron/*.ts` CLIs declare multi-word booleans with `default: false`, so their kebab spellings are dead in the same way. Only the two this PR already touches are fixed here.
- **Wiring vs decision.** `troncast send`'s tests drive the command's real `run` body, so guard placement in the live path is asserted by execution. The three deploy-side suites drive the extracted wrapper, because reaching the enclosing function needs `out/` forge artifacts and deployment files; each wrapper is the only `.send()` in its file (grep-verified), but that link is read, not run.
- **`.agents/rules/202-tron-scripts.md:40** pointed at `script/deploy/tron/helpers/estimateContractEnergy.ts`, deleted when that helper moved into `@lifi/tron-devkit`. Fixed, along with two stale constant names in the same bullet group, plus a new section for the pre-flight convention.


script/troncast/commands/send.ts
```diff
@@ -17,6 +25,35 @@ import {
...
+ *
+ * Read once and passed to both the send and the pre-flight: a guard that
+ * recomputed it could compare against a different number than the send uses.
+ * Rejects a non-integer rather than letting `NaN` reach the comparison, where
+ * it surfaces as a `BigInt` conversion error instead of a usable message.
+ */
+function resolveFeeLimitSun(tronWeb: TronWeb, feeLimit?: string): number {
+  if (!feeLimit) return TRONCAST_DEFAULT_FEE_LIMIT_SUN
+
+  const sun = Number(tronWeb.toSun(parseFloat(feeLimit)))
```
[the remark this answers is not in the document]
0xDEnYO: Fixed in 1de68fde0, before this review landed. `resolveFeeLimitSun` no longer uses `parseFloat`: it converts with `Number()`, so `10junk` is `NaN` and refuses with `Invalid --fee-limit: "10junk" (must be a positive TRX amount)`. It also rejects the valueless `--fee-limit` that citty passes through as `true`, which `Number(true)` would have capped the send at 1 TRX. Covered by `rejects a valueless --fee-limit rather than capping at 1 TRX` in `send.test.ts`.

.agents/rules/202-tron-scripts.md
```diff
@@ -37,9 +37,17 @@ paths:
 
 ### Energy estimation
 
-- Use `estimateEnergyAndFeeLimit()` from `script/deploy/tron/helpers/estimateContractEnergy.ts` for fee limit calculation.
-- Always apply the safety margin from constants (`ENERGY_SAFETY_MARGIN`, default 1.2).
-- Pricing uses cached values via `tronPricing.ts` with TTL-based refresh.
+- Use `estimateEnergyAndFeeLimit()` from `@lifi/tron-devkit` to derive a fee limit from an estimate.
```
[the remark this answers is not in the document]
0xDEnYO: Good catch on the rule text — the two bullets did read as compatible, and a fee limit derived from the estimate it is compared against makes the comparison unfailable. The rule now says the limit must be fixed before the estimate (flag, constant or env var) and the same value passed to both the guard and the `broadcast` callback, and that `estimateEnergyAndFeeLimit()` is the alternative to a pre-flight rather than a companion to one — that is why `deploy-safe-tron.ts:312,564` need no guard. All four guarded sites in this PR already take a fixed limit: `--feeLimit`/`resolveFeeLimitSun`, `TRANSFER_OWNERSHIP_FEE_LIMIT_SUN`, `DEFAULT_FEE_LIMIT_TRX`, `DIAMOND_CUT_FEE_LIMIT_SUN`.

0xDEnYO:
@coderabbitai review

0xDEnYO:
@coderabbitai review

script/troncast/utils/tronweb.ts
```diff
@@ -91,7 +91,10 @@ export function parseValue(value: string): string {
   // Handle formats like "0.1tron", "100sun", "1000000"
   if (value.endsWith('tron')) {
     const amount = parseFloat(value.replace('tron', ''))
-    return (amount * 1_000_000).toString() // Convert to SUN
+    // Rounded, because the TRX→SUN multiplication leaves float artefacts on
+    // ordinary amounts: 4.1 * 1e6 stringifies as 4099999.9999999995, which is
+    // not a SUN amount and which TronWeb's own integer validator rejects.
+    return Math.round(amount * 1_000_000).toString()
```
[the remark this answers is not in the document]
0xDEnYO: Correct, and worse than the finding says: `Math.round` was itself the fix I had just made for the float artefact (`4.1 * 1e6` is `4099999.9999999995`), so it traded one wrong answer for a quieter one.

The TRX form no longer multiplies at all — it shifts the decimal string digit-wise, which is exact for every amount and needs no float:

- `4.1tron` -> `4100000`
- `0.000001tron` -> `1`
- `0.0000004tron` -> refused: `is finer than one SUN (6 decimal places); TRX has no smaller unit`
- `1000000000000000tron` -> `1000000000000000000000`, not the string `1e+21`
- `abctron`, `-1tron`, `1e6tron` -> refused before reaching the node

A side effect worth noting: because the value no longer loses precision on the way through, `estimateTronEnergyBySelector`'s `MAX_SAFE_INTEGER` guard can now actually fire, where from this call site it was previously unreachable.

Covered by a new `script/troncast/utils/tronweb.test.ts` plus a case in `send.test.ts` asserting the broadcast spy is never reached for a sub-SUN `--value`. Reverting to `Math.round` fails 8 of them.

0xDEnYO:
@coderabbitai review

0xDEnYO:
@coderabbitai review

0xDEnYO:
@coderabbitai review

0xDEnYO:
@coderabbitai review

0xDEnYO:
@coderabbitai review

0xDEnYO:
@coderabbitai review

0xDEnYO:
**Policy (kept fail-closed):** a TronGrid 429 / dead HTTP after `estimateTronEnergy`'s retries is treated the same as "unaffordable" or "simulation revert" — no usable estimate, so no broadcast. That includes `emergencyPauseBreakGlass` (`troncast send`). The EVM side of that script can fail open on a gas-price read because the failure mode is overpay, not a part-applied pause.

Override remains `ALLOW_GAS_ESTIMATE_FALLBACK=tron` (scope it). No transport-only bypass. `script/emergency/README.md` is intentionally untouched so this does not re-open `protect-critical-code`.

0xDEnYO:
@coderabbitai review

gvladika (commented):
Peer review — 5 findings total; posting the top 3 inline. Two low-severity ones (no retry on the now-mandatory `estimateTronEnergyBySelector`, and `parseValue` refusing trailing-zero TRX amounts like `1.5000000tron`) are in the thread rather than here.

Findings 1 and 2 were verified by executing this branch's code against the repo-pinned TronWeb/citty, not by reading. Note the citty claims behind your defects #1/#4/#6 hold on 0.1.6 but do *not* reproduce on 0.2.2 — anyone re-verifying has to run inside the repo.

script/troncast/utils/abi.ts
```diff
@@ -42,3 +42,119 @@ export function formatOutput(type: string, value: any): string {
...
+  if (declared.length === 0 || bound === undefined || bound === '')
+    throw new Error(
+      `The ABI in use declares no function named "${name}", so there is nothing ` +
+        `to estimate or to send. Pass the call as --calldata when the contract ` +
+        `does not publish an ABI carrying it.`
+    )
+
+  // TronWeb binds `contract[name]` to the last ABI entry of that name, so that
+  // is the entry the broadcast uses whatever spelling the operator typed.
+  const boundEntry = declared[declared.length - 1] as IAbiFunctionEntry
```
gvladika: **The TronWeb binding rule is inverted here, so for an overloaded name this prices one function and sends another.**

The comment and the line below it assume `contract[name]` resolves to the *last* ABI entry. TronWeb populates the two maps in opposite directions in one loop (`node_modules/tronweb/src/lib/contract/index.ts:135-142`):

```js
this.methodInstances[name] = method     // plain assign -> LAST wins
if (!this.hasProperty(name)) {
    this[name] = methodCall             // guarded     -> FIRST wins
}
```

`methodInstances[name]` is last-wins, so `bound` is right. But `contract[name]` — which is what `send.ts:688` actually broadcasts through — is **first**-wins, because `hasProperty` blocks the second assignment. So `boundEntry` (and the selector taken alongside it) points at the wrong entry.

Verified against a real `Contract` over `[withdraw(uint256), withdraw(address)]`: `contract.withdraw === methods['withdraw(uint256)']` is `true`.

Two consequences:

- Typing `withdraw(uint256)` — the overload the broadcast *does* reach, and which works on `main` — throws "Refusing to guess … the broadcast resolves the bare name to `withdraw(address)`". That sentence is inverted, and it's a regression.
- Typing `withdraw(address)` passes, the estimate prices `withdraw(address)`, and then `contract[funcSig.name]` broadcasts `withdraw(uint256)`. The fee-limit gate cleared on a different function's energy — the half-applied-send failure this PR exists to close, now reached *through* the guard.

`abi.test.ts`'s "allows the overload the bare name does reach" asserts the inverted direction, so the suite ratifies it rather than catching it.

Fix: `declared[0]`, and take the selector off `boundEntry` rather than `methodInstances[name]` so both agree. Cleaner alternative: broadcast through the exact signature key `contract[\`${name}(${types})\`]` (TronWeb registers `this[signature]` too), which removes bare-name resolution from the picture entirely. Either way the test case needs flipping.

One thing I did not check: whether any contract we reach this way actually declares overloaded externals. If none do this is latent rather than live, which is the difference between a blocker and a follow-up — your call.

script/troncast/commands/send.ts
```diff
@@ -546,10 +649,47 @@ export const sendCommand = defineCommand({
...
+      )
+
+      const txId = await sendGuardedTronContractCall({
+        networkName: networkKey,
+        operation: `${args.signature} on ${args.address}`,
+        feeLimitSun,
+        estimateEnergy: () =>
+          estimateTronEnergyBySelector({
+            tronWeb,
+            contractAddress: args.address,
```
gvladika: **`args.address` goes to the estimate unconverted, so a `0x…` target now hard-refuses a send that works today.**

`isValidAddress` (`utils/parser.ts:90`) accepts `0x…`, and the broadcast tolerates it because `contract().at()` overwrites `this.address` with the node's base58. But `estimateTronEnergyBySelector` → `transactionBuilder.triggerConstantContract` validates with `TronWeb.isAddress`, which is `false` for `0x…` — verified: it throws `Invalid contract address provided` before any request goes out.

So `troncast send 0x0a38…35 "pause()"` fails the pre-flight and is refused as "Energy estimation failed". Worth noting `pause()` on a hex address is a plausible break-glass shape.

This is the same defect as #2 in your list, which you fixed on the `--calldata` branch — that one does `fromHex(toHex(...))` and I verified it's correct. The signature branch needs the same conversion, or just pass `contract.address` (already normalised by `.at()`).

script/deploy/tron/transfer-ownership-to-timelock.ts
```diff
@@ -235,16 +243,18 @@ const main = defineCommand({
         'Run only step 1 (transferOwnership to Timelock) or step 2 (Safe proposal for confirmOwnershipTransfer). Omit step 1 only; add --confirm to also run step 2 in one invocation.',
       default: undefined,
     },
+    // No citty `default` on the multi-word flags below. For a multi-word
+    // argument citty resolves the spelling the caller did NOT type to the
+    // default, so `--dry-run` left `args.dryRun` at `false` and transferred
+    // ownership for real. The fallbacks are applied in the body instead.
     noPropose: {
```
gvladika, line 250: **`--no-propose` is still silently ignored, so the PR body's claim that both spellings carry doesn't hold for this one.**

mri (under citty) treats a `--no-` prefix as *negation*, not as part of the name. On the repo-pinned citty 0.1.6, with no `default` here, `--no-propose` parses to `{ propose: false }` — verified. So `args.noPropose` stays `undefined` and `flagIsOn` returns `false`.

Net effect: `bun script/deploy/tron/transfer-ownership-to-timelock.ts --step 2 --no-propose` still creates the MongoDB Safe proposal the operator explicitly asked to skip. Only `--noPropose` works.

Dropping the `default` fixed `--dry-run` (no `--no-` prefix there, so mri doesn't rewrite it) but it can't fix this flag, because the problem is the negation rewrite rather than default resolution.

Either read it through `readBooleanFlag`/`argv`, which already handles the negated spelling explicitly, or drop the both-spellings claim for `--no-propose` specifically.

script/troncast/utils/abi.ts
0xDEnYO: Confirmed, and the direction was inverted exactly as you describe — `methodInstances[name]` is last-wins, `contract[name]` is first-wins behind `hasProperty`. Reproduced against a real `Contract` over `[withdraw(uint256), withdraw(address)]`:

```
methodInstances['withdraw'].functionSelector = withdraw(address)
contract.withdraw === methods['withdraw(uint256)'] : true
```

Took your cleaner alternative rather than `declared[0]`. `resolveBroadcastCall` now picks the ABI entry from the typed argument types, finds its method instance by object identity (`instance.abi === chosen`, which holds because `Method` stores the entry by reference), and returns *that* entry's canonical selector; `send.ts` broadcasts through `contract.methods[selector]`. Bare-name resolution is gone from both sides, so the estimate and the send cannot pick different entries, and both overloads are now reachable instead of one being refused.

The inverted test case is replaced by two: each overload resolves to its own selector, and the resolved selector keys a method whose ABI entry matches the typed types.

On your open question — no contract we reach this way declares an overloaded external today, so it was latent rather than live. Fixed as a blocker anyway since the fix removes the ambiguity rather than papering over it.

script/troncast/commands/send.ts
0xDEnYO: Fixed — same `fromHex(toHex(...))` conversion as the `--calldata` branch, so the signature path no longer hands `triggerConstantContract` a form `TronWeb.isAddress` rejects. Verified `TronWeb.isAddress('0x…')` is `false` before changing it.

Covered by a parameterised test over both the `0x…` and `41…` forms, asserting the estimate receives base58 and the broadcast still happens.

script/deploy/tron/transfer-ownership-to-timelock.ts
0xDEnYO: Fixed, and you were right that dropping the `default` could not reach it — the problem is mri's negation rewrite. Confirmed on the repo-pinned citty: `--no-propose` parses to `{ propose: false }` and `args.noPropose` stays `undefined`.

New `readOptOutFlag` in `cli-flags.ts` resolves both spellings from argv, so `--noPropose`, `--no-propose` and `--propose=false` all skip the proposal. It also refuses the two spellings contradicting each other (`--propose --noPropose`), which land on different parser keys and so slip past the per-flag duplicate check that already refuses `--propose --no-propose`. `noPropose` is the only `no`-prefixed citty argument in the repo, so I did not extend the `cittyArgDefaults` validator to the class — worth doing if more appear.

Your two thread items are in as well: the selector estimate is retried now (and classified by the shape of the throw — a revert arrives as a plain `Error` from TronWeb, an axios failure with no response / 429 / 5xx is what gets another attempt), and `parseValue` trims trailing zeros before the sub-SUN check, so `1.5000000tron` is 1500000 SUN rather than a refusal.

gvladika (approved):

PR #2329, opened 2026-09-07
feat(codehash): the sign-time gate's decision modules (EXSC-906)

0xDEnYO:

Ref EXSC-906 — WP-2.1.
[EXSC-906](https://linear.app/lifi-linear/issue/EXSC-906/wp-21-verifycodehash-primitive-sign-time-gate)

Deliberately `Ref`, not `Fixes`: the ticket also tracks the `confirm-safe-tx.ts`
wiring, and closing it on merge would lose that.

The five decision modules the gate is built from, plus the wiring into
`confirm-safe-tx.ts` and the process documentation. The ticket's sequencing note
asked for it in this order — pure modules with real fixtures first, wiring once
#2297 merged.

**Look hardest at `codehash-sign-gate.ts` and its call site.** One invariant
there took six attempts, five of them defeated by adversarial review, and the
dead ends are worth knowing before reviewing the sixth:

| Attempt | How it was defeated |
|---|---|
| Pin the call site's spelling | A comment reciting the correct spelling above the wrong code |
| The same pin on comment-stripped source | A string literal doing the same; and `'/*'` inside a string, which deleted the real call site and anchored the scan on a decoy |
| An interface naming only `safeTransaction` | Cast-free: `ISafeTxDocument.safeTx` is the *same type*, so `{ safeTransaction: row.safeTx }` type-checked |
| A type-level brand on the struct | A spread keeps the brand and swaps the bytes; and the judging function still took a plain structural type, so a second call site skipped the selector |
| A runtime identity check on a value the selector then discarded | Reassigning or spreading the detached `{ data, network }`; and grading one pending proposal while signing another, since every row carries a struct the check accepts |

Each of the first four was also a **false red** — an alias, a rename or a
hoisted `const` failed the test it satisfied in spirit.

What holds now is two things. The gate's input carries the struct **by
reference** and reads `struct.data.data` when it judges, so swapping the bytes
means mutating the struct Safe will sign. And the verdict is **bound to the
transaction** through `proposalKeyOf` (`to|value|data|operation|nonce`), with
`createGatedSigner` taking a required `keyOf` — so a passing verdict cannot
authorise a different row, or a mutated one.

Evidence deliberately lives where CI looks: `bun test`, and `tsc-files` over
changed production files. A `@ts-expect-error` is **not** evidence in this repo —
CI type-checks changed production files only, so a test file is not in the
program, and full `tsc --noEmit` is red at baseline.

**Residuals, stated rather than buried.** A MATCH says nothing about bytes
excluded as immutables, so one renders grey until WP-2.3 can price them.
Calldata this decoder cannot open renders **no claim** and names the frames it
could not read, rather than reporting a pass — except that unopened calldata is
byte-scanned for the `diamondCut` selector, and a hit blocks. And the rebuild is
at the commit each deployment record names: per D3 the verifier asserts that
commit is present and fetchable, **not** that it is an ancestor of `main`, so
code deployed honestly from an unmerged branch reads MATCH. `§6` of
[MultisigSigningProcess.md](docs/MultisigSigningProcess.md) now says all of
this.

The through-line is D19's finding: **every value read out of deployed bytecode is
written by whoever proposes the cut.** Measured on a real facet — one code byte
flipped plus three trailer bytes claiming solc 0.8.99 moved the verdict from
MISMATCH to UNVERIFIABLE, which is the softer bucket. So each module here is
built so a proposer-controlled value can never widen what it is checked against.

**`lineage-scope.ts` — where the legitimate toolchain set comes from (D19a).**
Supplies the `scope` that `compareToAttestedSet` requires with no default, from
`config/networks.json` plus the profiles parsed out of `foundry.toml`. Never the
trailer. Profiles are parsed rather than tabled, because a hardcoded pair keeps
passing after someone retunes a profile — the same silent drift the gate exists
to catch, one layer up. Asserted across the real fleet: every active network
resolves to a closed set, so a config change that would hand authority back to
the trailer fails the suite.

Two defects in my own first draft, each now pinned by a test that fails without
the fix. A version→profile table mapped every unrecognised hardfork onto
`default`, reporting a closed set against the wrong compiler pair — a false
GREEN, unreachable today, which is precisely why it needed a test rather than a
reading. Matching on `evm_version` instead then admitted the **zk** profile for
every cancun network, since `zksync` also declares cancun; disqualification is
keyed on the zksolc pin, not the profile name, so a rename cannot re-admit it.

Two things it refuses rather than guesses. `"n/a"` is not a missing value — it is
`abstract`, `lens` and `zksync`, three **active mainnets** that compile through
zksolc, and a fallback there would grade them all UNVERIFIABLE, which is the grey
fail-open #2301 was fixed to remove. And when `targetEvmVersion` and `isZkEVM`
contradict each other it ERRORs, because "closed" would refuse an honest deploy
while "open" consults the trailer. Measured: the two agree for all 73 networks
today.

**`bytecode-trailer.ts` — the third trailer format (D19b).** zksolc writes the
same length-word and CBOR envelope solc does, so the strip already worked; only
the `solc` value differs — a three-byte release versus a
`zksolc:1.5.15;solc:0.8.29;llvm:1.0.2` text triple. Extended the existing reader
rather than adding a zk one, because one field with two readers is how the bytes
vouched for and the bytes compared come apart. The triple is surfaced whole,
since D19(b) requires the fork/LLVM sub-version compared **unmasked** — a reader
exposing only `solc` would make that comparison silently incomplete while looking
finished. All-or-nothing, anchored, ordered and fully numeric: a half-read triple
shown to a signer as fact is worse than none. Dropping the anchors kills exactly
one case, `llvm:notaversion`, which is what makes those negatives load-bearing.

**`zk-immutables.ts` — layer 2 on zkEVM (F11).** Immutables are not inlined
there, so offset masking has nothing to bite on; the values live in
`ImmutableSimulator` and are addressed by ordinal. **The ordinal is an input and
its absence is a refusal.** The spike measured index = declaration-ordinal × 32
on three real contracts, and its own caveat is that a constructor assigning out
of declaration order breaks that — reading the wrong slot then checks one
immutable against another's expectation, and *passes*. So the fallback to source
order is deliberately absent rather than merely unused. A failed read is reported,
never defaulted to zero, because `SymbiosisFacet` legitimately holds two zero
immutables on zksync.

**`commit-availability.ts` — fetch by SHA before concluding anything (T3/D3).**
The fleet sweep measured 56 of 98 audit commits unreachable from any local ref
and retrievable by SHA, so "not in this checkout" is evidence of nothing.
Concluding unverifiable without trying the fetch grades honest deploys grey — and
a grey that is usually wrong is a grey signers learn to click through. The
opposite failure is worse and is a separate outcome: a commit that genuinely
cannot be fetched ERRORs rather than falling through to "absent". A prefix is
refused before git is touched, and presence is asked as `<sha>^{commit}` so a
tree or blob sharing the SHA cannot answer for a commit.

**`cut-classification.ts` — what the gate must vouch for (A1).** Per FacetCut
element, never per operation: a batch pairing an Add with a Remove still gates
the Add. `_init` is a target of its own because it is delegatecalled in the
diamond's storage context — gated on an additive cut, and **refused** on a
removal-only one, where init calldata is arbitrary code framed as a deletion.

# One reversible choice, named

The zksolc pin, not the profile name, is what disqualifies a profile from a
non-zk network. Keyed that way so renaming the `zksync` profile cannot silently
re-admit it; the cost is that any future profile carrying a zksolc pin is
excluded from EVM networks by construction, which is the direction I wanted to
fail in.

# Not covered here, deliberately

The `confirm-safe-tx.ts` wiring, the attested-set construction, and the local
rebuild these compare against. The zk **rebuild** leg has a real dependency
worth flagging: `.foundry-version` pins `forge`, not `./foundry-zksync/forge`, so
nothing yet guarantees the toolchain a zk rebuild runs under —
[EXSC-933](https://linear.app/lifi-linear/issue/EXSC-933/toolchain-gate-residuals-zk-pin-ungated-forge-build-sites-tron-deploy) (a).

- [ ] For new facets: not applicable — no Solidity in this PR

- [ ] No new contracts — TypeScript deploy tooling only, no audit applicable


0xDEnYO:
@coderabbitai review

script/deploy/safe/safe-decode-utils.ts
```diff
@@ -935,3 +942,183 @@ export async function formatDecodedTxDataForDisplay(
...
+
+  // A cut in bytes nothing above decoded is the failure mode a wrapper list
+  // cannot cover, so it stops the signature instead of going unmentioned. Only
+  // reached for an unrecognised outer selector: a decodable call whose
+  // arguments carry these four bytes — a whitelist entry for `diamondCut`
+  // itself — is not a hidden cut.
+  const refusals =
+    calls.length === 0 &&
+    !DECODABLE_SELECTORS.has(topSelector) &&
+    data.toLowerCase().includes(DIAMOND_CUT_SELECTOR.slice(2).toLowerCase())
```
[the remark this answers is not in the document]
0xDEnYO: Reviewed against `cff3cb0ea` — this review ran on a SHA about 11 hours older, and the recording half landed in between.

**Every abandonment path you list is now recorded**, each with its own reason, in `collectDiamondCutTargets`: a truncated payload (`a N-byte payload`), the depth limit (`<selector> nested more than N deep`), a decode failure (`<selector> (<message>)`), and an unreadable cut entry (`<selector> (a cut entry could not be read)`).

**The fail-open you describe is closed, and not by the `calls` result.** The refusal is keyed on *any* unopened frame plus the `diamondCut` selector appearing anywhere in the raw calldata — deliberately a byte scan rather than an outer-selector check, because a cut nested inside an unknown envelope beneath `scheduleBatch` is exactly what an outer-selector test cannot see. So a dropped nested payload that actually carries a cut refuses regardless of nesting depth or which branch abandoned it. Tests: `refuses a cut hidden one level below a known envelope`, `refuses envelopes nested deeper than it will walk`, `refuses a diamondCut whose own body will not decode`, `refuses a batch that pairs a readable cut with an unreadable frame`, plus two added just now for the truncated case in both directions.

**What is deliberately still non-blocking**, and stated as a residual rather than an oversight: an unopenable frame with *no* cut selector anywhere in the bytes. That renders `· NO CLAIM`, naming the frames it could not read. Blocking it would refuse every ordinary non-cut proposal — a fee change, a role grant — since this decoder opens only the envelopes it knows. That trade is documented in §6 of `docs/MultisigSigningProcess.md` and has been put to the repo owner as an open question; it is not something I want to decide silently inside a PR.

**One case cannot be tested honestly: the unreadable `_init`.** I probed it — hand-crafted calldata with the top 12 bytes of the `_init` word set to `0xff…` — and viem silently masks them, returning a clean `0x2222…2222`, so `getAddress` never rejects and that `catch` is unreachable through any input `decodeFunctionData` accepts. It stays as fail-closed defence, but I have not written a test that would only appear to cover it.

script/deploy/codehash/lineage-scope.ts
```diff
@@ -0,0 +1,191 @@
...
+
+  // zkEVM chains compile through zksolc, so "n/a" is a positive statement that
+  // no EVM hardfork applies — not an absent value to fall back on.
+  if (declaresZk) {
+    const zk = deps.profiles[ZK_PROFILE]
+    if (zk === undefined)
+      throw new Error(
+        `Toolchain scope: "${network}" is zkEVM, but foundry.toml pins no "${ZK_PROFILE}" profile. Either it was renamed or the network's config is stale.`
+      )
+    return { isClosedSet: true, profiles: [zk] }
```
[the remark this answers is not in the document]
0xDEnYO: Correct, and it fails open rather than loudly, which is what makes it worth the catch: the scope came back closed, `codehash-sign-gate-deps.ts` reads a missing `zksolcVersion` as "not zk", and zkEVM bytecode would then be normalised under EVM rules — a different trailer format and no immutable masking, so the comparison is unsound in whichever direction it lands.

Fixed as suggested, as a separate branch from the missing-profile one so the message can say which of the two states it found. Regression test added with the paired positive (the real profile, which does pin a version, still resolves), so the refusal is about the absent pin rather than about zk networks generally. Removing the guard fails that test.

Note this is the mirror of a check already in `parseBuildProfiles`, which refuses a zksolc pin with no `[profile.zksync]` to attach it to — the pin is attached by name, so a renamed profile would detach it silently. The two together cover both halves.

docs/MultisigSigningProcess.md
```diff
@@ -325,7 +325,24 @@ signer sees:
...
    "filmstrip"** (`renderLedgerFlexFlow`, `ledger-flex-preview.ts`), an ASCII
    replica of the device screens for the exact to-be-signed values.
-4. The action prompt: `Do Nothing` / `Sign` / `Sign & Execute` /
+4. **The sign-time codehash gate** (`codehash-sign-gate.ts`, deciding through
+   `script/deploy/codehash/`). Where the calldata decodes to a `diamondCut` —
+   timelock-wrapped and batched frames included — every address the cut would
+   install is compared against attested rebuilds of `main`, and the verdict
+   **blocks the signature**. `MATCH`, `MISMATCH` and `UNVERIFIABLE` stay three
+   separate buckets: both of the latter stop a signature but they are different
+   facts, and collapsing them is what teaches a signer to click through grey. A
```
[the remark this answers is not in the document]
0xDEnYO: Fair — the docs described the three verdict buckets and never said what happens when the gate cannot reach a verdict at all, which is the distinction the whole design rests on.

Added to §4.3 as a fourth outcome beside the buckets: an unreadable `foundry.toml`, an RPC that will not answer, an attestation store that is down. It blocks, and it is rendered as a refusal rather than as a verdict, because there is no verdict to render — "we could not check" and "we checked and it is fine" are the pair the gate exists to keep apart. The §5 row now says the same, so the table does not imply that only a MISMATCH blocks.

0xDEnYO:
@coderabbitai review

0xDEnYO:
@coderabbitai review

0xDEnYO:
@coderabbitai review

script/deploy/safe/codehash-sign-gate-deps.ts
```diff
@@ -0,0 +1,643 @@
...
+    // The commit comes from a Mongo row and reaches both a path join and git's
+    // argv. `ensureCommitAvailable` checks the same shape, but it runs in a
+    // different module on a different call, so this does not rely on ordering.
+    if (!FULL_SHA.test(request.commit))
+      throw new Error(
+        `refusing to rebuild at "${request.commit}": a commit must be a full 40-character lowercase SHA before it reaches a path or a git argument.`
+      )
+    const checkout = join(deps.checkoutRoot, request.commit)
+    if (!deps.exists(checkout)) {
+      deps.git(['worktree', 'add', '--detach', checkout, request.commit])
```
gvladika, line 364: **The rebuild does not reproduce deployed bytecode — verified locally.**

`git worktree add --detach` does not populate submodules, so `lib/` is empty in the new checkout. `forge build` then prints `Missing dependencies found. Installing now...` and clones the libs itself — at the **wrong revisions**.

I ran exactly this sequence against `b27e3c74d`:

| submodule | repo pin | fresh worktree got |
|---|---|---|
| openzeppelin-contracts | `e50c24f5` (v4.9.2) | `bbf3600d` (v4.8.0-1252) |
| forge-std | `620536fa` (v1.16.1) | `16cb9c99` (v1.16.2-20) |
| solady | `678c9163` (v0.0.208) | `2afba69b` (v0.1.26-36) |
| solmate | `2001af43` | `89365b88` |
| sgn-v2-contracts | `1782ba66` | `dd60f956` |
| create3-factory | `0c1f3889` | `heads/main` |

Six of nine drifted (`git submodule status` marks them `+`). `src/` imports `@openzeppelin/`, `solmate/` and `solady/`, so the rebuilt runtime bytecode cannot equal what was deployed. Every gated address grades MISMATCH, and the gate blocks essentially every production diamond cut — with a failure that looks exactly like the attack it exists to catch.

Suggested fix: run `git -C <checkout> submodule update --init --recursive` after `worktree add` and before building, and put forge in offline mode so its auto-install can never silently substitute a revision. Worth asserting the worktree's `submodule status` is clean before the artifact is trusted at all — a rebuild against unpinned libraries is not an attestation.

script/deploy/safe/codehash-sign-gate-deps.ts
```diff
@@ -0,0 +1,643 @@
...
+        ? join(deps.repoRoot, 'foundry-zksync', 'forge')
+        : 'forge'
+      const args = [
+        'build',
+        '--out',
+        outDir,
+        ...(isZk ? ['--zksync'] : []),
+        '--skip',
+        'test',
+        'script',
```
gvladika: **`--skip test script` doesn't mean what it reads as.**

From `forge build --help`:

> `--skip <SKIP>...` — Skip building files whose names contain the given filter. **`test` and `script` are aliases for `.t.sol` and `.s.sol`.**

So this skips `*.t.sol` and `*.s.sol` only. The repo's own `[profile.solc_floor]` uses the correct form — `skip = ['test/**', 'script/**']`.

As written, the gate compiles the whole `test/solidity/` tree on every rebuild. In my local run that is exactly where the build died (`test/solidity/utils/TestHelpers.sol`, `TestToken.sol`, `TestWrappedConverter.sol` …), turning into `rebuild-failed` → `AttestationSourceError` → UNVERIFIABLE → blocked signature. Even once #1 is fixed it is needless compile time inside an interactive signing session.

This is the failure mode `.agents/rules/000-global-standards.md` names directly — CLI flags verified from `--help` or source rather than from analogy. Worth sweeping the other invocations this PR adds for the same thing.

script/deploy/safe/codehash-sign-gate-deps.ts
```diff
@@ -0,0 +1,643 @@
+/**
+ * The three real dependencies `verifyCutTargets` needs, wired at the sign-time
+ * call site.
+ *
+ * They live here rather than in `script/deploy/codehash/` because they are the
+ * untested glue that owns credentials — an RPC endpoint, the deployment-record
```
gvladika, line 6: **"the untested glue" is the leg that breaks, and there is no evidence it has ever reached MATCH.**

The five decision modules are genuinely well covered against fixtures — that part of the PR is solid. But `createForgeRebuildRunner`, `createMongoRecordSource` and `createRuntimeCodeObserver` are what actually decide whether anyone can sign, and both blocking findings I filed live here, in the code this header calls untested.

Before merge I'd want one recorded end-to-end run: a real facet, a real network, the gate returning MATCH against production. As it stands nothing in the PR shows the gate has ever produced a green verdict against a real deployment — only that it refuses correctly in unit tests.

Related, and cheap to answer at the same time: the whole thing hinges on `gitCommitHash` being present and correct in production deployment records. `script/deploy/repair-deployment-records.ts` existing suggests it is not uniformly so. A count of production facet records carrying a usable commit would say what fraction of cuts grade UNVERIFIABLE on day one.

gvladika (commented):
Three findings from review, filed inline. The design is right and I want this gate to exist — but as it stands the rebuild leg cannot reproduce deployed bytecode, so merging turns every production cut into a MISMATCH. (1) and (2) are small fixes; (3) is what I would gate the merge on. Remaining lower-severity notes are in the review thread rather than inline.

script/deploy/safe/codehash-sign-gate-deps.ts
0xDEnYO: Fixed in 4b9850e0b: after `worktree add` we now `git -C <checkout> submodule update --init --recursive`, assert `submodule status` is clean (no `+`/`-`/`U` rows), and forge with `--offline` so auto-install cannot substitute a tip revision.

script/deploy/safe/codehash-sign-gate-deps.ts
0xDEnYO: Fixed in 4b9850e0b: `--skip test/**` / `--skip script/**` (same form as `[profile.solc_floor]`), verified against `forge build --help` — the bare `test`/`script` aliases only drop `*.t.sol`/`*.s.sol`.

script/deploy/safe/codehash-sign-gate-deps.ts
0xDEnYO: Acknowledged — (1)+(2) are fixed in 4b9850e0b. Leaving this open: still need a recorded end-to-end MATCH against a real production facet + a count of production records with usable `gitCommitHash`. Will pick that up next (needs Mongo + a live rebuild).

gvladika (approved):

PR #2331, opened 2026-09-07
feat(deploy): record branch, scoped dirty tree and actor on deployment logs (EXSC-695)

0xDEnYO:

Fixes [EXSC-695](https://linear.app/lifi-linear/issue/EXSC-695/s9-deploy-log-twin-provenance-fields) (S9, under EXSC-686 · Multisig Signing Process 2.0)

A deployment record answered "which commit?" and nothing else, so a deploy from an
uncommitted edit — the tell-tale of the rogue-deployer scenario — was invisible to anyone
reviewing the record afterwards. This adds `gitBranch`, `dirtyTreeScoped`,
`dirtyTreeTruncated` and `actor` next to `gitCommitHash` on `IDeploymentRecord`, read
through `captureGitProvenance`, the same helper the Safe proposal document is built from
(EXSC-691/692/693), so a record and the proposal that installs it describe branch, tree and
actor by one definition rather than two.

Three decisions worth a reviewer's attention:

- **Provenance stays out of the record's identity.** The upsert filter remains the four
  identity fields, and the new fields are deliberately not added to
  `DEPLOYMENT_QUERY_EQ_KEYS`. A branch in either place would insert a second row for the
  same deployed address on every deploy from a new branch. Both are pinned by tests.
- **An empty dirty list is a claim, so it is only made when its own probe ran.** The shared
  capture returns an empty list both for a clean tree and for a `git status` it could not
  execute, and its error collector is shared with every other probe — so neither the list
  nor the collector separates the two. The dirty tree is therefore read through its own
  collector, and the field is written *absent* when that probe failed. Recording a failed
  probe as "clean" would suppress exactly what the field exists to show.
- **The dirty scope is broader than the one `assertTreeRecordable` blocks on.** That gate
  refuses a deploy on build-affecting paths only; this record is a reviewer's view of the
  whole tree minus the artefacts the deploy pipeline rewrites during its own run. Both
  scopes are intentional and the docstring says which is which.

The two summary lines print at info level on purpose: `logContractDeploymentInfo` runs this
command with stderr discarded unless `DEBUG` is set, so a *warning* about the tree a deploy
came from would never have been seen — which is the visible half of A1.2.

`add --dryRun` (or `--dry-run`) prints the upsert without connecting to MongoDB. It is what
makes the capture demonstrable without writing to the deployment store, and what the
placement tests drive.

## Evidence

Every new behaviour was broken deliberately and the failure recorded, then restored from
git. **Ten mutations, ten caught:**

| Mutation | Tests that fired |
|---|---|
| `gitBranch` dropped from `provenanceUpdate.set` | 5 (3 CLI, 2 unit) |
| dirty list written only when non-empty | 2 |
| branch/actor written unconditionally (a JSON sync would erase Mongo's copy) | 4 |
| `gitBranch` added to `DEPLOYMENT_QUERY_EQ_KEYS` | 1 |
| the `add` command stops capturing provenance at all | 4 (placement) |
| an unreadable tree recorded as clean | 1 |
| dirty readability re-derived from the whole capture's error bag | 1 (the clean-tree row) |
| the kebab `--dry-run` spelling loses its alias | 1 |
| an absent capture rendered like a clean tree | 3 |
| the actor hardcoded instead of read from the capture | 1 |

Two of those are worth reading. `--dry-run` is the spelling every sibling flag on this
command teaches, and citty resolves a kebab spelling through its own fallback only for an
arg carrying no `default` — so a `--dry-run` read as `false` made a dry run a real Mongo
write. Both spellings are now probed against the real CLI. And the dirty-readability
mutation survived the first version of its test: a dirty tree is recorded under both the
right rule and the wrong one, so only a *clean* tree forces the question of whether the
probe ran. The test now carries that row.

Placement is proven by running the real `add` CLI in throwaway git repos whose dirtiness the
test controls (`update-deployment-logs.cli.test.ts`), not by calling the pure functions
alone: dirty tree, clean tree, the dirty-list cap, CI actor with a detached HEAD, the kebab
flag, and a directory that is not a git checkout. Those probes point every store URI at an
unparseable value and assert no child ever reached a store. They also clear `CI` /
`GITHUB_*` from the child environment — without that the capture prefers the workflow's
branch and actor over the throwaway repo, so the suite would have passed locally and failed
on a runner. Verified by running both files with a full `pull_request` environment set: 36
pass, 0 fail.

Real-repo run, this branch with one source file deliberately dirty:

```text
ℹ Deployment provenance: branch feature/exsc-695-deploy-log-provenance, actor human, dirty 5 path(s)
ℹ Deployed from a dirty tree: script/deploy/shared/mongo-log-utils.ts, …, src/Facets/AcrossFacetV4.sol
"gitBranch": "feature/exsc-695-deploy-log-provenance",
"actor": "human",
"dirtyTreeScoped": [ …, "src/Facets/AcrossFacetV4.sol" ],
"dirtyTreeTruncated": false
```

`bun test script/`: 2610 pass, 0 fail. `bunx tsc --noEmit` clean for the touched files,
eslint clean, markdownlint 0 errors. Nothing was written to any deployment store.

## Deliberately not in this PR

- **The new fields stay out of `DEPLOYMENT_QUERY_EQ_KEYS`.** EXSC-695 says not to add them
  without a reason, and a test holds them out. Filtering records by `actor` or `gitBranch`
  is a separate, arguable feature.
- **`batchUpsertDeployments` still builds its own upsert.** It normalises absent fields to
  `''` where `buildDeploymentUpsert` writes them through, so folding the two together would
  change what a JSON sync writes. `RecordTransformer` emits none of the provenance fields,
  so the sync path cannot erase them either way — pinned by a test.
- **`gitCommitHash`'s `'UNKNOWN'` still reaches `$set`,** unlike the other sentinels: it is
  a plain truthy string and predates this work. The docstring now names it as the exception
  rather than implying otherwise.
- **`dirtyTreeScoped` inherits the shared capture's `git status` parsing,** which reads
  porcelain without `-z`, so a dirty path containing a space or non-ASCII bytes is recorded
  git-quoted. Changing that changes the proposal document too.

### DIFF
--- a/docs/MultisigSigningProcess.md
+++ b/docs/MultisigSigningProcess.md
@@ -78,9 +78,22 @@ verification, and calls `logContractDeploymentInfo` →
 time, from the dev's machine**. The record (`IDeploymentRecord` in
 `script/deploy/shared/mongo-log-utils.ts`) carries name, network, version,
 address, constructor args, salt, a self-reported `verified` boolean, and the
-deployer's HEAD `gitCommitHash` — no branch, dirty-tree flag, or human
-identity. File logs (`deployments/{network}.json`) only land in git at PR
-merge; the deploy scripts never commit.
+provenance of the run that produced it: the deployer's HEAD `gitCommitHash`,
+the `repo` it was cloned from, the `gitBranch` checked out, the `actor`
+(`human`, `bot`, `ci`), and `dirtyTreeScoped` — the working-tree paths that
+differed from that commit, with the artefacts the deploy pipeline rewrites
+during its own run excluded. An **empty `dirtyTreeScoped` means the capture
+ran and the tree was clean; an absent one means no capture ran**, which is
+what every record written before the field existed looks like. Branch, tree
+and actor come from the same `captureGitProvenance` pass the Safe proposal
+document uses, so a deployment and the proposal that installs it cannot
+disagree about where they came from. The capture is self-reported context, not
+a control: it makes an honest mistake such as a deploy from an uncommitted
+edit visible, while `assertTreeRecordable` — not this record — is what refuses
+such a deploy, on the narrower set of build-affecting paths. Add `--dryRun` to
+the `add` command to print the record and the upsert without writing. File
+logs (`deployments/{network}.json`) only land in git at PR merge; the deploy
+scripts never commit.
 
 ### 4.2 Propose
 
--- a/docs/MultisigSigningProcess.md
+++ b/docs/MultisigSigningProcess.md
@@ -80,21 +80,23 @@ time, from the dev's machine**. The record (`IDeploymentRecord` in
 address, constructor args, salt, a self-reported `verified` boolean, and the
 provenance of the run that produced it: the deployer's HEAD `gitCommitHash`,
 the `repo` it was cloned from, the `gitBranch` checked out, the `actor`
-(`human`, `bot`, `ci`), and `dirtyTreeScoped` — the working-tree paths that
-differed from that commit, with the artefacts the deploy pipeline rewrites
-during its own run excluded. An **empty `dirtyTreeScoped` means the capture
-ran and the tree was clean; an absent one means no capture ran, or ran and
-could not read the tree** — the same shape every record written before the
-field existed has. Branch, tree
-and actor come from the same `captureGitProvenance` pass the Safe proposal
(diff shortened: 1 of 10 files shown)


0xDEnYO:
@coderabbitai review

gvladika (approved):

gvladika (approved):

PR #2333, opened 2026-09-08
fix(ci): corroborate a stale cron verdict before alerting (EXSC-939)

0xDEnYO:

Fixes [EXSC-939](https://linear.app/lifi-linear/issue/EXSC-939/cron-liveness-watchdog-false-positives-on-githubs-stale-eventschedule)

The watchdog paged us about `Check Active Networks for Missing RPC Endpoints` being 26.4 days
dead ([run 34106671878](https://github.com/lifinance/contracts/actions/runs/34106671878)). The
cron had fired at `2026-09-07T00:16:59Z` — nine hours before the alert — and has fired daily
without a gap. The same watchdog said `114.4d` for the same workflow on 2026-09-03.

`checkCronLiveness.ts` read the last scheduled run from
`/actions/workflows/{id}/runs?event=schedule&per_page=1`. That event-filtered listing is
index-backed and can serve a snapshot days-to-weeks behind the unfiltered listing. Reproduced
live against this repo: the same URL returned `total_count: 160`, newest run `2026-08-04`, and
seconds later `total_count: 410`, newest `2026-09-08` — 35 days of lag, on a query whose newest
result cannot legitimately move backwards. The unfiltered listing was current at that moment.

The lag is **episodic, not steady-state**: 1 stale response in 159 sampled today. And the fix is a
best-effort second opinion, not a guarantee — nothing establishes that the unfiltered listing is
immune to the same replica lag, only that it was fresh when the filtered one was not. A lagging answer there is indistinguishable from a dropped schedule, which is
precisely the signal this job exists to raise.

**The fix, and what it deliberately does not do:**

- The filtered query stays the primary source. Only a **stale** verdict pays for a second
  opinion, so the healthy path still costs one request per workflow and the run's API budget is
  unchanged in the common case.
- The second opinion scans the **unfiltered** run listing and selects `event === 'schedule'`
  client-side. `workflow_dispatch` still never counts as evidence of life — that rule is the
  reason the filter was there in the first place, so it moved rather than went away.
- Only a **newer** timestamp overrides the verdict. Staleness can only make a run look older, so
  there is no direction in which this can invent liveness.
- When neither source has a scheduled run, the alert fires exactly as before. Same when the
  corroborating request itself fails: a second opinion we could not obtain is not evidence of
  life, so the stale verdict stands.
- Paging is bounded (5 × 100) and stops at the first page carrying a scheduled run. For a cron
  whose workflow has no other trigger, page one holds every run it ever had. The budget only
  matters where a workflow`s other triggers bury its schedule: `networkRpcsChecker.yml` has a
  bare `push:` on every branch, and one 100-run page there has covered as little as **11.2h** —
  less than its 39h grace window. Measured worst case is 208 runs in any 39h window, so 500 runs
  clears every scheduled workflow in the fleet. Exhausting the budget is reported separately from
  "no scheduled run" and logs a warning naming the constant to raise, so this degrading later is
  loud rather than a silent return to false-paging.

**Verified against the live API, not just unit tests** (`--dry-run`, real repo, real workflows):

1. Unmodified: 9/9 workflows classified as before; only the manually-disabled clear-signing sync
   alerts. No behaviour change on the healthy path.
2. Primary lookup forced to return nothing (a stand-in for a maximally stale index): all 9
   workflows were rescued by the corroboration scan and reported alive, each with its real
   scheduled-run timestamp.
3. Both lookups forced to return nothing: all 9 report stale and the Slack alert fires. The
   change cannot silence a genuinely dead cron.

A re-run cannot rejuvenate a dead cron: verified on a real re-run (`30192494683`, `run_attempt: 2`)
that `created_at` stays pinned to the original trigger while `run_started_at` moves — the code
reads `created_at`.

Unit tests cover the extracted `newestScheduledRun` selector and the paging rules of
`scanForNewestScheduledRun` (paging stops at the first page with a scheduled run, a short page ends
the listing, an exhausted page budget is reported rather than folded into a null): other events ignored, newest of
several scheduled runs, empty page returns null so the caller keeps paging, and an unparseable
timestamp dropped rather than returned as an `Invalid Date` (which would flow into the age
arithmetic as `NaN`, compare false against every grace window, and silently mark the workflow
alive).


0xDEnYO:
@coderabbitai review

gvladika (approved):

PR #2335, opened 2026-09-08
feat(safe): check ledger with per-anchor results and a rolled-up report (EXSC-698)

0xDEnYO:
# Which Linear task belongs to this PR?

[EXSC-698](https://linear.app/lifi-linear/issue/EXSC-698/s12-check-ledger-rolled-up-report) — S12 / WP-4.1, the check ledger and its rolled-up report.

# Why did I implement it this way?

Today a pre-signing check that **could not run** — flaky RPC, unreachable store, unreadable anchor — looks exactly like one that passed. This adds the structure that makes those two states impossible to confuse, as two pure modules plus tests. **Nothing is wired to a call site yet:** migrating the existing checks onto the ledger is a separate package, so this PR changes no behaviour on any signing path.

`check-ledger.ts` — the model.

- Four statuses, `pass` / `fail` / `error` / `needs-ack`, and every result carries **expected value, actual value and which anchor it came from** (`A-CI`, `A-LOCAL`, `A-MAIN`, `A-AUDIT`, `A-CHAIN`, `A-MONGO`, `A-PROPOSAL`, `A-UNRESOLVED`).
- The **coverage denominator is fixed up front**. A registered check that reported nowhere is the most important row in the report, so it is counted as `missing` rather than being absent, and a ledger with no networks or no checks is refused outright — `0/0` would render as a fully green run that verified nothing.
- `error` and `missing` **block on either class**: there is nothing for a human to acknowledge about a check that was not shown to have run. An integrity `fail` blocks too, because T3 gives the integrity class no acknowledgement path at all. Only a semantic non-pass is acknowledgeable.
- Two coercions live in `recordCheck`, so no check can opt out of them by how it reports: a `needs-ack` on an integrity check becomes `fail`, and a `pass` sourced from a reporting-only anchor (`A-MONGO`, `A-PROPOSAL`, `A-UNRESOLVED`) becomes `error`. The second is N3 made structural — the deployment record and the proposal document are proposer-writable, so they may report but never decide a green.
- `--triage` is narrowed per T2: a relaxation needs **both** a purely subtractive op **and** a semantic check, so the codehash gate is never relaxed on any profile.
- `buildReviewAttestation` produces the record a caller stores on the proposal document: a keccak digest over the denominator and every result (values, status, anchor, detail), plus the verdict, the triage profile, and how many acknowledgements were answered versus dropped. It is a reconstruction record, not an oracle (T5).

[...]

PR #2336, opened 2026-09-08
feat(deploy): bound the deployer key's power to a documented inventory (EXSC-878)

0xDEnYO:

Fixes [EXSC-878](https://linear.app/lifi-linear/issue/EXSC-878/wp-73-deployer-key-power-inventory-config-assertion) — WP-7.3 / R7.1, deployer key-power inventory + config assertion.

The deployer wallet is the one key that appears at every stage of a production change. Measured from the repo at `95fdb620d`, it: deploys (`PRIVATE_KEY_PRODUCTION`), writes the deployment record at deploy time, creates the Safe proposal (`propose-to-safe.ts` `runPropose`), holds one of the `safeOwners` slots — `safeOwners[0]` **is** `deployerWallet` — holds `CANCELLER_ROLE` on every timelock (the `_cancellerWallet` constructor arg in `LiFiTimelockController`), and broadcasts the Safe execution once the threshold is met. R7.1 asks for that set to be enumerated, bounded, and alerted on when it widens.

**`script/deploy/deployer-key-power.ts` carries the inventory and the assertion, with three legs, all collected before throwing so one violation never hides another.**

- **The config leg** walks the whole of `config/global.json` and `config/networks.json` and refuses any slot holding a deployer identity that is not in `DOCUMENTED_DEPLOYER_CONFIG_SLOTS`. It is a generic walk rather than a scan of known field names, because the widening R7.1 worries about is precisely a *new* field pointing at the deployer — a scan of today's field names would not see it.
- **The bound leg** re-derives the claim the inventory rests on instead of asserting it in prose: the deployer occupies exactly one owner slot, and one slot cannot reach `SAFE_THRESHOLD`. That is why it cannot schedule or push an operation. Injecting a threshold of 1 refuses; 2 passes, so the leg is not a blanket refusal.
- **The inventory leg** refuses if any power marked as held in production steady state is classed `integrity`. If a future change adds one, the check fails and forces the premise to be re-argued rather than quietly widened.

**Two scoped exceptions are in the inventory rather than hidden by it**, because both are real and the honest reading of "DoS only, never an integrity break" is production-mainnet-scoped:

- On **testnet and staging** the deployer owns the diamond outright and can `diamondCut` unilaterally — those networks have no Safe or timelock by design (`healthCheckInvariants.ts` asserts exactly that ownership; `deployAllContracts.sh` stage 12 skips the timelock transfer there).
- During **production bring-up** the deployer owns a production diamond between `transferOwnership(timelock)`, which it sends directly as the then-current owner, and the Safe-executed `confirmOwnershipTransfer()`. A network left in that state is already reported unhealthy by the `diamond-owner` invariant.

The deploy-time record write is classed `untrusted-input`, not `no-authority`: a lying record is the G7 threat model, and the premise holds only because the sign-time gate re-derives rather than trusting it.

**Timelock execution is carried as `pending`, not current.** `EXECUTOR_ROLE` is granted to `address(0)` on `main`, so execution is permissionless and the deployer holds no executor grant. It becomes a deployer power when the F7 restriction lands (WP-6.2 / EXSC-872).

**Where the check runs.** `bun deployer-key-power` prints the inventory and exits non-zero on a violation; `script/deploy/deployer-key-power.test.ts` is the CI leg, and it is reached on a config-only change because `ts-unit-tests.yml`'s path filter already covers `config/**`. It is deliberately *not* wired into the proposal funnel: those files are owned by open PRs #2329 / #2334 / #2335. Reversible — moving it into the funnel later is a one-line call.

**Asserted against the committed config, not fixtures**, and every falsification mutates the real file on disk and is reverted with `git checkout`:

| Mutation | Result |
|---|---|
| unmutated committed config | `CLI exit code: 0` |
| `pauserWallet` reassigned to the deployer | refuses, names `global.json:pauserWallet`, exit 1 |
| a field that did not exist when the check was written (`someNewOperatorWallet`) | refuses, exit 1 |
| the deployer standing in for `networks.json:mainnet.safeAddress` | refuses, exit 1 |
| a duplicated `safeOwners` entry | refuses (`occupies 2 safeOwners slots, expected exactly 1`), exit 1 |
| `pauserWallet` **and** the duplicate together | both violations printed in one message |
| injected `SAFE_THRESHOLD` of 1 / of 2 | refuses / passes |

`bun test:ts` 2596 pass / 0 fail against a 2582 / 0 baseline on the same worktree. Repo-wide `bunx tsc --noEmit` shows 88 pre-existing errors, none in the new files; `bunx tsc-files` and `bunx eslint` are clean on both.

**Measured, worth a reviewer's eye:** the `safe-config` health-check invariant only asserts config owners ⊆ on-chain owners, so an *extra* on-chain Safe owner — a genuine widening of the signer set — is invisible to it. Out of scope here (this package bounds the deployer key, not the owner set) but it is the same class of gap.

### DIFF
--- a/docs/MultisigSigningProcess.md
+++ b/docs/MultisigSigningProcess.md
@@ -43,6 +43,39 @@ Status: **current state**, verified against the repo. Author: Daniel B. (SC).
 | Executor (Safe leg) | Any owner may execute once the threshold is met, but in practice the **deployer wallet** broadcasts: it is the only owner funded on every chain, whereas the signer hardware wallets are not | The "…With Deployer" execute variants inside `bun confirm-safe-tx`, which broadcast with `PRIVATE_KEY_PRODUCTION` |
 | Executor (timelock leg) | The **"Timelock Auto Execution" GitHub cron** (`.github/workflows/runPendingTimelockTXs.yml`), every 10 minutes, gated on repo var `ENABLE_TIMELOCK_AUTO_EXECUTION` | `script/deploy/safe/execute-pending-timelock-tx.ts --executeAll`, signing with `TIMELOCK_EXECUTOR_PRIVATE_KEY` — a pure gas-payer EOA with no protocol authority (`EXECUTOR_ROLE` is open). Manual fallback: `bun execute-timelock` |
 
+### 2.1 What the deployer key can do
+
+The deployer wallet (`config/global.json` `deployerWallet`, key
+`PRIVATE_KEY_PRODUCTION`) is the only key that appears at every stage: it
+deploys, writes the deployment record, creates the proposal, holds one of the
+Safe owner slots, and holds `CANCELLER_ROLE` on every timelock. Its power is
+nevertheless bounded to **DoS and griefing** on production mainnets — it cannot
+schedule or push a bad operation, because both need the Safe threshold, and it
+holds one signature of `SAFE_THRESHOLD` (`script/deploy/shared/constants.ts`).
+
+`bun deployer-key-power` prints the full inventory and refuses if the config
+grants the wallet anything outside the documented set. The check walks the whole
+of `config/global.json` and `config/networks.json` rather than a list of known
+field names, so a newly added field pointing at the deployer is caught too, and
+it runs in `bun test:ts` (whose path filter includes `config/**`).
+
+Two scoped exceptions are part of the inventory rather than hidden by it. On
+**testnet and staging** the deployer owns the diamond outright — those networks
+have no Safe or timelock — so the bound above is a production-mainnet claim.
+And during **production bring-up** the deployer owns the diamond between
+`transferOwnership(timelock)` and the Safe-executed `confirmOwnershipTransfer()`;
+a network left in that state is reported unhealthy by the `diamond-owner`
+invariant.
+
+Timelock execution is **not** a deployer power today: `EXECUTOR_ROLE` is granted
+to `address(0)`, so execution is permissionless. It becomes one when the F7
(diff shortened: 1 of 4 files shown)


0xDEnYO:
Two changes since the last review, both in `9fab7137e`, plus a merge of `main`.

**The unprefixed-address finding was real, and the suggested fix was not enough.** Normalizing the two operands closes `<body>` with no prefix, but `isHexAddressForm`'s `0x` was matched **case-sensitively** — so a `0X` form still failed the form test and fell through to the verbatim comparison, escaping by a different spelling of the same idea. Both halves are fixed. Seven spellings now refuse where one did before: unprefixed, either digit case, `0X`, and both Tron `41`-hex forms.

Only *recognition* widens, so nothing that passed for a good reason stops passing — asserted, not assumed: the committed config still passes, an unrelated address still passes, and removing the deployer from `safeOwners` (a reduction of its power) still passes.

**`main` is merged in** because `protect-critical-code` failed here naming 13 `.github/**` files this PR does not touch — its whole diff is four files under `docs/`, `package.json` and `script/deploy/deployer-key-power.*`. Bringing the branch up to date is what cleared the same failure on #2334. Worth knowing that the workflow's own diff already uses three-dot, so this is not the two-dot bug its source comment warns about; tracked as EXSC-947, along with the sharper problem that this gate is *required*, fires only on review submission, and skips on drafts — so a ready PR can satisfy every required check while this one has never evaluated its head.

gvladika (approved):

PR #2340, opened 2026-09-08
feat(deploy): apply the Safe threshold floor to Tron deployments (EXSC-945)

0xDEnYO:

[EXSC-945](https://linear.app/lifi-linear/issue/EXSC-945) — **part (a) only**. Part (b), the hand-typed Safe address in `script/deploy/safe/deploy-safe.ts`, is out of scope here and being handled separately; this PR does not touch that file.

**Stacked on #2338 (`feat/exsc-944-safe-deploy-guards`) and must merge after it** — it reuses `script/deploy/safe/safe-deploy-guards.ts`, which lands there.

`script/deploy/tron/deploy-safe-tron.ts` checked only `threshold < 1 || threshold > ownersEvm.length`, so `--threshold 1` produced a 1-of-6 Safe on Tron mainnet, where `config/networks.json` names a live governance Safe. `script/deploy/healthCheckInvariants.ts` is the only other `SAFE_THRESHOLD` consumer and is EVM-only, so Tron's threshold was floor-checked neither at deploy time nor in the daily sweep.

**Reuse, not a second constant.** The Tron script now calls `assertSafeThresholdFloor` from `safe-deploy-guards`, the same function `deploy-safe.ts` calls, judged against the same `SAFE_THRESHOLD`.

**Placement.** The floor sits after the owner-count range check and before both `getEnvVar('PRIVATE_KEY_PRODUCTION')` and the `--setupOnly` dispatch. After the range check so nothing is swallowed: the range check keeps every input it already refused (`0`, and anything above the owner count), and the floor only refuses inputs that previously passed. Before the key read and the setup-only branch so a refusal happens before any credential is loaded and on both paths, since `--setupOnly` calls `setup()` with the same threshold.

**Tron testnet.** `TRON_DEPLOY_NETWORK` is the compile-time constant `'tron'`, and `config/networks.json` types it `mainnet`, so the exemption is unreachable today — `tronshasta` exists but is `type: "testnet"`, `status: "inactive"`, and is not what this script targets. The answer is still read from repo config via `isTestnetNetwork` rather than hardcoded `false`: retargeting the script is one edit to that constant, and the floor must follow config rather than a literal that would then be wrong. No env var is involved.

**One source for the threshold.** `script/deploy/safe/add-safe-owners-and-threshold.ts` carried its own `EXPECTED_THRESHOLD = 3` beside `SAFE_THRESHOLD = 3` in `script/deploy/shared/constants.ts`. The local const is gone and its 13 references now read `SAFE_THRESHOLD`; the value and behaviour are unchanged. This collapse has no test that can distinguish it — it is a rename onto an identical value — so it rests on review.

**Tests.** `safe-deploy-guards.test.ts` already covers what the guard decides; `script/deploy/tron/deploy-safe-tron-threshold.test.ts` covers the two things only the call site can get wrong — the inputs the Tron script feeds the guard (read from committed `config/networks.json` and `config/global.json`, not fixtures) and where in `run()` the refusal happens. Every refusal is paired with the real current configuration still passing, so none of it is satisfied by a blanket refusal.

The script is never executed, not even with `--help`: its default path deploys a Safe with the production key. The ordering claims are therefore read off the source, with each anchor asserted to occur exactly once so a rename fails the file loudly.

Run it with:

```bash
bun test script/deploy/tron/deploy-safe-tron-threshold.test.ts script/deploy/safe/safe-deploy-guards.test.ts
bun test script/
bunx tsc-files --noEmit script/deploy/tron/deploy-safe-tron.ts script/deploy/tron/deploy-safe-tron-threshold.test.ts script/deploy/safe/add-safe-owners-and-threshold.ts
```

Eight mutations of the call site and of the floor itself — deleting the guard, moving it after the key read, moving it after the setup-only dispatch, hardcoding `isTestnet: true`, downgrading the assert to a discarded evaluate, lowering the CLI default, flattening the floor to 1, and excluding the floor from its own comparison — were each killed with the test total held constant. Harness and per-mutation counts are in the review thread rather than the body, since a score in a body goes stale.

### DIFF
--- a/script/deploy/safe/add-safe-owners-and-threshold.ts
+++ b/script/deploy/safe/add-safe-owners-and-threshold.ts
@@ -3,7 +3,7 @@
  *
  * Proposes Safe transactions to add owners (from `config/global.json:safeOwners`
  * and/or `--owners`) and, when current threshold differs, propose a change to
- * the expected value (`EXPECTED_THRESHOLD`).
+ * the expected value (`SAFE_THRESHOLD`).
  * Supports a single network (`--network <name>`) or every active EVM mainnet
  * (`--all-networks`, excludes `networks.json` entries with `type: "testnet"`).
  * Signs with a Ledger by default; falls back to a private
@@ -26,6 +26,7 @@ import {
 import globalConfig from '../../../config/global.json'
 import networksData from '../../../config/networks.json'
 import { getViemChainForNetworkName } from '../../utils/viemScriptHelpers'
+import { SAFE_THRESHOLD } from '../shared/constants'
 
 import type { ILedgerAccountResult } from './ledger'
 import { assertTicketPresent } from './proposal-intent'
@@ -39,8 +40,6 @@ import {
   storeTransactionInMongoDB,
   type ISafeTxDocument,
 } from './safe-utils'
-
-const EXPECTED_THRESHOLD = 3
 
 const ansi = (code: string, text: string) => `\x1b[${code}m${text}\x1b[0m`
 const green = (text: string) => ansi('32', text)
@@ -88,7 +87,7 @@ interface ICheckResult {
 const main = defineCommand({
   meta: {
     name: 'add-safe-owners-and-threshold',
-    description: `Proposes transactions to add SAFE owners from global.json (and/or --owners) and sets threshold to ${EXPECTED_THRESHOLD}. Single network via --network; --all-networks skips networks where type is testnet.`,
+    description: `Proposes transactions to add SAFE owners from global.json (and/or --owners) and sets threshold to ${SAFE_THRESHOLD}. Single network via --network; --all-networks skips networks where type is testnet.`,
   },
   args: {
     network: {
@@ -155,16 +154,12 @@ const main = defineCommand({
       ) as Address[]
 
       consola.info(
-        `Checking ${networks.length} network(s) against ${expectedOwners.length} expected owner(s) + threshold=${EXPECTED_THRESHOLD}`
+        `Checking ${networks.length} network(s) against ${expectedOwners.length} expected owner(s) + threshold=${SAFE_THRESHOLD}`
       )
 
       const checkResults: ICheckResult[] = []
       for (const network of networks) {
-        const r = await checkNetwork(
-          network,
-          expectedOwners,
-          EXPECTED_THRESHOLD
-        )
+        const r = await checkNetwork(network, expectedOwners, SAFE_THRESHOLD)
         checkResults.push(r)
       }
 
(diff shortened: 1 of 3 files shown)


0xDEnYO:
@coderabbitai review

0xDEnYO:
Mutation run against the two test files this PR's guard is covered by (`deploy-safe-tron-threshold.test.ts` + `safe-deploy-guards.test.ts`), baseline 37 pass / 0 fail. Each mutation was applied to the committed implementation, its search anchor asserted to match exactly once, and the file restored via `git checkout --` with the restore asserted byte-identical. Test total held at 37 in every round.

| # | mutation | pass/fail | verdict |
|---|---|---|---|
| M1 | delete the `assertSafeThresholdFloor` call from `run()` | 33 / 4 | killed |
| M2 | move the guard after `getEnvVar('PRIVATE_KEY_PRODUCTION')` | 36 / 1 | killed |
| M3 | move the guard after the `--setupOnly` dispatch | 35 / 2 | killed |
| M4 | `isTestnet: true` instead of `isTestnetNetwork(TRON_DEPLOY_NETWORK)` | 36 / 1 | killed |
| M5 | `assertSafeThresholdFloor` → `evaluateSafeThresholdFloor`, verdict discarded | 34 / 3 | killed |
| M6 | CLI `--threshold` default lowered from 3 to 1 | 36 / 1 | killed |
| M7 | `const floor = 1` for every network | 31 / 6 | killed |
| M8 | `threshold >= floor` → `threshold > floor` | 33 / 4 | killed |

No survivors. Two caveats worth stating rather than hiding: M1 and M4 also fail `tsc-files` (they strand a now-unused import), so they are not purely behavioural mutants — both were killed by the tests independently of that. And the `EXPECTED_THRESHOLD` → `SAFE_THRESHOLD` collapse is a rename onto an identical value, so no mutation can distinguish it; it rests on review.

The refusals, run against committed `config/networks.json` (`tron` → `type: mainnet`, `safeAddress: TG7unADV9rXUNzH1Q1FkEnqhQ3xxJdXeBs`) and `config/global.json` (6 `safeOwners`):

```text
isTestnetNetwork('tron') = false  owners=6

--threshold 1 -> allowed=false floor=3
   REFUSAL: --threshold 1 is below the 3 confirmations required on tron. A Safe deployed under the floor can act on fewer signatures than the fleet's governance assumes.
--threshold 2 -> allowed=false floor=3
   REFUSAL: --threshold 2 is below the 3 confirmations required on tron. …
--threshold 3 -> allowed=true floor=3
--threshold 6 -> allowed=true floor=3
--threshold 3.5 -> allowed=false floor=3
   REFUSAL: --threshold 3.5 is not a whole number of confirmations. …
```

3 is the CLI default, so the configuration this script actually runs with still passes.

0xDEnYO:
@coderabbitai review

0xDEnYO:
@coderabbitai review

PR #2341, opened 2026-09-08
feat(timelock): cancel-decision matrix for the pre-execute check (EXSC-872)

0xDEnYO:

[EXSC-872](https://linear.app/lifi-linear/issue/EXSC-872/wp-62-cancel-decision-matrix-close-the-open-executor-gap) · WP-6.2

The unattended timelock runner needs a policy for what to do with a queued operation once the
pre-execute re-derivation (EXSC-701) has produced its signals. This PR is that policy and nothing
else: one pure module, `script/deploy/safe/timelock-cancel-decision.ts`, with `evaluateCancelDecision`
for one operation, `evaluateCancelPass` for the pass-level circuit-breaker, two renderers, and
`assertDecisionPermitsExecution` as the broadcast-path guard. It imports nothing, so the policy is
readable and testable without a chain, a Safe, or the queue.

**Cancelling is the last resort, not the response to "red".** A cancelled operation has to be
re-proposed and re-signed by a threshold of humans, so the only input that reaches `cancel` is a
divergence that is *proven*: derived from an authoritative anchor rather than a stored value, and
agreed by at least two independent providers. Everything else — a transient RPC failure, a
sub-quorum observation, a verdict recomputed from a stored codehash, a leg the re-derivation cannot
evaluate at all, an operation that would simply revert — lands on a named non-destructive verdict.
The full matrix, the order of the checks and why that order is the safety property are in
`docs/TimelockCancelDecisions.md`.

Two design points worth a reviewer's attention:

- **Order.** Proven divergence is evaluated *before* the error paths, so an error on one leg can
  neither cause a cancel nor suppress one that another leg proved. The final `execute` branch
  requires all four legs to be affirmative, so an unrecognised signal value blocks rather than
  falling through — tested per field.
- **The circuit-breaker is gated on the executor posture.** Above two cancels in one pass the likelier
  cause is one bug in our own checker, so the cancels are withheld and a human is paged. That is only
  safe where declining to execute is itself a control. `EXECUTOR_ROLE` is granted to `address(0)` by
  both timelock deployment scripts
  (`script/deploy/facets/DeployLiFiTimelockController.s.sol:63`,
  `script/deploy/zksync/DeployLiFiTimelockController.zksync.s.sol:63`), so today anyone can execute a
  matured operation and withholding a cancel would leave a provably-divergent operation executable by
  a stranger. With an `open` or `unknown` posture the breaker therefore pages *without* withholding.

**Deliberately out of scope, and why.** Restricting `EXECUTOR_ROLE` on the live fleet needs no
contract change — the role's admin is `TIMELOCK_ADMIN_ROLE`, which the Safe holds directly, so
`revokeRole`/`grantRole` on each timelock are ordinary Safe-signed calls. That fleet-wide ceremony
and the deployment-script edit that stops new timelocks deploying open are recorded on the ticket,
not started here. No Solidity is touched by this PR. Wiring the module into
`execute-pending-timelock-tx.ts` waits for the re-derivation that produces its signals: this ships
the decision, not a caller, and the doc says so rather than implying it is live.

**Evidence.** `bun test script/deploy/safe/timelock-cancel-decision.test.ts` covers every row of the
matrix, both records (a missing *deployment* record blocks; a missing *sign-time verdict* record
alerts and still executes), the breaker under both executor postures, and each refusal paired with a
passing counterpart so none of them can degrade into a blanket refusal. The deploy-script parse runs
against both committed scripts, asserting the executor role is *still* open, with the falsifying pair
being the same file with the grant renamed to a wallet — so the day the ceremony's script half lands,
that test says so. A mutation round over the module killed every mutation, including the off-by-one on
the provider quorum, the breaker limit and the revert threshold, dropping the anchor-provenance
conjunct, treating `error` or `unsupported` as a mismatch, and letting a non-`execute` decision past
the broadcast guard.

No contracts and no on-chain calls in this PR: it adds one pure TypeScript module, its tests, and a
doc page.


0xDEnYO:
@coderabbitai review

0xDEnYO:
@coderabbitai review

0xDEnYO:
Fixed the `op-not-yet-matured` gating in e7dd6c79b — the branch now re-states `integrity`/`opIdentity`/`deploymentRecord`/`executability` rather than inferring them from the refusals above, which name only the values they reject. Regression test covers all four fields with an unrecognised value; it fails without the fix.

The companion finding on the same branch — that a verified `pending` operation reaches `unclassified-signals`, pages, and is never retried — does not hold at this head. The `op-not-yet-matured` branch sits immediately before that fallback, so a verified pending operation returns `hold`/`notice`/`retry: true` and never reaches it. That branch *is* the fix for the behaviour described, landed in an earlier commit on this PR; `timelock-cancel-decision.test.ts` covers it ("holds a verified pending operation instead of calling it unclassified").

@coderabbitai review

0xDEnYO:
@coderabbitai review

0xDEnYO:
@coderabbitai review

0xDEnYO:
@coderabbitai review

0xDEnYO:
@coderabbitai review

gvladika (approved):

PR #2342, opened 2026-09-08
feat(safe): resolve calldata addresses against the deployment record (EXSC-699)

0xDEnYO:

[EXSC-699](https://linear.app/lifi-linear/issue/EXSC-699/s13-verify-calldata-addresses-against-the-deploy-log) — S13, under [EXSC-686](https://linear.app/lifi-linear/issue/EXSC-686/multisig-signing-process-20) (WP-3.3 / R1.2).

Two new files, no production call site. `script/deploy/safe/calldata-address-check.ts` grades
the addresses a Safe proposal's calldata references against deployment entries the caller
supplies, and `confirm-safe-tx.ts` is not touched: #2329 owns that file and the funnel it will
be wired into. **Wiring waits for #2329 to merge.**

**Pure, and not importing `safe-utils.ts`.** That module runs `dotenv.config()` and pulls in the
Mongo driver at import time, so a decision module importing it would drag both into every
consumer. The four record fields the check reads are declared locally instead; the test asserts
at compile time that `IDeploymentRecord` still satisfies that shape, so the two cannot drift
without CI noticing.

**It grades, it does not decode.** `collectInstalledFacetAddresses`
(`script/deploy/shared/funnel-deploy-gate.ts`) already walks the timelock `schedule` /
`scheduleBatch` envelopes, byte-scans for a cut selector it could not open, and reports what it
could not read through. Duplicating that walker would give the repo two snapshots of the wrapper
list to keep in sync. Instead the input carries an `undecodable` list — the exact shape that
function returns — and any entry in it makes the verdict an **error**. A subset of a proposal's
addresses cannot be graded green, which is how an envelope becomes a bypass.

**Sign time is not propose time, and this does not sit in front of the funnel gate.**
`assertFunnelDeployGate` already refuses an unattributable facet address at propose time. It
proves nothing about what a signer is asked to approve: `confirm-safe-tx.ts` reads pending rows
out of Mongo, and a proposal can be created from the Safe UI, a teammate's script, or a
compromised proposer key without passing any funnel in this repo. The two are disjoint by route
and by time; neither swallows the other.

**Absence from a policy set is never permission.** Every policy lookup in the module is an
allowlist: only `DeploymentIndexSourceEnum.DeploymentRecord` may decide a verdict, and only the
roles named in `REFUSAL_BEARING_ROLES` or `WARN_ONLY_ROLES` have a policy at all. A source or a
role the module has no argument about produces an **error**, so extending either enum without
extending the policy blocks signing rather than quietly deciding addresses. The same rule covers
the `expectations` map: a refusal-bearing reference with no anchor-supplied identity **errors**
rather than grading `identity-unchecked` and passing — an unanswerable identity gets the same
treatment as an address nobody looked up, because a wiring PR that forgets to build the map must
not produce a verdict indistinguishable from a verified one. Keys are normalised inside the
module and validated, so the checksummed form that `_targetState.json` and the selector registry
both carry reaches the name comparison instead of silently matching nothing.

**The source of the entries is itself judged**, because `deployments/<network>.json` and
`<network>.diamond.json` merge only after the proposal has executed (OQ7's corollary), which at
sign time is every address a proposal is about, and
`deployments/_deployments_log_file.json` lags the record and omits recent contracts. One real
facet, `SymbiosisFacet` on mainnet, shows both gaps in the committed data:

- `0x23Fc1b73e66Cd13e988170CB94e252Cb7FF88185` is `SymbiosisFacet@1.0.0` in the deploy-log
  export and appears in **neither** repo file. Those files are current-address maps, so omitting
  a superseded version is correct of them — and it is also exactly what a repo file that has not
  caught up with an address looks like. Trusting one as the record grades a facet the record
  holds on mainnet unknown.
- `0xa0353221443CA4E2e6A040F30A57B47F5A6d479D` is `SymbiosisFacet@2.0.0`, the version mainnet
  currently runs, in both repo files and **nowhere** in the export. So the export cannot rule an
  address out either.

So `DeploymentIndexSourceEnum` names all three sources and only the MongoDB record may decide;
the other two produce an error naming the lag rather than a refusal blaming the address. That
turns a documented corollary into a check that fires.

**What it refuses, what it errors on, and what it deliberately does not.** Refuses on: an
address the record holds nowhere; an address the record holds only on other networks; a recorded
name or version that is not the one an anchor expects; the zero address in a role that must name
a contract; a value that is not a 20-byte address. Errors on: an unreadable record; a source that
may not decide; a call the extractor could not open; an address the store was never asked about;
an expectations key that is not an address or that contradicts another; an identity no anchor
named; a role the module has no policy for. Warns without refusing on: a removal target, in
every one of those states — T2 puts subtractive operations outside the gate entirely, so a
removal target nobody looked up warns exactly as a queried-and-unknown one does. A narrower
query must not false-red the rollback path.

Out of scope, stated rather than left as a gap: **an EOA where a contract is expected** needs
`eth_getCode` and belongs to the codehash gate (WP-2.1), which reads live code anyway;
**third-party addresses** in `batchAddDex` / whitelist calls are not LI.FI deployments by
construction and refusing them would block every dex-approval proposal; **a Remove cut carrying
a non-zero facet address** reverts in `LibDiamond` and belongs to the executability simulation
(S11), not to a second guard in front of it; and **a record the deployer wrote falsely** stays
out — the log records `git rev-parse HEAD` with no dirty flag, so this is a mistake-catcher, not
a lie-catcher. Only S20's attestation makes it the latter. The module header says so.

Run `bun test script/deploy/safe/calldata-address-check.test.ts` for the check's own cases and
`bun test script/` for the suite it sits in.

No Solidity in this PR.


0xDEnYO:
@coderabbitai review

0xDEnYO:
@coderabbitai review

0xDEnYO:
@coderabbitai review

0xDEnYO:
@coderabbitai review

0xDEnYO:
@coderabbitai review

0xDEnYO:
Fixed in `900f46b5b` and resolved the thread. Renamed to the outcome it asserts and added `expect(verdict.refuses).toBe(false)` beside it, so the `error` vs `refuses` distinction is checked rather than implied by the title — the name had claimed the wrong one of the two outcomes this module is built on.

Flipping to ready. The bot review at `38c85848c` covered all of the production logic and `900f46b5b` changes one test file and nothing else (`git diff --name-only` is a single path), so no production code on this PR is unreviewed. Requesting a review at the new head anyway for the record, but not gating the flip on it — the account is on roughly one review per 87 minutes right now and this diff cannot change behaviour.

@coderabbitai review

gvladika (approved):

PR #2343, opened 2026-09-08
feat(safe): RPC quorum for integrity reads (EXSC-870)

0xDEnYO:

[EXSC-870](https://linear.app/lifi-linear/issue/EXSC-870/wp-42-rpc-quorum-for-integrity-reads) · WP-4.2 · R4.4 / R2.7 / T3

An integrity read — a codehash, an owner set, a role holder, a stored value — is
today believed on one endpoint's word, and the endpoint list comes out of a
writable store the deployer influences. `script/deploy/safe/rpc-quorum.ts` adds
the agreement layer: two independent providers, reading the same block, must
return the same value before a read is green. `#2228` built failover
(availability); this is quorum (agreement).

**Not wired into any production call site.** WP-4.2 ships the decision module
only. Three packages are already built-and-unwired behind #2329 and the wiring
sits with that.

## What counts as agreement

Two observations agree when their **value, block number and block hash are all
the same**, and they count as two only when they come from **two independent
providers**. Provider identity is derived from the endpoint host's last two
labels, so several endpoints of one provider count once.

**The quorum is 2** (`MIN_INDEPENDENT_PROVIDERS`), and a caller asking for less
is refused rather than obeyed — at 1 a lying endpoint is the whole evidence
base.

**Any disagreement blocks, whatever the split.** There is no majority rule: a
provider that lies about one value is not evidence about the others, so 2-of-3
does not resolve to the 2. Per T3 every non-green status is an ERROR that blocks
like a FAIL — `assertRpcQuorum` throws on all of them, transient ones included,
and there is no acknowledgement path.

**Absence never reads as agreement.** A provider that errored, that was never
asked, or that answered without saying which block it read is counted toward
neither the responses nor the agreement, and a shortfall gets its own status
(`insufficient-providers` / `insufficient-responses`) rather than falling through
to the green branch.

## Cross-provider failure modes and where each one lands

| Failure mode | Disposition |
|---|---|
| One provider lies | `disagreement` — blocks, never retried; retrying a divergence is a second chance for the liar |
| One provider lags a block | `heights-not-aligned` — values from different blocks are not comparable, so this is *not* graded as a disagreement. Transient, bounded retry |
| Reorg / two chains at one height | `fork-divergence` — same block number, different hash; no comparison between them carries evidence. Blocks |
| Two endpoints, one upstream | Collapsed to one provider by host, then `insufficient-providers`. Identity derivation deliberately over-collapses: merging two providers can only lower the count and refuse, while splitting one would invent a quorum. An operator can declare a shared upstream across unrelated hosts, and that declaration can only ever *merge* — it cannot split two endpoints the URLs already call one provider |
| A provider errors vs answers differently | Strictly separate: an error is a non-response, not a vote |
| Every provider errors | `no-responses` — unverified, not verified-as-unchanged. Transient, bounded retry |
| Fewer providers than the quorum | `insufficient-providers` — never retried, since another attempt cannot add an endpoint |
| Well-formed but empty answer | `agreed-absent` — agreement that nothing is there, which is not a verified read of something that should exist |
| Results at different heights | Same as the lag row: alignment is settled *before* values are compared |
| Quorum set below the minimum | `quorum-misconfigured` — the read is refused rather than run |

Bounded retry (R4.4 / F4) covers only the transient statuses, with exponential
backoff off a capped policy, and the module still blocks when the bound is
exhausted.

**Scoped out, deliberately:** two providers behind one upstream that share no
domain are undetectable from a URL and stay a named residual, closed by an
operator declaration. Endpoint *health tracking* and reordering already live in
`rpcEndpoints.ts`/`viemScriptHelpers.ts` and are not duplicated here. The
ticket's two acceptance criteria that are not code — **≥2 independent providers
on every production chain** and **write-hardening on
`blockchain-configs.RpcEndpoints`** — are infrastructure work this PR cannot
land. **`renderQuorumCoverage` does not measure the first one as wired, and the
0/73 figure an earlier version of this body quoted is a schema property rather
than a fleet fact.** `config/networks.json` stores `rpcUrl` as a *singular
string* — 73 networks, 73 strings, no array field on any of them — so a count
taken from that file reports zero at any quorum above one, today and after any
amount of real remediation. The operational endpoint inventory is the Mongo
`RpcEndpoints` collection (`IRpcEndpoint = { url, priority, isActive?,
environment?, network? }`, one row per endpoint), which is also the writable
store this module's own header names as the threat surface.

So the real coverage number is **unknown from here** and deliberately not
guessed at: querying the production store is out of scope for the session that
built this. `evaluateQuorumCoverage` now takes the source it measured as a
**required** argument and prints it in the rendered line, so a bare number
cannot be quoted as a fleet fact again. Feeding it from `RpcEndpoints` belongs
with the wiring.

## Evidence

Refusals shown firing on live chain state and real repo config, reproducible:

```bash
bunx tsx script/deploy/safe/rpc-quorum-demo.ts
```

It prints, at a pinned mainnet block: the green two-provider read; the same read
with one provider lying; a single-provider shortfall; providers answering at
different heights; a total outage running its bounded attempts and still
blocking; and the fleet coverage report naming every network that cannot reach
quorum.

Unit tests:

```bash
bun test script/deploy/safe/rpc-quorum.test.ts
```

## After the gate round

Two blocking findings, both in the identity periphery rather than the grader, and both fixed:

- **An IP-form endpoint could not share an identity with the hostname form of the same node**,
  so adding an IP alias of a node you already control derived a second provider and greened a
  read from one upstream — the splitting direction, which is the only one that can invent a
  quorum. A name may resolve to that address, so independence is not establishable either way:
  every IP host now collapses onto one identity, and a usable IP endpoint refuses the read with
  `provider-identity-unverifiable`, non-transient, ahead of every verdict that reads the
  provider count. No committed endpoint is an IP literal, so it costs nothing today.
- **A trailing-dot host split from its rooted form**, and nothing asserted the `.filter(Boolean)`
  that prevents it — the last two labels of `rpc.example.com.` are `['com', '']`, which matches
  no hostname identity.

The two identity tests had asserted the earlier intent (an address kept whole rather than sliced
to its last two octets). That concern is real and still holds — an address is not a hostname —
but it is not a reason to make the address an identity of its own, so they now assert the
collapse and state the octet case as the reason an address is handled apart.

Came back clean under direct attack: `agreeingProviders` cannot be nonzero without real
agreement and is not what greenness depends on; `providerId` can only merge and never split
(union-find, `union` the sole mutator); height alignment is settled before values; a same-height
hash divergence blocks and is never retried; an error is never counted as a vote; a quorum below
the minimum is refused rather than obeyed; and `MIN_INDEPENDENT_PROVIDERS = 2 → 1` is killed by
ten tests, so that constant is pinned by value and not through its own symbol.

No Solidity in this PR, so the facet checklist does not apply. No user-visible
behaviour changes yet either — nothing calls the module — so
`docs/MultisigSigningProcess.md` is updated by the wiring package, not this one.

No contracts and no on-chain calls: the module is pure and the demo issues only
`eth_getBlockByNumber` and `eth_getCode`.


0xDEnYO:
@coderabbitai review

0xDEnYO:
@coderabbitai review

0xDEnYO:
Pushed `938f9ead2` + `e38698b5e`. A gate round on the previous two commits found that **the fix for the false red had opened a false green**, so the merge logic is reworked:

An empty host derived the identity `""`, which is not a sentinel, so it won the group's canonical name over the IP sentinel — in every input order. A bare-IP endpoint plus any host-less URL (`file:`, `unix:`, `data:`) sharing a declared `providerId` therefore counted as a named provider, and one honest endpoint alongside those two reported `agreed` on a quorum whose second arm may be the first arm's own address. Both the URLs and the `providerId` come from the same writable store.

Three changes: a host with no labels reaches a sentinel; countability is named positively (non-empty, non-sentinel) instead of as absence from a list of the unverifiable identities already known, which is what let `""` through; and the refusal now asks each IP endpoint whether **its own group** reached a countable identity, which no group name can launder and which does not depend on arrival order. That also closes a second bypass — between the two sentinels the name was first-write-wins, so an unparsable endpoint declared to be the IP endpoint refused in one order and agreed in the other.

Also aligned `independentProviders` in `evaluateRpcQuorum` with `evaluateQuorumCoverage`, which already excluded sentinels: the pre-flight report and the live gate disagreed about the same list.

Falsified each change separately. The emptiness half of the countability test turned out to be observed by nothing — reverting it alone left the suite green — so rather than drop it and be back to a denylist, the deriver's contract is now asserted where a test can hold it.

@coderabbitai review

0xDEnYO:
@coderabbitai review

0xDEnYO:
@coderabbitai review

0xDEnYO:
Review at `e38698b5e` came back with zero actionable comments — it covers the false-green rework in `938f9ead2` and the contract test in `e38698b5e`, which is the part I most wanted a second pair of eyes on.

One nitpick, and I am not taking it: separating action from assertion (Arrange-Act-Assert) across five ranges in `rpc-quorum.test.ts`. It is labelled Trivial / Low value, three of the five ranges are the loop-driven tests where inlining the call is what makes the both-orders assertion readable, and one range is pre-existing. Mostly, though, I cannot find the rule: there is no `.coderabbit.yaml` in the repo and nothing in `.agents/rules/` or `.cursor/rules/` mentions AAA or test structure — the review info says "Configuration used: Repository UI", so the guideline appears to live in CodeRabbit’s web UI.

If the squad does want AAA enforced, it belongs in `.agents/rules/` where humans and every agent can read it, rather than only in the CodeRabbit UI — that is the whole point of the agent-agnostic rules layout. Happy to apply it if someone confirms it is a real house rule; I would rather not reformat tests on a bot default I cannot see.

For contrast, I did act on the naming finding raised on #2342 — that one was substantively wrong (a test called "still refuses" asserting `verdict.error`), not a style preference.

Flipping to ready.

gvladika (approved):

PR #2344, opened 2026-09-08
feat(safe): tell a signer whether a proposal would revert if executed now (EXSC-697)

0xDEnYO:

[EXSC-697](https://linear.app/lifi-linear/issue/EXSC-697/s11-tier-0-executability-simulation) · S11 / WP-3.3 · the executability half of acceptance row A3.3 (the address-lookup half is #2342).

Tier-0 answers one question — *would this proposal revert if it were executed now?* — from `eth_call` plus cheap state reads, with no fork and no archive node. The module is pure: it grades payloads a decoder already recovered and chain reads a collector already made. That is what lets every refusal be exercised against a real diamond without a Safe, a signer or a broadcast.

**Not wired into any production call site.** Packages are stacking behind #2329, and wiring waits for it. `evaluateExecutability` / `renderExecutability` / `assertProposalWouldExecute` follow the shape of `delegatecall-gate.ts` and `safe-deploy-guards.ts`, and the module deliberately does not import `safe-utils.ts` — measured, that drags a top-level `dotenv` `config()` and the Mongo driver into anything importing it.

## The verdict says what kind of answer it is holding

Every finding carries a certainty, and the reason line counts them:

- **Proven** — follows from the payload's own bytes. No state this proposal could meet makes it execute, so the finding cannot be wrong however stale the reads are.
- **Predicted** — follows from state read a moment ago. A cut queued ahead of this one, a later deployment or an advancing nonce can each turn the answer over.

Nothing here is evidence a proposal is *benign*. A clean simulation says the transaction lands, not that it should; that stays the codehash gate's question, and the rendered pass line says so in as many words.

## Enumerated revert modes, and what this claims about each

Everything `LibDiamond.diamondCut` can revert on, plus the proposal-level cases:

| Revert mode | LibDiamond site | Classification |
|---|---|---|
| `Remove` cut carrying a non-zero facet address | `removeFunctions`, `FacetAddressIsNotZero` | **prove** |
| `Add`/`Replace` cut with the zero facet address | `addFunctions` / `replaceFunctions`, `FacetAddressIsZero` | **prove** |
| Any cut with an empty selector array | all three, `NoSelectorsInFace` | **prove** |
| Cut action outside `Add`/`Replace`/`Remove` | `IncorrectFacetCutAction` | **prove** |
| `_init` zero with non-empty `_calldata` | `initializeDiamondCut`, `InitZeroButCalldataNotEmpty` | **prove** |
| `_init` non-zero with empty `_calldata` | `initializeDiamondCut`, `CalldataEmptyButInitNotZero` | **prove** |
| A selector conflict the batch creates **within itself** | `FunctionAlreadyExists` / `FunctionDoesNotExist` | **prove** |
| A nonce the Safe has already passed | Safe `execTransaction` | **prove** — a chain read, but a Safe nonce only rises, so no reachable state undoes it |
| `Add` of a selector the diamond already serves | `addFunctions`, `FunctionAlreadyExists` | predict |
| `Replace` of a selector nobody serves | `removeFunction`, `FunctionDoesNotExist` | predict |
| `Replace` onto the facet that already serves the selector | `replaceFunctions`, `FunctionAlreadyExists` | predict |
| `Remove` of a selector nobody serves | `removeFunction`, `FunctionDoesNotExist` | predict |
| `Replace`/`Remove` of a selector defined on the diamond | `removeFunction`, `FunctionIsImmutable` | predict |
| Facet address holding no code | `addFacet` → `enforceHasContractCode` | predict |
| `_init` target holding no code | `initializeDiamondCut` → `enforceHasContractCode` | predict |
| Caller is not the diamond's owner | `DiamondCutFacet`, `OnlyContractOwner` | predict |
| The inner payload's own `eth_call` reverting | any | predict — the mechanism the ticket specifies |
| A function call whose target holds no code | — | predict, and **not** a revert: the call succeeds and does nothing, which is why it still blocks |
| Two pending proposals at one nonce, or a nonce beyond an unfilled gap | Safe | predict, **warn only** — the queue resolves both |
| Executor balance below the estimated cost | — | predict, **warn only** — both sides move and the remedy is a top-up, not a new proposal |
| `InitReverted` / a reason bubbled from the `_init` delegatecall body | `initializeDiamondCut` | **scoped out** as a static claim; reached only through the real `eth_call` |
| Timelock delay not elapsed, operation not `Ready` | `TimelockController` | **scoped out** — the timelock queue modules own it |
| The outer `execTransaction` envelope | Safe | **scoped out** — it needs signatures that do not exist at sign time; Tier-0 simulates the inner payload, as specified |
| Any other payload's own revert conditions | — | **scoped out** by design, and named per payload in `notSimulated` so a proposal cannot be counted as simulated when it was not |

## Two rulings this had to respect

**T3 — a simulation that could not run is not a simulation that passed.** `error` is separate from `refuses`, `assertProposalWouldExecute` throws on either, and there is no acknowledgement path. It errors when chain state was unreadable, when no `eth_call` was attempted at all, when one payload of several was left unsimulated, when an `eth_call` could not be sent, when the decoder could not read a call through, and on every single fact a finding would have needed and did not have — an unasked question never grades as an answer of no.

It also errors when `eth_call` reports success for a payload a proven finding says cannot execute. One of the two is then wrong about what was simulated, and neither is preferred over the other.

**T2 — a removal is warn-only.** The carve-out turns on certainty, not action. A *predicted* revert on a `Remove` cut warns: the reason such a cut fails is that the selector is not where the map said it was, which is exactly the unverifiability of what is being removed that T2 refuses to block a rollback over. A *proven* revert on a `Remove` blocks — it is a cut that cannot execute whatever anyone verifies, and letting it through spends a threshold of human signatures on a transaction that reverts. An observed `eth_call` revert also blocks, because it is an observation rather than an inference.

## The interlock with #2342

#2342 deliberately scoped out one case for S11: a `Remove` cut carrying a non-zero facet address, which reverts in `LibDiamond.removeFunctions`. That case lands here, as `FacetAddressIsNotZero`, proven and blocking.

It does not duplicate or shadow anything in `calldata-address-check.ts`. That module asks whether an address is accounted for in the deployment record; this one asks whether the cut's shape can execute. `calldata-address-check.ts` is not on this branch at all — this branch is off `origin/main` and touches no existing file — so no guard here can sit in front of one there by construction. When both land the two are independent: neither imports the other, and neither is wired yet.

## Fit with `timelock-cancel-decision.ts`

`evaluateCancelDecision`'s `executability` field takes `'ok' | 'would-revert' | 'error'`, so `toCancelDecisionExecutability` projects a verdict onto exactly those three. `error` wins over `would-revert` deliberately: the cancel matrix holds on an error and cancels on nothing, while a `would-revert` it believes drives a row towards a destructive block — so a verdict that could not decide must not arrive there wearing a decision.

## Evidence

Every refusal was exercised against Ethereum mainnet's real `LiFiDiamond` (`0x1231DEB6f5749EF6cE6943a275A1D3E7486F4EaE`), its real `owner()`, the real facet addresses and selectors `facets()` reports, and the real production Safe's `nonce()` — with a real `eth_call` beside each one whose decoded `LibDiamond` custom error corroborates the finding. The passing pair is a `Replace` of two selectors the diamond really serves onto another facet it really carries: `refuses=false`, and its real `eth_call` succeeds. The reads are `eth_call` and `eth_getCode` against the public endpoint `config/networks.json` names; nothing was signed, sent or broadcast.

Cite the suite by running it rather than by a number here: `bun test script/deploy/safe/executability-simulation.test.ts`, and `bun test --coverage` on the same file for the coverage figure. `bunx tsc-files --noEmit` and `bunx eslint` are clean on both changed files.

No documentation change: nothing is wired, so no operator-visible behaviour changes yet. `docs/MultisigSigningProcess.md` gets this check when the wiring package lands.

No Solidity in this PR — TypeScript only, and no contract is deployed or called with anything but `eth_call` / `eth_getCode`.


0xDEnYO:
@coderabbitai review

0xDEnYO:
@coderabbitai review

0xDEnYO:
Pushed `19c48af83`. A gate round on `ca3dce5bc` found the case-insensitive comparison consistent on every side and the composed-proof detection intact under mixed-case forms — no false green. One real item, pre-existing rather than introduced:

`isEvmAddress` anchored on a lowercase `0x`, and its only caller skips a cut it rejects **without recording anything**. So a facet given as `0X…` was dropped from the code check silently: no `FacetContainsNoCode` finding and no unchecked note either, which is the one combination that tells the signer nothing. The same escape was closed in `deployer-key-power.ts`. Fixed, with a test that fails without it.

Also dropped a docblock line that asserted a caller precondition as a fact and restated the interface docstring twelve lines above.

Known and not fixed: the delta moved the immutability fixture from a lowercase observed value to a checksummed one rather than covering both, so the lowercase twin is no longer exercised. Harmless — `normalise` on a lowercase string is the identity — but worth restoring if this PR takes another round.

@coderabbitai review

0xDEnYO:
@coderabbitai review

0xDEnYO:
@coderabbitai review

0xDEnYO:
Review at `19c48af83` returned zero actionable comments, over the range `b37c782b3..19c48af83` — so it covers both the case-insensitive comparison work and the `0X` prefix fix.

Three nitpicks from the earlier review at `8f3e08eb7`, all Trivial, none taken, and the reasoning is the same for each: they cite "coding guidelines" that are not in this repo. There is no `.coderabbit.yaml`, and `.agents/rules/` has nothing on explicit return types, nested ternaries or test structure — the review info says "Configuration used: Repository UI", so they live in CodeRabbit’s web UI.

- **Nested ternaries** (three sites, two pre-existing): this codebase uses them widely — `confirm-safe-tx.ts` picks its nonce colour that way — and eslint is green, so no repo-enforced rule is violated. Rewriting them here would make this file inconsistent with the rest.
- **Explicit return type on `findingFor`**: the fairest of the three, since every other helper in that file does declare one. It is a test helper and a one-line change; I have not taken it only because the review budget is at zero for the next window and this PR is otherwise finished. Worth folding into any future round on this PR.
- Also outstanding from my own gate round, and deliberately not fixed for the same reason: the delta swapped the immutability fixture from a lowercase observed value to a checksummed one rather than covering both, so the lowercase twin is no longer exercised. Harmless (`normalise` on a lowercase string is the identity).

If the squad wants any of these enforced, they belong in `.agents/rules/` where humans and every agent can read them, rather than only in the CodeRabbit UI.

Flipping to ready.

gvladika (approved):

PR #2347, opened 2026-09-09
feat(script): gate the bash sendOrPropose direct-broadcast route (EXSC-956)

0xDEnYO:

EXSC-956 — https://linear.app/lifi-linear/issue/EXSC-956

`SEND_PROPOSALS_DIRECTLY_TO_DIAMOND=true` (and staging, and testnets) route the bash
`sendOrPropose` chokepoint to `universalCast sendRaw`, which enters neither proposal funnel.
`assertFunnelDeployGate` runs inside `propose-to-safe.ts` / `propose-to-safe-tron.ts`, and
`assertDirectBroadcastDeployGate` covers only `diamondUpdateFacet.sh`'s own direct cut — so on a
production mainnet in the bring-up window this route could broadcast a cut from the production
deployer key with nothing vouching for the code. `docs/MultisigSigningProcess.md` said so in as many
words ("is **not** gated today; it never was").

This route hands over **calldata**, exactly what the funnels hand over — so it reuses
`assertFunnelDeployGate`'s cut recovery through a thin CLI
(`script/deploy/shared/assert-direct-broadcast-gate.ts`) rather than a second implementation. It is
`diamondUpdateFacet.sh` that has facet *names* and therefore needs its own; that gate is untouched,
so every route it covered is still covered by it.

Disjoint by construction, per D9: a proposal is gated on its calldata inside the funnel, and only
the route that never reaches a funnel is gated here.

Four details worth a look:

- **The refusal sits before the broadcast.** `assertDirectBroadcastCalldataGate` is called on the
  line above `universalCast "sendRaw"`, and a test moves it below to prove the ordering is load-bearing.
- **The exemption set is named by what may proceed:** exactly `ENVIRONMENT == "staging"`, plus
  testnets. `getPrivateKey` hands out the production key for every ENVIRONMENT that does not
  *contain* "staging", so the exact-string match keeps the gate at least as broad as the key it
  protects — `staging2` is a staging key and is gated anyway. Both skips print their reason; no
  silent skip.
- **Exit 0 is not consent.** A CLI that never reached `runMain` also exits 0, so the gate would
  print that it passed without having run — the failure mode @gvladika blocked on this PR. The CLI
  now writes an explicit `DIRECT_BROADCAST_GATE_ALLOWED` line on every allow path, and
  `assertDirectBroadcastCalldataGate` requires it as a *whole line on stdout*: `CALLDATA` and
  `NETWORK` are the caller's, so a substring match over a `2>&1`-merged stream would let a CLI that
  merely echoed its arguments stand in for the gate. Its entrypoint check realpaths both sides,
  because `[CONV:NODE-RUNTIME-APIS]`'s plain compare is false through any symlinked path — see the
  measurement on EXSC-971.
- **Non-cut calldata is the common case on this route** (whitelist sync, periphery registration,
  ownership). `assertFunnelDeployGate` already skips, with a logged reason, when no call installs
  facet code — so the gate cannot degrade into refusing everything.

Falsified, each mutation killing exactly the intended assertion and the tree restored byte-identical
after each: call site deleted (5 fail); gate moved after the broadcast (1 fail — the ordering case);
`staging` matched as a substring (1 fail — the `staging2` case); call site passing `$TARGET` instead
of the calldata (1 fail); the CLI's `process.exit(1)` turned into `exit(0)` (2 fail — both real-data
refusals); the token check relaxed from `grep -Fxq` to `grep -Fq` (1 fail — the token-inside-a-line
case); the streams re-merged with `2>&1` (1 fail — the stderr case); the entrypoint compare stripped
of `realpath` (1 fail — the symlink case).

Deliberately out of scope: the divergence policy itself is `verify-approvals.ts`'s, already covered
by its own suite, and this PR only routes into it. The new CLI tests therefore exercise the real
entry point against real repo data up to the attribution step (a `diamondCut` whose facet address no
production log records is refused; a testnet skips; a non-cut passes) and do not spin up the
network-dependent `main`-equivalence leg.


0xDEnYO:
@coderabbitai review

0xDEnYO:
@coderabbitai review

0xDEnYO:
@coderabbitai review

0xDEnYO:
@coderabbitai review

script/deploy/shared/assert-direct-broadcast-gate.ts
```diff
@@ -0,0 +1,57 @@
...
+      process.exit(1)
+    }
+
+    consola.success(
+      `Direct-broadcast deploy gate passed for ${args.network} - nothing has been broadcast yet.`
+    )
+  },
+})
+
+if (import.meta.main) runMain(main)
```
gvladika: **Blocker: this guard makes the gate a silent no-op on the machines that run deploys.**

`tsx` runs on Node, where `import.meta.main` does not exist before Node 22.23. It is `undefined`, the guard is falsy, `runMain` never runs, and the process exits **0**. `assertDirectBroadcastCalldataGate` reads exit 0 as "allowed" and prints `[info] direct-broadcast deploy gate passed`.

Verified on a Node v18.20.8 checkout:

```
$ bunx tsx script/deploy/shared/assert-direct-broadcast-gate.ts --network arbitrum --calldata not-hex
EXIT=0        # no output at all; this input must be refused
```

The four real-CLI cases in `script/helperFunctions.gate.test.ts` fail there too (15 pass / 4 fail) — and the 15 that pass stub `bunx`, so they would pass against a CLI that does nothing. CI is green only because `ts-unit-tests.yml` uses `setup-bun` with no `setup-node` and inherits the runner's Node 24. Deploys do not run in CI; they run in an operator's shell, where this is inert while printing that it passed.

This is `[CONV:NODE-RUNTIME-APIS]` (`.agents/rules/200-typescript.md`), which names this exact failure mode, and #2358's guard test is deliberately scoped to `import.meta.dir` so it will not catch it.

Suggested fix — the portable form that rule names (needs `import { fileURLToPath } from 'node:url'` added above):

```suggestion
if (process.argv[1] === fileURLToPath(import.meta.url)) runMain(main)
```

Worth pairing with it: the bash side treats CLI exit 0 as consent, so a gate that never ran is indistinguishable from a gate that passed. Having the CLI emit an explicit allow token that `assertDirectBroadcastCalldataGate` requires would turn this class of bug from fail-open into fail-closed.

0xDEnYO: Fixed in 2a8843f45.

Replaced `import.meta.main` with the portable `fileURLToPath` + `realpath` entrypoint this repo already uses (`proposePeripheryWithWhitelist.ts`), and paired it with an explicit `DIRECT_BROADCAST_GATE_ALLOWED` token. `assertDirectBroadcastCalldataGate` now refuses a CLI that exits 0 without that token, so a guard that never ran cannot print `[info] direct-broadcast deploy gate passed`.

Staying draft until Linear is `Fixes`.

gvladika (approved):

PR #2348, opened 2026-09-09
feat(safe): pinned expected state — grade the target state against origin/main at sign time (EXSC-704)

0xDEnYO:

[EXSC-704](https://linear.app/lifi-linear/issue/EXSC-704/s18-pinned-expected-state-kills-the-wrong-branch-problem) — WP-3.2, pinned expected state.

The ticket was auto-closed on 2026-09-07 by the misnamed branch of #2324 (WP-1.4) and reopened to Backlog; nothing of WP-3.2 existed before this PR.

**The problem.** `_targetState.json` was read from `process.cwd()` (`facet-version-utils.ts`), so the version a signer was shown as "expected" came from whatever branch their checkout happened to be on. A reviewer sitting on the proposer's branch got a proposer-controlled anchor (F17), and everyone else got false red banners — the worst possible conditioning, because it teaches signers that red means branch noise.

**The anchor is now pinned.** `pinned-target-state.ts` reads `git show origin/main:script/deploy/_targetState.json` after a fresh `git fetch`, from a repo root resolved off the module rather than the working directory. One fetch and one blob read per process. The proposer's branch is never read at all: nothing a proposer controls can turn this check green. A fetch that fails refuses rather than falling back to a stale local `origin/main` — a stale ref is not an anchor.

**R4.6 is graded per case, not as one equality.** That single equality is the defect this package exists to remove, because the target state is a post-hoc record in today's workflow — the PR updating it merges *after* sign and execute:

- **Upgrade of a facet `main` already targets** — the mechanical anchor, and the majority of proposals. A downgrade refuses; so does a version pair that cannot be ordered, and a proposed version no deployment record resolves (with an anchor present, an unresolvable proposed version cannot rule out a downgrade). First mechanism against OQ6.
- **First-time add** — no entry on `main` by construction. Labeled `not previously targeted`, corroborated by the cross-fleet count of networks already declaring that contract at that version, and it does **not** block. Blocking here would refuse every honest new-chain rollout with no remedy; intent rests on the mandatory ticket link (WP-1.2) and the linked PR.
- **Removal** — reported only (T2).

**An address with no record is a third question, not a first-time add.** Self-review found this fail-open: without a contract name there is nothing to look the anchor up by, and clearing that as "not previously targeted" would have handed a free pass to any address a proposer named. It is now `contract-unidentified` and refuses. Absence of a *record* and absence of an *entry on main* are different facts and get different statuses.

**A deployment record can contradict itself, and the first match is not an answer** (CodeRabbit, round 1). `(network, address)` is not unique in the production cache — measured against the live mirror: **6 duplicate pairs**, two contradicting on version (`tron` `TokenWrapper` 1.1.0/1.2.1, `tron` `AllBridgeFacet` 2.1.1/2.1.2) and one on name (`metis`, two contracts at one address). Taking whichever the unsorted scan reached first could have produced a **false downgrade refusal** — the exact false red this package exists to remove. The lookup now returns `resolved` / `ambiguous` / `unrecorded`, and a contradiction refuses as `deployment-record-ambiguous`, naming both values. One refinement on the suggestion: a record with **no** version cannot contradict one that has a version, so those are filtered first — the remaining 3 duplicate pairs are exactly that shape, `PolymerCCTPFacet` on base and polygon among them, and calling those ambiguous would have false-red'd real fleet addresses.

**The policy is named by what may proceed.** `STATUSES_CLEARED_TO_PROCEED` names the five statuses that may be signed; every other status refuses, so a status added later has to be admitted deliberately rather than falling through a denylist.

**Look hardest here — two deliberate calls a reviewer should weigh.** The gate is placed after `Do Nothing` has already `continue`d, so it refuses **Execute** as well as **Sign**: executing a downgrade is the harm, and a refused proposal can be cancelled and re-proposed. And an anchor that cannot be refreshed refuses rather than falling back, which means an offline reviewer cannot sign or execute a cut at all — the emergency path (§7) is separate tooling, but if that trade is wrong the fix is a named status admitted to the may-proceed set, not a silent fallback.

**Placement.** The refusal sits after the pre-existing nonce gates — which only gate execute actions, and whose messages stay intact — and before the acknowledgement prompt, so a refused proposal cannot be acknowledged and nothing has been signed or broadcast at that point. It `continue`s to the next proposal, so one bad network does not cost the rest of a fleet run.

**One walker, not two.** Getting per-entry cut *actions* (Remove included) needed a decoder that `collectInstalledFacetAddresses` did not expose. Rather than write a second one, the envelope enumeration moved to `script/deploy/shared/diamond-cut-calls.ts` and the funnel deploy gate now delegates to it — so the wrappers that can be seen through (`diamondCut`, timelock `schedule` and `scheduleBatch`, the depth bound, the byte-aligned-selector refusal) are enumerated once. #2324's 45 gate tests pass unchanged, including the address-ordering ones.

**Removed rather than left available:** `getTargetStateFacetVersion` and the collapsed `VERSION MISMATCH` warn it fed. Leaving a cwd-reading target-state helper in the tree leaves the footgun.

**Tron.** A Tron deployment record stores the base58 address, so the hex address the cut carries matches nothing there — every Tron facet would have graded as "never targeted" and a Tron downgrade would have passed silently. The deps expand the address to both forms, and a test pins the base58 value rather than deriving it from the conversion it checks.

## Review round 2

**A blank-version sibling must not mask a *name* disagreement.** The blank-version filter above was applied before *both* the versions and the names were collected, so `{AcrossFacetV4, 1.0.0}` beside `{SomethingElse, ""}` at one address resolved as `AcrossFacetV4` and the anchor was looked up under a wrong-but-plausible name. Names now come from every match; versions still ignore blanks. Latent, not live — the three blank-version pairs on today's mirror all agree on name, and the old-vs-new resolution diff over all 4,438 on-network pairs is **0**.

**`contract-unidentified` is a record gap, and widening the lookup is the wrong remedy — I tried it and withdrew it.** Measured: **13 installed facet addresses** across the 71 non-staging diamond logs (1,262 installed facets) have no deployment-record row on their own network — `avalanche` `DiamondCutFacet`, and the whole of `robinhood`'s core set. The suggested "same address on any network, name must agree" fallback resolves all 13, so I built it, then measured what it actually does and reverted it:

- **Its premise is false on this repo's own data.** `0xae77c9ad4af61faec96f04bd6723f6f6a804a567` is `CalldataVerificationFacet` v1.1.0 on 17 networks and `LiFuelFeeCollector` v1.1.0 on `opbnb`. Same address, same version, different contract. Cross-chain address identity is a deployment convention here, not a property — deployment goes through CREATE3 with a salt of `keccak(DEPLOYSALT ‖ contractName)`, which binds the *name*, and binds the bytecode only for deployments that went through the script.
- **12 of the 13 flipped from a refusal to a clearance**, and each of those `robinhood` clearances rested on **exactly one** foreign row, from `outlaw`. No corroboration, and the ambiguity check cannot fire on a single-row shape. That is turning a refusal into a pass on weaker evidence than the refusal.
- **It bought no signability.** The sign-time codehash gate reads the record strictly per `(address, network)` (`rebuild-attestations.ts`), so those same 13 addresses are refused one gate over as `no-record` → UNVERIFIABLE regardless.

So the refusal stands and now names its remedy — write the missing MongoDB record — and says explicitly that the same address elsewhere is not consulted, so nobody re-derives the idea from the message.

**`proposed-version-unresolved` has a measured blast radius, and it is not zero.** `DiamondCutFacet` at `0xf7993A8d…` (11 networks) and `DiamondLoupeFacet` at `0xF5ba8Db6…` (10) carry `version: ""` in **every** row they have anywhere in the cache, while `origin/main` declares both at `1.0.0`. A proposal that Adds or Replaces either address — re-registering a cut or loupe facet that fell out of a diamond — is unsignable **and unexecutable** on those chains, with no override; one already at threshold would have to be cancelled. The refusal is kept: a blank proposed version genuinely cannot be ordered against a known anchor, same class as `version-not-comparable`, and admitting it reopens the downgrade hole on exactly the facets that control the diamond's upgrade surface. What changed is that it names the remedy — the blank is in the deployment record, not in the cut.

**`origin` is not an identity, a short ref name is not the ref you fetched, and a fetched ref is not always fetched.** Three separate ways the anchor could be authored by whoever supplied the clone:

- The read now refuses unless `git remote get-url origin` is `github.com/lifinance/contracts`, as its own reason outside the may-proceed set. `get-url` rather than `config remote.origin.url` on purpose: it applies any `insteadOf` rewrite, so it reports where a fetch would really go — verified against a repo whose configured URL is canonical and whose `insteadOf` points elsewhere; `get-url` prints the rewritten target and the check refuses.
- The blob is read through `refs/remotes/origin/main` in full. Git resolves the short `origin/main` through `refs/`, tags and heads *before* remote-tracking refs, so a tag of that name in the clone shadowed the ref the fetch had just written. Falsified: restoring the short name fails exactly the new shadowing test.
- The fetch is explicit (`+refs/heads/main:refs/remotes/origin/main`). `git fetch origin main` relies on git's *opportunistic* ref update, which a clone whose fetch refspec does not cover main skips — exit 0, no error, stale anchor. That is not exotic: `actions/checkout` sets a single-branch refspec. The test falsifies the old behaviour rather than asserting the argv — it proves `git fetch origin main` leaves `origin/main` where it was, then reads the moved anchor through the explicit refspec.

**A transient fetch failure no longer pins the process,** and neither does a `git` exec that could not run. Only the deterministic outcomes are memoized — the same distinction the sibling cache reader in `facet-version-utils.ts` already documents. A remote that could not be *read* is also no longer reported as the *wrong* remote, which would have pointed the signer at a problem that does not exist.

## Falsification

16 mutations. Each production change was mutated, the specific new test confirmed failing, then restored byte-identical (`git checkout --` from a commit, verified with `git diff --exit-code`):

| Mutation | Tests killed |
|---|---|
| first-time add dropped from the may-proceed statuses | 3 |
| removal dropped from the may-proceed statuses | 2 |
| `proposed-version-unresolved` admitted to the may-proceed statuses | 1 |
| `pinned-state-unavailable` admitted to the may-proceed statuses | 1 |
| `no-diamond-cut` dropped from the may-proceed statuses | 2 |
| downgrade comparison direction flipped | 4 |
| anchor read from the working tree instead of `origin/main` | 4 |
| failed fetch ignored and the read attempted anyway | 1 |
| anchor read eagerly instead of only for install actions | 1 |
| `Remove` action rewritten to `Add` in the walker | 1 (walker) + 2 (evaluator) |
| gate moved after the acknowledgement prompt | 1 |
| Tron base58 address form dropped from the record lookup | 1 |
| `contract-unidentified` admitted to the may-proceed statuses | 1 |
| record lookup back to first-match-wins | 3 |
| a blank-version sibling counted as a contradiction (the false-red direction) | 2 |
| `deployment-record-ambiguous` admitted to the may-proceed statuses | 1 |
| a blank-version sibling allowed to hide a name disagreement | 1 |
| the anchor read through the short `origin/main` instead of the remote-tracking ref | 1 |
| the explicit fetch refspec back to `origin main` | 1 |

Five earlier mutation attempts applied nothing (indentation mismatch) and reported 0 failures, which would have read as a coverage gap; the harness now checks that the edit actually landed before running anything.

The `origin/main` property is proven against a real git repository built inside the test: the anchor is committed to `main`, then a `proposer-branch` commits a `9.9.9` version and is checked out — the verdict still reads `1.2.0`, and the same test asserts the working tree really does say `9.9.9`, so the absence is paired with a present.

The gate's *placement* is a source-order assertion, not an executed one: `confirm-safe-tx.ts` signs and broadcasts, so it cannot be spawned from a test. Its docblock says so. It is falsified by actually moving the block.


0xDEnYO:
@coderabbitai review

0xDEnYO:
@coderabbitai review

0xDEnYO:
@coderabbitai review

0xDEnYO:
@coderabbitai review

script/deploy/safe/facet-version-utils.ts
```diff
@@ -107,43 +110,51 @@ export function getDeployedFacetVersionFromLog(
...
 
-    return typeof version === 'string' ? version : null
+    const networkLower = network.toLowerCase()
+    const found = records.find(
+      (r) =>
+        r.network?.toLowerCase() === networkLower &&
+        typeof r.address === 'string' &&
+        normalizedCandidates.includes(r.address.toLowerCase())
+    )
+    if (!found) return null
```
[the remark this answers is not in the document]
0xDEnYO: Valid, and worse than "can" — measured against the live production mirror, `(network, address)` is not unique: 6 duplicate pairs, and `records.find(...)` returning the first unsorted one could resolve either way.

Two of them contradict on version — `tron` `TokenWrapper` 1.1.0/1.2.1 and `tron` `AllBridgeFacet` 2.1.1/2.1.2 — so a cut installing 2.1.2 against a `main` declaring 2.1.2 could have been graded a false `downgrade`, which is exactly the false-red this package exists to remove. One contradicts on name (`metis`: `LiFuelFeeCollector` vs `TokenWrapper` at one address, both 1.0.1), leaving no key to look the anchor up by.

Fixed in adecc6ab8: `resolveDeployedContractByAddress` now returns a discriminated `resolved` / `ambiguous` / `unrecorded`, and a contradiction reaches the signer as a new `deployment-record-ambiguous` status that is **not** in the may-proceed set, so it refuses and names both values rather than picking one.

One refinement on top of the suggestion: a record carrying **no** version cannot contradict one that does, so those are filtered before the comparison. The other 3 duplicate pairs are exactly that shape — including `PolymerCCTPFacet` on base and polygon, a facet — and treating them as ambiguous would have false-red'd real fleet addresses. Both directions are falsified: reverting to first-match-wins fails 3 tests, and counting a blank sibling as a contradiction fails 2.

0xDEnYO:
@coderabbitai review

0xDEnYO:
Answering a question raised in this PR's build trail: **the flip out of draft at 09:07:51Z was me**, deliberately, as part of getting the queue's finished PRs ready. Nothing anomalous, and no need to convert it back.

Worth recording why it looked anomalous. The building session's brief was to open this as a draft and leave it there, and it correctly reported that none of its own commands could have flipped it. But `gh api user` is `0xDEnYO` for every session on this machine, so the timeline attributes my `gh pr ready` and its own calls to the same actor — the API cannot distinguish them. It was right to flag it rather than assume, and right not to revert a promotion it could not attribute.

Independently verified the Major fix in `adecc6ab8` from the source rather than the report: `deployment-record-ambiguous` refuses when a record contradicts itself, and a record carrying *no* version is handled separately as `proposed-version-unresolved` with the reasoning that a downgrade cannot be ruled out. That separation is the part that matters — treating the no-version duplicates as ambiguous would have false-red'd real fleet addresses, `PolymerCCTPFacet` on base and polygon among them, which is the exact false red this package exists to remove.

gvladika (commented):
Review of the pinned expected-state gate. The design is sound and the falsification table is the strongest part of this PR — the `origin/main` property proven against a real git repo, and the harness now checking that a mutation actually landed, are both the right instinct.

I simulated the new gate against the **live fleet** to see what it would do today: all 1262 installed facet entries in `deployments/*.diamond.json`, graded against the real `.cache/deployments_production.json` (4442 rows) and `origin/main`'s target state. Distribution:

| status | count |
|---|---|
| `matches-main` | 839 |
| `downgrade` | 244 |
| `not-previously-targeted` | 110 |
| `ahead-of-main` | 36 |
| **`proposed-version-unresolved`** | **21** |
| **`contract-unidentified`** | **14** |

The last two rows are hard refusals on addresses installed in production right now, and they drive the two most significant comments below. Five inline comments: one high, four medium.

Verified green on this head: 119/119 tests across the five affected suites (including #2324's 45 gate tests unchanged), `tsc --noEmit` clean on every changed file, full CI green. `eslint` can't run locally (`ERR_REQUIRE_ESM` in `@typescript-eslint/parser`) — pre-existing and unrelated; `validate-scripts` covers it in CI.

Also: the CodeRabbit round-1 thread is substantively resolved (the bot replied ✓ and logged a learning) but stayed **open** because it couldn't self-resolve — worth closing manually before pinging `#dev-sc-review`.

script/deploy/safe/pinned-target-state.ts
```diff
@@ -0,0 +1,571 @@
...
+        crossFleetCount: proposedVersion
+          ? countNetworksDeclaring(read.state, contractName, proposedVersion)
+          : null,
+        status: 'not-previously-targeted',
+        detail: `${contractName} is not previously targeted on ${network} in ${PINNED_REF} — expected for a first deployment, since the target-state update merges only after execution. Intent rests on the linked ticket and PR.`,
+      })
+      continue
+    }
+
+    if (!proposedVersion) {
```
gvladika, line 385: **high — `proposed-version-unresolved` blocks core-facet cuts on 11 chains today.**

`DiamondCutFacet` (`0xf7993A8df974AD022647E63402d6315137c58ABf`) and `DiamondLoupeFacet` (`0xF5ba8Db6fEA7aF820De35C8D0c294e17DBC1b9D2`) carry `version: ""` in **every** production cache row they have — 21 blank rows across mainnet, polygon, optimism, bsc, gnosis, cronos, boba, mode, fuse, moonbeam and arbitrum (`DiamondCutFacet` only). Not one has a non-blank version anywhere in the cache. Meanwhile `origin/main` declares both at `1.0.0`.

So: `nonBlank(r.version)` → null, `mainVersion` → `"1.0.0"`, status → `proposed-version-unresolved`, which is absent from `STATUSES_CLEARED_TO_PROCEED`.

Failure scenario: a proposal that Adds or Replaces either address — re-registering a loupe or cut facet that fell out of a diamond — becomes unsignable **and unexecutable** on any of those chains, with no override. A proposal of that shape already at threshold turns unexecutable and has to be cancelled; re-proposing doesn't fix the blank Mongo row.

Refusing is defensible in isolation — a blank proposed version genuinely can't be ordered against a known anchor, same class as `version-not-comparable`. Two things would make it safe to ship: the PR description should own the measured blast radius, and this `detail` string should name the remedy (fix the deployment record), since right now it tells the signer a downgrade can't be ruled out without saying what to do about it. Note the contrast with the ambiguous-record case you fixed for CodeRabbit: there a filter accounts for blank versions, here nothing does — despite the blank rows provably existing.

script/deploy/safe/pinned-target-state.ts
```diff
@@ -0,0 +1,571 @@
...
+    const contractName =
+      deployed.kind === 'resolved' ? deployed.contractName : null
+    const proposedVersion =
+      deployed.kind === 'resolved' ? deployed.version : null
+
+    // An address with no record is not the same question as a contract with no
+    // entry on `main`. Without a name there is nothing to look the anchor up
+    // by, so clearing it as a first-time add would hand a free pass to any
+    // address a proposer chose to name.
+    if (!contractName) {
```
gvladika, line 357: **medium — `contract-unidentified` is a systematic core-facet record gap, not a rare edge.**

`resolveDeployedContractByAddress` keys on `(network, address)`, and 14 of the 1262 installed addresses have no row at all:

- `avalanche` `DiamondCutFacet` — at the *same* `0xf7993A8d…` that resolves fine on ten other chains, just no `avalanche` row
- `robinhood` — all 12 core facets (DiamondCut, Loupe, AccessManager, CalldataVerification, EmergencyPause, GenericSwapV3, Ownership, PeripheryRegistry, WhitelistManager, Withdraw, RelayDepository, AcrossFacetV4)
- `tron` `EmergencyPauseFacet` — installed at `TNDAp17M3vKJ…` while the only recorded address is `TGyX2ysGYjKG…`, so the base58 expansion below doesn't rescue this one; it's genuinely unrecorded

`robinhood` isn't an empty chain — it has 15 cache rows — but none for its core facets. The gap clusters on facets installed at diamond creation, so this is a repeatable hole in whatever path writes those records.

That collides with this PR's own rationale. The description argues blocking a first-time add "would refuse every honest new-chain rollout with no remedy" — but `contract-unidentified` does exactly that whenever the record row is absent, and `robinhood` shows absence happening for a whole chain's core set. Turning the fail-open into a refusal was right; the gap is that the refusal has no remedy path either.

Widening the lookup to "recorded at this address on any network, name must agree" would cover `avalanche` cleanly and keep the fail-closed property, since a name disagreement would still go `ambiguous`.

script/deploy/safe/pinned-target-state.ts
```diff
@@ -0,0 +1,571 @@
...
+  const git = options?.git ?? defaultGit(repoRoot)
+  let memo: PinnedTargetStateRead | undefined
+
+  return () => {
+    if (memo) return memo
+
+    try {
+      git.fetch()
+    } catch {
+      memo = { ok: false, reason: 'fetch-failed' }
```
gvladika: **medium — a transient fetch failure is memoized for the life of the process, against this package's own stated policy.**

`fetch-failed` is written into `memo` and returned forever. Failure scenario: on a 70-network fleet run the first install-action proposal hits a DNS blip or a momentarily dropped tunnel; every remaining proposal then refuses `pinned-state-unavailable`, and since the anchor never retries the operator must kill the run and redo the whole startup — Ledger transport, blind-signing check, deployment-cache warm, reconcile sweep.

The sibling reader in this same package deliberately rules the other way. `loadProductionDeploymentRecords` (`facet-version-utils.ts:59`):

> Transient read/parse faults (e.g. a partially-written cache during a concurrent refresh) must not permanently disable facet-version display — only the deterministic paths above (path escape, missing file, invalid shape) memoize null.

Same distinction applies here: memoize `ok:true`, `blob-unreadable` and `invalid-shape` (all deterministic for the run), but let `fetch-failed` retry. That keeps the "a stale ref is not an anchor" property — a retry still refuses if the remote is genuinely unreachable — while not converting one packet loss into a restarted fleet run.

script/deploy/safe/facet-version-utils.ts
```diff
@@ -107,44 +110,85 @@ export function getDeployedFacetVersionFromLog(
...
+    const candidates = versioned.length > 0 ? versioned : matches
+
+    const versions = [
+      ...new Set(
+        candidates
+          .map((r) => nonBlank(r.version))
+          .filter((v): v is string => v !== null)
+      ),
+    ]
+    const contractNames = [
```
gvladika, line 179: **medium — a blank-version sibling masks a *name* contradiction.**

`contractNames` is derived from `candidates` (post-`versioned`-filter) rather than from `matches`. The refinement you made over CodeRabbit's suggestion is right for **versions** — a blank version can't contradict a present one — but it also silently drops a **name** disagreement.

Failure scenario: rows `{name: 'AcrossFacetV4', version: '1.0.0'}` and `{name: 'SomethingElse', version: ''}` at the same `(network, address)`. `versioned` keeps only the first, so `contractNames` has one entry, the result is `resolved` as `AcrossFacetV4`, and the anchor is looked up under a wrong-but-plausible name that clears the gate. Today's `metis 0x851450e3…` pair (LiFuelFeeCollector / TokenWrapper, both at `1.0.1`) correctly refuses — add a blank version to one side and it stops refusing.

I measured the current mirror: the three blank-version pairs (`moonbeam GasZipPeriphery`, `base`/`polygon PolymerCCTPFacet`) all agree on name, so **nothing is masked today** and this is latent rather than live.

One-line fix — take the name set from all matches, the version set from `candidates`:

```ts
const contractNames = [
  ...new Set(
    matches
      .map((r) => nonBlank(r.contractName))
      .filter((n): n is string => n !== null)
  ),
]
```

script/deploy/safe/pinned-target-state.ts
```diff
@@ -0,0 +1,571 @@
...
+
+/** Injectable git reads, so tests never touch a real remote. */
+export interface IPinnedStateGit {
+  fetch: () => void
+  show: (revSpec: string) => string
+}
+
+const defaultGit = (repoRoot: string): IPinnedStateGit => ({
+  fetch: () => {
+    execFileSync('git', ['fetch', '--quiet', 'origin', 'main'], {
```
gvladika: **medium — the anchor is whatever this clone happens to call `origin`.**

`defaultGit` fetches and reads `origin/main` with no assertion about what `origin` *is*. The module header claims "nothing a proposer controls may turn this check green", and that holds only while `remote.origin.url` is `lifinance/contracts`. A fork-based checkout, or a `git remote set-url origin <fork>`, silently re-points the anchor at a branch the proposer owns — the same defect class this PR exists to remove, moved one level out.

Separately, `git fetch origin main` relies on git's *opportunistic* update of `refs/remotes/origin/main`, which is skipped when the clone's fetch refspec doesn't cover it (`--single-branch` clones). If `origin/main` already exists from an earlier clone you then read a **stale** anchor with no error, which contradicts the docblock's "a stale ref is not an anchor".

Both close cheaply:

```ts
// assert the remote identity before trusting the ref
const url = execFileSync('git', ['remote', 'get-url', 'origin'], { cwd: repoRoot, encoding: 'utf8' })
// ... check it resolves to lifinance/contracts, else a named refusing status

// and guarantee the ref is written rather than relying on opportunistic update
execFileSync('git', ['fetch', '--quiet', 'origin', '+refs/heads/main:refs/remotes/origin/main'], { ... })
```

A wrong-remote refusal wants its own status admitted deliberately to the may-proceed set's complement, in the spirit of the allowlist design.

script/deploy/safe/pinned-target-state.ts
0xDEnYO: Fixed in 4b2ca0fbd: a failed `fetch` is no longer written into `memo`. Next call retries. Pinned by `retries a failed fetch instead of pinning the refusal for the process`.

script/deploy/safe/pinned-target-state.ts
0xDEnYO: Seen — 21 production cache rows for Cut/Loupe with `version: \"\"` while `origin/main` declares `1.0.0`, so a re-register of those CREATE2 addresses hard-refuses today. That is a policy call, not a one-liner: treat blank-as-unversioned-cleared vs keep refuse until the cache is backfilled. Holding draft until that is named. Not going to fail this open from the chair.

script/deploy/safe/pinned-target-state.ts
0xDEnYO: Agreed this is a record gap (avalanche Cut, robinhood's 12 core facets, tron EmergencyPause at a different address), not something the resolver should invent. Holding with the high: either backfill the cache or we need a named policy for `contract-unidentified` on known CREATE2 cores. Staying draft.

script/deploy/safe/facet-version-utils.ts
0xDEnYO: Agreed — `contractNames` should be taken from `matches`, not from the versioned subset, so a blank-version sibling cannot hide a name contradiction. Will do that in the next pass on this PR; staying draft until the high is decided.

script/deploy/safe/pinned-target-state.ts
0xDEnYO: Agreed — `origin` must be `lifinance/contracts` (or we fetch a hardcoded URL) and the refspec has to actually update `refs/remotes/origin/main`. Will pin that in the next pass; staying draft.

script/deploy/safe/pinned-target-state.ts
0xDEnYO: Kept the refusal, took both conditions you named.

Blast radius is in the PR description now, measured rather than asserted: `DiamondCutFacet` `0xf7993A8d…` on 11 networks and `DiamondLoupeFacet` `0xF5ba8Db6…` on 10, blank in **every** row they have anywhere in the cache — so the cross-network idea from your `contract-unidentified` comment does not rescue them either, there is no versioned row to fall back to.

The `detail` now names the remedy: the blank is in the MongoDB deployment record, not in the cut, so the fix is to backfill that record. Admitting the status instead would reopen the downgrade hole on exactly the facets that control the diamond's upgrade surface, which is the one place I am least willing to trade it.

In 411b765eb.

script/deploy/safe/pinned-target-state.ts
0xDEnYO: Agreed it is a record gap — 13 addresses, and I built the widening you suggested before measuring what it does. Reverted it in 411b765eb; three reasons, in order of weight:

**The premise does not hold on this repo's own data.** `0xae77c9ad4af61faec96f04bd6723f6f6a804a567` is `CalldataVerificationFacet` v1.1.0 on 17 networks and `LiFuelFeeCollector` v1.1.0 on `opbnb`. Same address, same version, different contract. Deployment is CREATE3 with salt `keccak(DEPLOYSALT ‖ contractName)`, which binds the *name*, and binds the bytecode only for deployments that went through the script — so cross-chain address identity is a convention here, not a property.

**It converted refusals into clearances on thinner evidence than the refusal.** 12 of the 13 flipped `contract-unidentified` → `matches-main` (cleared), and every one of those `robinhood` clearances rested on **exactly one** foreign row, from `outlaw`. The "name must agree" safeguard cannot fire on a single-row shape, so the fail-closed property you were preserving is only preserved where a contradicting row happens to exist — 6 of 550 multi-network addresses.

**It bought no signability anyway.** `rebuild-attestations.ts` reads the record strictly per `(address, network)`; those same 13 addresses come back `no-record` → UNVERIFIABLE → block at the codehash gate regardless. So the relaxation costs this gate its independence and moves nothing.

So the refusal stands, and — same as the high — it now names its remedy (write the missing record) and says explicitly that the same address elsewhere is *not* consulted, so the idea does not get re-derived from the message.

script/deploy/safe/facet-version-utils.ts
0xDEnYO: Done in f52525e78 — names from `matches`, versions still from the versioned subset.

Confirmed latent, not live, the way you had it: old-vs-new resolution over all 4,438 on-network `(network, address)` pairs differs in **0** cases, and the three blank-version pairs all agree on name. New test pins the masked shape directly.

script/deploy/safe/pinned-target-state.ts
0xDEnYO: Both in ec2489f9a / 411b765eb, plus a third hole in the same family that fell out of chasing these two.

**Remote identity.** Refuses unless `origin` is `github.com/lifinance/contracts`, as its own reason outside the may-proceed set. `git remote get-url` rather than `git config remote.origin.url` deliberately: `get-url` applies `insteadOf`, so it reports where a fetch would really go. Verified against a repo whose configured URL is canonical and whose `insteadOf` points elsewhere — `get-url` prints the rewritten target and the check refuses. Reading the raw config would have passed it.

**Refspec.** Now `+refs/heads/main:refs/remotes/origin/main`. The test falsifies rather than asserting the argv: in a clone with its fetch refspec unset it proves `git fetch origin main` leaves `origin/main` at its old sha (exit 0, no error), then reads the moved anchor through the explicit refspec. Worth noting `actions/checkout` sets a single-branch refspec, so this was live for any CI-shaped invocation.

**Third one:** the blob was read as `git show origin/main:…`, a short name git resolves through `refs/`, tags and heads *before* remote-tracking refs — so a tag named `origin/main` in a clone the proposer supplied shadows the ref the fetch just wrote, which is the same authorship the remote check exists to deny. Read is now `refs/remotes/origin/main` in full; restoring the short name fails exactly one test.

Also split `remote-unreadable` off from `remote-unexpected` and stopped memoizing it — a `git` exec that could not run was being reported to the signer as "your origin is not lifinance/contracts", a diagnosis pointing at a problem that does not exist.

0xDEnYO:
@coderabbitai review

0xDEnYO:
@coderabbitai review

gvladika (approved):

PR #2349, opened 2026-09-09
feat(safe): close the remaining proposer-controlled routes into the signer's terminal (EXSC-951, EXSC-949)

0xDEnYO:

- [EXSC-951](https://linear.app/lifi-linear/issue/EXSC-951/four-routes-into-the-signers-terminal-survived-exsc-948-the-ledger)
- [EXSC-949](https://linear.app/lifi-linear/issue/EXSC-949/the-decoded-calldata-display-renders-proposer-controlled-text-one)

Stacked on #2334 (`feat/exsc-937-refuse-delegatecall-proposals`), where the EXSC-948 code this
extends lives. Not on `main` yet.

> **⚠ Not ready to review as evidence. Three claims in the body below were false and are now marked
> inline with corrections.** An independent falsification pass found: a **regression this PR
> introduced** — the periphery deployments check keyed on the sanitised name, so it printed a green
> "matches deployments" for a name the calldata does not contain (fixed in `0986ba9d5`, reproduced by
> reverting the fix); **eight surviving mutations**, five on guards added here; and **two
> pre-existing raw routes this PR claims to have closed** — `safe-utils.ts:3182` prints a
> 4byte-supplied name raw on the commonest proposal shape (`decodeDiamondCut`, which has no test case
> here at all), and `safe-decode-utils.ts:606`/`:609` print `target=` and `selector=` raw inside the
> very function the body says is covered. Those two are open holes on the base, not regressions —
> but the claim was the defect. Everything is itemised on
> [EXSC-951](https://linear.app/lifi-linear/issue/EXSC-951).

**Correcting the record:** EXSC-948's merged commit subject reads "sanitise every stored field in
the signer's transaction detail block". That was not true of the merged code. The Ledger filmstrip
and the decoded-calldata display both printed stored row text, one screen either side of the block
that was hardened. The subject is in history and cannot be edited, so the correction is here.

`confirm-safe-tx.ts` prints a proposal to a human immediately before they sign it. `tx.safeTx` is a
MongoDB row written by whoever proposed the transaction, so every field in it is attacker-controlled
text. Two threats, and the second needs no escape sequence at all: terminal injection, and a value
that renders identically to a different value.

## What was closed

**1. The Ledger Flex filmstrip** (`ledger-flex-preview.ts`). `dataRows` uppercased and wrapped the
stored calldata; `frameLine` clipped each row at 20 characters while stripping nothing, and `ESC[2J`
is four. This is the artefact the signer is instructed to compare against their physical device, so
text injected here attacks the verification step itself. `data` is now typed `unknown` — it reached
the module through a cast, so it holds whatever the row holds — and is rendered through the shared
primitive. Two disclosures print *below* the panels, never inside one, because a notice threaded
through a box would break the geometry the character-by-character comparison depends on: what had to
be repaired, and separately that the value is not `0x`-prefixed hex, since which bytes the row holds
is the whole question and nothing here may quietly change them. `to` and `value` were left alone —
`getAddress` and `bigint` already settle them.

**2. The decoded-calldata display, EXSC-949** (`safe-decode-utils.ts`). Six sites, all now through
the primitive:

- `Periphery Name:` printed `String(decoded.args[0])`, so a **cleanly decoding, legitimately
  encoded** `registerPeripheryContract(string,address)` call injected a screen clear and a
  fabricated `To:` line. Reproduced against the real formatter before the fix; the payload and the
  before/after are below.
- `Data (raw):` on both the no-decode arm and the catch arm — a raw slice of the stored calldata.
- `Failed to decode data:` echoed viem's message, which quotes its input back.
- `formatDecodedArg`'s bare `return s`, the route every generic decoded argument takes. The notice
  is appended at the top level only: inside the JSON branch it would land in a JSON string, where
  `JSON.stringify` escapes its own colour codes into visible text and the disclosure reads as part
  of the value.
- The signature `resolveSelectorsViaFourByte` returns. Remote text, and a proposer picks which of it
  is fetched by picking the selector.
- `formatTimelockScheduleBatch`'s scalars and payload preview, and `formatRoleChange`'s role.

**Correction — not every caller is covered by execution.** A later falsification pass reverted each
new guard in turn and **eight mutations survived**, five of them on guards added here: the 4byte
signature at `:413`, `Predecessor:` at `:581`, the `payload=` preview at `:622`, `formatRoleChange`
at `:779`/`:782`, the generic `Function:` at `:911`, and the catch arm at `:738`. `:581` and `:622`
are reachable with hostile text through an unvalidated timelock row, so those are real routes rather
than defence in depth. The nested recursion at `:593`/`:833` **is** covered — the scheduleBatch test
asserts the nested frame was entered. Full list on
[EXSC-951](https://linear.app/lifi-linear/issue/EXSC-951).

**3. Length.** `safe-tx-detail-display.ts` had no bound, where both sibling modules do and say why
(`delegatecall-gate.ts`'s `MAX_RENDERED = 80`, `provenance-display.ts` clipping the reason "so a
hand-edited Mongo row cannot scroll the prompt away"). Every stored field **in that module** is now
bounded at `MAX_FIELD_CHARS = 120` — the longest legitimate value in the prompt is a 66-character hash or a PR
URL — and `parkedTaskRefs` is bounded at 20 elements with a counted-overflow line, because the array
comes off the row and can scroll the block away without any single field being long. The measured
line for a 500,000-character `safeTxHash` went from 500,032 characters to 208.

**One deliberate exception: `data` stays unbounded.** It is the payload the signature covers and the
only place a signer can read it in full — the decoded display above it prints a 66-character preview
when nothing decodes, and clipping the detail line too would leave no full rendering anywhere. Its
length is also its own disclosure: a wall of hex reads as one, where a 66-character field silently
grown to 500,000 does not. Named as a residual rather than closed.

**Correction — `data` is not the only unbounded field.** The decoded display in
`safe-decode-utils.ts` bounds nothing: a 500,000-character periphery name renders as two
half-million-character lines with **no notice at all**, since nothing was sanitised and nothing
clipped, and the flood lands below the name it belongs to and scrolls the target address and the
✅/❌ verdict away. That is the failure this bound exists to prevent, and it is still open —
[EXSC-951](https://linear.app/lifi-linear/issue/EXSC-951).

**4. Confusables — disclosed, not disclaimed.** `parkedTaskRefs[].facet` of `AcrоssFacetV3` with
Cyrillic U+043E was byte-different, glyph-identical and drew no warning, in the block whose only job
is naming the facet being removed. I chose disclosure over a docblock caveat because the check is
cheap and exact here: every field this primitive renders is ASCII by construction — an address, a
hash, a decimal, a facet name, a URL — so a bare non-ASCII count has no false positives on real
data, needs no script-mixing table, and a benign row therefore still renders byte-identically. A
homoglyph also clears `identityPreserved`, so the target name and explorer link are withheld rather
than vouching for an address that is not the one stored.

**5. `confirm-safe-tx.ts:194`.** The throw echoed the stored `txDoc.safeTxHash`. Lowest severity —
it fires only after signing, so it cannot corrupt the decision, but it can repaint a success line
over a real failure to persist and hide that the signature was never stored. `:223` and `:298` are
untouched: they read a local computed by `safeClient.getTransactionHash`.

## The structural question: yes, closed by construction

`asPrintable` returns a branded `Printable`, and `color` in the display module accepts nothing else.
A field added later as a plain `string` — the natural default, and what three existing fields used —
is now a compile error rather than a review question. There is exactly one hatch, `trustedMarkup()`,
and its call sites are greppable; the intentional exceptions are `nonceWarning` (chain-read
`bigint`), `operationLabel` (already sanitised by `describeOperationValue`) and the
`signatureCount`/`threshold` pair, all of which carry colour codes of their own that sanitising
would strip. `nonceColor` stays a closed literal set, being the only field interpolated *inside* an
escape sequence rather than beside one.

The brand is not decorative: adding it turned three call sites in `safe-tx-detail-display.test.ts`
and two in `confirm-safe-tx.ts` into type errors that had to be answered one way or the other. That
is the mechanism the ticket asked for.

## The rule-201 question, answered

`.agents/rules/201-safe-decode-scripts.md` says not to log raw `Data:` hex when a decoded display is
shown, and prescribes it for `execute-pending-timelock-tx.ts`. **It should not be extended to
`confirm-safe-tx.ts`,** and the rule now says so with the reason: the two scripts are deciding
different things. `execute-pending-timelock-tx` runs an operation whose calldata was reviewed at
schedule time; `confirm-safe-tx` is where a human commits a signature to exact bytes, and the hex is
the artefact they compare against the Ledger screen. The rule gained a second line requiring every
stored or decoded value to go through `printable-field` before it is interpolated into a log line,
with the notice outside the colour code.

## Evidence

**The payload that drives the terminal today.** `registerPeripheryContract(string,address)` encoded
with `args: ['GasZipPeriphery' + ESC + '[2J' + ESC + '[H  To:  0x00…01', '0x11…11']` — a call this
repository encodes the same way, decoding cleanly, whose entire attack surface is the `string`
argument. Before, run against the real formatter:

```text
Periphery Name: ESC[33mGasZipPeripheryESC[2JESC[H  To:  0x0000000000000000000000000000000000000001ESC[0m
```

The two live sequences clear the screen and home the cursor, and the fabricated `To:` line then
stands alone above the sign prompt. After:

```text
Periphery Name: ESC[33mGasZipPeriphery[2J[H To: 0x0000000000000000000000000000000000000001ESC[0m
                ESC[33m ⚠ sanitised for display — stored 71, printable 67ESC[0m
```

The same payload in the filmstrip's `data` field rendered `0xESC[2JESC[HDEADBEEF` inside a panel that
still measured the right width; it now renders `0x[2J[HDEADBEEF` with both notices below the panels.

**Benign output is byte-identical.** A benign run of all three surfaces — four decoded calls
(`registerPeripheryContract`, `grantRole`, a nested `scheduleBatch`, `transfer`), the full detail
block with provenance and a parked ref, and the filmstrip — was captured before and after the change
and diffed: 70 lines, zero difference. Re-checked after the falsification pass restored every
mutation.

**Falsification.** Fourteen mutations, each reverting one production change, each confirmed to turn
the specific test red and then restored from git (not from an editor buffer):

| Mutation | Test that went red |
|---|---|
| clip disabled in `asPrintable` | clips a field that needs no escape at all |
| non-ASCII remark removed | discloses a homoglyph that renders identically |
| `confusable === 0` dropped from `identityPreserved` | withholds the target name for a homoglyph address |
| `parkedTaskRefs` slice removed | bounds the number of parked refs |
| filmstrip renders `String(data)` | leaves no escape from the row anywhere in the output |
| filmstrip notice dropped | discloses below the panels what it had to repair |
| hex-shape check disabled | flags calldata that is not hex |
| `renderScalarArg` returns raw | renders a hostile decoded string inert |
| `Periphery Name` back to `String(args[0])` | hostile `registerPeripheryContract` name (and the nested recursion case) |
| `rawDataPreview` back to a raw slice | raw preview when nothing decodes, and via the catch arm |
| catch-arm message un-sanitised | sanitises the error message on the catch arm |
| ~~4byte signature un-sanitised~~ | **this row is false — see the correction below** |

The 4byte case was **green on its first falsification run** — the stub returned a name that does not
hash back to its selector, which `resolveSelectorsViaFourByte` correctly drops, so the test had been
observing the raw-preview arm instead. The selector is now derived from the hostile signature and
the test asserts the name actually reached a line.

**Correction — that was not the only vacuous assertion, and the 4byte row above is still false.**
A later pass reverted the guard at `:413` and the suite stayed green: the test named beside it
actually covers `:989`. The guard itself works — it was observed firing — but nothing asserts it.
Same defect class, same feature, same mechanism as the one this section congratulates itself on
catching, which is the more useful finding than either instance.

**Not proven by execution: one site.** `confirm-safe-tx.ts:194`. `confirm-safe-tx.ts` has `runMain`
at module scope and cannot be imported, and spawning the CLI is prohibited for this script in any
mode. The site is a one-expression change reviewable by reading, and the primitive behind it is
tested directly, but the placement itself carries no executed assertion. Stated rather than papered
over.

**Assertions are executed, never scanned.** No new test reads source text. Three source-scanning
guards on EXSC-948 were each defeated by behaviour-preserving rewrites; the existing
`safe-tx-detail-display-placement.test.ts` says so in its own header, and finding 1 of EXSC-951 sat
fourteen lines below its assertion.

`bunx tsc-files --noEmit` clean on all ten touched files (`bun test` does not typecheck), `eslint`
clean, `bun test script/deploy/safe/` → 1310 pass / 0 fail, 42 of those tests new.


0xDEnYO:
@coderabbitai review

0xDEnYO:
@coderabbitai review

0xDEnYO:
@coderabbitai review

0xDEnYO:
Collision to resolve before this lands: **#2350** (`feat/exsc-hash-compare-flex-and-tron-rpc`, EXSC-963, based on `main`) modifies `ledger-flex-preview.ts` and `confirm-safe-tx.ts` too, and changes `frameLine`'s signature to take a `ranges` parameter.

Two reasons that is coupled rather than merely adjacent:

1. This PR routes the filmstrip's stored calldata through `asPrintable` inside `dataRows`. A conflict resolved in favour of #2350's version of that file would restore an unsanitised path to the signer's terminal — on the artefact they are told to compare against the physical device.
2. `ranges` looks like the feature that starts styling dynamic text. The guard comment this PR adds at the `String.replace` in `frameLine` records that the substitution hazard there — `$&` and friends in the replacement operand — is unreachable **only** because every styled row is a hardcoded label. If a range can style row-derived text, that stops being true.

Flagged on #2350 as well, so neither side resolves it blind.

0xDEnYO:
@coderabbitai review

0xDEnYO:
@coderabbitai review

PR #2350, opened 2026-09-09
fix(safe): preview the hash mode Ledger screens, and stop routing every Tron RPC to /jsonrpc (EXSC-963)

0xDEnYO:

Fixes EXSC-963 — https://linear.app/lifi-linear/issue/EXSC-963/hash-mode-signer-display-ledger-flex-preview-with-the-16-compared

Three findings from walking `bun confirm-safe-tx` end to end.

## 1. Every non-TronGrid Tron RPC was being routed to a 404

`normalizeRpcUrlForNetwork` appended `/jsonrpc` to every Tron RPC URL. That path is
TronGrid-specific — its stored root serves Tron's native HTTP API — but a provider that
scopes the chain onto the path already speaks JSON-RPC there. Measured against the
configured dRPC endpoint:

```text
base            -> 200 {"jsonrpc":"2.0","result":"0x5219afe"}
base + /jsonrpc -> 404 Not found
```

In a real run that surfaced as `Failed to read Safe state on tron: HTTP request failed.
Status: 404`, dropping Tron out of the actionable set entirely. The same function feeds
the RPC reachability probe, so Tron was misreported as unreachable there too.

The rewrite is gated on the TronGrid host family alone, so it is strictly narrower than
`main`: every non-TronGrid endpoint is left exactly as stored, whether it carries a path
or is a bare host. A TronGrid URL with a native-API path (`/wallet`) still gets the route
appended, as before. Gating on "has a path" alone was the first attempt and was wrong in
the other direction — it would have silently stripped the route from a keyed per-project
endpoint whose JSON-RPC sits under its token path, a shape the URL cannot be told apart
from dRPC's. Such an endpoint must be stored as its full JSON-RPC URL, which the docstring
now says.

Seven tests pin the rule, each failing against a different wrong one: three path-scoped
shapes, a bare non-TronGrid host and a `trongrid.io.example.invalid` lookalike all fail
against `main`; the lookalike also fails against a substring host match; and a
TronGrid-with-path shape fails against a rule that keys on the path instead of the host.

## 2. The signing mode we actually use had no preview of the device

`resolveSignerVerificationDisplay` returns `filmstrip` only under
`ENABLE_SAFE_EIP712_SIGNING=true`. Hash mode has been the default since EXSC-871, and it
got a paragraph of prose — so the one mode with a picture of the Ledger was the one nobody
signs in.

`renderLedgerFlexHashFlow` renders the three message screens the device shows
(Review message → Message → Sign message / hold to sign), with the first and last eight
hex characters highlighted on the screen itself and spelled out in a column beside it.
That makes the locked 16-character comparison the action the layout asks for rather than a
sentence to read; the rationale (four-and-four is grindable) stays in the prose below.

**The previewed hash is read from the Safe's own `getTransactionHash`, never the stored
`safeTxHash`.** The proposer writes that field, so previewing it would show the operator a
picture of what the proposer *claims* the device will display — the exact substitution the
anchor rule exists to prevent, and one no unit test on the renderer can see, since both
values are well-formed hashes. There is a source-shape guard for it in
`confirm-safe-tx-ack.test.ts` that says what it does and does not catch. A stored hash
that disagrees with the computed one is reported, report-only, naming only the computed
value.

The renderer rejects anything that is not `0x` + 64 hex rather than framing it: the value
lands in the operator's terminal, and a string carrying its own escape codes or box
characters could redraw the screens it is printed inside.

## 3. The second review prompt is gone

`Confirm you reviewed this change` fired *after* the operator selected `Sign`, with the
payload, the provenance and the device screens already on screen. Selecting an action is
the acknowledgement. The ledger still records and rolls up per target + value + operation
+ payload, so the run summary keeps its N/N — relabelled `acted on` from `reviewed`,
because nothing observes a review and the old word claimed something the count could not
back. The failing-nonce verdict is still warned about at the same point; execute-shaped
actions on a stale nonce were already refused earlier and still are.

`shouldPromptForAcknowledgement` and `isChangeAcknowledged` went with the prompt rather
than staying as dead exports; the ledger tests now use a test-local read helper.

## Where to look hardest

**The Tron rule cannot be derived from the URL.** dRPC serves Tron JSON-RPC at
`/tron/<key>`; a keyed per-project endpoint may serve it at `<token>/jsonrpc`. The two are
indistinguishable as strings, so the rule rewrites only TronGrid hosts and leaves every
other Tron endpoint exactly as stored.

The stored data is now enumerated (`bun audit-rpc-priorities`, production `RpcEndpoints`):
three endpoints across both Tron networks — a keyed path-scoped `lb.drpc.live` primary on
`tron`, plus the two bare TronGrid roots. Each one through the real function:

```text
tron primary   (keyed, path-scoped)   main -> https://lb.drpc.live/tron/<KEY>/jsonrpc
                                      PR   -> https://lb.drpc.live/tron/<KEY>          CHANGED
tron fallback1 (bare TronGrid)        main -> https://api.trongrid.io/jsonrpc          unchanged
tronshasta primary (bare TronGrid)    main -> https://api.shasta.trongrid.io/jsonrpc   unchanged
```

Exactly one resolved URL changes, and it is the one that was 404ing. No stored Tron
endpoint is a bare non-TronGrid host.

`deviceHash` adds one `getTransactionHash` read per proposal on the display path. It sits
before the action prompt, so it does not delay the Ledger, but it is one more RPC call per
network on a fleet run.

**The preview shows only what has been observed on a Flex** — the three screen titles, the
prompt on each, and the hash. Page counters and a `Skip` header were in an earlier revision,
carried over from the typed-data filmstrip rather than read off a device, and are gone: a
detail the operator is invited to check against a screen that may not show it teaches them
to shrug at the mismatch the preview exists to raise. A test fails if either comes back.

**Collision with #2349**, which rewrites `ledger-flex-preview.ts` and `confirm-safe-tx.ts`
to route proposer-controlled row text through `asPrintable`. This PR replaces `frameLine`'s
`String.replace`-based styling with slice-based `applyStyleRanges`, which removes the
`$&` / `` $` `` / `$'` substitution hazard #2349's comment describes — measured on the old
construction, a 20-character row came out 32 and 33 characters wide, breaking the frame.
There is a regression test pinning it. Whoever resolves the conflict must keep both sides:
#2349's `asPrintable` on the `data` rows, and slice-based styling here.


0xDEnYO:
Heads-up on a collision with **#2349**, which is stacked on #2334 and rewrites the same two files — worth knowing before either conflict is resolved.

#2349 sanitises the Ledger filmstrip: `confirm-safe-tx.ts` passes `tx.safeTx.data.data` (the stored MongoDB row, so proposer-controlled) into `renderLedgerFlexFlow`, and `dataRows` now routes it through `asPrintable` from the new `printable-field.ts`. Before that, `frameLine` clipped each row at 20 characters and stripped nothing, so a four-character `ESC[2J` passed through inside a row that still looked the right width. That matters here specifically because the filmstrip is the artefact the signer is told to compare against their physical device.

**The part that bears on your change:** you add a `ranges` parameter to `frameLine`. The styling there is applied as a `String.replace` whose replacement operand is the row text itself — so `$&`, a backtick-dollar, and `$'` inside that text are substitutions rather than literals. That is unreachable today only because every styled row is a hardcoded label and all row-derived content is pushed with no `style`. #2349 adds a comment at that line saying exactly this. If `ranges` lets a screen style text that came off the proposal, the invariant is gone and the frame can be rewritten by the row.

Not asking for a change — you may already handle it. But whoever resolves the conflict between these two should know that dropping #2349's side would restore an unsanitised path to the signer's terminal, and that the two changes are coupled through this one function rather than merely adjacent. Detail in [EXSC-951](https://linear.app/lifi-linear/issue/EXSC-951).

script/mongoDb/rpcEndpoints.ts
```diff
@@ -110,10 +120,12 @@ export function normalizeRpcUrlForNetwork(
     return rpcUrl
   }
 
+  const path = parsed.pathname.replace(/\/+$/, '')
+  if (path !== '' && !isTronGridHost(parsed.hostname)) return rpcUrl
```
gvladika: **Blocker** — not this line's correctness, but what it is gated on.

Measured on this branch:

```
https://tron.drpc.org            -> https://tron.drpc.org/jsonrpc
https://tron-rpc.publicnode.com  -> https://tron-rpc.publicnode.com/jsonrpc
https://lb.drpc.org/tron/KEY     -> https://lb.drpc.org/tron/KEY          ✓ fixed
```

The bare-host branch is still Tron-wide, so a non-TronGrid endpoint stored as a bare host still gets TronGrid's route appended — the same 404 this PR fixes, on the same provider in its keyless form. The docstring's own framing ("a fact about TronGrid, not about Tron") argues against it.

It is not a regression from `main`, and `config/networks.json` stores only TronGrid roots, so the fallback path is unaffected either way. Which means the decision rides entirely on what is actually in the MongoDB `RpcEndpoints` collection — the enumeration the description flags as unfinished.

**That enumeration is the blocker, not this line:** without it, the PR's stated result (Tron stops 404ing) is unverified. If it turns out any stored Tron endpoint is a bare non-TronGrid host, this becomes `if (!isTronGridHost(parsed.hostname)) return rpcUrl` plus a test.

script/deploy/safe/ledger-flex-preview.ts
```diff
@@ -362,3 +428,164 @@ export const renderLedgerFlexFlow = (
...
+  {
+    header: '',
+    content: [
+      { text: '[=]', align: 'center' },
+      { text: '', align: 'center' },
+      { text: 'Review message', align: 'center', style: BOLD },
+      { text: '', align: 'center' },
+      { text: 'Swipe to review', align: 'center' },
+    ],
+    footer: navFooter(1, 3),
```
gvladika: **Blocker** — the three-screen shape and these page counters cannot be verified in software.

`LEDGER_FLEX_HASH_NOTE` covers case and line breaks, but not the page count or screen 2's `Skip` header. The 4–5 wrapped rows do fit the 6-row page budget the `data` model uses, so one content page is internally consistent — but internal consistency is not the bar for an artefact whose whole function is that an operator treats a mismatch against the physical device as an alarm. If the counters are visibly wrong, it trains people to shrug at mismatches, which is worse than no preview at all.

You have the device and already walked the flow end to end: a photo of the Flex beside the terminal, in the description, closes this. Otherwise I would drop the counters and the `Skip` header and keep only what is certain.

gvladika (commented):
Reviewed the whole diff. Merged `origin/main` in locally: clean, no conflict with #2349's `asPrintable` — `buildSafeTxDetailLines` still runs immediately above the new filmstrip. All 1493 tests across `script/deploy/safe/` + `script/mongoDb/` pass post-merge, and `tsc-files --noEmit` is clean on all seven files.

The computed-vs-stored hash discipline checks out: `signTransactionWithHash` signs the output of the same `getTransactionHash` call on the same struct, so the preview really is the signed value. And the slice-based `applyStyleRanges` does remove the `$&` hazard #2349 described, with a test that pins it.

Three blockers. Two are inline; the third has no line to hang on:

**Removing the acknowledgement prompt needs a named yes from whoever owns the signing flow.** I think the reasoning holds — the action select is a deliberate choice made with the payload, the provenance and the device screens all on screen, and the one thing the prompt actually gated (a forced re-prompt on a failing nonce verdict) never blocked anything, since execute-shaped actions on a stale nonce are refused separately and a bare `Sign` on a stale proposal cannot execute anyway. `acted on` is the honest label for what the counter measures. But it is a deliberate removal of a speed bump from the multisig signing flow, which under `[CONV:ARCH-GOVERNANCE]` wants an explicit sign-off recorded on the PR rather than a rationale in the description.

Not a blocker, but worth doing here because it works against what the PR is for: `SafeClient.getTransactionHash` carries a bare `console.log('Generated transaction hash:', …)` (`safe-utils.ts:770`), so hash mode now prints an unlabelled hash immediately above the filmstrip — between the stored `Safe Tx Hash` row and the artifact the operator is told to read. Deleting that line is a one-liner in a `consola` codebase.

The rest of my notes (the second `getTransactionHash` read per proposal, `style` silently swallowing `ranges` in `frameLine`, three nits, and a suggestion to split the Tron fix into its own PR so it can ship against the live outage) are follow-ups — happy to file them separately.

script/mongoDb/rpcEndpoints.ts
0xDEnYO: Fixed in d6a7ac40b: `/jsonrpc` is appended only on TronGrid hosts. A bare non-TronGrid host (`tron.drpc.org`, PublicNode, …) is left as stored — same as a path-scoped dRPC URL. Pinned by `leaves a non-TronGrid bare host untouched`.

script/deploy/safe/ledger-flex-preview.ts
0xDEnYO: Agreed — software cannot certify the Flex page count or the Skip header. Holding this until we have a photo of the device beside the terminal for the hash-mode three-screen walk. Not going to mark this resolved from the chair.

0xDEnYO:
@gvladika (1) TronGrid-only `/jsonrpc` is in d6a7ac40b; also dropped the unlabelled `getTransactionHash` `console.log`.

(2) Flex photo — needs the device, holding.

(3) Dropping the acknowledgement prompt: will not treat the rationale as a yes. Holding for a named sign-off on the signing-flow speed-bump removal.

PR stays draft until (2)+(3) and Linear `Fixes`.

script/mongoDb/rpcEndpoints.ts
0xDEnYO: Enumeration done — `bun audit-rpc-priorities` against the production `RpcEndpoints` collection. Both Tron networks, every stored endpoint:

```text
tron        primary   p=2  keyed   lb.drpc.live
            fallback1 p=1  NO KEY  api.trongrid.io
tronshasta  primary   p=1  NO KEY  api.shasta.trongrid.io
```

Three endpoints, two hosts. No stored Tron endpoint is a bare non-TronGrid host, so the case you measured does not occur in the stored data — and under the current rule it resolves correctly anyway.

Each stored endpoint through the real function, `main` vs this branch:

```text
tron primary   (keyed, path-scoped)   main -> https://lb.drpc.live/tron/<KEY>/jsonrpc
                                      PR   -> https://lb.drpc.live/tron/<KEY>          CHANGED
tron fallback1 (bare TronGrid)        main -> https://api.trongrid.io/jsonrpc
                                      PR   -> https://api.trongrid.io/jsonrpc          unchanged
tronshasta primary (bare TronGrid)    main -> https://api.shasta.trongrid.io/jsonrpc
                                      PR   -> https://api.shasta.trongrid.io/jsonrpc   unchanged
```

Exactly one resolved URL changes, and it is the one that was 404ing (base returns 200, base+`/jsonrpc` returns 404 — measured on the live endpoint). The two TronGrid endpoints are byte-identical to `main`. So the PR's stated result now has evidence behind it rather than an assumption.

script/deploy/safe/ledger-flex-preview.ts
0xDEnYO: You are right, and I have taken the second option rather than the photo.

The page counters and screen 2's `Skip` header were reproduced from the typed-data filmstrip's conventions, not read off a Flex. Only the three titles and their prompts were actually observed. Both are gone in 9a3ade474 — the preview now shows the three screen titles, the prompt on each, and the hash, and nothing else. A test pins it: re-adding either a counter or the `Skip` header fails `shows nothing that has not been read off the device` (verified both ways).

That leaves the preview making only claims someone has seen, so a difference from the device stays worth treating as an alarm. If the counters do turn out to match the device, they can come back later with the photo as evidence — but they should not ship ahead of it.

gvladika (approved):

PR #2351, opened 2026-09-09
feat(codehash): the attestation key and the audit bridge (EXSC-952)

0xDEnYO:

- Fixes [EXSC-952](https://linear.app/lifi-linear/issue/EXSC-952/wp-52-ci-attestation-minting) · WP-5.2 · R2.1

## Why this is the schema half, on its own

WP-5.2's spec carries an ordering constraint: **the `sourceClosureHash` field must land before the first mint**, or early manifests cannot carry the audit bridge (E1). So this PR is the key and the bridge as pure modules with golden tests — no CI workflow, no minting, and no overlap with the open codehash PRs (#2301 merged, #2329 awaiting approval), which is what EXSC-906's own sequencing note recommends for this surface.

It also does not need anything merged. The minter writes attestations and the gate consumes them; they are independent surfaces. Until attestations exist the shipped gate grades everything grey, which is why minting is what turns it into a working control rather than an inert one.

## What the key has to identify

Every input that can change the bytecode. Anything the key leaves out lets two different builds share one entry — and the set-membership compare in `attested-set.ts` is then answering about a build the deployed code is not.

- **The source closure is in the key.** A contract's own file and its declared version do not move when something it inherits does. Commit `4cfabbf22` bumped `SwapperV2` from 1.0.2 to 1.2.0 (alongside `WhitelistManagerFacet` and `IWhitelistManagerFacet`, neither of which inherits it, and deleting `DexManagerFacet`). **33 facets inherited `SwapperV2` at that commit**, 31 do today. None of them was touched by it and none bumped its own version — `AcrossFacet` stayed at 2.0.0 — yet recompiling `AcrossFacet`'s 14-file closure at both commits gives two different masked hashes under a byte-identical key. `closureHash` closes that; it was already computed and already on the attestation, one field short of the key.
- **The settings are hashed whole, never enumerated.** A field list has to be kept complete against every compiler release, and an incomplete one fails silently in the unsafe direction. `artifactSettingsHash` hashes `metadata.settings` as the compiler emitted it, less two fields that name the source rather than the settings: `compilationTarget` (carried as `sourceId`) and `remappings` (only decides what an import resolves to, and the resolved sources are now named by `closureHash`). Keys are sorted at every depth and nulls dropped, so emitter key order and a reader that fills absent keys cannot split one build across two keys.
- **`solcVersion` and the zk toolchain are named separately**, because they are not in `metadata.settings`. The zk section carries the solc fork and LLVM version alongside zksolc, since those float independently of the zksolc release (F14, 10c §3.3).
- **A contract's name does not identify its source**, and neither does its closure — `LiFiDEXAggregator.sol` declares `IPool` beside the aggregator, and `ReceiverStargateV2.sol` declares a different `IPool` of its own — so `contractName` and `sourceId` are both in the key. (Both `IPool`s are interfaces with empty deployed bytecode; the collision matters for the concrete contracts that share a closure with them.) On the real tree, 190 artifacts span only 169 distinct closures.
- **An absent zk section is written as a marker, not skipped.** A key that shortens when a field is missing is a key two builds can share.
- **Where the settings came from is part of the key.** zksolc writes no `metadata` object (0 of 280 `zkout/` artifacts), so a zk build's settings come from repo config — a claim about what we asked for, not a report of what ran. `configuredSettingsHash` tags those so they can never answer a lookup for a self-reported build, and the tagging helper is unexported, so the tag always follows from which reader was used rather than from what a caller chose.

`profile` and `repo` are deliberately **not** in the key. `[profile.ci]` inherits `[profile.default]` and differs only in fuzz settings, so keying on the profile filed one build under two keys; two repos with genuinely different code already differ in `closureHash`. Both ride on the attestation as provenance alongside the hashed settings, so a refusal a signer has to act on still names the settings rather than only their digest.

## The audit bridge

`sourceClosureHash` is taken over the source closure only, never the settings, so the same sources compiled under different settings bridge to the same audit record. That is what makes it usable as an audit key: an audit is of source, not of a build (E1). It is now also a key field, which does not disturb that property — the bridge value is unchanged, the key merely names it.

It **refuses** a closure naming one path twice with different hashes rather than picking a winner — either choice hashes a closure that was never compiled, so the bridge would vouch for source the build did not use. An agreeing duplicate is accepted, so the refusal is about contradiction rather than repetition.

## Two design choices worth a reviewer's attention

**Fields are length-prefixed, not delimited.** A delimiter only works while no field can contain it, which is an assumption about every present and future input — and source paths are outside this module's control. Length-prefixing needs no such assumption: `['A', 'B C']` and `['A B', 'C']` cannot collide. I started with a delimiter and replaced it after the falsification pass showed the property rested on that assumption.

**Conflicts are returned, not resolved.** Two attestations under one key mean the key does not identify the build, so a lookup would return whichever was filed first and a gate would vouch for bytecode on the strength of an attestation of something else. The caller has to refuse. The source-closure arm of that check is gone: with `closureHash` in the key, two attestations under one key necessarily name the same closure, so the check keeps the masked-hash arm only.

## Falsification

Every check, one mutation at a time, each failing only its own test and each restored byte-identical. The harness asserts the mutation actually landed before trusting a green run — three times in this PR a mutation silently failed to apply, or broke parsing rather than mutating, and read as a passing test until that assert caught it.

| Mutation | Test that caught it |
| --- | --- |
| `closureHash` dropped from the key | separates a facet whose bytecode moved because a helper it inherits did |
| key hashes reach the serialiser unvalidated | refuses a key whose hashes are not keccak digests |
| settings hashed without their source tag | separates configured settings from self-reported ones |
| nested settings objects not recursed (order unnormalised) | does not separate on the order of a nested settings object |
| canonical JSON stops sorting top-level keys | does not separate on the order the compiler emitted the settings |
| canonical JSON stops dropping nulls | reads an absent setting and an explicitly null one as the same build |
| strip list emptied | ignores the fields that name the source rather than the settings |
| strip list also drops `optimizer` | covers a setting no field list names |
| strip list also drops `evmVersion` | separates on every input that can change the bytecode |
| empty-settings refusal removed | refuses settings that identify no build |
| compilation-target count check removed | refuses an artifact that does not name exactly one compilation target |
| `sourceId` dropped from the key | separates on every input that can change the bytecode |
| conflict payload shallow-spread instead of cloned | hands back a refusal a caller can annotate without touching the input |
| conflict keeps the last attestation's provenance | reports the provenance of the first attestation, not the last |
| conflict hashes sorted rather than first-seen | reports the hashes in the order found, not sorted |
| zk key names only zksolc | separates two zk builds that differ only below zksolc |
| absent-zk marker removed | gives an EVM build a key a zk build cannot reach |
| length prefix dropped | separates two closures a delimiter-joined hash would merge |
| digest length check removed | refuses a source hash that is not a keccak digest |
| empty-path refusal removed | refuses an entry that names no path |

A third survivor turned up in the last round and was a real gap, not a defensive one: flattening the zk triple into plain concatenation passed the entire suite, and zksolc 1.5.15 / fork 0.8.29 then collides with zksolc 1.5.1 / fork 50.8.29. The outer join protected the zk blob from its neighbours; nothing protected the three versions from each other. Now pinned.

Two survivors remain and are recorded rather than papered over: removing the `canonicalJson` array branch (arrays fall through to the object branch and still separate on index, so behaviour is preserved), and dropping the length prefix from the settings-hash join (`artifact` and `config` differ at the first character and the JSON always follows, so no collision is constructible for a two-value tag). Both are defensive rather than load-bearing.

Run against the real `out/` tree: **190 artifacts, 190 distinct keys, zero collisions.** `zkout/`: 280 artifacts, 0 with a `metadata` object, which is why the zk path takes its settings from config and tags them.

Three premises from earlier rounds turned out to be wrong and are worth recording, because all three are easy to re-derive:

- **solc does not write `viaIR: null`.** It emits `viaIR` only when the IR pipeline ran and omits the key otherwise — absent in all 190 artifacts. `jq '{viaIR}'` prints a missing key as `null`, which is where the reading came from.
- **solc drops `optimizer.enabled` when `optimizer.details` is supplied.** An earlier commit here read that absence as "the optimizer did not run", so a tuned optimized build keyed as `noopt` and collided with an unoptimized one. Hashing the settings whole removes the branch that could get this wrong.
- **A contract's declared version does not track its dependencies.** This is the `SwapperV2` case above, and it is why naming the closure is not optional.

## What is left of WP-5.2

Not in this PR, and none of it blocked by it:

- The CI workflow: one mint per key on merge (contract, source, settings hash), both repos, `--ast` builds, `attest-build-provenance`, compiler binaries pinned by hash, the Sigstore trust root pinned.
- **The zk producer, which needs two inputs that exist nowhere on disk.** zksolc writes no `metadata` object at all (0 of 280 `zkout/` artifacts), so a zk build has neither a settings source nor a `metadata.sources` closure. `configuredSettingsHash` covers the settings half by taking whatever the producer configures; the closure half has no reader here, and this PR does not add one. Until both are designed, the zk arm of the key rests on caller discipline — the `config` tag stops a configured claim answering an artifact lookup, but it does not stop two zk builds under different settings sharing a key.
- Fail on unexplained immutables at mint time.
- The live mint → `gh attestation verify` round trip, which has never been performed and cannot be from a local session.
- The gate consuming attestations as its source of truth with local rebuild as an explicit fallback *state*, visible in the report as `A-CI` vs `A-LOCAL`.
- The constitution's CI-gate demo: that the gate still fires **when its producing step did not run**, not only on bad input.

The sidecar's remaining fields — resolved names, types, offsets, expectation sources — attach to `IMintedAttestation` when the minter exists; the bridge field they had to precede is here.


script/deploy/codehash/attestation-key.ts
```diff
@@ -0,0 +1,211 @@
...
+import { keccak256, stringToHex } from 'viem'
+
+/** A toolchain and settings combination a build can be produced under. */
+export interface IAttestationKey {
+  /** Contract name as the artifact records it. */
+  contractName: string
+  /** Version the contract declares, e.g. `1.2.0`. */
+  version: string
+  /** Foundry profile the build ran under. */
+  profile: string
```
gvladika: **The key omits `evmVersion` and the optimizer settings.**

The module header states the key has to name every input that can change the bytecode. It names `solcVersion` and `viaIR`, but not `evm_version`, `optimizer.enabled` or `optimizer_runs`, all of which change the compiled output. `profile` is the stand-in for them.

`lineage-scope.ts:16-27` already rejects that proxy: a lineage there is (profile, solcVersion, **evmVersion**), parsed out of `foundry.toml` rather than hardcoded, because "a hardcoded pair would keep passing after someone retunes a profile, which is the same silent-drift failure the gate downstream exists to catch."

The same drift applies to the profile *name*. Change `[profile.default] evm_version` from `cancun` to `prague`, or change `optimizer_runs`, and every serialised key stays byte-identical while every masked hash moves. Two different builds then share one key, which is the collision `findAttestationConflicts` exists to detect and cannot see, because the collision is in the key itself.

Both values are available at no cost from the same artifact the closure hash already reads. From `out/AccessManagerFacet.sol/AccessManagerFacet.json`, under `.metadata.settings`:

```json
{ "evmVersion": "cancun", "viaIR": null, "optimizer": { "enabled": true, "runs": 1000000 } }
```

Suggest adding `evmVersion` and the optimizer fields, or a hash over `metadata.settings` with the source-dependent parts stripped.

script/deploy/codehash/attestation-key.test.ts
```diff
@@ -0,0 +1,200 @@
...
+  serialiseAttestationKey,
+  sourceClosureHash,
+  type IAttestationKey,
+  type IMintedAttestation,
+} from './attestation-key'
+
+const EVM: IAttestationKey = {
+  contractName: 'CBridgeFacet',
+  version: '1.2.0',
+  profile: 'production',
```
gvladika: **`profile: 'production'` contradicts the field's own docstring.**

`'production'` here and `'staging'` at line 47 are deployment environments. The field is documented at `attestation-key.ts:28` as "Foundry profile the build ran under", and `foundry.toml` defines `default`, `ci`, `solc_floor` and `zksync` — there is no `production` profile.

Whichever is intended, the other is wrong. It also changes the scope of my comment on `attestation-key.ts:29`: if `profile` is the deployment environment then it implies no compiler settings at all, and the key's coverage gap is wider than the settings-drift case.

script/deploy/codehash/attestation-key.ts
```diff
@@ -0,0 +1,211 @@
...
+ * ambiguous and hashing either reading would vouch for source that was not
+ * compiled
+ */
+export const sourceClosureHash = (sources: readonly ISourceEntry[]): string => {
+  const byPath = new Map<string, string>()
+  for (const entry of sources) {
+    const seen = byPath.get(entry.path)
+    // Two hashes for one path is not something to pick a winner from: whichever
+    // is dropped, the resulting hash claims a closure that was never compiled.
+    if (seen !== undefined && seen !== entry.keccak)
```
gvladika: **Hash comparison is raw string equality, unlike the rest of this directory.**

Verified against the branch: two attestations of the *same* masked hash differing only in case are reported as a conflict, and the docstring at :174 says a caller must refuse on a non-empty result — so that is a blocking false positive. The check here has the same shape: `0xAA…` and `0xaa…` for one path throw "twice with different hashes".

```
conflicts on case-drifted identical hash:
  maskedHashes: ["0xaaaa…aaaa", "0xAAAA…AAAA"]
closure THREW: source closure names src/A.sol twice with different hashes
```

Latent rather than live today, since viem's `keccak256` and solc both emit lowercase. But `attested-set.ts:74` already carries a `normalizeHash` for exactly this, and `hex.ts` opens with "Import this rather than re-deriving the checks: a module that masks or hashes unvalidated hex produces a plausible-looking result from nonsense input." This module does neither — `keccak: ''` or `keccak: 'nope'` hashes into a closure without complaint. `strip0x` and `frameFault` would cover both, and `findAttestationConflicts` (:192, :194) needs the same treatment.

script/deploy/codehash/attestation-key.ts
```diff
@@ -0,0 +1,211 @@
...
+ * source, not of a build.
+ *
+ * Paths are sorted, so the hash does not depend on the order solc happened to
+ * emit them, and each entry carries its separator so a path ending in the
+ * delimiter cannot impersonate the start of the next entry.
+ * @param sources - Every source in the closure, in any order
+ * @returns `0x`-prefixed keccak over the canonical closure
+ * @throws When two entries claim the same path, since the closure is then
+ * ambiguous and hashing either reading would vouch for source that was not
+ * compiled
```
gvladika: **Two claims in this docstring no longer match the code.**

- :123-124, "each entry carries its separator so a path ending in the delimiter cannot impersonate the start of the next entry" — this describes the delimiter design the PR body says was replaced. There is no delimiter and no per-entry separator; `joinFields` length-prefixes.
- :127, `@throws When two entries claim the same path` — it throws only when they *disagree*. Agreeing duplicates are accepted, and the test at `attestation-key.test.ts:169` asserts that.

script/deploy/codehash/attestation-key.ts
```diff
@@ -0,0 +1,211 @@
...
+  /** Version the contract declares, e.g. `1.2.0`. */
+  version: string
+  /** Foundry profile the build ran under. */
+  profile: string
+  /**
+   * Whether the build went through the IR pipeline. Required rather than
+   * optional: absent and `false` are the same build only by coincidence, and a
+   * key that omits it silently merges an IR build with a legacy one.
+   */
+  viaIR: boolean
```
gvladika: **Four minor ones, grouped here.**

- **`viaIR` is `null`, not `false`, in real solc metadata** — `.metadata.settings.viaIR` in the artifact. The reasoning in this docstring is right, but a required boolean pushes the null→false mapping onto every producer with no shared helper, and there will be at least two of them (the CI mint and a local rebuild). That is the case `hex.ts` warns about: "the two modules disagreeing about what counts as valid is the same bug twice." A `viaIRFromMetadata(settings)` here would keep it in one place.
- **`IAttestationConflict.key` (:164) is the serialised string.** Callers must refuse, so a human ends up reading `1:A5:1.0.010:production5:viaIR6:0.8.293:evm` in the failure. Carrying the structured `IAttestationKey` alongside would make the refusal legible.
- **`entry.masked.includes()` inside the loop (:192, :194) is O(n²).** A `Set` preserves insertion order and produces identical output.
- **`it('is stable for the same inputs')` (`attestation-key.test.ts:88-94`)** compares `{ ...EVM }` with `EVM` through a pure function, so it holds for any implementation.

gvladika (commented):
Reviewed at `68a09603`. Ran the suite (13 pass), `eslint` and `tsc-files` on both files (clean), and checked the assumed solc shape against a real artifact in `out/`.

Confirmed the length-prefix encoding is genuinely injective: the digit run always terminates at the first `:`, so decoding is deterministic, and the nested zk section is safe because the inner result is itself one length-prefixed field. The argument in the PR body for replacing the delimiter holds. Conflicts-returned-rather-than-resolved and the settings-independent closure hash both look right to me.

Six findings. The two on the key itself (`attestation-key.ts:29` and the fixture at `attestation-key.test.ts:23`) are about what the key *means* and I would settle those before merge; the docstring and hash-comparison ones are quick; the grouped batch at `:35` is suggestions. The sixth — the vacuous assertion at `attestation-key.test.ts:85` — is a reply in CodeRabbit's existing thread rather than a duplicate comment here; I confirmed it by mutation.

script/deploy/codehash/attestation-key.test.ts
```diff
@@ -0,0 +1,200 @@
...
+    expect(forkMoved).not.toBe(serialiseAttestationKey(ZK))
+    expect(llvmMoved).not.toBe(serialiseAttestationKey(ZK))
+    expect(forkMoved).not.toBe(llvmMoved)
+  })
+
+  it('gives an EVM build a key a zk build cannot reach', () => {
+    // The absent zk section is written as a marker, not skipped: a key that
+    // shortens when a field is missing is a key two builds can share.
+    expect(serialiseAttestationKey(EVM)).toContain('evm')
+    expect(serialiseAttestationKey(ZK)).not.toContain(' evm')
```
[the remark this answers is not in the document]
gvladika: Confirmed by mutation on the branch. Replacing `NO_ZK` with `''` fails line 84 only — line 85 passes for every possible implementation, because `joinFields` emits `<length>:<field>` and never a space:

```
84 |     expect(serialiseAttestationKey(EVM)).toContain('evm')
error: expect(received).toContain(expected)
(fail) serialiseAttestationKey > gives an EVM build a key a zk build cannot reach
 12 pass, 1 fail
```

So the "absent-zk marker removed" row of the falsification table in the PR description is carried by line 84 alone, and the property line 85's comment states is untested. Asserting the exact `3:evm` form fixes both, and tightens 84 against a contract name that happens to contain `evm`.

script/deploy/codehash/attestation-key.ts
0xDEnYO: Fixed in 51f3012e0: masked and closure hashes are compared lowercased. Case-drift of the same hash is no longer a conflict. Pinned by `treats case-drifted identical hashes as the same source` / `does not report the same hash written in different case`.

script/deploy/codehash/attestation-key.ts
0xDEnYO: Fixed in 51f3012e0: the docstring now matches `joinFields` and only claims a throw on disagreeing duplicates.

script/deploy/codehash/attestation-key.test.ts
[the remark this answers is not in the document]
0xDEnYO: Fixed in 51f3012e0: the EVM key must contain the exact `3:evm` field; the zk key must not.

script/deploy/codehash/attestation-key.test.ts
0xDEnYO: Fixture now uses `default` / `ci` (real `foundry.toml` profiles) in 51f3012e0. The open question remains yours on :29: whether `profile` is the Foundry profile (and then evmVersion / optimizer must be named too) or a deployment environment.

script/deploy/codehash/attestation-key.ts
0xDEnYO: Agreed this is the merge-gating question. Not going to add evmVersion / optimizer from the chair — that changes what the key *means*. Holding draft until it is named: Foundry profile plus the settings `lineage-scope.ts` already parses, or a different identifier.

script/deploy/codehash/attestation-key.ts
0xDEnYO: Noted — viaIR-from-metadata helper, structured conflict key, and the rest of the grouped nits. Will take them in the next pass after the key-shape question on :29 is settled.

script/deploy/codehash/attestation-key.test.ts
0xDEnYO: Fixed in 09379b9dd. `profile` is the Foundry profile, and the fixtures are now real ones: `default` / `ci`, with the zk fixture on `zksync` — a zk build cannot have run under `[profile.default]`. `EVM.viaIR` is `false` too, since no profile in `foundry.toml` sets `via_ir`, so the previous `true` fixture was a build this repo cannot produce.

script/deploy/codehash/attestation-key.ts
0xDEnYO: Fixed in 51f3012e0.

script/deploy/codehash/attestation-key.ts
0xDEnYO: Both halves fixed. 51f3012e0 lowercased the compare; 09379b9dd adds the validation you asked for, as `digestFault` in `hex.ts` — `frameFault` alone was not enough here. It was written for bytecode, whose length is not known in advance, so it accepts `0xdeadbeef` and a 20-byte address sitting in a hash slot, which are the corruptions that actually happen. `digestFault` adds the 32-byte length and both call sites use it. `sourceClosureHash` also refuses an entry with no path.

script/deploy/codehash/attestation-key.ts
0xDEnYO: All four are in — `buildSettingsFromMetadata` for the metadata read, structured `key` plus `serialisedKey` on the conflict, `Set` instead of `includes`, and the stability test replaced by one that pins field order against the object literal order.

One correction on the first bullet, because it changed what the helper had to tolerate: solc emits `viaIR` **only when the IR pipeline ran** and omits the key otherwise. It is absent in all 193 artifacts of a real `out/` tree — never `null`. `jq '{viaIR}'` prints a missing key as `null`, which is where the `null` reading came from (I reproduced it before checking `has("viaIR")`). Your conclusion holds and gets stronger: three shapes reach a producer, not two, so the mapping belongs in one place. The docstring and fixture said `null`; both now say absent, with `null` and `false` kept as tolerated inputs.

script/deploy/codehash/attestation-key.ts
0xDEnYO: Took your second option: `artifactSettingsHash` hashes `metadata.settings` whole, less `compilationTarget` (carried as `sourceId`) and `remappings`. Enumerating was the wrong bet — I had named 3 of ~7 bytecode-affecting fields and was confident, and one of the ones I did name went wrong in exactly the way you'd expect: solc drops `optimizer.enabled` when `optimizer.details` is supplied, so a tuned optimized build keyed as `noopt` and collided with an unoptimized one.

`profile` is out of the key entirely rather than patched around. `[profile.ci]` inherits `[profile.default]` and differs only in fuzz settings, so keying on the name filed one build under two keys. It rides on the attestation as provenance.

Two things your question surfaced that went past it:

**The closure is now in the key.** Your drift argument applies to source as well as settings, and worse. `4cfabbf22` bumped `SwapperV2` 1.0.2 → 1.2.0; 33 facets inherited it at that commit, none was touched, none bumped its own version, and recompiling `AcrossFacet`'s closure at both commits gives two different masked hashes under a byte-identical key. `closureHash` was already computed and already on the attestation — one field short of the key.

**The zk arm is still unconstrained, and I'd rather you see that than not.** zksolc writes no `metadata` object at all (0 of 280 `zkout/` artifacts), so a zk build has neither a settings source nor a `metadata.sources` closure. `configuredSettingsHash` covers the settings half by hashing whatever the producer configures and tagging it `config`, which stops a configured claim answering an artifact lookup — but it does not stop two zk builds under different settings sharing a key, and the closure half has no reader here at all. Written up under "What is left of WP-5.2" rather than papered over. If you think that blocks the schema landing, say so and I'll do it in this PR.

Also worth flagging since it started as your correction: solc does not write `viaIR: null`. It emits the key only when the IR pipeline ran and omits it otherwise — absent in all 190 artifacts. `jq '{viaIR}'` prints a missing key as `null`, which is where we both got it. Moot now that the settings are hashed whole.

0xDEnYO:
@coderabbitai review

script/deploy/codehash/attestation-key.ts
```diff
@@ -0,0 +1,415 @@
...
+ * free to pick the tag could mint an `artifact` hash from configured values,
+ * which is the one thing the tag exists to prevent.
+ */
+const settingsHashTagged = (
+  source: SettingsSource,
+  settings: Record<string, unknown>
+): string => {
+  const hashable = hashableSettings(settings)
+  if (Object.keys(hashable).length === 0)
+    throw new Error('build settings are empty, so they identify no build')
```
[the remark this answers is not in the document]
0xDEnYO: Fixed in e240e6d1e — confirmed reachable and worth the change. Reproduced before fixing: `{evmVersion: null, optimizer: null}` and `{viaIR: null, libraries: null, metadata: null}` both hash to `0x466fd64f…`, so two unrelated settings objects shared one hash under the very check meant to prevent it.

Took the canonical-form version of your fix. As you note it is unreachable from solc's `metadata.settings`, but `configuredSettingsHash` takes a caller-shaped object whose form this PR deliberately leaves open for the zk producer, so that is exactly where it would have landed. Pinned by `refuses settings that canonicalise away to nothing`, which fails if the check moves back ahead of canonicalisation.

gvladika (approved):

PR #2353, opened 2026-09-09
feat(safe): persist the signed set, and re-derive before broadcast (EXSC-953)

0xDEnYO:

Fixes [EXSC-953](https://linear.app/lifi-linear/issue/EXSC-953/wp-61-persist-the-signed-set-pre-broadcast-gate) · WP-6.1 · R3.1, R3.2

Follow-up split out on review: [EXSC-1004](https://linear.app/lifi-linear/issue/EXSC-1004/wp-61b-re-land-the-pre-broadcast-codehash-check-as-an-attestation) · re-lands the codehash half as an attestation lookup, blocked on EXSC-952 (#2372).

Two things: a sign-time record of what each signer's machine saw (R3.1), and a gate that
re-derives its own verdict before the auto-executor broadcasts (R3.2). The gate reports
but does not block — `PRE_BROADCAST_GATE_ENFORCE` must be exactly `'true'` to make a
refusal binding, and it is not set anywhere.

**This PR previously carried a third thing — a broadcast-time codehash comparison — and
Goran's review argued it out.** That reasoning is the main thing worth reading here, so it
is first.

## Why the codehash half is not in this PR

Scheduling takes 3 signatures, each from a signer who ran the sign-time gate on their own
machine. A single tricked signer or lying machine therefore cannot produce a malicious
proposal — quorum covers that. What quorum does not cover is **correlated** failure: three
machines running the same repo code, against the same deployments file, comparing against
artifacts from the same `forge build`, reach the same wrong verdict if `main` is wrong.
Three observers sharing an oracle is one observation.

The cut check re-derived from that same repo, the same deployments file and the same local
build. It was fully correlated with the thing whose correlation is the weakness: on the
honest path it re-asked a settled question, and on the compromised path it was fooled the
same way.

Running it earlier would not have helped either. Nothing else in the delay window can
change — parameters are chain-enforced by `executeBatch`'s own `hashOperationBatch` and
`isOperationReady`, and deployed code is immutable.

What breaks the correlation is a CI-minted attestation: a build nobody's laptop produced.
That is EXSC-952 / #2372. Once it lands the check becomes a *lookup*, cheap enough to sit
exactly where this PR already put the gate, with no rework of the placement. So the cost of
waiting is low, and the cost of shipping now was ~950 lines written against a stand-in
anchor that is known-wrong today (`LiFiDiamond` graded MISMATCH on 6 of 6 mainnets, all
honest deploys).

Gone with it: `readArtifactAnchor`, `deriveLineageScope`, and a second copy of
`normalizeRuntimeCode`. That last one is the clearest argument for the cut — the canonical
implementation's own docstring says both sides of a comparison must go through one
function, and the copy had already diverged in exactly the way it warns about (it masked
the body then stripped the trailer, reversing the order, so an immutable pointing into the
trailer would be *refused* by the canonical function and *silently masked-then-discarded*
by the copy). Its tests passed because non-overlapping immutables agree.

EXSC-1004 carries that finding and four more of Goran's, so the re-land starts from them.

## What the gate checks now

**Storage authorities**, and why they earn the RPC: of the things this gate could
re-derive, they are the only ones whose subject can change during the delay window.

`LiFiDiamond.owner`, `LiFiDiamond.pauserWallet` and `ERC20Proxy.owner`, read live and
compared against what `main` declares. Verified with no false positives on mainnet,
arbitrum, optimism and polygon. The declared-getter table is typed against the ABI the gate
calls through, so declaring a getter the ABI does not carry fails to compile rather than
becoming a run-time read error — which the gate cannot tell apart from an unreachable node,
and would therefore swallow as a HOLD.

Coverage is exactly those three getters, which is three of the five `healthCheckInvariants`
knows. A contract absent from the table produces no row and no finding — and, new in this
round, **the PROCEED reason says so**: it now carries `N of M calldata address(es) resolve
to a contract main names, and K declared storage authority value(s) were read`, on every
disposition. The previous wording ("every target's live code matches … and every storage
authority matches config") claimed more than the gate looked at.

**What the timelock has scheduled.** This replaces the `hashOperationBatch` re-derive the
previous revision carried. That call was `pure`, so it could only ever confirm that viem and
solc agree on ABI encoding — worth knowing once in a test, not once per operation per tick,
and the executor's own trust check 1 had already proven the same equality locally.
`getTimestamp(id)` reads what the controller actually stored instead, for the same one call:
a row whose parameters were edited consistently in both places hashes to an id the
controller has never scheduled, and comes back zero. It does not make the broadcast safe on
its own — `executeBatch` enforces `isOperationReady` regardless — it turns that revert into
a pre-flight refusal an operator can act on.

Zero is not the only sentinel. `TimelockController` writes `_DONE_TIMESTAMP` (1) over the
entry once an operation has run, so a bare `=== 0n` test cannot see an already-executed
operation — and the first revision's test pinned exactly that as PROCEED. One now blocks on
its own branch. It is not reachable today: the executor retires a done row before the gate
runs, and `executeBatch` would revert regardless. The point is that the gate's own stated
invariant is true of the code rather than resting on an upstream guard it does not
reference.

Addresses come from a 4-byte-stride word scan of the payload rather than a decoder: the set
of functions a timelock operation may carry is open, so a decoder covers whatever it was
taught and silently omits the rest. The stride is 4 and not 32 because a call nested in a
`bytes` argument — `diamondCut`'s init `_calldata` — shifts its own words by its own
selector. On a real viem-encoded `diamondCut` over real `deployments/mainnet.json`
addresses with a nested `initX(address,address)`, the 32-byte stride found 3 of 5 addresses
and the 4-byte stride finds 5 of 5. Widening cannot invent a target, because a candidate
only counts if the deployments file already names it: fed all 188 compiled artifacts'
bytecode as payloads against a deliberately over-approximated 1381-address known set, it
produced zero addresses the 32-byte stride did not already find.

### Still blind to the facet a rollout installs

A payload-scanned address is kept only when `knownAddresses.has(candidate)`. A facet a
rollout is *installing* is not in `deployments/<network>.json` yet, so it produces no row.
Top-level `targets` are added unconditionally, so this is payload-only addresses.

Holding on every unresolvable payload address is not the fix — a 4-byte stride over a blob
yields noise nothing can hold on. The fix is decoding, which `classifyCut` /
`verify-cut-targets` on `main` already do. The **reporting** half is closed in this round
(the counts above), so a PROCEED no longer reads as full coverage; the coverage gap itself
is on EXSC-1004.

## The record is an audit trail, never the oracle — and now comparable between signers

`signed-set-record.ts` persists the address→raw-codehash set plus the R2.6
storage-authority values. Every value in it was produced by the proposer's own run, so
treating it as the execute-time reference would mean comparing live code against something
the proposer wrote.

That property is structural, not intended. `deriveGateInput` reduces the whole record to a
boolean; `IObservationDependencies` is split out so the record writer cannot be handed a
record at all; the gate's input type has no field for a stored verdict, hash or authority
value. One test reaches the same verdict from an honest record, a forged one, and none at
all, and another feeds a sentinel string through the record and asserts it appears nowhere
in the gate's input — paired with a positive control so the absence is a fact about the
seam, not about the search.

**The key now carries the signer.** Previously `(network, operationId)` upserted with `$set`
over the whole document, so on a 3-of-N Safe the last signer overwrote the first two and the
stored record claimed one machine's view with nothing indicating the others existed. That
was the record's main reason to exist: one observation is a log line, three agreeing
observations are evidence no single machine can manufacture. `fetchSignedSetRecord` asks the
two-field filter, which is all the gate's presence boolean needs.

Two consequences follow, and both are deliberate:

- The stored hash is **raw**, not masked. Masking needs `immutableReferences` from a local
  artifact, and folding anything machine-local into the hash is exactly what would stop two
  signers' copies from being comparable. `maskedHash` is gone from the document.
- `signed-sets` now gets an index on `(network, operationId, signer)`, created on connect
  and degrading with a warning on a `createIndex` authorization failure — the shape
  `ensureParkedTasksIndexes` already uses, and for the same reason (readWrite-only
  credentials in CI must still be able to write the audit trail).

## The authority verdict goes on the check ledger

`main` moved under this branch: #2369 made `check-ledger.ts` the sign-time composition point
three commits ago. The authority observations now land there as a `storage-authority` row
rather than as a second block printed beside the ledger.

The interesting part is the anchor. The row is anchored on the **weaker of the two sides**,
because a comparison is only as good as its expectation:

| Authority | Expectation from | Anchor | May grade green? |
|---|---|---|---|
| `LiFiDiamond.pauserWallet` | `config/global.json` | `A-LOCAL` | yes |
| `ERC20Proxy.owner` | `config/global.json` | `A-LOCAL` | yes |
| `LiFiDiamond.owner` | deployment record | `A-MONGO` | no — coerced to `error` |

The deployment record is proposer-writable, so a "match" against it is the executor agreeing
with a value the proposer supplied one side of. The ledger's existing reporting-only
coercion turns that into `error` rather than `pass` on its own; a test drives the real
`recordCheck` and asserts the green never appears.

The row is also reduced per network rather than written per proposal. `recordSignedSet`
runs once per proposal, and `rollUpChecks` only lets a `fail` block supersession — so a
later clean proposal would have erased an earlier proposal's unread authority with no
trace, on a check whose own header says a check that could not run is not a check that
passed. It now goes through `proposalChecks` and the worst-first reduction, which is what
the target-state row already did.

The placement test could not see that, because it scanned the loop's own text for
`recordCheck(` and the call sits lexically above the loop. It took three further rounds to
get the guard right, and the route there is worth one paragraph because each round was
blind in a way the previous round's controls could not express.

Two regex versions came first. The second missed `recordNothingToGrade` and
`recordCouldNotGrade` — the file's own recorders — because a brace hunt reads a concise
arrow's returned object literal as its body. Parsing the file fixed the enumeration but not
the direction: walking outwards from the loop resolves a callee by the name it is called
under, so passing a recorder to `forEach`, aliasing it, or hanging it off a property all
dropped out of the walk silently, an unresolved name being indistinguishable from a library
call.

It is inverted now. The set of functions reaching `recordCheck` is computed over the whole
file to fixpoint, and the loop is asked which of those names it mentions at all — which no
choice of invocation can hide. Seventeen shapes are verified in both directions: ten
declaration forms and four indirections fail it, and a string literal or a comment carrying
the name does not. It reads `confirm-safe-tx.ts` alone, so a recorder moved into another
module is outside it; that limit is in the docblock rather than left to be discovered.

The codehash observations have no verdict to record — nothing to compare against until
EXSC-1004 — so they stay document-only. A ledger row needs a comparison.

## Why it ships in shadow mode

Changed reason from the previous revision, because the old one was about the anchor that is
now gone. What remains is that a **HOLD refuses the broadcast**: an authority read that
fails intermittently would stop honest rollouts on every chain it touches. The gate has to
demonstrate it does not hold on honest traffic before an unattended ten-minute cron is
allowed to act on it.

The exactness of the opt-in is deliberate — unset, empty, `'1'`, `'TRUE'`, `'yes'` and
`' true '` all read as shadow, so enforcing cannot be entered by a near-miss in a workflow
file.

### Shadow mode is now log-only

Goran flagged alert volume as a merge blocker, and he was right about the volume even
without the codehash half: every operation queued before this merges has no sign-time
record, so `buildGateGapAlert` would fire on every honest operation on every chain on the
first rollouts after merge, plus every Tron operation. The channel it would post to is the
one carrying blocked-operation and revert-threshold escalations that a human is expected to
act on.

So **the same flag gates the alerting**. Shadow mode logs and posts nothing; turning
enforcement on turns the alerting on with it. Output nobody may act on does not belong on
the channel carrying output somebody must.

### Shadow mode is total, and that is a deliberate fail-open

Three paths in `enforcePreBroadcastGateOrAbort` — `getDeployments` throwing,
`fetchSignedSetRecord` throwing, and any throw inside `runPreBroadcastGate` — route through
`unverifiedGateOutcome`, which consults the flag: enforcing leaves the row queued for the
next tick, shadow mode reports the gap and clears the operation, as the uncovered-Tron path
already did.

**The call site now catches too.** The previous revision's totality rested on every alert
call being unable to throw — a property of `sendNotificationWithRetry`'s default argument in
another module, which a regex over this file could not see. Goran's point: a throw from
`alertGap` or `alertShadowRefusal` would leave the row mid-flight with no outcome recorded
and no alert, the exact failure the in-function `try` was written to prevent one level down.
It now holds by construction rather than by promise.

Worth naming plainly: **three fail-closed paths are fail-open** under shadow mode. Before,
an unreadable deployment record or record store stopped the broadcast; now it proceeds and
logs. That is the right trade only while the gate is non-binding. Under
`PRE_BROADCAST_GATE_ENFORCE=true` all three are fail-closed again.

`retry` still has no attempt budget — tracked as EXSC-972, out of scope.

## One review finding I am reporting differently than it was raised

Goran flagged `REFUSING_RETURN`'s `g` flag in `prebroadcast-gate-placement.test.ts`: passed
to `.not.toMatch` inside a loop, a persisted `lastIndex` could make a genuine match read as
no-match, so the assertion would pass on exactly the refusal it guards.

**The mechanism is real and I confirmed it** — in bun, `toMatch` with a `g`-flagged regex
does advance `lastIndex`, so a second call on the same string reports no match. Fixed with a
separate non-global copy.

**But I could not make it produce a false green in this file.** The poisoning needs an
earlier catch block that also matches, and that earlier block fails the assertion first, so
the test aborts before a later missed match can hide anything. I tried the arrangement
designed to exploit it (a bare refusal late in catch 1, early in catch 2) and both the
buggy and the fixed regex fail. So: latent hazard removed, not a live false-green closed. I
would rather say that than claim a catch I did not verify.

## Mongo client churn — pushing back on half of one finding

The same comment flagged `getSignedSetCollection` opening a client per call, noting the rest
of the directory uses longer-lived handles. I checked, and that premise does not hold:
`getTimelockQueueCollection` (`timelock-queue.ts:183`) and `getParkedTasksCollection`
(`parked-tasks.ts:318`) both construct a `MongoClient` per call with the caller closing it —
it is the house pattern here, and `execute-pending-timelock-tx.ts` already makes 8 such
calls per operation. One more is ~12% on an existing pattern, not a new class of problem.
Pooling is a worthwhile change for the whole directory and a bad one for this file alone.

The index half of that finding was a real gap and is fixed above.

## Nothing the gate observes may carry the endpoint that produced it

viem puts the full node URL — API key included — in the message it throws, and the gate
turns that message into `observationError` / `readError`. Those strings are persisted to the
sign-time record *and*, under enforcement, published to the escalation channel.

The two halves want different helpers, and `[CONV:REDACT-RPC-URL]` says so: `describeError`
returns `redactUrls(...)`, redacted but uncapped, because the record is the durable forensic
artifact and `redactErrorReason`'s 180-char Slack cap lands mid-`Details:` on a real
rate-limit error — it would drop the cause while looking like it kept it. `buildGateGapAlert`
and `buildShadowRefusalAlert` apply the cap per bullet, where the text is actually published.

Sweeping the same value found three sinks the gate routes into that were not the gate's own:
`SlackNotifier.extractErrorMessage` returns `errorObj.message` raw into the Slack block, so
`alertFailure` in both pre-execute guards and the broadcast-failure catch in `executeOperation`
published unredacted viem and Mongo-driver errors. The last is live today, not gated by
`PRE_BROADCAST_GATE_ENFORCE`. `guard-error-redaction.ts` carries the two shapes with its own
tests; the call sites are one line each.

The six older sinks elsewhere in `execute-pending-timelock-tx.ts` are the same class and are
closed here too, rather than left for a follow-up that would only conflict with this file:
the per-network catch and the two guard catches carry viem and Mongo-driver errors, and the
three notification-send catches carry a failed POST to the webhook — whose URL is itself the
credential. The three `abortUnverified` call sites pass the error onward untouched on purpose;
that function redacts at each of its own two sinks.

`script/utils/rpc-url-log-scan.ts` cannot regress-guard any of this: its own docblock lists
"an endpoint viem embeds in an error object's message" as a gap it does not reach. The tests
are the only backstop, so a future `describeError`-shaped site leaks silently.

## The executor workflow moves onto the foundry pin

`runPendingTimelockTXs.yml` called `foundry-rs/foundry-toolchain` directly. That was a missed
migration, not an exception: the step landed three days before the commit that introduced
`./.github/actions/setup-foundry` and the `[CONV:FOUNDRY-SETUP]` rule, and the sweep never
touched this file.

It is a behaviour change, not a refactor — the old step installed an unpinned v1.6, the
composite installs `.foundry-version` (1.7.1). It also adds three cache restore/save steps to a
job whose next step is `bun typechain` (`forge clean && forge build src`), which discards the
restored `out/` immediately; the cost is restore time on every ten-minute tick, and the cache
key is namespaced by workflow so it cannot affect anything else.

## Scope this PR does not cover

EXSC-953 asks for the re-derive to run early (right after scheduling, so the delay window is
available as cancel runway), repeatedly, and on an independent third machine. This runs
in-process in `execute-pending-timelock-tx.ts` at broadcast time: no cancel runway, and no
independence from the machine it gates. The repeating cron is not here. Shadow mode makes
that acceptable for now — a reporting-only gate on the executor is strictly more than the
nothing that is there today — but the placement the ticket asks for is still outstanding.

Goran's separate question — whether the *continuous* authority check belongs in
`healthCheckInvariants.ts`, since authority drift is an emergency whether or not an
operation is queued — is a good one and is not answered here.

Tron is uncovered: reading code there needs a TronWeb client the executor does not carry.
The gate reports the gap explicitly rather than returning a verdict, tracked as EXSC-954.

## Not claimed

The WP-6.1 acceptance rows are unclaimed: no row has been shown to fail at the commit before
this lands. The two executor call sites are not proven by an executing test — the placement
test asserts where the gate sits in the source, not that a real run takes that path; the
only harness that would prove that spawns the executor for real.


0xDEnYO:
@coderabbitai review

0xDEnYO:
@coderabbitai review

script/deploy/safe/prebroadcast-anchors.ts
```diff
@@ -0,0 +1,393 @@
...
+  }
+
+  for (const payload of payloads) {
+    const body = strip0x(payload ?? '')
+    // Skip the 4-byte selector so words line up with ABI encoding.
+    for (
+      let offset = 8;
+      offset + EVM_WORD_HEX_CHARS <= body.length;
+      offset += EVM_WORD_HEX_CHARS
+    ) {
```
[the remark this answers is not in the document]
0xDEnYO: Fixed in `6e7e96e6d`, and the finding is right about more than the stride: the test that pinned it asserted a misaligned pair finds nothing, so the defect had been written down as a property. Replaced with the case that a nested frame's address IS found, plus the paired present that eight times as many windows still name nothing the deployments file does not hold — that half matters because a spurious row would make the gate read code for an address the operation never touches, and a failed read holds the operation. Falsified: reverting to `offset = 8` with the 32-byte stride fails the nested-frame test and only that one, restored byte-identical. 46 tests pass.

script/deploy/safe/execute-pending-timelock-tx.ts
```diff
@@ -1659,6 +1672,153 @@ async function revalidateFoldedRemovalsOrAbort(
...
+      targets: operation.targets,
+      payloads: operation.payloads,
+    },
+    {
+      ...viemGateReaders(publicClient),
+      deployments,
+      globalConfig: globalConfig as unknown as Record<string, unknown>,
+      networkConfig: (data as INetworksObject)[networkName] ?? {},
+      artifactRoot: REPO_ROOT,
+      lineage: `local build of ${getGitCommit()}`,
```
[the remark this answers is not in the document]
0xDEnYO: Fixed in `6e7e96e6d`. I confirmed the boundary claim before changing anything — the guard is called at :1888 and `executeOperation`'s own try starts at :1950, so a throw did leave by a path that returns no GuardOutcome, records no status and sends no alert. It now holds and alerts like every other failed read in that function, matching `revalidateFoldedRemovalsOrAbort`'s contract.

Stated plainly: this fix carries **no executed test**. Making `runPreBroadcastGate` or `getGitCommit` throw inside the real CLI means spawning the executor, which must not run from a test — so the evidence is the sibling contract plus the static call-site check, not a red-to-green.

0xDEnYO:
Post-fix adversarial pass over `6e7e96e6d` — no blocking defect, but it found my own justification untrue, corrected in `db5e65bd7`.

**The comment named the wrong throw sources.** `getGitCommit` cannot throw: `shared/git-provenance.ts` is fail-soft by contract (provenance must never be why a deploy fails) and yields `PROVENANCE_UNKNOWN`; `readArtifactAnchor` returns undefined on every failure. The genuinely uncaught sources are the derivation helpers — `deriveLineageScope` and `resolveExpectedAuthority` are called outside the per-row try blocks (`prebroadcast-gate.ts:112`, `:176-180`), plus what the `Promise.all` surfaces. The catch stays and is right; it is now defence in depth on honest grounds rather than on two fail-soft calls.

**I also understated the cost of a spurious address row.** It is not merely a hold: the address is name-resolved and its live code compared, so a referenced-but-never-called contract on an older build would MISMATCH → BLOCK, a false red an operator has to clear. What makes the widening acceptable is likelihood, and that is now measured rather than asserted — all 188 compiled artifacts' bytecode fed in as payloads against a deliberately over-approximated 1381-address known set produced **zero** addresses the 32-byte stride did not already find. A 20,000-case fuzz found 0 invented addresses, 0 lost versus the old stride, 0 duplicates, stable order.

The stride fix is load-bearing on real data: a viem-encoded `diamondCut` over real `deployments/mainnet.json` addresses with a nested `initX(address,address)` in `_calldata` finds **3 of 5** addresses under the old stride and **5 of 5** now — the init target and both nested arguments.

The comment now also says plainly that four bytes is not every byte: nesting shifts by a selector, always a multiple of four, but an address packed at an arbitrary offset inside a `bytes` blob is still missed. Deliberate.

Two things filed rather than fixed here, as [EXSC-972](https://linear.app/lifi-linear/issue/EXSC-972): a guard `retry` has no attempt budget, so a **durable** cause (a malformed config, say) alerts every 10 minutes forever with no analogue of `REVERT_BLOCK_THRESHOLD`; and none of that guard's three catches is testable, because the file is deliberately non-importable — its own footer prescribes the fix (move the guard to a sibling module with the dependencies injected), which would also bring the two pre-existing catches under test.

0xDEnYO:
## Review: not ready — the gate refuses every honest operation on the fleet

Merged `origin/main` in (`e58ff8d33`) to clear the conflict; the only collision was the import block in `confirm-safe-tx.ts`, resolved by keeping both sides. Staying a draft, because the re-derivation's attestation source is wrong in a way that breaks the auto-executor fleet-wide.

### Measured, on production data

I ran the gate's own path — `readArtifactAnchor` → `deriveLineageScope` → `compareToAttestedSet` — against live code at the real `LiFiDiamond`, `LiFiTimelockController` and `ERC20Proxy` addresses on six chains, using a fresh `forge build src` for `out/` exactly as the cron's `bun typechain` step produces it:

| network | LiFiDiamond | LiFiTimelockController | ERC20Proxy |
| --- | --- | --- | --- |
| mainnet | **MISMATCH** | MATCH | MATCH |
| arbitrum | **MISMATCH** | **MISMATCH** | **MISMATCH** |
| base | **MISMATCH** | **MISMATCH** | **MISMATCH** |
| optimism | **MISMATCH** | **MISMATCH** | MATCH |
| polygon | **MISMATCH** | **MISMATCH** | **MISMATCH** |
| bsc | **MISMATCH** | MATCH | **MISMATCH** |

`LiFiDiamond` is MISMATCH on 6/6. Every one of those reads `deployedSolc=0.8.17 builtSolc=0.8.29 closedSet=true` — the deployed code is an honest older build, and the closed-set branch does not consult the compiler version at all, so it lands on MISMATCH rather than the softer UNVERIFIABLE.

Every timelock schedule batch targets the diamond, so `evaluatePreBroadcastGate` returns **BLOCK** for effectively every queued operation. `enforcePreBroadcastGateOrAbort` then calls `blockTimelockOp`, which flips the queue row to `blocked` — durable, never auto-retried, one manual `requeue-timelock-op.ts` per row to clear. That is a fleet-wide outage of the ten-minute executor on the first tick after merge, plus a Slack alert per operation per chain.

So the falsification bar is met in the wrong direction: the new check demonstrably fires on real data, and it fires on 100% of honest traffic. It is a false-refusal generator, not a drift detector.

### Why: the attested set is one fresh local build

`prebroadcast-anchors.ts` reads a single artifact out of the **current checkout's** `out/` and hands `compareToAttestedSet` a one-element set, with a local `deriveLineageScope` that closes the set whenever `networks.json`'s `targetEvmVersion` equals the artifact's `metadata.settings.evmVersion` (62 of 73 networks declare `cancun`, which is the default profile — so closed nearly everywhere). That assumes what is deployed was built from today's `main` with today's toolchain, which is false for anything not redeployed since the compiler moved.

`main` already solves this, and its own header names this exact failure — `script/deploy/codehash/rebuild-attestations.ts`:

> The record supplies the commit, never the toolchain. Fourteen production slots record `0.8.29` for code the compiler stamped `0.8.17`, so a profile taken from the record would grade honest deploys rogue.

The sign-time gate consumes that: `codehash-sign-gate-deps.ts` uses `deriveToolchainScope` (from `lineage-scope.ts`) plus `rebuild-attestations`, which rebuilds **at the deployment record's commit under every legitimate toolchain**. This PR re-implements a one-profile, current-checkout version of both and inherits the bug they were written to avoid.

### The part that needs a decision, not a patch

Pointing the gate at `attestationsFor` is not a mechanical swap: a rebuild means fetching the deployment commit and compiling, per contract, per toolchain — inside a 10-minute cron with a 20-minute cap that sweeps 60+ chains. Either the pre-broadcast gate wants the WP-5.2 attestation *lookup* rather than a rebuild, or it needs a cached store, or the gate has to run somewhere other than the executor tick. That is a design call, so I have not guessed at it.

### Other findings, worth fixing in the same pass

1. **`blocksBroadcast` is dead where it matters.** `DISPOSITIONS_THAT_MAY_BROADCAST` exists so a disposition added later refuses by default, and nothing in production reads it — the only assertions are in `prebroadcast-rederive.test.ts`. The executor branches on `result.disposition === 'PROCEED'`, which is precisely the pattern the set was built to prevent. Have `enforcePreBroadcastGateOrAbort` use `result.blocksBroadcast`.
2. **`readSignTimeRecordPresence` / `ISignTimeRecordPresence`** wrap `record != null` in an exported interface plus function for one internal call site in `deriveGateInput`. Inline it; the T5 seam is `deriveGateInput`'s signature, which does the work on its own.
3. **`viemGateReaders.readAuthority` casts `getter as 'owner' | 'pauserWallet'`.** The cast makes an unknown getter a type-level non-event and a runtime `readError` — i.e. a silent HOLD. `DECLARED_STORAGE_AUTHORITIES` should be typed against `AUTHORITY_ABI`'s function names so adding a getter without extending the ABI fails to compile.
4. **`persistSignedSetRecord` `$set`s the whole record including `_id`.** Harmless at today's only call site because `buildSignedSetRecord` never sets it, but an upsert that ever carries one throws on an existing document. Omit `_id` from the `$set`.
5. **Comment discipline** (`.agents/rules/000-global-standards.md`) — several comments carry measurement notes, rejected alternatives and PR rationale rather than a non-obvious *why*. Worst offenders: the 24-line block in `extractCalldataAddresses` ("Measured against all 188 compiled artifacts' bytecode as payloads…", "Worth being precise about the cost if it ever did…"), and the 12-line block above the `runPreBroadcastGate` try in `execute-pending-timelock-tx.ts` ("The two obvious-looking sources are NOT the reason…"). Both belong here, in the PR body.
6. **PR body** does not follow `.github/pull_request_template.md` — it keeps only the Linear-task heading and replaces the rest with custom sections, including process narration ("Added after the first push", "Rebased", "This work was written by a session that was killed by a usage limit"). Worth rewriting against the template once the design question is settled.

### What holds up

The re-derivation-not-record property is real and structurally enforced: `deriveGateInput` reduces the stored record to a boolean, `IObservationDependencies` genuinely cannot be handed one, and the gate input has no field for a stored verdict. A `MISMATCH` refuses rather than warns, and a `HOLD` is distinguished from it correctly. The problem is not the shape of the decision — it is what the decision is anchored to.

0xDEnYO:
## Shadow mode landed (`b20249ab6`) — and the re-anchor half has an unmet precondition

Per the ruling on EXSC-953 (shadow-mode first, then anchor onto the WP-5.2 attestation store), the first half is in. The second half **cannot be built yet**, and that is the finding worth reading.

### What shipped

The gate still reaches the same verdict; it no longer acts on it.

- `PRE_BROADCAST_GATE_ENFORCE` gates enforcement, off unless the value is exactly `'true'`. Unset, empty, `'1'`, `'TRUE'`, `'yes'` and `' true '` all read as shadow — the fail-closed direction here is *not enforcing*, because the failure this default protects against is an outage, not a bad broadcast.
- A refusal in shadow mode logs the full verdict, alerts Slack, and returns `ok`. `blockTimelockOp` is never reached, so no queue row is flipped to `blocked`.
- The policy is a pure function of the environment (`isPreBroadcastGateEnforcing`) rather than a module constant, so both modes are reachable from a test **without spawning the executor** — spawning it would run the real thing.
- `enforcePreBroadcastGateOrAbort` now branches on `result.blocksBroadcast` instead of `disposition === 'PROCEED'`. `DISPOSITIONS_THAT_MAY_BROADCAST` exists so a disposition added later refuses by default, and nothing in production was reading it.
- The shadow alert is its own builder. Reusing `buildGateGapAlert` would have said the gate "proceeded without a complete verdict" — false here, and it is the difference between "nothing was checked" and "something was checked and found wrong".

11 new tests, 2031 pass / 0 fail across `script/deploy/safe/`. Falsified by mutation: flipping the policy to fail-open kills 6; giving the shadow alert the gap wording kills 2. Both restored via `git checkout` and verified.

### Why the re-anchor is blocked

The ruling says to anchor on a WP-5.2 attestation **lookup** instead of a rebuild. Measured on `origin/main` at `3b592f68c`:

- `git grep -l attestation origin/main -- .github/workflows/` → **no matches**. Nothing mints attestations in CI.
- The only `attestationsFor` implementation is `createAttestationSource` in `rebuild-attestations.ts`, whose own header says a rebuild "is minutes of compile". Its two consumers — `verify-cut-targets.ts` and `codehash-sign-gate-deps.ts` — are sign-time paths where a human is waiting, so minutes are affordable there.

Both EXSC-952 PRs (#2351, #2363) are merged, but they shipped the *key and grading* half. **There is no persisted attestation store to look up**, so "lookup instead of rebuild" has nothing to read. EXSC-952 is still In Progress and that is the missing piece.

So the dependency is the other way round from what the ruling assumed: the re-anchor is gated on WP-5.2's minting half existing, not merely on choosing to use it.

### What that leaves

Shadow mode is not a stopgap here — it is the only correct posture until there is something honest to anchor to, and it closes the fleet-outage risk completely. This PR can go ready on that basis, with the enforcement flip tracked separately against EXSC-952.

Still open from the earlier adversarial pass, not touched by this commit: inline `readSignTimeRecordPresence`, type `DECLARED_STORAGE_AUTHORITIES` against `AUTHORITY_ABI`, omit `_id` from `persistSignedSetRecord`'s `$set`, strip the measurement-note comments, and rewrite the body against the PR template.

script/deploy/safe/execute-pending-timelock-tx.ts
```diff
@@ -1721,11 +1958,29 @@ async function executeOperation(
     decodeContext
   )
 
+  // Ahead of the prompt on purpose: an operator must see the integrity verdict
+  // before choosing, and must not be offered Execute against a refusal. Reject
+  // stays offered, because a refused operation is exactly one a human may want
+  // to cancel.
+  const gate = networkName
```
gvladika, line 1960: **Stands regardless of the codehash decision.**

Shadow-mode totality is asserted here where it could be structural. `alertGap` and `alertShadowRefusal` are on every path out of `enforcePreBroadcastGateOrAbort`, and this call site has no `try`/`catch` — a throw from either leaves the row mid-flight with no outcome recorded and no alert, which is the failure `6e7e96e6d` was written to prevent one level down.

The PR is explicit that totality rests on `sendNotificationWithRetry`'s `throwOnFailure` defaulting to false (`script/utils/slack-notifier.ts:175-193` — correct today, I checked). But that is a property of another module's default argument, and the thing asserting it is a regex over this function's own source, which cannot see that module at all.

```ts
const gate = networkName
  ? await (async () => {
      try {
        return await enforcePreBroadcastGateOrAbort(/* … */)
      } catch (error) {
        consola.error(`${networkPrefix} pre-broadcast gate threw outside its own handling:`, error)
        return unverifiedGateOutcome(process.env)
      }
    })()
  : 'ok'
```

That makes it unfalsifiable and lets most of `prebroadcast-gate-placement.test.ts` go.

script/deploy/safe/confirm-safe-tx.ts
```diff
@@ -282,6 +315,90 @@ const processTxs = async (
...
+   * Records what this machine saw at every address in the signed calldata
+   * (WP-6.1 / R3.1): the address→codehash set plus the declared storage
+   * authorities.
+   *
+   * A G6 reconstruction trail, not a check — nothing here can refuse a
+   * signature, and the pre-broadcast gate never reads these values back. Every
+   * failure is therefore a warning: the gate alerts on a record that never
+   * landed, and blocking here would turn a write error into a signing outage.
+   */
+  async function recordSignedSet(
```
gvladika, line 406: **Stands regardless of the codehash decision.**

This is the half I would land, but not as a second surface at the same prompt.

`main` moved under this branch: #2369 (`ab7e5f3ed`, three commits ago) made the check ledger *the* sign-time composition point. `check-ledger.ts` already carries `pass | fail | error | needs-ack`, `integrity | semantic`, the anchor set (`A-LOCAL`, `A-CHAIN`, `A-MONGO`, `A-UNRESOLVED`), reporting-only anchor coercion, and this rule in its own header:

> A check that could not run is not a check that passed: it is recorded as `error`, counted as unverified, and the verdict comes back blocked with no acknowledgement path.

`formatSignedSetForDisplay` re-implements exactly that rule by printing `NOT READ` rows, and `recordSignedSet` prints its own block beside the ledger's. The merge conflict this file currently has is the symptom rather than the problem.

Suggest: the codehash and authority observations become `recordCheck` rows on `A-CHAIN` (live value) against `A-LOCAL`/`A-MAIN` (expectation), and the document keeps only what the ledger has no place for. That also gets the reporting-only coercion for free, which matters here — every value in the record is proposer-produced.

script/deploy/safe/signed-set-record.ts
```diff
@@ -0,0 +1,299 @@
...
+
+/**
+ * Filter for the natural key. `$eq`-wrapped so a value arriving as an object
+ * cannot become a query operator.
+ *
+ * @param network - Lowercased network name.
+ * @param operationId - Timelock operation id.
+ * @returns The filter.
+ */
+export const bySignedSetKey = (
```
gvladika, line 301: **Stands regardless of the codehash decision.**

The key has no signer dimension, and `persistSignedSetRecord` upserts with `$set` over the whole document including `signer` — so on a 3-of-N Safe the last signer overwrites the first two, and the stored record then claims one signer and one machine's view with nothing indicating the others existed.

The PR names this:

> That is a real limit on the record as an adversarial artifact — comparing what two signers' machines each saw is not possible today

It is also the record's main reason to exist. Three machines that each observed the same set is the evidence; one machine's observation is a log line. Adding `signer` to the key is a one-line change now and a migration later:

```ts
export const bySignedSetKey = (
  network: string,
  operationId: string,
  signer: string
): Record<string, unknown> => ({
  network: { $eq: network.toLowerCase() },
  operationId: { $eq: operationId },
  signer: { $eq: signer.toLowerCase() },
})
```

`fetchSignedSetRecord` then becomes "any record for this operation" (`findOne` on the two-field filter, or a count), which is all the gate's `signTimeRecordPresent` boolean needs.

script/deploy/safe/signed-set-record.ts
```diff
@@ -0,0 +1,299 @@
...
+  return lines
+}
+
+/**
+ * Opens a short-lived client and returns the signed-set collection.
+ *
+ * @returns The connected client (caller must `close()`) and the collection.
+ * @throws When `MONGODB_URI` is not set.
+ */
+export const getSignedSetCollection = async (): Promise<{
```
gvladika, line 268: **Stands regardless of the codehash decision.** Two operational notes.

A fresh `MongoClient` is opened and closed per call, and both callers are in per-operation paths — `fetchSignedSetRecord` runs inside the executor's gate, so that is one connect/close per operation per network per tick of a 10-minute cron sweeping 60+ chains. The rest of this directory reaches Mongo through longer-lived handles; worth matching rather than opening a client per read.

Second: no index is declared on `signed-sets`, so `findOne(bySignedSetKey(...))` is a collection scan that grows with every operation ever signed. Whatever creates the `queue` collection's indexes should create `{ network: 1, operationId: 1 }` here too (and `signer` if that joins the key — see the comment below).

gvladika (commented):
The PR body already names nearly every limitation below — the new-facet blind spot has its own heading, the `deriveLineageScope` divergence is disclosed, the op-id check is called weaker than it first looks. So this isn't a list of discoveries. The disagreement is narrower: **which of those acknowledged limitations are follow-ups, and which are the shape of the change.**

Verified locally at `eb99393c6`: 121 tests pass, `tsc-files` and `eslint` clean on the new files. Currently conflicting with `main`.

## What I'd change

Land two of the three halves now; cut the third.

**1. The sign-time record — land it, on the check ledger.** `recordSignedSet` prints via `formatSignedSetForDisplay` and writes its own store, in parallel with `recordCheck`/`renderCheckLedger`, which #2369 made the sign-time composition point three commits ago. The ledger already has `pass|fail|error|needs-ack`, `integrity|semantic`, anchors including `A-LOCAL`/`A-CHAIN`/`A-MONGO`, and the rule that a check which could not run is not a check that passed — which `formatSignedSetForDisplay` re-implements by printing NOT READ rows. The conflict in `confirm-safe-tx.ts` is the symptom. These want to be ledger rows plus a side-car document.

**2. The authority check — land it where it is.** Of the three checks, only storage authorities have a subject that can change during the delay window. That one earns its place at broadcast time, and it is the cheapest thing here. The `AuthorityGetter` extraction off `AUTHORITY_ABI` is a good touch — a getter declared but absent from the ABI fails to compile instead of becoming a read error the gate would swallow as a HOLD.

Separately worth asking: should the *continuous* version live in `healthCheckInvariants.ts`? Authority drift is an emergency whether or not an operation is queued, and tying the observation to the queue means we only look during rollouts.

**3. The codehash re-derive — cut it from this PR.**

## Why the codehash half should wait

Scheduling takes 3 signatures, each from a signer who ran the sign-time gate on their own machine. So a single tricked signer or lying machine cannot produce a malicious proposal — quorum covers that. What quorum does not cover is **correlated** failure: three machines running the same repo code, against the same deployments file, comparing against artifacts from the same `forge build`, reach the same wrong verdict if `main` is wrong. Three observers sharing an oracle is one observation.

The gate re-derives from the same repo, the same deployments file, the same local build. It is fully correlated with the thing whose correlation is the weakness. On the honest path it re-asks a settled question; on the compromised path it is fooled the same way.

That is not an argument for running it earlier either. A post-schedule sweep would share the same fate, and nothing else in the window can change: parameters are chain-enforced by `executeBatch`'s own `hashOperationBatch` + `isOperationReady`, and deployed code is immutable.

The one thing that breaks the correlation is a CI-minted attestation — a build nobody's laptop produced. That is #2372, and it is in flight. Which means `readArtifactAnchor` / `deriveLineageScope` / `normalizeRuntimeCode` (~250 lines, plus 546 lines of tests) are a stand-in for a store that lands shortly, and once it does the check becomes an attestation *lookup* — cheap enough to sit exactly where this PR already put it, in the executor, with no rework of the placement. The PR's own framing supports this:

> There is no persisted attestation store to read, so "lookup instead of rebuild" has nothing to look up.

Until there is, the anchor is the load-bearing input and it is known-wrong — `LiFiDiamond` at 6 of 6 on mainnets. Shipping the harness before the anchor means its shape was chosen without the anchor's constraints visible.

Cutting it also resolves the blind spot the PR names, without needing the decoder follow-up:

> A facet a rollout is *installing* is not in that file yet, so it produces no target row, no finding and no alert, while PROCEED reads as though every address in the calldata was checked.

## Findings that stand regardless

Inline below. The ones that apply to the two halves I would land — the call-site `try`/`catch`, the ledger composition, the record's signer key, the Mongo client churn, and the alert-volume decision — want fixing either way. The rest are marked as applying only if the codehash half ships anyway.

The conflict with `main` needs resolving regardless. The unclaimed WP-6.1 acceptance rows and the two unproven call sites are already disclosed; leaving them unclaimed is the right call.

One thing worth keeping whatever happens to the surrounding code: the `unverifiedGateOutcome` fix in `23e7163e6`. Three fail-closed paths reachable with the flag unset would have reported honest operations as failed fleet-wide. That is the most valuable commit in the branch.

script/deploy/safe/prebroadcast-gate.ts
```diff
@@ -0,0 +1,475 @@
...
+ * The message for a refusal the run declined to act on.
+ *
+ * Must not share wording with {@link buildGateGapAlert}: that one says nothing
+ * was checked, this one says something was checked and found wrong, and the
+ * two warrant different responses.
+ *
+ * @param input - The network and operation, the disposition, and its findings.
+ * @returns The message, or null when there is no refusal to report.
+ */
+export const buildShadowRefusalAlert = (input: {
```
gvladika, line 458: **Stands regardless — and I would treat it as a merge blocker rather than a follow-up.**

With `LiFiDiamond` grading MISMATCH fleet-wide, this fires on every honest operation on every chain. A 60-chain rollout is 60 of these, plus a `buildGateGapAlert` on every operation queued before this merges (no sign-time record exists for any of them) and on every Tron operation.

It is bounded at roughly one per operation rather than per tick — shadow mode returns `ok`, so the row executes and does not recur — but that is still the entire rollout volume landing on the timelock webhook.

The PR asks the right question:

> Whether that alert is throttled or held until #2372 is in wants a decision before this runs for long.

I would answer it before turning this on rather than after. Alert fatigue on this channel is expensive to undo, and the channel is the same one carrying the blocked-operation and revert-threshold escalations that actually need a human. Options: route shadow-mode output to a separate low-traffic webhook, reuse the `BLOCKED_ALERT_INTERVAL_MS` throttle shape, or hold the alerting entirely until the anchor is real and log-only until then.

script/deploy/safe/prebroadcast-gate.ts
```diff
@@ -0,0 +1,475 @@
...
+ * the row's parameters hash to the id the row claims, and nothing more. A row
+ * edited consistently in both still agrees here; what refuses that one is the
+ * timelock itself, which has no schedule entry under the new id.
+ *
+ * @param publicClient - Client for the network.
+ * @param timelockAddress - The controller to ask.
+ * @param params - Operation parameters from the queue row.
+ * @returns The reader dependency.
+ */
+export const viemOperationIdReader =
```
gvladika: **Stands if anything remains at execute time.**

Agreeing with the PR's own assessment here rather than disputing it:

> in the current wiring this cannot reach a verdict that check has not already reached. What it is, honestly, is a viem-versus-chain encoding cross-check.

Trust check 1 (`execute-pending-timelock-tx.ts:1139`) already proves `computeOperationIdBatch(params) === row.operationId` with a strict `!==` before any of this runs, and `executeBatch` re-derives the id on-chain and checks `isOperationReady` regardless of what we do. So this RPC can only ever confirm that viem and solc agree on ABI encoding — which is worth knowing once, in a test, not once per operation per tick.

Sourcing `id` from `row.operationId` rather than the local derivation is right and worth keeping either way.

`getTimestamp(id)` reads what the controller actually stored, is already used in the cancel path a couple of hundred lines up, and costs the same one call. If the RPC stays, I would spend it there.

script/deploy/safe/prebroadcast-anchors.ts
```diff
@@ -0,0 +1,398 @@
...
+ * claiming a signed attestation exists. `solcVersion` comes from the
+ * artifact's own trailer, never from its metadata block, so it is read the same
+ * way the deployed code's version is.
+ *
+ * @param contractName - Contract to read, as named in the deployments file.
+ * @param artifactRoot - Directory holding forge's `out/`.
+ * @param lineage - Label for this build, e.g. the commit it was built from.
+ * @returns The anchor, or undefined when no artifact exists for the name.
+ */
+export const readArtifactAnchor = (
```
gvladika: **This is the function the "cut the codehash half" argument is about.**

`readArtifactAnchor` is a stand-in for the attestation store #2372 is building. Everything downstream of it inherits that: the anchor is one local build of the current checkout under the default profile, so anything not redeployed since the compiler moved grades MISMATCH, which the PR measures at 6 of 6 for `LiFiDiamond`.

The `provenance: 'A-LOCAL'` marking is the right call and honestly done — a match here is the executor agreeing with its own build and can never be presented as CI-attested. But that is also the whole problem: the executor agreeing with its own build is not independent evidence, and independence is the only thing this check could add over the three sign-time runs that already happened.

Once #2372 lands, this function, `deriveLineageScope`, `normalizeRuntimeCode` and `prebroadcast-anchors.test.ts` are all replaced by a lookup — which is cheap enough to sit right where the gate already is, with no change to the placement you have chosen. So the cost of waiting is low and the cost of shipping now is ~800 lines that get deleted plus a re-anchoring pass.

Not a correctness objection. A sequencing one.

script/deploy/safe/prebroadcast-anchors.ts
```diff
@@ -0,0 +1,398 @@
...
+}
+
+/**
+ * Turns raw runtime bytecode into the normalised form the comparison takes.
+ *
+ * @param runtimeHex - Bytecode as read, `0x`-prefixed.
+ * @param refs - `immutableReferences` from the matching artifact, if any.
+ * @returns The normalised observation, or why it could not be produced.
+ */
+export const normalizeRuntimeCode = (
```
gvladika: **Applies if the codehash half ships anyway.**

This name already exists. `script/deploy/codehash/rebuild-attestations.ts:180` exports `normalizeRuntimeCode(runtimeHex, refs, { isZk })`, and its docstring is specifically about not doing this:

> Both sides of the gate MUST go through this one function. The attested and observed hashes are only comparable if the same bytes were removed from each, and a second implementation of "strip then mask" is how two sides come to normalise differently while both look finished.
>
> The trailer comes off before masking, so an immutable occurrence pointing into it is refused rather than zeroing bytes the hash no longer covers.

This copy reverses that order: it masks the full body, then slices `trailer.totalStrippedBytes` off the end. For non-overlapping immutables the two agree, which is why the tests pass — but `maskImmutables` bounds-checks against the code it is handed (`immutable-offsets.ts:104`), so an occurrence pointing into the metadata trailer is **refused** by the existing function and **silently masked-then-discarded** by this one. That is precisely the adversarial input the original comment was written for.

Two more deltas from the original:

- No `isZk` branch. zksolc keeps immutables in `ImmutableSimulator`, so an offset into zk runtime code points at real codegen; masking it blinds the comparison to 32 bytes per immutable. The original refuses to do this deliberately.
- No `frameFault` validation, so a non-hex code string reaches `keccak256` and throws out of `observeTarget`, which only wraps `readCode`.

Fix is to import the existing one.

script/deploy/safe/prebroadcast-anchors.ts
```diff
@@ -0,0 +1,398 @@
...
+ * Closed only when the artifact was built with the exact profile the network
+ * declares and the network is not a zkEVM, where a different compiler produces
+ * different bytecode from the same source. Derived from repo configuration on
+ * both sides; the deployed bytecode has no say, because the proposer writes it.
+ *
+ * @param network - The network's `config/networks.json` entry.
+ * @param artifactEvmVersion - The profile the artifact declares.
+ * @returns The scope to hand the comparison.
+ */
+export const deriveLineageScope = (
```
gvladika: **Applies if the codehash half ships anyway.**

`deriveToolchainScope` in `script/deploy/codehash/lineage-scope.ts` already does this, from `config/networks.json` + the `foundry.toml` profiles, and it is strictly stronger: it throws on contradictory config, resolves `targetEvmVersion: "n/a"` to the zk profile, and returns a *closed* set with the profile list for every well-configured network.

This version compares the network's declared EVM version against the local default-profile artifact's and opens the set otherwise. Measured against `config/networks.json` as it stands, that opens the set on:

```
fuse, mantle, nibiru, telos, viction   london   → deriveToolchainScope closes on solc_floor
abstract, lens, zksync                 n/a      → closes on the zksync profile
localanvil                             ""       → deriveToolchainScope throws; this opens silently
```

The PR discloses the count:

> The local `deriveLineageScope` opens the attested set on 11 of 73 networks where `lineage-scope.ts` would close it, which downgrades a real tamper from BLOCK to a silently-retrying HOLD on those chains.

Worth adding that on the five `london` networks it is not only the scope that is wrong. `bun typechain` runs `forge build src` under the default profile (0.8.29/cancun), so the single artifact this gate reads is not the toolchain those chains ship — the anchor is wrong, not just open.

This one is config-only, no rebuild, so the "a ten-minute cron cannot pay for a compile" argument does not reach it. Reusing `deriveToolchainScope` is a strict improvement whenever this code runs.

script/deploy/safe/prebroadcast-gate.ts
```diff
@@ -0,0 +1,475 @@
...
+ */
+export type GateCoverage = 'covered' | 'uncovered-tron'
+
+/**
+ * Whether the gate can read code on a network.
+ *
+ * @param network - Network name as it appears in `config/networks.json`.
+ * @returns The coverage class.
+ */
+export const resolveGateCoverage = (network: string): GateCoverage =>
```
gvladika, line 55: **Applies if the codehash half ships anyway.**

Tron gets an explicit coverage class, a gap alert and a ticket. zkEVM gets a verdict.

On `zksync`, `abstract` and `lens` this returns `covered`, so the gate goes on to read `out/<Name>.sol/<Name>.json` — the EVM artifact; the zk build lives in `out/zksync` — and applies EVM immutable masking to zksolc codegen. It lands on UNVERIFIABLE because `deriveLineageScope` opens the set for `isZkEVM`, so nothing unsafe happens. But "we could not tell" and "we compared the wrong bytes" reach an operator as the same grey line, and the second one is the kind of thing that stays unnoticed for a year.

Given the file already has the vocabulary for this, an `uncovered-zk` member costs almost nothing and makes the gap legible the same way the Tron one is.

script/deploy/safe/prebroadcast-rederive.ts
```diff
@@ -0,0 +1,238 @@
...
+  const disposition: PreBroadcastDisposition =
+    blockFindings.length > 0
+      ? 'BLOCK'
+      : holdFindings.length > 0
+      ? 'HOLD'
+      : 'PROCEED'
+
+  const reason =
+    disposition === 'PROCEED'
+      ? `re-derived from main: every target's live code matches an attested build and every storage authority matches config${
```
gvladika: **Applies if the codehash half ships anyway.**

This is the sentence an operator reads, and it claims more than the gate checked.

`extractCalldataAddresses` keeps a payload word only when `knownAddresses.has(candidate)`, so a payload address the deployments file does not name produces no target row at all — no finding, no alert, and no trace in this string. Likewise a contract absent from `DECLARED_STORAGE_AUTHORITIES` contributes no authority row, so "every storage authority matches config" is true of the three that have rows and silent about the rest.

The PR body is careful about both:

> Top-level `targets` are added unconditionally, so this is payload-only addresses.
> ... a contract absent from the table produces no row and no finding, so PROCEED is not a statement about its authorities.

Neither qualification survives into what gets printed. Putting the counts in the reason closes the gap without waiting on the decoder work — something like `n of m payload addresses resolved, k authorities read`. A PROCEED that says how much it looked at is much harder to misread than one that says "every".

script/deploy/safe/prebroadcast-gate-placement.test.ts
```diff
@@ -0,0 +1,110 @@
...
+  const end = SOURCE.indexOf('\n}\n', start)
+  expect(end).toBeGreaterThan(start)
+  return SOURCE.slice(start, end)
+}
+
+/** Both spellings of "only when a refusal is binding". */
+const ENFORCING_CHECK =
+  /unverifiedGateOutcome\(process\.env\)|isPreBroadcastGateEnforcing\(process\.env\)/gu
+
+const REFUSING_RETURN = /return '(?:retry|blocked)'/gu
```
gvladika, line 50: Minor, but it affects whether this file catches what it exists to catch.

`REFUSING_RETURN` carries the `g` flag and is passed to `expect(after).not.toMatch(REFUSING_RETURN)` inside the three-iteration loop below. `matchAll` clones the regex so those uses are fine, but if the matcher implementation calls `.test()`, `lastIndex` persists between iterations and a genuine match can read as no-match — the assertion then passes on exactly the revert it is guarding.

Simplest fix is a separate non-global copy for the `toMatch` assertions:

```ts
const REFUSING_RETURN = /return '(?:retry|blocked)'/gu
const REFUSING_RETURN_TEST = /return '(?:retry|blocked)'/u
```

Worth noting the whole file becomes mostly unnecessary if the call site gets the `try`/`catch` in my other comment — the property would then hold by construction rather than by statement order.

0xDEnYO:
@coderabbitai review

script/deploy/safe/execute-pending-timelock-tx.ts
0xDEnYO: Done in `ba397aaea`, in the shape you wrote — the call site catches and falls back to `unverifiedGateOutcome(process.env)`, so totality holds by construction rather than by a promise about another module's default argument.

Mutation check: removing the `.catch` fails `prebroadcast-gate-placement.test.ts` (1 fail / 5 pass), so the property is pinned rather than asserted.

script/deploy/safe/confirm-safe-tx.ts
0xDEnYO: Taken. `confirm-check-registry.ts` is new in `ba397aaea`: the authority observation now lands on the ledger as a `storage-authority` row instead of `recordSignedSet` printing its own block beside it, and the target-state verdict goes through the same mapping.

The anchor is the point you made about proposer-produced values, so each row is anchored on the weaker of the two sides:

| Authority | Expectation from | Anchor | May grade green? |
|---|---|---|---|
| `LiFiDiamond.pauserWallet` | `config/global.json` | `A-LOCAL` | yes |
| `ERC20Proxy.owner` | `config/global.json` | `A-LOCAL` | yes |
| `LiFiDiamond.owner` | deployment record | `A-MONGO` | no — coerced to `error` |

Mutating the `A-MONGO` anchor to `A-LOCAL` fails `confirm-check-registry.test.ts` (3 fail), so the coercion is pinned and not incidental.

One thing your comment made visible that I had not fixed with it: the row was going onto the ledger once per *proposal*, because `recordSignedSet` is called from inside the proposal loop. Since `rollUpChecks` only lets a `fail` block supersession, a later clean proposal would have erased an earlier one's unread authority — on a check whose own header says a check that could not run is not a check that passed. It now goes through `proposalChecks` and the worst-first reduction, the way the target-state row already did (`7f77dcd9a`).

script/deploy/safe/signed-set-record.ts
0xDEnYO: Done — `bySignedSetKey` takes the signer and the index is `(network, operationId, signer)`. `fetchSignedSetRecord` asks `byOperationKey`, the two-field filter, which is all the gate's presence boolean needs.

Dropping `signer` from the key fails `signed-set-record.test.ts` (2 fail).

script/deploy/safe/signed-set-record.ts
0xDEnYO: Index: fixed. `ensureSignedSetIndexes` creates `(network, operationId, signer)` on connect and degrades with a warning on a `createIndex` authorization failure, matching `ensureParkedTasksIndexes` — CI's readWrite-only credentials must still be able to write the audit trail.

Client-per-call: pushing back, because I could not confirm the premise. `getTimelockQueueCollection` (`timelock-queue.ts:183`) and `getParkedTasksCollection` (`parked-tasks.ts:318`) both construct a `MongoClient` per call with the caller closing it — that is the pattern in this directory, not the exception, and `execute-pending-timelock-tx.ts` already makes several such calls per operation. Pooling looks right for the directory as a whole and wrong for this one file, which would then be the odd one out. Happy to take it as a follow-up across all of them if you read it differently.

script/deploy/safe/prebroadcast-gate.ts
0xDEnYO: Agreed, and fixed by making the same flag gate the alerting: `mayAlert()` now requires `isPreBroadcastGateEnforcing(process.env)`, so shadow mode logs and sends nothing. Turning enforcement on turns the alerting on with it.

That covers the gap alert and the shadow-refusal alert, not just the codehash volume — output nobody may act on does not belong on the channel carrying output somebody must.

Mutation check: dropping the enforcing term from `mayAlert` fails the placement test.

script/deploy/safe/prebroadcast-gate.ts
0xDEnYO: Taken — the `hashOperationBatch` re-derive is gone and `readScheduledAt` is `getTimestamp(id)`, for the same one call. `viemOperationIdReader` no longer exists.

It does not make the broadcast safe on its own, since `executeBatch` enforces `isOperationReady` regardless; what it buys is turning that revert into a pre-flight refusal an operator can act on. Mutating the zero-timestamp branch so it no longer blocks fails `prebroadcast-rederive.test.ts`.

script/deploy/safe/prebroadcast-anchors.ts
0xDEnYO: Taken — the codehash half is cut. `prebroadcast-anchors.ts` and its test are deleted, along with `readArtifactAnchor`, `deriveLineageScope` and the local `normalizeRuntimeCode` copy.

The sequencing argument is the one that decided it: three machines sharing one oracle is one observation, and the check was fully correlated with exactly that oracle. What remains is the sign-time record plus the storage-authority check — the only subject in the delay window that can actually change.

The re-land is [EXSC-1004](https://linear.app/lifi-linear/issue/EXSC-1004/wp-61b-re-land-the-pre-broadcast-codehash-check-as-an-attestation), blocked on EXSC-952 (#2372). It carries this and four more of your findings so the re-land starts from them rather than repeating them.

script/deploy/safe/prebroadcast-anchors.ts
0xDEnYO: The copy is deleted with the rest of the codehash half, so nothing to fix here — but the finding is the clearest argument for the cut, so it is carried as finding 1 on [EXSC-1004](https://linear.app/lifi-linear/issue/EXSC-1004/wp-61b-re-land-the-pre-broadcast-codehash-check-as-an-attestation), including the reversed order, the missing `isZk` branch and the missing `frameFault` validation. The re-land imports the canonical function.

script/deploy/safe/prebroadcast-anchors.ts
0xDEnYO: Deleted with the codehash half. Carried as finding 2 on [EXSC-1004](https://linear.app/lifi-linear/issue/EXSC-1004/wp-61b-re-land-the-pre-broadcast-codehash-check-as-an-attestation) — the re-land uses `deriveToolchainScope` rather than a second derivation — along with your point that on the five `london` networks the local default-profile artifact was the wrong anchor, not merely an open set.

script/deploy/safe/prebroadcast-gate.ts
0xDEnYO: Moot as raised, and worth saying why rather than just closing it: the gate no longer reads an artifact or masks anything. `observeTarget` hashes exactly the bytes it read from the chain, with no comparison — so there is no longer a path on which zksolc codegen meets EVM immutable masking.

The `covered | uncovered-tron` vocabulary stays as is for that reason. The zkEVM coverage class is carried as finding 4 on [EXSC-1004](https://linear.app/lifi-linear/issue/EXSC-1004/wp-61b-re-land-the-pre-broadcast-codehash-check-as-an-attestation), where the comparison comes back and the distinction starts to matter again.

script/deploy/safe/prebroadcast-rederive.ts
0xDEnYO: Fixed, in the shape you suggested. `describeCoverage` is spelled out on every disposition, PROCEED included:

`N of M calldata address(es) resolve to a contract main names, and K declared storage authority value(s) were read`

Mutating the reason so it drops the counts fails `prebroadcast-rederive.test.ts` (3 fail).

The coverage gap itself — a facet a rollout is installing is not in the deployments file yet, so it produces no row — is finding 5 on [EXSC-1004](https://linear.app/lifi-linear/issue/EXSC-1004/wp-61b-re-land-the-pre-broadcast-codehash-check-as-an-attestation). This round closes the reporting half only.

script/deploy/safe/prebroadcast-gate-placement.test.ts
0xDEnYO: Fixed with the separate non-global copy — `REFUSING_RETURN_TEST`.

Reporting the verification honestly, because it came out in two parts. The mechanism is real and I confirmed it directly: in bun, `toMatch` on a `g`-flagged regex advances `lastIndex`, and a second `.not.toMatch` against a string that genuinely matches then passes. So the hazard you describe exists.

What I could not do is make it produce a false green *in this file*. The poisoning needs an earlier catch block that also matches, and that earlier block fails the assertion first, so the test aborts before a later missed match can hide anything — I built the arrangement designed to exploit it (a bare refusal late in catch 1, early in catch 2) and both the old and the fixed regex fail. Latent hazard removed, not a live false-green closed.

On the file becoming unnecessary: the call site now has the `try`/`catch`, so you are right that the property holds by construction. I have kept the file because it still pins the ordering property — that no refusal precedes an enforcing check — which the `catch` does not give us. Mutating the call site to stop catching does fail it.

0xDEnYO:
@coderabbitai review

0xDEnYO:
@coderabbitai review

gvladika (dismissed):

script/deploy/safe/confirm-safe-tx.ts
```diff
@@ -362,6 +387,100 @@ const processTxs = async (
...
```
melianessa, line 406: `recordSignedSet` builds a fresh viem client per proposal and, without `--rpcUrl`, falls back to viem's bundled public endpoint instead of the run's configured RPC.

script/deploy/safe/prebroadcast-gate.ts
```diff
@@ -0,0 +1,474 @@
...
+  readScheduledAt: () => Promise<bigint>
+}
+
+export interface IGateOperation {
+  operationId: string
+  targets: readonly string[]
+  payloads: readonly string[]
+}
+
+const describeError = (error: unknown): string =>
```
melianessa, line 88: viem RPC error messages embed the full node URL including the provider API key, and `describeError` writes them verbatim into `readError`/`observationError`, which are persisted to MongoDB, printed, and posted to Slack under enforcement.

0xDEnYO: Confirmed and fixed in 3114ab93e / 95feabc42 — and it was worse than the write path: the raw message also reached `buildShadowRefusalAlert`, so under enforcement the key would have gone to the escalation channel, not just to Mongo.

Split across the two helpers per [CONV:REDACT-RPC-URL], because `describeError` feeds both a durable record and a Slack line:

- `describeError` now returns `redactUrls(...)` — redacted but uncapped. `redactErrorReason`'s 180-char cap is sized for a Slack line and lands mid-`Details:` on a real rate-limit error, which would have quietly eaten the cause from the forensic record.
- `buildGateGapAlert` and `buildShadowRefusalAlert` apply `redactErrorReason` per bullet, so the capping happens where the text is published.

Three tests pin it (unreadable target, unreadable authority, and the published verdict a refusal alert is built from); all three fail on the pre-fix code with `dkey=…` present.

The sweep turned up three more sinks of the same value that the gate routes into, now closed in 95feabc42: `notifyOperationFailed` renders whatever it is handed straight into the Slack block with no redaction, so the executor's `alertFailure` and its broadcast-failure catch both published raw viem/Mongo errors — the latter is live today, not gated behind the enforce flag. Extracted as `guard-error-redaction.ts` with its own tests (100% cov) rather than inlined, so the two sinks are a one-line call each.

script/deploy/safe/confirm-safe-tx.ts
0xDEnYO: Pushing back on the public-endpoint half, and I checked it rather than assuming.

`getViemChainForNetworkName` throws unless the network's RPC env var is set (`viemScriptHelpers.ts:314-317`), and viem's `http()` with no URL uses `chain.rpcUrls.default.http[0]` — which is that same env-derived endpoint. So `rpcUrl ? http(rpcUrl) : http()` was already the run's configured RPC in both branches; verified empirically that `http(undefined)({chain}).value.url === http()({chain}).value.url === <the ETH_NODE_URI-derived primary>`. There is no viem-bundled fallback to reach here, so a warning would fire on a condition that cannot occur.

The fresh-client-per-proposal half stands, and the ternary is what made it read as a fallback. Swapped for `buildReadOnlyClient(networkKey, rpcUrl)` in 3114ab93e — the same helper every other read-only Safe query goes through, byte-identical behaviour in both the override and no-override cases (only difference is the `as PublicClient` cast `viemGateReaders` needs), with the non-obvious part now stated in one comment instead of implied by a ternary.

Pooling the clients themselves is real and directory-wide — `getTimelockQueueCollection` and `getParkedTasksCollection` have the same shape — tracked as EXSC-1008 and now named in the docblock.

0xDEnYO:
All four addressed, one with a correction. `3114ab93e` + `95feabc42`.

**1 — API key leakage. Fixed, and the blast radius was larger than reported.**

The raw message also reached `buildShadowRefusalAlert`, so under enforcement the key went to the escalation channel, not only to Mongo.

Split across the two helpers per `[CONV:REDACT-RPC-URL]`, which distinguishes them: `describeError` now returns `redactUrls(...)` — redacted, uncapped — because the string is persisted as the sign-time record, and `redactErrorReason`'s 180-char cap is sized for a Slack line and lands mid-`Details:` on a real rate-limit error, dropping the cause from the forensic artifact. Capping moved to the publish boundary: `buildGateGapAlert` and `buildShadowRefusalAlert` apply `redactErrorReason` per bullet.

Three tests pin it (unreadable target, unreadable authority, published verdict); all three fail on the pre-fix code with `dkey=…` present.

Sweeping the same value found three more sinks the gate routes into, closed in `95feabc42`: `SlackNotifier.extractErrorMessage` returns `errorObj.message` raw and renders it into the Slack block, so `alertFailure` in both pre-execute guards and the broadcast-failure catch in `executeOperation` published unredacted viem/Mongo errors. The last one is live today — it is not behind `PRE_BROADCAST_GATE_ENFORCE`. Extracted as `guard-error-redaction.ts` with its own tests at 100% coverage.

Not fixed here: six older sinks elsewhere in `execute-pending-timelock-tx.ts` (notification-send failures, per-network catch, parked-queue open). Same class, none introduced or routed into by this PR; a file-wide sweep belongs in its own change.

**2 — Public RPC fallback. The premise does not hold; fixed the thing that does.**

`getViemChainForNetworkName` throws unless the network's RPC env var is set (`viemScriptHelpers.ts:314-317`), and viem's `http()` with no URL resolves to `chain.rpcUrls.default.http[0]` — that same env-derived endpoint. Verified: `http(undefined)({chain}).value.url === http()({chain}).value.url === <the ETH_NODE_URI-derived primary>`. So `rpcUrl ? http(rpcUrl) : http()` was the run's configured RPC in both branches, there is no public endpoint to fall back to, and an auditor cannot be looking at a public-endpoint observation. A warning would fire on a condition that cannot occur, and requiring `--rpcUrl` would reject a run that is already correct.

What was real is that the ternary read as a fallback — to two reviewers. Replaced with `buildReadOnlyClient(networkKey, rpcUrl)`, the helper every other read-only Safe query uses; byte-identical behaviour in both cases.

**3 — MongoClient per call. Ticket filed, docblock points at it.**

[EXSC-1008](https://linear.app/lifi-linear/issue/EXSC-1008/pool-mongoclient-handles-across-scriptdeploysafe-instead-of-one-client), scoped directory-wide rather than to this file — `getTimelockQueueCollection` and `getParkedTasksCollection` have the same shape, and fixing one of the three makes it the odd one out for no measurable gain.

**4 — Foundry pin. Fixed, and it is not a no-op.**

Now `uses: ./.github/actions/setup-foundry`. History: the direct `foundry-toolchain@v1.6` step landed in `05e34c476` three days *before* `1fe19fdf1` introduced the composite action and the `[CONV:FOUNDRY-SETUP]` rule, and that sweep never touched this file — a missed migration, not a carve-out. After this commit `foundry-rs/foundry-toolchain` appears nowhere in `.github/` outside the composite action.

Worth stating plainly since it changes what runs: the replaced step installed an unpinned toolchain at v1.6; the composite installs the `.foundry-version` pin (1.7.1). That is the point of the rule, but it is a version change on the executor cron, not a refactor. It also adds three cache restore/save steps to a job that runs `bun typechain` (= `forge clean && forge build src`), which discards the restored `out/` immediately — cost is restore time every tick, no correctness effect, and the cache key is namespaced by workflow so it cannot poison others.

gvladika (approved):

melianessa (approved):

PR #2354, opened 2026-09-09
feat(deploy): record the deployed codehash, and assert commit presence not ancestry (EXSC-959)

0xDEnYO:

Fixes [EXSC-959](https://linear.app/lifi-linear/issue/EXSC-959) — WP-2.2 residual: codehash on the deployment record + commit-presence verifier. Parent [EXSC-686](https://linear.app/lifi-linear/issue/EXSC-686).

Closes the last two of EXSC-907's five WP-2.2 items. The other three had landed elsewhere and were verified against `origin/main` before starting, not taken from the ticket: `repo` on `IDeploymentRecord` is at `mongo-log-utils.ts:63` (#2318), `.foundry-version` is checked on the deploy path (#2325 / EXSC-932), and `record-codehash.ts` / `commit-presence.ts` are absent on `main` with `git grep codehash` returning zero hits across the record path — so both items here were genuinely unbuilt.

**Presence, not ancestry.** `origin/main` is squash-merged, so a commit that was genuinely built and deployed is an ancestor of nothing once its PR lands. Measured over the 99 unique commit SHAs in `audit/auditLog.json` — the real record of commits we deployed and audited — the old predicate `git branch -r --contains <sha> --list 'origin/*'` refuses **75**; in a `--single-branch` clone that only tracks `origin/main`, it refuses **79**. That spread is the actual defect: the answer depends on which refs the checkout happens to hold, so it is a fact about the clone rather than about the repository. Every one of those commits is `PRESENT` when the repository is asked.

The local refs are therefore read as a **positive only** — a remote-tracking ref holding the commit means this clone fetched it from that remote, which settles presence offline and for free, and is the only path the happy case takes. Everything the local read cannot settle asks the declared repository through `gh api`, read-only, inheriting the deployer's own credentials and rewriting no refs. `--hostname` is passed explicitly because `GH_HOST` applies wherever a hostname was not given, and an exported `GH_HOST` would otherwise answer about a different server's repository of the same name.

**Nothing that used to pass can now fail.** Before this change, empty local refs were a hard refusal with no query at all. Now that same state gets a query whose `PRESENT` answer lets the deploy through. The set of refused deploys is a strict subset of what it was, so the new `gh` dependency lives entirely inside a case that previously always failed — which is why it is a relaxation and not a new gate. `gh` missing, unauthenticated, or pointed at a non-`github.com` remote lands on `UNKNOWN`, and `UNKNOWN` refuses exactly as before.

Only `PRESENT` proceeds. `UNKNOWN` is this check failing to get an answer, and reading it as a pass would turn an unreachable API into a green light; `ABSENT` is reserved for the API's own "no commit found" (HTTP 422). A `404` grades `UNKNOWN` deliberately — it covers both a repository that does not exist and one the token cannot see, neither of which is an answer about the commit.

Two consequences of the offline positive are worth a reviewer's attention, because both are things it does **not** establish. It does not establish that the record names the remote the ref came from: where the remote URL cannot be resolved to an identity at all, presence rests on the local ref alone. And it is a historical fact — a branch force-pushed or gc'd upstream leaves a ref that still contains the commit. The message, `ITreeState`'s docs and `docs/Deploy.md` all state what was established rather than the stronger claim, and a refusal prints the verdict's own reason. Whether an unresolvable declared repository should refuse a production deploy outright is a separate question from this PR: it is unchanged behaviour, and tightening it would be a new refusal on the deploy path.

One deliberate loosening: the old refusal carried the caveat "a commit carried only by a tag does not count", which was an artifact of `git branch -r --contains` being branch-only. The commits API answers `PRESENT` for a tag-reachable commit, which is correct under the D3 standard this is built on — present and fetchable is the whole requirement.

**The codehash group is four values or none.** The metadata trailer carries its own length word, so equal masked hashes mean equal code only with the deployed length pinned alongside them; a masked hash stored without it is exactly the shape that let appended bytes normalise to an honest prefix. A partly-passed group is refused rather than stored.

**It reports; it never decides.** The run that writes the record chose the bytes it hashed, so a later check has to recompute from the chain and use this field only to say what was observed at deploy time and by whom. Nothing reads it today — the sign-time gate rebuilds from the commit instead, as it must.

**Refusal ordering, and why the exit code is zero.** `add` runs *after* the deploy, so refusing over a codehash it cannot accept would trade a record missing one field for a broadcast deployment with no record at all. The command writes the rest of the record, prints why the codehash was dropped **on stdout**, and exits **zero** — `logContractDeploymentInfo` discards this command's stderr unless `DEBUG` is set and reads any non-zero status as "the record did not land", so an error here would abort a deploy over a field it never asked for, with the reason hidden. The ordering and the zero exit are what the tests assert, not just the message.

**The seam is not wired yet.** No deploy path passes the four new flags, so records written today carry no codehash, and `codehashFromSelfCheck` has no production caller. `docs/MultisigSigningProcess.md` says so rather than implying the linkage is live. The flags are the seam the post-deploy self-check will write through; the CLI tests spawn the real command so the seam is proven reachable rather than merely present. `evaluateDeploySelfCheck` is equally unwired on `main` — this package is landing module by module, with the wiring to follow.

**Falsification.** Sixteen non-equivalent single edits to the production files, each applied alone and restored byte-identically, all sixteen caught by a named test: dropping the `codehash` key from the upsert, letting an `undefined` codehash reach it, relaxing the digest regex, `parseInt` for the byte counts, removing the four-or-none check, accepting a zero length, dropping `Number.isSafeInteger`, widening the self-check outcome set, letting `UNKNOWN` presence proceed, grading a `404` as `ABSENT`, reading empty local refs as `PRESENT`, abbreviating the SHA regex, loosening the bash staging match, dropping the flag prefix from a refusal, and removing the git timeout.

Two of them are only caught by one specific leg each. `codehash: undefined` in the upsert is invisible through the dry-run seam, because JSON erases an undefined value — measured against this repo's own driver (mongodb 6.17.0, `ignoreUndefined` resolving to `false`), it would be written as a **null** and erase what a run that did observe the code had stored; only the own-key assertions on `buildDeploymentUpsert` catch it, and the CLI-level test alone lets it survive. The git timeout is only observable through a `git` stub that never returns, which without the timeout runs until the case times out.

Against the real repository rather than a fixture: a fabricated SHA returns `ABSENT` end-to-end with the CLI exiting 1, a real commit returns `PRESENT`, a repository the token cannot see returns `404` → `UNKNOWN` → refusal, and with `gh` stubbed to log its own invocations the happy path never calls it.

Neither `deployments/<network>.json` nor `deployments/_deployments_log_file.json` is read anywhere in this change; MongoDB is the record.


0xDEnYO:
@coderabbitai review

0xDEnYO:
@coderabbitai review

0xDEnYO:
@coderabbitai review

0xDEnYO:
@coderabbitai review

gvladika (approved):

PR #2355, opened 2026-09-09
feat(codehash): bridge an audit to the build that was deployed (EXSC-876)

0xDEnYO:

- Fixes [EXSC-876](https://linear.app/lifi-linear/issue/EXSC-876/r210-bridge-the-audit-record-to-the-build-that-was-deployed) · R2.10

#2351 has merged, so `sourceClosureHash` is on `main` and this branch adds only the bridge.

## The gap, verified rather than assumed

`audit/auditLog.json` holds 104 audits with five fields — `auditCompletedOn`, `auditedBy`, `auditorGitHandle`, `auditReportPath`, `auditCommitHash` — and **nothing about the compiler**. `auditedContracts` maps 109 contracts by version to audit ids, also compiler-free. So a contract audited under one compiler and deployed under another is invisible in the log today, and that is routine rather than exotic: facets ship under the `solc_floor` profile (0.8.17/london) while their networks declare cancun.

## Three moves, all forward-only

Per the ruling: no fleet re-audit, no redeploy.

**Audits stay version-agnostic.** An audit is of source, and a different compiler does not invalidate it. What needs review is the *compiler set*, vetted once per profile against solc and zksolc advisories — so `VETTED_COMPILER_SETS` is the review surface, and editing it **is** the review event.

That map is deliberately **not** derived from `foundry.toml`. Reading the profile's current pin would make it agree with whatever the repo happens to pin today, which is the opposite of a review record: bumping the pin would silently bless the new compiler. A vetted profile whose versions have since moved therefore reports rather than passing — the profile name agreeing is not the compiler set agreeing.

**`sourceClosureHash` carries the other half**, proving the build compiled the audited source over the full transitive closure. The closure and not the file is the unit because a swapped library changes the closure without changing the facet file — the `AcrossV4SwapFacet`/`LibAsset` case. **No audit records a closure today** — the field is not in `audit/auditLog.json`, and `/add-audit` does not extract one — so on the fleet as it stands the answer is always that the bridge *cannot be proved*. That is its own note rather than being conflated with the two closures disagreeing, and it is the half of this module that only becomes informative once the audit side records a closure.

**An unchecked half is neither a match nor a mismatch.** Nothing outside a bytecode trailer records the zk fork and LLVM versions, so a zk build assembled from the deployment record carries no zk section — and calling that a mismatch would brand every zk deploy while calling it a match would bless a toolchain nobody checked. It reports `zk-toolchain-unverified` instead. For the same reason a closure is validated as a whole keccak digest before it is compared: two empty strings are equal, so an empty or sliced hash on both sides otherwise came back with no notes at all, which is a proof of nothing rendered as a clean bill.

**Everything is a NOTE and the return type carries no blocking state at all.** A per-deploy version gate would brand every london-chain deploy unaudited, which is a false red on ordinary work. The failure this module must avoid is refusing honest deploys; refusing dishonest ones is the codehash gate's job. Notes accumulate rather than short-circuit, so a build that is both unaudited and built under an unvetted compiler says both — one note hiding another is how a signer comes to believe the lesser problem is the only one.

## Falsification

Six behaviours reverted one at a time, each failing only its own tests, the tree restored byte-identical and the file re-checked as valid text afterwards:

| Mutation | Tests that failed |
| --- | --- |
| the no-audit note dropped | 2 |
| an unvetted profile accepted | 3 |
| version drift under a vetted profile ignored | 1 |
| the zk section not compared | 2 |
| an absent zk section treated as a match | 1 |
| a closure disagreement muted | 1 |
| closures compared without `normalizeHash` | reads one closure in two hex spellings as one closure |
| versions compared byte-for-byte | does not report a compiler set that differs only in case or whitespace |
| a profile read straight off the object | treats a profile named after an Object prototype member as unvetted |
| the mismatch detail dropping its zk half | names the zk toolchain in the mismatch detail |
| the vetting dates rewritten | names the vetting date in the mismatch detail |
| the compiler compare skipped when the zk half is unreported | reports a zk build whose toolchain nothing recorded, without calling it a mismatch |
| the zero digest accepted as a closure | does not read two unusable closures as a proved bridge |
| the zk versions compared strictly | does not report a compiler set that differs only in case or whitespace |
| an unreported zk toolchain called a mismatch | reports a zk build whose toolchain nothing recorded, without calling it a mismatch |
| a reported zk toolchain never compared | names the zk toolchain in the mismatch detail |
| the closure digest check dropped | does not read two unusable closures as a proved bridge |

`bunx tsc-files --noEmit` and `bunx eslint` clean; `bun test script/deploy/codehash/audit-build-bridge.test.ts` green.

## Not claimed

No call site yet — this is the bridge as a pure module, and the signer-facing render belongs with whichever surface consumes it. `isFullyBridged` is gone with this round: nothing called it, and with no audit recording a closure it was false for all 742 attested builds, so a single marker could only mislead. The note list is the whole surface.

Two fields no deployment record carries are the caller's to derive, and both are documented as such on `IDeployedBuild`: `profile`, and an `evmVersion` read from the profile rather than from the deploy log's `EVM_VERSION` — the log says `zkevm` where the profile says `cancun`, and this module does not translate between the two vocabularies. The vetting dates say a human read the advisories for those versions; I set them to today because the versions are the ones this repo currently pins, and that claim is the one thing here a reviewer has to actually check rather than read. `zksync.zk.solcForkVersion` (0.8.29) and `llvmVersion` (1.0.2) need the same check and get less help from the repo: neither value appears in `foundry.toml` or any config file, so nothing but a bytecode trailer can corroborate them.


0xDEnYO:
@coderabbitai review

0xDEnYO:
@coderabbitai review

0xDEnYO:
Added the two mismatch cases that were missing: the suite only ever drifted `llvmVersion`, so `zksolcVersion` and `solcForkVersion` were compared but unpinned. Now a table over the three, each verified to fail when its own comparison is removed.

On the signer detail: `describeCompilerSet` renders both zk tuples when either side has one, and the vetted-has-zk / build-has-none direction is no longer a mismatch at all — it reports `zk-toolchain-unverified`, since nothing outside a bytecode trailer records the fork and LLVM versions.

0xDEnYO:
@coderabbitai review

gvladika (approved):

PR #2356, opened 2026-09-09
feat(safe): route every Safe proposal through one blessed wrapper (EXSC-957)

0xDEnYO:

Fixes [EXSC-957](https://linear.app/lifi-linear/issue/EXSC-957) — WP-7.5 of
[EXSC-686](https://linear.app/lifi-linear/issue/EXSC-686) (Multisig Signing Process 2.0),
acceptance row A7.8. The docs half of R6.5 landed in #2126; this is the funnel half.
Residue tracked by [EXSC-984](https://linear.app/lifi-linear/issue/EXSC-984).

**One seam instead of nine.** `proposeSafeTx()` (`script/deploy/safe/propose-safe-tx.ts`) is
now the only thing that turns a call into a stored Safe proposal: it checks Safe ownership,
signs, hashes **the signed transaction**, and stores. Nine call sites move onto it —
`propose-to-safe.ts`, the TypeScript `sendOrPropose`, the five chain-id-mapping /
registration tasks, `unpauseAllDiamonds.ts` and `add-safe-owners-and-threshold.ts` — each of
which previously repeated the build/sign/hash/store sequence and its own reading of the
store's return value.

The seam is only worth having if the set of paths through it is closed, which is what the
fence is for. Before this, that set was a convention plus a grep.

**The fence is structural, not textual.** `.eslintrc.funnel-fence.cjs` is
`no-restricted-syntax` keyed on the AST — `Identifier[name='storeTransactionInMongoDB']`,
plus `Literal[value=…]` and `TemplateElement[value.cooked=…]` for a computed lookup. Not
`no-restricted-imports`: an import restriction sees only `import` / `export … from`, so a
namespace member access (`safeUtils.storeTransactionInMongoDB`), a re-export, an alias, a
dynamic import or a bracket lookup would reach the funnel unflagged. Not a text scan
either: a mention in a comment is not a call site, and a file scan cannot tell the two
apart. Every one of those shapes has a test (`propose-funnel-fence.test.ts`), including
the negatives — a comment mention and a compliant call site both pass.

Two shapes the first cut of the fence missed, both now closed and both tested by planting
a real file and running the CI command: a file-level `/* eslint-disable
no-restricted-syntax */`, which the standalone run now refuses because it passes
`--no-inline-config`; and a route written at any extension outside `.ts`/`.js`/`.tsx`,
which the directory sweep did not visit at all — `.mjs` is live in this tree
(`script/claude/install.mjs`), and the repo-wide lint globs miss it too, so such a file
was linted by nothing. `lint:funnel` now sweeps `.ts,.tsx,.mts,.cts,.js,.mjs,.cjs`, which
also makes the fence see its own config, so `.eslintrc.funnel-fence.cjs` joins the
allowlist as the file that names the funnel to build the rule. A test asserts the swept
extensions cover every module extension actually present under `script/`, measured off the
tree rather than off a list in the test.

**It is named by what may proceed.** The rule denies the name everywhere; a five-entry file
allowlist in `overrides` is the closed positive set, and a test asserts its length and
membership so it cannot be widened unnoticed. A propose route added next month is
refused by default, and making it lint-clean means editing the allowlist — a reviewable act
with a reason and a ticket attached, not an omission. The entries are exact paths, not globs.

**Only one route is still exempt.** `propose-to-safe-tron.ts` hand-rolls its own signature
instead of signing through a `SafeClient`, which is what the wrapper signs with, so it cannot
pass through it as written — [EXSC-984](https://linear.app/lifi-linear/issue/EXSC-984).
`add-safe-owners-and-threshold.ts` needed no exemption: `proposeSafeTx` takes a prebuilt
`safeTx` with a caller-owned nonce, which is exactly what `createAddOwnerTx` and
`createChangeThresholdTx` produce, so it is migrated here rather than grandfathered.

**Where it still fails open, stated plainly.** Two shapes are out of reach and both are noted
in the config: a name assembled by concatenation, which no static selector can see; and — the
one that matters — the fence covers the funnel's storage *function*, not a hand-rolled
`insertOne` into the `pendingTransactions` collection that never names it. Closing that needs
a type-aware rule; the read-only reporting scripts legitimately open the same collection by
name, so widening the fence to the collection name would refuse them and train people to grow
the allowlist — the exact erosion this rule exists to prevent.

**Three behaviour changes worth a reviewer's eye:**

- `unpauseAllDiamonds.ts` hashed the **unsigned** transaction
  (`getTransactionHash(safeTransaction)` where the signed value was `signedTx`). Going through
  the wrapper fixes that; the stored hash and the stored signature can now only describe the
  same bytes.
- The owner check in the TypeScript `sendOrPropose` moved from before the Mongo connection to
  inside the wrapper, so its refusal message is now the wrapper's. The connection is closed on
  that path.
- `add-safe-owners-and-threshold.ts` now pays one extra `getOwners()` per proposed transaction
  (the wrapper's own check, on top of the script's up-front one). It runs once per Safe;
  `propose-to-safe.ts` keeps its earlier check too, deliberately, because it fires before the
  drain claims any parked task.
- The TypeScript `sendOrPropose` and `add-safe-owners-and-threshold.ts` used to print the
  Safe tx hash immediately after signing, so it appeared even when storage then failed.
  Both now print it after `proposeSafeTx` returns, so a nonce collision or a rejected
  insert reports the error without the hash. Kept: the hash of a proposal that was never
  stored identifies nothing the error message does not, and reconstructing it at the call
  site would mean rebuilding the transaction outside the seam.

**A missing check is not a green one:** `.github/workflows/enforceProposalFunnel.yml` is
guarded by the `protect-critical-code` required check, which only evaluates on
`pull_request_review: submitted`. It will show as pending until someone reviews — that is
expected here, not a failure.

**Two gaps in the enforcement around the fence, neither of which this PR closes.** Both are
repo settings rather than code, so they are called out rather than changed here:

- `enforce-proposal-funnel` is **not** in the `main protection` ruleset's required checks,
  so a red fence does not block a merge today. The job has no path filter and no skip
  condition, so it is safe to require directly.
- `protect-critical-code` matches `^\.github/|^\.husky/pre-commit|^script/emergency/|…`,
  which does **not** match `.eslintrc.cjs`, `.eslintrc.funnel-fence.cjs` or `package.json`.
  The fence's rule, the `extends` line that wires it in, and the `lint:funnel` script can
  therefore each be weakened with one ordinary approval — the same "a reviewer might
  notice" state this PR replaces everywhere else. Adding `^\.eslintrc` to that regex
  mirrors the immutable-registry precedent (#2308), but it changes what needs ISM/CTO
  sign-off repo-wide, so it is a call for the reviewer rather than a drive-by edit here.

**Draft CI is not evidence.** `run-ts-tests` and `deploy-smoke-test` skip on drafts while
their `*-required` aggregators report green, so `ts-tests-required` passing in 3 s on the
draft did not run either new suite. `enforce-proposal-funnel` (~29 s) is the only guard
that actually executed before this PR left draft.

`docs/MultisigSigningProcess.md` §4.1, §4.2, §5 and §8 are updated: §4.2 said "there is **no
single chokepoint**" and pointed at a `grep -rn storeTransactionInMongoDB` for the route list,
both of which this PR makes false. `docs/DeferredDiamondCleanupQueue.md` facts 3 and §6 said
the storage function is called from ~9 sites rather than one wrapper — also now false.

- [ ] For new facets: n/a — no Solidity in this PR

- [ ] I have ensured that any new contracts have had AT A MINIMUM 1 preliminary audit conducted

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -21,6 +21,8 @@ module.exports = {
     'plugin:import/recommended', // provides the base import rules
     'plugin:import/typescript', // provides the typescript-specific import rules
     'prettier',
+    // Safe-proposal funnel fence; also run on its own by `bun lint:funnel`
+    './.eslintrc.funnel-fence.cjs',
   ],
   rules: {
     // General JavaScript rules
--- a/.eslintrc.funnel-fence.cjs
+++ b/.eslintrc.funnel-fence.cjs
@@ -33,6 +33,9 @@ const FENCE = [
   // that had only the two selectors above.
   { selector: `Literal[value='${FUNNEL}']`, message: MESSAGE },
   { selector: `TemplateElement[value.cooked='${FUNNEL}']`, message: MESSAGE },
+  // Not covered, deliberately: a name assembled by concatenation. No static
+  // selector can see it, and unlike the three above it is not a rewrite anyone
+  // reaches for by accident.
 ]
 
 module.exports = {
--- a/.eslintrc.funnel-fence.cjs
+++ b/.eslintrc.funnel-fence.cjs
@@ -13,7 +13,11 @@
  *
  * Extended by `.eslintrc.cjs` so a commit is checked by lint-staged, and run on
  * its own by `bun lint:funnel` (and `.github/workflows/enforceProposalFunnel.yml`)
- * so the fence does not depend on repo-wide lint being green.
+ * so the fence does not depend on repo-wide lint being green. That standalone run
+ * passes `--no-inline-config`, without which a file-level `eslint-disable` would
+ * turn the rule off, and sweeps every module extension the repo can hold — the
+ * repo-wide globs stop at `.ts`/`.js`/`.tsx`, so a `.mjs` or `.cts` route would
+ * otherwise be linted by nothing at all.
  */
 
 const FUNNEL = 'storeTransactionInMongoDB'
@@ -32,9 +36,11 @@ const FENCE = [
   // and as a template literal it is neither.
   { selector: `Literal[value='${FUNNEL}']`, message: MESSAGE },
   { selector: `TemplateElement[value.cooked='${FUNNEL}']`, message: MESSAGE },
-  // Not covered, deliberately: a name assembled by concatenation. No static
-  // selector can see it, and unlike the three above it is not a rewrite anyone
-  // reaches for by accident.
+  // Not covered, deliberately: a name assembled by concatenation, which no
+  // static selector can see, and an alias re-exported from an allowlisted file,
+  // which carries the funnel to a consumer under a name this rule never reads.
+  // Neither is a rewrite anyone reaches for by accident; the allowlisted files
+  // export no such alias today.
 ]
 
 module.exports = {
@@ -51,6 +57,8 @@ module.exports = {
   overrides: [
     {
       files: [
+        // names it to build the rule
+        '.eslintrc.funnel-fence.cjs',
         // declares it
         'script/deploy/safe/safe-utils.ts',
         // the blessed wrapper: the only caller
--- a/.eslintrc.funnel-fence.cjs
+++ b/.eslintrc.funnel-fence.cjs
@@ -1,0 +1,70 @@
+/**
+ * Safe-proposal funnel fence.
+ *
+ * `script/deploy/safe/propose-safe-tx.ts` is the blessed way to create a Safe
+ * proposal. Every check the signing process sites "in the funnel" only covers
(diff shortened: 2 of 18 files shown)


0xDEnYO:
@coderabbitai review

0xDEnYO:
@coderabbitai review

0xDEnYO:
@coderabbitai review

0xDEnYO:
@coderabbitai review

0xDEnYO:
@coderabbitai review

gvladika (approved):

melianessa (approved):

PR #2362, opened 2026-09-09
feat(safe): re-land the signer-display routes with #2349's open holes closed (EXSC-951, EXSC-949)

0xDEnYO:

Fixes [EXSC-951](https://linear.app/lifi-linear/issue/EXSC-951) — four routes into the signer's terminal left open by EXSC-948
Fixes [EXSC-949](https://linear.app/lifi-linear/issue/EXSC-949) — the decoded-calldata display, one screen above the sanitised block

## What it does

Every value the signing prompt shows off a MongoDB proposal row is text the proposer wrote, and the rows are not all written by this repository. EXSC-948 closed the detail block; five routes past it stayed open. This adds one primitive, `script/deploy/safe/printable-field.ts`, and puts all of them through it.

The routes closed:

| Route | Where |
|---|---|
| Ledger Flex filmstrip renders the stored calldata unsanitised | `ledger-flex-preview.ts` |
| `safeTxHash` echoed raw in a throw after signing | `confirm-safe-tx.ts` |
| `scheduleBatch` predecessor / salt / delay / selector / payload | `safe-decode-utils.ts` |
| `decodeDiamondCut`'s 4byte-resolved function name | `safe-utils.ts` |
| decoded `string` ABI arguments, incl. inside tuples | `safe-decode-utils.ts` |

Three properties beyond stripping control characters, each reachable through a row that needs no escape sequence at all:

- **Length.** A 500,000-character `safeTxHash` is a valid row shape; the line wraps for thousands of terminal rows and pushes the proposal off the top, leaving the prompt. Every stored field **in the three modules this PR puts on the primitive** carries a bound, so does every decoded argument, and the element count of `parkedTaskRefs` is bounded too. Two exceptions, one deliberate and one not: the `Data:` field itself, by design (see below), and the provenance block's `proposerHandle` / `gitBranch` / `prUrl`, which are stored row text rendered in the same prompt and are **not** bounded — that block is EXSC-986 and is named under "Not in this PR".
- **Invisibles.** `sanitizeProvenanceText` keeps U+200D and the Hangul fillers by design, so two values differing only by those printed identically. Counted and disclosed.
- **Confusables.** A Cyrillic `о` in a facet name is glyph-identical to the ASCII one. No repair is possible without changing which value it is, so the count is reported. Every field in the prompt is ASCII by construction, which is what makes a bare non-ASCII count precise enough to act on.

Nothing is cleaned quietly: each repair yields a `⚠` notice naming what it was. A signer who cannot see that a field was repaired cannot tell a normal proposal from a hand-edited one.

## EXSC-951 asks which way the funnel is closed. **By construction.**

`Printable` is a branded string; `color`, `concatPrintable` and `detailLine` accept nothing else, and only `asPrintable` and `trustedMarkup` produce one. The intentional exceptions — values built from chain reads, which carry colour codes of their own that sanitising would strip — are greppable as `trustedMarkup` call sites.

**Where that holds, stated precisely, because it is the ticket's actual question.** The brand closes the detail block: in `safe-tx-detail-display.ts` a field added later as a plain `string` is a compile error rather than a review question, and the brand survives the obvious ways round it — a template literal, `String(x)` and `[a, b].join('')` are each rejected. `Printable` appears as a *type* in that module and in the primitive, nowhere else. The other four modules on the primitive — `safe-decode-utils.ts`, `safe-utils.ts`, `ledger-flex-preview.ts`, `confirm-safe-tx.ts` — call `printableField()` and interpolate the result into ordinary `consola` lines, so a new raw interpolation there compiles clean and is held by review, exactly as before. So this narrows the source-scanning placement test EXSC-951 records as defeated five separate ways to the four modules it still governs; it does not replace it everywhere.

## EXSC-949 asks about the rule-201 `Data:` question. **The two funnels stay different, and the rule now says so.**

`execute-pending-timelock-tx.ts` must not print raw `Data:` hex when a decoded display is shown; `confirm-safe-tx.ts` deliberately does. That is where a human commits a signature to exact bytes, and the hex is the artefact they compare against the Ledger screen. Rule 201 previously prescribed one behaviour and the code did the other; the carve-out is now written down.

## Two calls that decide rather than repair

The load-bearing piece, if a reviewer reads one thing: **the deployments lookup keys on the stored periphery name while printing the sanitised one.** Sanitising can turn a name the record does not hold into one it does — a zero-width space inside it is simply removed — and the check would then print a ✅ about a value the calldata does not contain. What is shown and what is decided come off the same argument, but not off the same string.

Same shape for `target=` in `formatTimelockScheduleBatch`: a value that does not normalise to an address for the network yields a notice and no explorer link, rather than being repaired into a different address that still looks like one.

The `To:` line asked the same question and got it wrong, which the review gate caught: it gated the target's name and explorer link on `identityPreserved`, which asks whether sanitising *changed* the text — not whether the value was ever an address. A value that was never one survives sanitising untouched, and neither callback refuses it: `formatAddressForNetworkCliDisplay` passes an unrecognised shape through and `buildExplorerAddressUrl` interpolates what it is handed. So a plain-ASCII non-address rendered as `this-is-not-an-address https://etherscan.io/address/this-is-not-an-address` — a real-looking link to a value no chain holds, one line above the sign prompt, with no notice. It is now checked by asking the network to resolve the value — the same call the signing path already makes — and shown as stored with a notice and no link. The two notices stay distinct: a repaired address still reads "target name withheld", because for a value that was never an address "the stored value is not what is shown" would be the false half of the explanation.

## Why this PR exists at all

The same content was merged on #2349 and reverted twelve minutes later on the stack base `feat/exsc-937-refuse-delegatecall-proposals` (`74b8f4040`, 1,436 deletions). **That revert was intentional and #2349's own body explains it:** the PR carried a "not ready to review as evidence" banner naming three false claims, a regression it had introduced, eight surviving mutations and two pre-existing raw display routes it claimed to have closed but had not. #2334 squash-merged the base branch, so `main` took the 937 work and not this, while both tickets read Done. `git ls-tree origin/main` shows no `printable-field.ts`.

So this PR is not a re-land of the reverted content as it stood. It is that content with the holes named in #2349 closed, and the sections below say which are closed by execution and which are not.

Worth knowing for anyone checking this class of thing: `main` does carry a same-named `asPrintable`, module-private inside `safe-tx-detail-display.ts` and with neither a `maxChars` bound nor a non-ASCII count — so grepping the symbol confirms a premise that is false.

## What #2349 left open, and where each item stands

**Two pre-existing raw display routes — closed, and each asserted by execution.** `safe-utils.ts`'s `decodeDiamondCut` printed a 4byte-supplied name raw on the commonest proposal shape and had no test case at all; `formatTimelockScheduleBatch` printed `target=` and `selector=` raw inside the very function #2349's body said was covered. All three now go through the primitive, and removing any one of the three guards fails a test.

**Surviving mutations — this is where the work went.** #2349 recorded eight mutations that survived, five on guards it had added. I re-ran that pass over every guard site in this diff rather than trusting the claim: mutate one site to the pre-PR raw behaviour (`printableField(x)` → `String(x)`, `asPrintable` → an identity record), run the five affected suites, record whether anything fails. **13 of 26 sites survived** — including every route #2349 named, so the defect had been carried across intact and the body's "each route is asserted by executing the real render path" was false.

Seven new tests close that. **23 of 26 sites are now killed by execution.** The three that remain are stated rather than claimed closed (two more, added in the last round, are named further down):

| Site | Why no execution assertion |
|---|---|
| `safe-decode-utils.ts:973` generic `Function:` | The branch is gated on `decoded.functionName === 'registerPeripheryContract'`, a literal equality — the value printed is provably that literal. Defence in depth against a future edit, not a route. |
| `confirm-safe-tx.ts:279` `safeTxHash` in a throw | The module has **no exports** — it is a CLI entrypoint — so the site is not reachable from a unit test without refactoring it, which is out of scope here. Lowest severity: it prints after signing, so it cannot corrupt the decision. |
| `safe-tx-detail-display.ts:172` `describeThrown` | Effectively unreachable: `formatProvenanceLines` is total by construction — its whole body is wrapped in a `try` that returns `unrenderableLines(error)` — so nothing escapes it into this catch, and the one value that could (an `Error` whose `.message` getter throws) fails the `String()` here too and lands in the inner catch, which returns a constant rather than reaching the primitive. |

**`safe-decode-utils.ts:1027` was in that list and should not have been.** I had recorded its `decoded.functionName` as repo-ABI-controlled. It is not: `decodeTransactionData` feeds a 4byte-sourced signature into `parseAbi` (`safe-decode-utils.ts:908`, "Dynamic signature from 4byte/Sourcify"), so a decoded name is remote, community-writable text and the proposer chooses which of it is fetched by choosing the selector. Proven by mutating that site alone and driving the real path: the line grows from 194 to 319 characters and loses its clip notice. `parseAbi`'s name grammar rejects escapes and non-ASCII, so the exposure is length only — but length is one of the three properties here, so the site is now asserted like any other.

The harness is a throwaway, not committed: it edits a source file in place, and a committed mutation runner that a suite could import is a worse hazard than the coverage gap it measures. Reproducing it is six lines of string replacement plus `bun test`.

**False claims — one found in this PR and fixed.** Rule 201 and this body both said that exactly two modules printing to the signing prompt still clean with a bare `sanitizeProvenanceText`, "without the bound, the invisible count or the notice". Both parts were wrong: `delegatecall-gate.ts` is a third such module and was unnamed, and bounds are not uniformly absent — `delegatecall-gate.ts` clips at 80 code points and the provenance block bounds its reason and commit, while only `render-check-ledger.ts` bounds nothing. Corrected in `976bc0c21`. What all three genuinely lack is the disclosure; bounds are absent only in places, and `render-check-ledger.ts` plus the provenance block's `proposerHandle` / `gitBranch` / `prUrl` are the unbounded ones.

## Every decoded argument is bounded, whatever it looks like

Worth a reviewer's attention because it is the one thing here that changed after the content was first written, and because the original shape was wrong in a way no test would have caught.

`scalarBound` used to pick a decoded argument's bound from the **value's shape** — `/^0x[0-9a-fA-F]*$/` meant unbounded, reasoning that a hex payload is what the signature covers and clipping it would hide what is being approved. That cannot tell a `bytes` payload from a `string` that merely looks like hex, and the difference was reachable on real functions: `PeripheryRegistryFacet.getPeripheryContract(string)`, `CalldataVerificationFacet.validateCalldata`'s `bridge` parameter, and any hex-shaped string nested in a struct argument (a bridge facet's `_integrator` / `_referrer`). A proposal carrying `'0x' + 'a'.repeat(500_000)` in one of those rendered in full, with no clip and no notice — the exact terminal flood `MAX_FIELD_CHARS` exists to prevent, needing no escape sequence at all.

So the exemption is gone and every decoded scalar is bounded. **What that costs, stated precisely:** the bytes under the signature are still disclosed in full by the unclipped `Data:` field, so nothing leaves the screen — but a reviewer wanting the exact value of one clipped `bytes` argument has to read it out of the raw blob rather than off the decoded line, since the decoded view is a convenience over calldata that is itself shown whole. I judged that a better trade than leaving a proposer able to choose how much of the prompt to occupy. `UNBOUNDED` now has exactly two callers, both of them that field — the detail block and the Ledger filmstrip.

The suite could not have caught this. Every crafted long value in it was `'x'.repeat(n)`, which never matches the hex shape, so the bug sat in a blind spot rather than behind a wrong assertion. There is now a test for the hex-shaped case, pinning 120 by literal value.

**The same blind spot had a second half, found by the review gate and fixed here.** Bounding every *element* still left the *aggregate* unbounded: the element count is encoded in the calldata, and with each element short no per-element notice fires either. A 20,000-element `address[]` reaching that branch — needing no escape sequence and no long value — rendered 900,001 characters silently, nearly twice the hex case above. Every crafted value in the suite was one long **scalar**; none was a long **array**, so the same class of miss sat one level up. `formatDecodedArg` now bounds the rendered aggregate at `MAX_ARG_JSON_CHARS` (2,000, pinned by literal value in the test), wide enough that a real argument shape still renders whole — a real `initFrax` argument built from `config/frax.json` measures 378 characters, and a bare `address[]` clips only past 44 elements. Removing the bound fails the test.

**What this bound does not cover, since an earlier draft of this body implied otherwise.** It governs the *generic* decoded-argument branch only. The proposal shapes that actually carry a proposer-sized array are routed away from `formatDecodedArg` by a dedicated formatter earlier in `formatDecodedTxDataForDisplay`: `diamondCut` by `decodeDiamondCut`, `scheduleBatch` by `formatTimelockScheduleBatch`, and the batch whitelist call by `formatBatchSetContractSelectorWhitelist` — each of which emits one line per element with no cap. So `diamondCut`, the dominant real proposal, remains an unbounded flood by element count, and this PR does not change that. (An earlier draft named a `batchAddDex`-shaped payload as the motivating case; that function no longer exists in this repo — the successor is `WhitelistManagerFacet`, whose batch entry point is the third formatter above.) Bounding the dedicated formatters is follow-up work, not something this PR claims.

## What the last gate round changed

Three fixes after the guards above were already written, so they need reading as new code rather than as part of the re-land.

**A target that was never a *string* got no explanation at all.** The `To:` address check reached only values `asPrintable` could report on as strings, because it keyed on `identityPreserved` — which is false both for a value the sanitiser repaired *and* for one that was never a string. So a row carrying `to: 42`, `undefined` or `true` rendered `To: 42` in plain green: no name, no link, and no notice saying why either was missing. It now keys on the notice being empty, which is what those two cases genuinely have in common, and the existing "never a string" test asserts the notice rather than only the withheld name. Reverting the condition fails that test. This is defence in depth rather than a live route — `initializeSafeTransaction` puts every pending row's `to` through `getAddress` before anything is displayed, so a non-string would have thrown upstream — but the comment above the check claimed address-ness was checked here, and for these values it was not.

**A test for the corruption that actually looks like an address.** The suite probed one input class, plain-ASCII garbage. A hex string one nibble short reads as an address to anyone scanning the prompt, and is exactly what the name and link were vouching for. Relaxing the check to "hex-shaped" fails the new case and nothing else.

**Two remote values reaching the prompt unsanitised, both raised by CodeRabbit and both confirmed by tracing rather than by accepting the finding.** `safe-utils.ts`'s facet-version line prints a version read from `.cache/deployments_production.json` — which `confirm-safe-tx.ts` refreshes from MongoDB, so it is remote row text, not a repo constant. And two provider/SDK error messages in `confirm-safe-tx.ts` are printed before the operator decides. Both now go through the primitive. **Neither is asserted by execution**, for the reason already in the table above: `confirm-safe-tx.ts` has no exports, and the facet-version site is module-private behind a cache keyed on the process working directory. So the count is 23 of 28 sites killed by execution, with five stated — the three in the table plus these two.

## What peer review changed

Three fixes from @gvladika's round, plus two docblock corrections from the gate that followed it. All five need reading as new code.

**A valid Tron proposal rendered a false "not a valid address".** The `To:` check above used viem's `isAddress`, but `initializeSafeTransaction` resolves the same field through `normalizeAddressForNetwork`, which accepts base58 on Tron. A Tron row storing base58 was therefore signable — and signed over the normalised hex — while the line above the sign prompt called it invalid and dropped both the target name and the explorer link. That is the one notice on the block that must never cry wolf, and it was firing on a correct row. The check now makes the same call the signing path makes, so the set of rows this reports on and the set that can reach the prompt are the same set. Two tests: one pins the base58 row keeping its name and link, one pins that a `T…` prefix alone is still refused — so the widening did not swallow the guard. Rows written by `propose-to-safe-tron.ts` store the hex form, so no live proposal hit this; it was reachable by any row that did not come from that script.

**`renderScalarArg`'s bound was justified by a field that exists on only one of its two callers.** The docblock said the clipped bytes are disclosed by the unclipped `Data:` field — true on `confirm-safe-tx.ts`, false on `execute-pending-timelock-tx.ts`, which reaches the same code through `formatTimelockScheduleBatch` and prints no `Data:` field before the Execute prompt, because rule 201 forbids it. So on that path a `bytes` argument over `MAX_FIELD_CHARS` is clipped with nowhere on screen to read it in full, and the operator decides on a partial value. The bound is unchanged and clipping stays the safe direction, but the docblock now names the path it reasons about and records this alongside the `diamondCut`/`scheduleBatch` element-count gaps above. **Closing it is follow-up work, tracked as [EXSC-999](https://linear.app/lifi-linear/issue/EXSC-999/a-clipped-bytes-argument-is-unreadable-on-the-timelock-execute-path)** — the fix is a caller-supplied bound, not a raw `Data:` line, which rule 201 rules out here.

**A notice painted in the colour of the value it warns about.** Rule 201 gains the criterion in this diff, and three sites defeated it: two wrapped a value in the notice's own yellow, and one embedded a notice-bearing string in a colour span the notice's own reset closed mid-line, so everything after the first repaired argument rendered uncoloured — a rendering glitch where a warning was intended. `colorAroundNotices` re-opens the value's colour after each reset, and takes the colour as a closed set that excludes yellow, so "paint a value in the notice colour" is a compile error rather than a judgement each call site makes again. The third site (`Periphery Name:`) is one @gvladika did not name.

**Two docblocks that described guarantees the code does not give**, both caught by re-running the gate on the fix commit and both verified by executing the call rather than reading it. `isAddressForNetwork`'s said "checksummed hex": the call it wraps resolves mixed-case hex by *recomputing* the checksum, so any hex of the right length is accepted and this is a shape check, not a corruption check. And the new `network` field's said the value never comes from the stored row — it does, whenever the operator passes no `--network`, since the network list is then `distinct('network')` over the pending rows. What actually holds is that both routes filter to an active `networks.json` key.

**Coverage of this round.** The `Periphery Name:` and decoded-argument sites are asserted by execution, as is each new address case. The two `call=` / `Call:` sites are **not**: they sit behind `tryFormatDiamondPayload`, which needs `diamond.json`, and that file is gitignored — so no test in CI can reach them. It is the same one-line substitution as the sites that are asserted, but it is stated, not proven. Bun's line-coverage report is not evidence to the contrary for the `call=` line: the `if (pretty)` guard marks the line hit for evaluating the condition even though the branch never runs.

**Sites deliberately left hand-wrapping their colour.** `colorAroundNotices` is applied where a value can carry a notice; four `Function:` lines and the `fallbackSignature` line are not converted. Three of the four are gated on an exact literal match (`registerPeripheryContract`, the `ACCESS_CONTROL_ROLE_FUNCTIONS` membership, `formatRoleChange`'s own label), so the value is one of a closed set of repo strings. The fourth is the generic fallback and can carry a signature resolved remotely, but it must first parse as a Solidity signature via `parseAbi` — ASCII identifiers and type lists only — so there is no shape for `asPrintable` to report on. `fallbackSignature` puts its notice at the end of the string with nothing following inside the colour span, so the mid-line truncation this helper exists to fix cannot occur there. Latent inconsistency rather than a live defect; naming it here rather than widening this PR to four more lines.

## Not in this PR

Three modules that print to this prompt still clean with a bare `sanitizeProvenanceText`, which discloses nothing — no invisible count, no confusable count, no notice: `render-check-ledger.ts` (merged with #2348), `delegatecall-gate.ts`, and the provenance block. They are [EXSC-986](https://linear.app/lifi-linear/issue/EXSC-986) rather than scope creep here.

**Not because their values are machine-reported — only `render-check-ledger.ts`'s are.** An earlier draft of this body and of rule 201 said all three, and that was wrong in the direction that matters: `delegatecall-gate.ts` documents its own operation value as proposer-controlled (it reaches the struct through a cast, so a row can carry a string), and the provenance block renders `proposerHandle`, `gitBranch`, `prUrl` and `reason` straight off the row. What those two lack is the disclosure, not a reason to need none — a `DelegateCall` carrying two ZWJs and a Cyrillic `е` renders on the Operation line today with no notice, while every other field in the same block would count it. Rule 201 now says this rather than the reassuring version, and deferring them is a scope decision, not a safety conclusion.

## Verification

- `bun test script/` — 4383 pass, 0 fail across 156 files. `bunx tsc-files --noEmit` and `bunx eslint` clean on every changed `.ts`; `markdownlint-cli2` clean on both prose files.
- **23 of 28 guard sites are asserted by executing the real render path against a crafted payload**, proven by mutating each site and observing a failure — not by scanning source, which EXSC-951 records as defeated three separate ways. The two guards added by the review gate (the aggregate argument bound, the `To:` address check) were verified the same way: each dies when mutated back to the previous behaviour. The five unasserted sites and the reason for each are in the table above and in "What the last gate round changed".
- A benign row's rendered block, a benign `registerPeripheryContract` call and a benign filmstrip are each asserted byte-identical to before the change.
- Merged with `origin/main` rather than rebased, three times as main moved. The first round's conflicts were both import blocks; `getTargetStateFacetVersion` no longer exists on `main` (removed by #2348) and its import went with it.
- **Not a pure re-land.** Beyond the reverted content, this PR carries fixes that were never on #2349: `formatDecodedArg` dropped every notice for object/tuple arguments, so a hostile or over-long string nested in a tuple was cleaned silently; `decodeTransactionData`'s catch labelled every `Error` "stored as an object, not a string"; `describeThrown` discarded the notice on a clipped error message; and a non-hex `data` was shown on the Ledger panel with a `0x` invented and the characters case-folded — on the one panel whose job is a literal comparison against the device. Those need reviewing as new code.
- **#2350 has since merged, and the collision flagged on #2349 is resolved.** It rewrote `frameLine`'s styling to take index `ranges`, and in doing so replaced the `body.replace(t, …)` call whose replacement operand made `$&`, `` $` `` and `$'` expand as substitution patterns — so that hazard is now structurally impossible rather than unreachable by convention, and the comment describing it is gone with it. The sanitised filmstrip path survived the merge intact; the hash-mode preview #2350 added reads a hash computed by `safe.getTransactionHash`, not the stored row, so it opens no new route.


0xDEnYO:
The #2349 revert on #2334 (`74b8f4040`) was intentional, not unexplained. #2349's own body marked itself not-ready (false claims, surviving mutations, two raw display routes still open). This re-land should stay draft until those holes are closed and Linear is `Fixes [EXSC-951](...)` / `Fixes [EXSC-949](...)`.

.agents/rules/201-safe-decode-scripts.md
```diff
@@ -27,7 +27,8 @@ paths:
...
+- **confirm-safe-tx.ts**: Uses `formatDecodedTxDataForDisplay(tx.safeTx.data.data, { chainId: chain.id, network })` for each Safe tx; uses `getTargetName` from safe-decode-utils for the "To:" line in Safe Transaction Details. Unlike `execute-pending-timelock-tx.ts` it also prints the full raw `Data:` hex alongside the decoded display, and that is deliberate: this is where a human commits a signature to exact bytes, and the hex is the artefact they compare against the Ledger screen. Do not extend the "no raw `Data:` hex" rule below to this script.
+- **Displaying a stored or decoded value**: every value that comes off a proposal row or out of a decode — including a `string` ABI argument and a signature resolved via 4byte — goes through `asPrintable` / `printableField` from [script/deploy/safe/printable-field.ts](script/deploy/safe/printable-field.ts) before it is interpolated into a log line. The notice must stay visually distinct from the value it warns about: it carries its own colour and reset, so it reads correctly wherever it is interpolated — do not strip those, and do not build a notice that inherits the surrounding colour.
```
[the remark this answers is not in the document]
0xDEnYO: Fixed in `8d3a8bb59` — and in all five links in that file, not only this one. Four predate this branch and are the same defect in the same list, so fixing one would have made the correct form look like the exception; the `../../docs/...` links in the same tree already used it. Each target is now checked to exist from the rule file's own directory (`.agents/rules/../../script/deploy/safe/*.ts`), and markdownlint is clean.

0xDEnYO:
@coderabbitai review

gvladika (commented):
Reviewed the diff and ran it locally off `502436f2b`: `bun test script/` green (the 5 `verify-immutable-registry` failures are my worktree missing built artifacts — `run-ts-tests` is green in CI), `bunx tsc-files --noEmit` and `bunx eslint` clean on all 7 changed `.ts` files.

The core work is sound, and the table of which guard sites are asserted by execution and which are only stated is genuinely useful. Three findings inline: one user-visible false alarm on Tron, one docblock justification that doesn't hold on the `execute-pending-timelock-tx.ts` path, and one cosmetic collision with the notice-visibility rule this PR itself adds.

script/deploy/safe/safe-tx-detail-display.ts
```diff
@@ -293,53 +233,107 @@ function formattedAddressField(
...
+ * value no chain holds, so address-ness is checked here rather than inferred
+ * from the text having come through intact.
  */
-function toLine(input: ISafeTxDetailInput): string {
+function toLine(input: ISafeTxDetailInput): Printable {
   const { text, identityPreserved, notice } = asPrintable(input.to)
   const { shown, failed } = renderAddress(text, input.formatAddress)
-  const resolvable = identityPreserved && !failed && shown !== ''
+  // Non-strict: a legitimately lower-case address is not a corrupt one.
+  const addressShaped = isAddress(text, { strict: false })
```
gvladika: **A valid Tron proposal now renders a false "not a valid address".**

`isAddress` is viem-only, but `initializeSafeTransaction` resolves `to` through `normalizeAddressForNetwork`, which accepts base58 `T…` on Tron networks (`script/utils/normalizeAddressStringForViem.ts`). So a Tron row storing base58 is signable and gets signed over the normalised hex — but the prompt says it isn't an address, and drops both the target name and the explorer link.

Confirmed by driving the real renderer with `network: 'tron'`:

```
To:  TXYZopYRdj2D9XRtbG411XZZ3kM5VkAeBf ⚠ not a valid address — shown as stored, and no explorer link
```

Rows written by `propose-to-safe-tron.ts` store the EVM hex form (`to: safeTxToEvm`, line 301), so the normal path is unaffected — but tolerating base58 upstream and rejecting it here is inconsistent, and a spurious red flag one line above the sign prompt is the exact failure mode this PR exists to prevent: it trains signers to skip the notice.

Suggest gating on `normalizeAddressForNetwork(network, text)` in a try/catch instead, which is what `formatTimelockScheduleBatch` already does for `target=` in this same diff (`safe-decode-utils.ts:640-652`). That keeps the two address checks on one rule and covers Tron for free.

script/deploy/safe/safe-decode-utils.ts
```diff
@@ -455,8 +465,41 @@ function getDiamondAbiItemForSelector(selector: string): Abi[number] | null {
...
+ * One decoded scalar as printable text plus whatever that cost.
  *
+ * A `string` ABI argument of a legitimately encoded call is arbitrary text the
+ * proposer chose — `registerPeripheryContract(string,address)` decodes cleanly
+ * with an ESC in its name — so it cannot reach the operator's terminal raw.
+ *
+ * Bounded unconditionally, and deliberately not by the value's shape: hex here
+ * is no more likely to be a payload than a `string` argument the proposer
+ * chose. The bytes under the signature are disclosed by the unclipped `Data:`
+ * field rather than by this line, which is what makes clipping safe.
```
gvladika: **This justification holds for `confirm-safe-tx.ts` but not for `execute-pending-timelock-tx.ts`.**

"The bytes under the signature are disclosed by the unclipped `Data:` field" is true on the signing path. It is not true on the timelock execute path, which reaches this code via `execute-pending-timelock-tx.ts:1711` → `formatTimelockScheduleBatch` → `formatDecodedTxDataForDisplay` → `formatDecodedArg` — and where rule 201, reaffirmed in this very diff, forbids printing raw `Data:` hex. Grepping that file, the only full-calldata print is `Data: ${primaryPayload}` at line 1975, inside the **reject** function, not before the Execute prompt.

Pre-PR, `formatDecodedArg` returned scalars unbounded (`const s = String(arg); … return s`). So on that path a `bytes` argument over 120 chars is now clipped with no way to read it anywhere on screen:

```
0xabab…abab ⚠ clipped for display — stored 402, shown 120
```

Not a security regression — clipping is the safe direction, and the notice fires — but it is a real loss of information on an execute decision, and the stated rationale doesn't cover it. Either take the bound from the caller (unbounded-ish where no `Data:` field accompanies the decode), or correct the docblock to say which path it reasons about and record the timelock gap alongside the `diamondCut`/`scheduleBatch` element-count gap already named in the PR body.

script/deploy/safe/safe-decode-utils.ts
```diff
@@ -469,24 +512,35 @@ export function formatDecodedArg(arg: unknown, network?: string): string {
...
+    // `JSON.stringify` escapes its colour codes into visible text and the
+    // disclosure reads as part of the value.
+    //
+    // Bounding each element leaves the aggregate unbounded: the element count
+    // is encoded in the calldata, so an array scrolls the proposal off the
+    // screen without any single value being long enough to clip.
+    const { text, notice } = asPrintable(json, MAX_ARG_JSON_CHARS)
+    return `${text}${
+      repaired
+        ? fieldNotice(
```
gvladika, line 548: **The notice isn't visually distinct at the consumer, which is the criterion rule 201 gains in this diff.**

The rule added here says "the notice must stay visually distinct from the value it warns about". Two places where the returned string defeats that:

- `safe-decode-utils.ts:1055` wraps the whole return in `\u001b[33m…\u001b[0m`, so the yellow notice sits inside a yellow value — warning and value are the same colour.
- `tryFormatDiamondPayload` (`:572`) embeds this string in `call=\u001b[34m${pretty}\u001b[0m`. The notice's own `RESET` closes the blue mid-expression, so everything after the first repaired argument renders uncoloured — which reads as a rendering glitch rather than a warning.

Cosmetic, and lower priority than the other two, but worth either giving `fieldNotice` a distinct colour from the fields it can land inside, or having the two call sites re-open their colour after the notice.

0xDEnYO:
@coderabbitai review

script/deploy/safe/safe-tx-detail-display.ts
0xDEnYO: Fixed in `8105db106`, as you suggested — `isAddressForNetwork` wraps `normalizeAddressForNetwork(network, text)` in a try/catch, so this check and `formatTimelockScheduleBatch`'s `target=` now answer to one call. `ISafeTxDetailInput` gains a `network` field for it; `confirm-safe-tx.ts` is the only production construction site.

Two tests: one pins your base58 row keeping its name and link, one pins that a `T…` prefix alone is still refused — the widening had to not swallow the guard.

One correction to my own docblock, from re-running the gate on the fix: it claimed the check accepts "checksummed hex". It does not. `getAddress` *recomputes* a mixed-case address's checksum rather than verifying it, so any hex of the right length passes on a non-Tron network — the same set the old `isAddress(…, { strict: false })` accepted. `76336a65b` says shape instead, so nobody reads this as a corruption check.

script/deploy/safe/safe-decode-utils.ts
0xDEnYO: Took your second option in `8105db106`: bound unchanged, docblock corrected. It now names `confirm-safe-tx.ts` as the path the justification holds for, names `execute-pending-timelock-tx.ts` as the one it does not, and records the clipped-`bytes` gap next to the `diamondCut`/`scheduleBatch` element-count ones. The PR body has it too — it was missing there entirely.

Confirmed your read of the path rather than taking it on trust: `executeOperation` goes from `formatTimelockScheduleBatch` straight to the Execute/Reject/Skip prompt, and the only full-calldata print in that file is `Data: ${primaryPayload}` inside `rejectOperation` — after the prompt, and only on the Reject branch.

I did not take the caller-supplied-bound option, and I want to be straight that this leaves a real gap rather than closes it: an operator can still approve a `bytes` argument they cannot read in full. Rule 201 rules out the easy fix here, so it wants its own ticket rather than a change in this PR.

script/deploy/safe/safe-decode-utils.ts
0xDEnYO: Fixed in `8105db106`, taking your second suggestion — the call sites re-open their colour after the notice, via a new `colorAroundNotices` in `printable-field.ts`. It takes the colour as a closed set that excludes yellow, so "paint a value in the notice colour" is a compile error rather than something each call site has to get right again.

There was a third site beyond the two you named: `Periphery Name:` had the same yellow-in-yellow collision. Converted too.

One honest caveat: of the four converted sites, `Periphery Name:` and the decoded-argument line are asserted by execution, but `call=` and `Call:` are **not** — both sit behind `tryFormatDiamondPayload`, which needs `diamond.json`, and that file is gitignored, so no test in CI can reach them. Same one-line substitution, not proven there. Worth knowing that bun's coverage report is not evidence otherwise for the `call=` line: the `if (pretty)` guard marks it hit for evaluating the condition even though the branch never runs.

script/deploy/safe/safe-tx-detail-display.ts
gvladika: Verified in `8105db106` — confirmed, and the widening is correctly scoped. Drove the real renderer across the matrix that mattered:

| network | `to` | result |
|---|---|---|
| tron | `TXYZop…AeBf` (valid base58) | name + explorer link restored ✅ |
| tron | `TZZZnotareal…` (bogus `T…`) | still refused ✅ |
| **mainnet** | `TXYZop…AeBf` | **still refused** ✅ |
| mainnet | `0x1234…7890` | name + link ✅ |
| mainnet | `this-is-not-an-address` | refused ✅ |
| mainnet | `0x` + 39 nibbles | refused ✅ |

Row 3 is the one I was watching for — routing through `normalizeAddressForNetwork` could have made base58 acceptable everywhere, and it doesn't, because the Tron branch is gated on `isTronNetworkKey`. So the accepted set widened by exactly Tron base58 and nothing else.

And your `76336a65b` correction is right — I checked viem rather than assuming. `getAddress` validates shape, then *recomputes*:

```
0x1234567890abcdef1234567890abcdef12345678 -> 0x1234567890AbcdEF1234567890aBcdef12345678
0x1234567890AbCdEf1234567890abcdef12345678 -> 0x1234567890AbcdEF1234567890aBcdef12345678
```

A deliberately wrong checksum is silently corrected, not rejected. Good catch to fix the docblock before it read as a corruption check — that's the kind of wrong comment that gets relied on later. Resolved from my side.

script/deploy/safe/safe-decode-utils.ts
gvladika: Docblock now says what's true — good, and I agree with the call not to change the bound here.

To be explicit about why I'm fine with it: the alternative (a caller-supplied bound) would mean `formatTimelockScheduleBatch` passing `UNBOUNDED` down, which puts the terminal-flood route straight back on the execute path. Trading a disclosed partial value for an undisclosed flood is the wrong direction, so "clip and say so, ticket the real fix" is right.

One ask: please put the ticket ID in the docblock next to the gap, the way the PR body does for the `diamondCut`/`scheduleBatch` element-count gaps. Right now a reader who hits the clip on the execute path has the problem named but no thread to pull — and this is exactly the kind of known gap that goes quiet once the PR is merged and the body stops being read.

Resolved from my side once the ID is in.

script/deploy/safe/safe-decode-utils.ts
gvladika: `colorAroundNotices` works — checked the output rather than the shape of the code. Input `abc<ESC>[33m WARN clipped<ESC>[0m` in blue gives:

```
ESC[34m abc ESC[33m WARN clipped ESC[0m ESC[34m ESC[0m      (spaces added for legibility)
```

Value blue, notice yellow, wrapper re-opened and closed, no orphan reset leaking into the rest of the line. Making `ValueColor` a closed set that excludes `33` is the better half of the fix — that's the part that stops this recurring.

Agreed on the `call=` / `Call:` caveat, and the note about bun's coverage report marking the `if (pretty)` line hit for evaluating the condition is correct and worth other people knowing. Fine to state it.

**One more site of the same class, which the fix didn't reach — `safe-tx-detail-display.ts`, `toLine`.** The name fragment is `color(YELLOW, targetName)` nested inside `color(GREEN, …)`, and `printableFragment` appends the fragment's own notice *inside* that yellow. Forcing a notice onto it:

```
ESC[32m 0x1234…7890  ESC[33m AAAA…(120)  ESC[33m ⚠ clipped for display — stored 200, shown 120 ESC[0m ESC[0m …
```

Yellow notice inside a yellow value, and the notice's reset closes the green wrapper early — both halves of what you just fixed elsewhere.

**Not asking you to fix it in this PR**, because it isn't a route: `toTargetName` comes from `getTargetName` off `deployments/<network>.json`, so it's repo-controlled, and no real contract name is over 120 chars or non-ASCII. Raising it because it's the same defect in the module the PR describes as closed by construction, and `colorAroundNotices` now exists to fix it in one line if you'd rather not leave the inconsistency. Your call — happy either way.

gvladika (approved):
Re-reviewed `76336a65b`. All three findings fixed; verified each by execution rather than from the replies.

- **Tron address gate** — `isAddressForNetwork` accepts base58 on Tron only; mainnet still refuses it, and a bogus `T…` is still refused on Tron. The accepted set widened by exactly the intended amount.
- **Decoded-argument bound** — bound unchanged, docblock now names the path the justification holds for and the path it doesn't. I re-confirmed the `execute-pending-timelock-tx.ts` path independently. Asked for the follow-up ticket ID to be inlined next to the gap.
- **Notice colour** — `colorAroundNotices` output checked directly; `ValueColor` excluding yellow is the part that stops the regression recurring.

Locally on the PR head: `bun test script/deploy/safe/` 2001 pass / 0 fail; `bunx tsc-files --noEmit` and `bunx eslint` clean on all changed files. CI 36 pass, 4 skipped (no Solidity changes), 0 fail.

Two things I want to note for the record rather than block on:

1. The `76336a65b` self-correction is the most valuable change in this round. The docblock had claimed the check accepts "checksummed hex"; viem's `getAddress` recomputes a mixed-case checksum rather than verifying it, so that would have read as a corruption guarantee the code never made. Catching your own wrong comment on a re-run of the gate is the behaviour I'd want on every security-adjacent PR.
2. One latent instance of the notice-in-colour collision remains in `toLine` (details on that thread). Not a route — the value is repo-controlled — so it's a consistency call, not a fix I'm requiring.

Approving. The remaining asks are a docblock ticket ID and an optional one-liner; neither needs another round from me.

Separately, and not gating: the PR body is still ~4,000 words of archaeology about #2349, mutation rounds and superseded drafts. `.agents/rules` asks PR descriptions to describe what the code does now, in plain language. The honesty is the good part — please keep the "5 of 28 sites not asserted by execution" table and the scope caveats — but the history belongs in EXSC-951. Worth trimming before merge so the next reader finds the diff instead of the defence.

gvladika (approved):

PR #2363, opened 2026-09-09
feat(codehash): grade a codehash match by who built what it matched (EXSC-952)

0xDEnYO:

- Ref EXSC-952 — [WP-5.2 · CI attestation minting](https://linear.app/lifi-linear/issue/EXSC-952/wp-52-ci-attestation-minting) · the `A-CI` / `A-LOCAL` state

Deliberately `Ref` rather than `Fixes`: WP-5.2 is a shared ticket and its remaining scope — the CI workflow, `attest-build-provenance`, compiler binaries pinned by hash, the Sigstore trust root, the live mint → `gh attestation verify` round trip — is not in this PR. #2351 auto-closing the ticket while landing only the schema half is the state this avoids repeating.

Independent of #2351 (the schema half of the same WP, now merged) — no shared code, and this branch has main merged in. Two cleanups it touches on are deferred to [EXSC-989](https://linear.app/lifi-linear/issue/EXSC-989/hoist-anchorid-to-a-shared-module-and-guard-hash-comparisons-with).

## The claim the gate cannot currently make

`compareToAttestedSet` answers whether the deployed code is a build this repo can vouch for. It does not answer **who built it**, and the two are not the same claim: a CI-minted attestation was produced on a machine the signer does not control, while a local rebuild was produced by the very host that might be compromised. Rendering both as one green tells the signer their own machine agreed with itself.

WP-5.2 asks for the fallback to be a visible *state* rather than a silent substitution. This is that grading, in its own module so the set-membership logic cannot come to depend on provenance or the reverse.

## The case that set the shape

Neither `A-CI` nor `A-LOCAL`. When CI has attested this contract and the deployed code matches **only** the local rebuild, CI and this host disagree about what the source compiles to — either CI built something else or this host did. That is not a weaker pass, so `ci-disagrees` is its own grade, and like a local match it is not presentable as attested. Only `ci-attested` is.

## The grading is handed the matched builds, and never re-derives them

This is the part worth a reviewer's attention, because two earlier rounds of this PR got it wrong in a way that inverted the module's whole purpose.

`gradeMatchProvenance` originally took a hash plus the attestation set and re-derived its own match on `maskedHash` alone. `compareToAttestedSet` requires the masked hash **and** the deployed length **and**, on a pinned attestation, the exact bytes. A weaker predicate makes the matched group a strict superset of the comparison's, so provenance was read off builds the comparison had **rejected** — a CI attestation that failed to match counted as one that matched. Two consequences, both bad: a local-only match was presented to a signer as `ci-attested`, and `ci-disagrees` could provably never fire, since the disagreeing attestation was always absorbed into the match.

Reached on zkEVM specifically, and not hypothetically: `rawHash` is pinned there and the zksolc/LLVM fork lives only in the metadata trailer that masking removes — the drift `rebuild-attestations.ts` measured on `LayerSwapFacet`, where a 1.0.1 to 1.0.2 bump moved 33 trailer bytes and no codegen. So a CI mint on a different fork shares the masked hash and differs in the exact bytes.

Accepting `matched` as a parameter fixed the module but **relocated the same predicate into the caller**, which is worth stating plainly because the first fix claimed to have removed it and had not. `ICodehashComparison` exposed the match only as `matchedLineages: string[]`, so the one way to obtain the builds was to filter the attestations by those strings — and a lineage is a human label, not a key. `describeLineage` carries no provenance marker, so a CI mint and a local rebuild of one `contract@version` at one commit under one profile produce the byte-identical string; the filter then re-admits the rejected build and the verdict is `ci-attested` again. Once A-CI attestations exist that is the ordinary case, not an edge one.

So `compareToAttestedSet` now returns the builds it matched — `ICodehashComparison.matched`, alongside `matchedLineages` for rendering — and the grading consumes that. No caller has to reconstruct the group, and identity never passes through a display string. `all` is still passed, consulted only to count CI attestations that did *not* match: the one question the matched builds cannot answer.

## One record type, not two

`provenance` is a required field on `IAttestedBuild`, and the parallel `IProvenancedAttestation` is gone. The parallel type carried only `maskedHash`, which is both what forced the weaker predicate above and what would have made every call site maintain two arrays derived from the same records — one for the comparison, one for the grading — with drift between them undetectable at compile time. Required rather than optional follows `IAttestedBuild`'s existing reasoning about `viaIR` and `rawHash`: absent and `A-LOCAL` are the same build only by coincidence. `attestationFrom` tags its rebuilds `A-LOCAL`, which is what they are.

## A malformed hash is not a match

A matched build whose `maskedHash` fails `digestFault` grades `malformed` and is not presentable. A malformed hash normalises and compares as readily as a real one, so without the guard an empty or truncated value on both sides produced a verdict resting on a comparison between two things that are not hashes — a fail-open on precisely the property this module exists to guarantee.

Scoped to the matched builds, and to this module. `compareToAttestedSet` has the same gap at `attested-set.ts:130` and is left alone deliberately rather than widening this diff into merged code — that, and the duplicate `A-CI`/`A-LOCAL` vocabulary already exported as `AnchorId` by `check-ledger.ts:35` (whose obvious fix is a `safe/` to `codehash/` layering back-edge), are both in EXSC-989.

## Falsification

The provenance defect was first reproduced against the shipped code, then pinned by `grades a CI attestation the comparison rejected as ci-disagrees`, which runs the real `compareToAttestedSet` and asserts the grading agrees with it. That test **fails on the pre-fix code** — `presentableAsAttested` came back `true` where it must be `false`. `holds when the two builds share a lineage` pins the caller-side variant, and asserts inline that the lineage filter really does return two builds where the comparison matched one.

Eight decisions reverted one at a time, each restored byte-identical (verified with `diff` against pre-mutation copies):

| Mutation | Tests that failed |
| --- | --- |
| grade on the whole set rather than the matched builds (the original defect) | 3 — incl. `grades a CI attestation the comparison rejected…` |
| `ciCount` taken over the matched builds instead of `all` | 3 — same three |
| the comparison returns every same-hash build instead of the exact ones | 2 — both integration tests |
| `attestationFrom` stamps `A-CI` instead of `A-LOCAL` | 1 — `attests every rebuild as A-LOCAL…` |
| the `malformed` guard widened from the matched builds to the whole set | 1 — `does not let a malformed attestation elsewhere…` |
| the `digestFault` guard removed | 1 — `refuses a matched build whose masked hash is not a digest` |
| an empty matched group treated as a match | 1 — `grades an empty matched set as unattested…` |
| a local rebuild presented as attested | 1 — `never presents a local rebuild as attested` |

The first two fail the same three tests, which is honest rather than ideal: they are two directions of losing the same matched-vs-`all` distinction, and only the `does not upgrade…` test pins the direction.

Two of these were green before the latest round, which is why the gate ran twice. The producer's `A-LOCAL` was written in one struct literal and read nowhere in the suite, so flipping it to `A-CI` — labelling every build this host produced as CI-attested — left the entire TS suite passing. The malformed guard's scope was unpinned in both directions.

9 tests here. Full `script/deploy/codehash/` suite: 298 pass. Whole TS suite (`bun test script/`): 3989 pass, 0 fail. Re-verified at the merged head, which carries #2354. `bunx tsc-files --noEmit`, `bunx eslint` and prettier (the repo's 2.8.8) clean on every changed file.

## Not claimed

**No call site.** Nothing sets `A-CI` yet, because nothing mints CI attestations — that is the rest of WP-5.2 (the workflow, `attest-build-provenance`, compiler binaries pinned by hash, the Sigstore trust root, the live `gh attestation verify` round trip). Until a minter exists every attestation is `A-LOCAL`, so `ci-attested` and `ci-disagrees` cannot fire and this grading answers `locally-rebuilt` on everything — the truthful answer for today's state rather than a defect. Same shape as #2351 landing the schema half ahead of the minter.

Nothing calls `gradeMatchProvenance` in production either; `verify-cut-targets.ts` and `codehash-sign-gate.ts` still consume `compareToAttestedSet` directly. Rendering the grade in the signer-facing report is the wiring half of the same work package.

I have not built the workflow, and deliberately: it cannot be demonstrated from a local session, and a check nobody has watched fire is not evidence. It belongs in a PR that can show a real mint and verify round trip.


0xDEnYO:
@coderabbitai review

0xDEnYO:
@coderabbitai review

gvladika (approved):

PR #2365, opened 2026-09-10
fix(security): redact RPC endpoints in script logs (EXSC-965)

0xDEnYO:

Fixes [EXSC-965](https://linear.app/lifi-linear/issue/EXSC-965/troncast-prints-the-full-rpc-url-so-a-keyed-endpoint-leaks-into-every)

`ETH_NODE_URI_*` is a dRPC URL with the API key in the path, so a log line naming the endpoint writes a **live credential** into every transcript of every run. troncast did exactly that and leaked the Tron key on 2026-09-09.

## The leaks

**TypeScript — five sites**, now through `redactUrls()`:

| file | was |
|---|---|
| `script/troncast/utils/tronweb.ts` | `Initializing TronWeb with … ${rpcUrl}` |
| `script/deploy/tron/deploy-core-facets.ts` | `RPC URL: ${rpcUrl}` |
| `script/deploy/tron/transfer-ownership-to-timelock.ts` | `Connected to: ${fullHost}` |
| `script/deploy/tron/register-facets-to-diamond.ts` | `Connected to: ${fullHost}` |
| `script/deploy/detect-evm-version.ts` | `via ${chain.rpcUrls.default.http[0]}`, plus its `consola.error(…, err)` — viem puts the endpoint in `error.message` |

**Bash — the scan reads no `.sh`, and it was leaking the same class.** Eighteen `echo`/`echoDebug`/`error` sites, thirteen of them **not debug-gated**, now through a new `redactRpcUrl` and its break-glass twin — including `parseExecuteCommandResult`, the shared executor that prints the captured stderr of every bash `forge` run, all three verbose-output prints in `diamondSyncWhitelist.sh`, and the pause-state error paths in both emergency scripts. A test pins the executor: removing the redaction fails it. `getRPCUrl`'s own stdout returns are deliberately untouched: redacting those breaks every caller.

**The bash class is not closed**, and nothing enforces it — these are the sites that were found, not all that exist. At least five more echo a variable that can hold `cast` error text (`playgroundHelpers.sh:1967/1973/2020/2027`, `helperFunctions.sh:3598/3696/3851`). They belong to [EXSC-974](https://linear.app/lifi-linear/issue/EXSC-974) with the enforcement, because the awkward part is that `getRPCUrl` legitimately returns the endpoint on stdout, so a naive bash scan false-reds on the function whose job is to return it.

Also the **exhaustion path** of `rpcCallWithRetry` in both break-glass scripts. Only the per-attempt retry line was redacted at first; the terminal return handed the raw `cast` error — which embeds `--rpc-url` — to callers that all echo it. That is the path that runs during a real incident. Redacted at the single return rather than at the eight call sites; the success path stays verbatim, because callers parse it.

**Shown by running it, not by reading it.** The real `initTronWeb` at `CONSOLA_LEVEL=5` with a *synthetic* key:

```text
before:  Initializing TronWeb with mainnet network: https://lb.drpc.org/ogrpc?network=tron&dkey=SYNTHETICKEY-NOT-REAL-abc123
after:   Initializing TronWeb with mainnet network: [redacted-url]
```

## The check parses the source

`script/utils/rpc-url-log-scan.ts` walks every shipped module under `script/` and `tasks/` and fails on a log call naming an endpoint outside a redaction.

It uses `ts.createSourceFile`. Every distinction that decides a case is a question about the grammar — a type annotation versus a value, a binding versus a read, a comment or string versus code, a regex literal versus division — and I first tried to approximate all four with a text scan. That cost three rounds: each version shipped a hole (one blinded the scan on 4 of 240 real files, including a troncast command), and each fix made the previous round's tests vacuous rather than failing. The parser answers all four for free and deletes ~200 lines of tokenizer plus the tests defending it.

Worth stating plainly since an earlier version of this description said otherwise: **the note that this repo's `typescript` had no usable compiler API was wrong.** The repo pins `^5.4.5` and resolves 5.9.2. I had read that in a worktree with no `node_modules`, where bun silently auto-installed a different major version.

Two limits are written into the docblock rather than implied: the reach is a fixed identifier vocabulary and log-method list, and it does not read bash or error-object embedding. Those two channels are [EXSC-974](https://linear.app/lifi-linear/issue/EXSC-974); the devkit's own `--verbose` leak is [EXSC-975](https://linear.app/lifi-linear/issue/EXSC-975).

## Evidence

- **Blind sweep**: the canonical leak line injected into a scratch copy of each of the 240 walked files, appended and prepended — **0 blind** either way (the one hit is the declared `EXEMPT` entry).
- **Mutation testing**: every decision the module makes was mutated one at a time and each killed by a named test — the type-node skip, the declaration rule and each of its exceptions, redactor detection (bare and namespaced), the receiver check, each vocabulary and log-method entry individually, the env prefixes, the walked roots, the exemption, the symlink skip, the test-file exclusion, the reported line, and both bash retry paths. Restore verified byte-identical with `cmp`, never `git checkout`.
  A first pass at this was itself vacuous: `it.each(RPC_IDENTIFIERS)` iterates the list being mutated, so deleting an entry deletes its own case. All three sets — the vocabulary, the log methods, and the exemption's identifiers — are now pinned by **value** as well; the exemption is per-identifier rather than whole-file, so the one exempted file is still scanned for everything it did not exempt.
- **False positives**: the shapes that must stay silent are asserted, including the three the lexer got wrong (an optional type property, an arrow parameter, a regex reached without a semicolon) and the six the first parser version got wrong (method shorthand, getter, setter, class property, named function expression, label).
- Every negative is pinned **beside a real leak in the same fixture**, so no case can pass because the scan stopped reading the file.
- `bun test script/` → **3489 pass / 0 fail** at `cba9381f3`. `tsc-files` and `eslint` clean; scanner coverage 100% funcs. (`tsc` caught two errors `bun test` did not, both times.)

### DIFF
--- a/script/deploy/tron/deploy-core-facets.ts
+++ b/script/deploy/tron/deploy-core-facets.ts
@@ -15,6 +15,7 @@ import { consola } from 'consola'
 import type { IDeploymentResult, SupportedChain } from '../../common/types'
 import { EnvironmentEnum } from '../../common/types'
 import { getPrivateKeyForEnvironment } from '../../demoScripts/utils/demoScriptHelpers'
+import { redactUrls } from '../../utils/redactUrls'
 import {
   getEnvVar,
   saveDiamondDeployment,
@@ -75,7 +76,7 @@ async function deployCoreFacetsImpl(options: {
   // Get RPC URL and API key configuration (automatically handles TronGrid API key)
   const { rpcUrl, headers } = getTronRPCConfig(networkName, options.verbose)
 
-  consola.info(`RPC URL: ${rpcUrl}`)
+  consola.info(`RPC URL: ${redactUrls(rpcUrl)}`)
 
   // Get the correct private key based on environment
   let privateKey: string
--- a/script/helperFunctions.sh
+++ b/script/helperFunctions.sh
@@ -1203,7 +1203,7 @@ function saveDiamondPeriphery() {
   echoDebug "ENVIRONMENT=$ENVIRONMENT"
   echoDebug "USE_MUTABLE_DIAMOND=$USE_MUTABLE_DIAMOND"
   echoDebug "FILE_SUFFIX=$FILE_SUFFIX"
-  echoDebug "RPC_URL=$RPC_URL"
+  echoDebug "RPC_URL=$(redactRpcUrl "$RPC_URL")"
   echoDebug "DIAMOND_ADDRESS=$DIAMOND_ADDRESS"
   echoDebug "DIAMOND_FILE=$DIAMOND_FILE"
 
@@ -2177,6 +2177,13 @@ function extractFromVerificationOutput() {
 # argument (never over a joined string) so a key containing whitespace or a
 # newline cannot spill its tail past the substitution, and remaining arguments
 # are quoted with %q so their boundaries stay visible in the log.
+function redactRpcUrl() {
+  # Endpoint URLs carry the provider key in the path or query, so the value must never reach a
+  # log line verbatim. Callers that need the URL itself (getRPCUrl returns it on stdout) must
+  # not use this.
+  printf '%s' "$1" | sed -E 's#[a-zA-Z][a-zA-Z0-9+.-]*://[^[:space:]]+#[redacted-url]#g'
+}
+
 function redactVerifyCmd() {
   local PLACEHOLDER='***REDACTED***'
   local OUTPUT=''
--- a/script/helperFunctions.sh
+++ b/script/helperFunctions.sh
@@ -2177,13 +2177,6 @@ function extractFromVerificationOutput() {
 # argument (never over a joined string) so a key containing whitespace or a
 # newline cannot spill its tail past the substitution, and remaining arguments
 # are quoted with %q so their boundaries stay visible in the log.
-function redactRpcUrl() {
-  # Endpoint URLs carry the provider key in the path or query, so the value must never reach a
-  # log line verbatim. Callers that need the URL itself (getRPCUrl returns it on stdout) must
-  # not use this.
-  printf '%s' "$1" | sed -E 's#[a-zA-Z][a-zA-Z0-9+.-]*://[^[:space:]]+#[redacted-url]#g'
-}
-
 function redactVerifyCmd() {
   local PLACEHOLDER='***REDACTED***'
   local OUTPUT=''
@@ -2209,6 +2202,13 @@ function redactVerifyCmd() {
   done
 
   printf '%s\n' "$OUTPUT"
+}
+
+# Endpoint URLs carry the provider key in the path or query, so the value must never reach a log
+# line verbatim. Only a scheme-bearing URL is matched, mirroring the TypeScript redactUrls().
+# Callers that need the URL itself — getRPCUrl returns it on stdout — must not use this.
+function redactRpcUrl() {
+  printf '%s' "${1:-}" | sed -E 's#[a-zA-Z][a-zA-Z0-9+.-]*://[^[:space:]]+#[redacted-url]#g'
 }
 
 function verifyContract() {
--- a/script/helperFunctions.sh
+++ b/script/helperFunctions.sh
@@ -5300,7 +5300,7 @@ function handleForgeScriptError() {
       warning "forge script returned exit code 0 but with unexpected/empty return data${NETWORK_MSG}${ATTEMPT_MSG}"
     fi
     if [[ -n "${STDERR_CONTENT:-}" ]]; then
-      error "stderr: ${STDERR_CONTENT}"
+      error "stderr: $(redactRpcUrl "${STDERR_CONTENT}")"
     fi
     if [[ -z "${RAW_RETURN_DATA:-}" || "${RAW_RETURN_DATA:-}" == "" ]]; then
       warning "No JSON output received. This usually indicates a connection/RPC error."
@@ -6072,7 +6072,7 @@ function estimatePauseCost() {
       return 2
     fi
     if [[ $ATTEMPT -ge $ESTIMATE_MAX_ATTEMPTS ]]; then
-      error "estimatePauseCost: cast estimate failed for $NETWORK after $ESTIMATE_MAX_ATTEMPTS attempts: ${CAST_ERR:-non-numeric gas estimate ($GAS_ESTIMATE)}" >&2
+      error "estimatePauseCost: cast estimate failed for $NETWORK after $ESTIMATE_MAX_ATTEMPTS attempts: $(redactRpcUrl "${CAST_ERR:-non-numeric gas estimate ($GAS_ESTIMATE)}")" >&2
       return 1
     fi
     sleep "$ESTIMATE_RETRY_SLEEP_SECONDS"
--- a/script/helperFunctions.sh
+++ b/script/helperFunctions.sh
@@ -5079,7 +5079,7 @@ function executeAndCapture() {
 
   # Debug: Show what we captured
   echoDebug "=== RAW_RETURN_DATA (stdout) ==="
-  echoDebug "$RAW_RETURN_DATA"
+  echoDebug "$(redactRpcUrl "$RAW_RETURN_DATA")"
   echoDebug "=== STDERR_CONTENT (stderr) ==="
   echoDebug "$(redactRpcUrl "$STDERR_CONTENT")"
 
@@ -5154,7 +5154,7 @@ function parseExecuteCommandResult() {
         error "stderr: $(redactRpcUrl "$STDERR_CONTENT")"
       fi
       if [[ -n "$RAW_RETURN_DATA" ]]; then
-        echoDebug "stdout: $RAW_RETURN_DATA"
+        echoDebug "stdout: $(redactRpcUrl "$RAW_RETURN_DATA")"
       fi
 
       case "$ON_ERROR_ACTION" in
--- a/script/helperFunctions.sh
+++ b/script/helperFunctions.sh
@@ -5081,7 +5081,7 @@ function executeAndCapture() {
   echoDebug "=== RAW_RETURN_DATA (stdout) ==="
   echoDebug "$RAW_RETURN_DATA"
   echoDebug "=== STDERR_CONTENT (stderr) ==="
-  echoDebug "$STDERR_CONTENT"
+  echoDebug "$(redactRpcUrl "$STDERR_CONTENT")"
 
   # Extract JSON if requested
   if [[ "$EXTRACT_JSON" == "true" ]]; then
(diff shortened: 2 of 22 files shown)


script/utils/rpc-url-log-scan.ts
```diff
@@ -0,0 +1,387 @@
...
+      if (depth === 0) return i
+    }
+  }
+  return src.length
+}
+
+/** Argument spans of every redaction call. */
+function redactedSpans(code: string): [number, number][] {
+  const spans: [number, number][] = []
+  const re = new RegExp(String.raw`\b(?:${[...REDACTORS].join('|')})\s*\(`, 'g')
```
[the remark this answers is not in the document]
0xDEnYO: Not fixing the binding resolution: distinguishing the real `redactUrls` from a local shadow of the same name needs a type-checked `ts.Program`, and this scanner runs per source file in the unit-test suite. Documented as a stated limit in the docblock instead (c040a01e4) alongside the other channels it does not reach, so a green run is not read as a closed class.

script/utils/rpc-url-log-scan.ts
```diff
@@ -0,0 +1,256 @@
...
+ * - A computed environment read, `process.env[name]` — the form three scripts here use.
+ * - Anything needing dataflow: a value renamed, destructured to another name, or handed to a
+ *   helper that logs it.
+ * - The specific identifiers a path lists in {@link EXEMPT}; everything else in such a file is
+ *   still scanned. And anything outside the roots the caller passes.
+ */
+import { type Dirent, readFileSync, readdirSync } from 'node:fs'
+import { join, relative } from 'node:path'
+
+import ts from 'typescript'
```
gvladika: **16 new ESLint warnings from this import, against the repo's zero-warnings policy.**

`import ts from 'typescript'` makes every `ts.isIdentifier` / `ts.createSourceFile` / `ts.forEachChild` access trip `import/no-named-as-default-member`:

```
✖ 16 problems (0 errors, 16 warnings)
```

`.agents/rules/000-global-standards.md` asks for every warning fixed, or an inline ignore carrying a justification. Named imports are the cheaper of the two — I confirmed they work here under both bun and `tsc-files --noEmit`:

```ts
import {
  createSourceFile,
  forEachChild,
  isCallExpression,
  isIdentifier,
  isPropertyAccessExpression,
  isQualifiedName,
  isShorthandPropertyAssignment,
  isTypeNode,
  ScriptTarget,
  type CallExpression,
  type Identifier,
  type Node,
} from 'typescript'
```

Either that or a file-level `/* eslint-disable import/no-named-as-default-member -- … */` with the reason.

script/helperFunctions.sh
```diff
@@ -2204,6 +2204,13 @@ function redactVerifyCmd() {
   printf '%s\n' "$OUTPUT"
 }
 
+# Endpoint URLs carry the provider key in the path or query, so the value must never reach a log
+# line verbatim. Only a scheme-bearing URL is matched, mirroring the TypeScript redactUrls().
+# Callers that need the URL itself — getRPCUrl returns it on stdout — must not use this.
+function redactRpcUrl() {
```
gvladika, line 2210: **Two of the seven deferred bash sites are ungated error paths that this new function closes in one line each — worth pulling into this PR rather than EXSC-974.**

The reason for deferring the *scan* is sound: a naive bash scan false-reds on `getRPCUrl`, whose job is to return the endpoint on stdout. But the reason doesn't transfer to two of the deferred *fixes*, which are the identical one-liner already applied at the five sites above and are not debug-gated:

- `script/deploy/deployAndStoreCREATE3Factory.sh:92-93` — `TX_OUTPUT=$(cast send --rpc-url "$RPC_URL" … 2>&1)`, then `error "cast send failed: $TX_OUTPUT"`. `error` always prints, and `cast` puts `--rpc-url` in its error text.
- `script/tasks/diamondSyncWhitelist.sh:1180` — `echo "$OUTPUT"` where `OUTPUT` is `universalCast … 2>&1`. I checked `script/universalCast.sh`: it passes `--rpc-url "$RPC_URL"` to `cast` on every path, so the same keyed URL rides in on failure.

Both run against production diamonds, and both are exactly the class the exhaustion-path fix in `rpcCallWithRetry` was made for — the failure path is the one that actually runs.

The other five deferred sites (`helperFunctions.sh:3598/3696/3851`, `playgroundHelpers.sh:1967/2020`) are genuinely subtler, since they echo a function's stdout return value, so leaving those to EXSC-974 reads right.

script/utils/rpc-url-log-scan.ts
0xDEnYO: Switched to named imports in 2585bb731 — matches `validateScripts.ts` and `cittyArgDefaults.ts`, which already import that way. eslint is 0 problems on the file now.

One note for the record: the warnings never showed up in CI. Nothing runs `lint:js` in a workflow, `eslint --fix` in lint-staged can't autofix this rule, and eslint exits 0 on warnings — so the pre-commit hook passed with all 15 present.

script/helperFunctions.sh
0xDEnYO: Both pulled in, and checking the surrounding paths turned up more of the same shape — thirteen sites now redacted rather than two:

- `diamondSyncWhitelist.sh` had **three** identical verbose-output prints, not one: `983` (removal) and `1243` (combined proposal) carry the same `universalCast`/`sendOrPropose` output under the same gate, only under a different variable name.
- `deployAndStoreCREATE3Factory.sh:117` — `error "... Last stderr: ${STDERR_CONTENT:0:500}"` prints forge's stderr from `--fork-url`, so it leaks on *every* failed factory deploy, not just the `cast send` fallback you pointed at.
- `helperFunctions.sh:5154` and `:5303` — `error "stderr: $STDERR_CONTENT"` in `parseExecuteCommandResult` and the forge-script retry handler. That is the shared executor the CREATE3 path goes through *before* reaching line 92, and the widest one of the set: every bash deploy prints through it.
- `emergencyPauseBreakGlass.sh:453/494` and `diamondEMERGENCYPause.sh:174` — pause-state errors from `bgOwnerCall` / `universalCast`, which don't go through the already-redacted `rpcCallWithRetry`. Incident-path prints.
- `helperFunctions.sh:6075` — `estimatePauseCost`'s cast stderr, on a fleet-wide sweep.

`parseExecuteCommandResult` is now pinned by a test in `script/redactRpcUrl.test.ts`; reverting the redaction fails it.

Your read on the other five holds — `playgroundHelpers.sh:1967/2020` and `helperFunctions.sh:3598/3696/3851` print a function's stdout return value, so telling a returned endpoint from a leaked one needs the EXSC-974 scan work. Ticket updated with what's left.

script/deploy/tron/deploy-core-facets.ts
```diff
@@ -75,7 +76,7 @@ async function deployCoreFacetsImpl(options: {
   // Get RPC URL and API key configuration (automatically handles TronGrid API key)
   const { rpcUrl, headers } = getTronRPCConfig(networkName, options.verbose)
 
-  consola.info(`RPC URL: ${rpcUrl}`)
+  consola.info(`RPC URL: ${redactUrls(rpcUrl)}`)
```
gvladika, line 79: **`displayNetworkInfo()` prints the raw endpoint 36 lines after this redacted one, and the scan reports green on it.**

This line redacts the URL; `115` hands the same `rpcUrl` in raw, and it lands in a `consola.box`:

```ts
// script/utils/utils.ts:966
const infoContent = `
Network: ${networkName}
RPC URL: ${rpcUrl}
...`.trim()
consola.box({ title: 'Network Information', message: infoContent, ... })
```

Run against the real `displayNetworkInfo` with a synthetic key:

```text
 │  RPC URL: https://lb.drpc.org/ogrpc?network=tron&dkey=SYNTHETICKEY-NOT-REAL-abc123  │
```

Same shape as the `detect-evm-version` fix in 74a2974 — a redacted line, then the identical URL one print later, in the same run. Seven Tron deploy scripts pass the raw endpoint into it: `deploy-core-facets:115`, `deploy-and-register-symbiosis-facet:101`, `-layerswap-:105`, `-allbridge-:104`, `-near-intents-:88`, `-eco-:86`, `-lifi-intent-escrow-facet-v2:90`.

**The scanner does not see it.** I ran `scanForRawRpcUrlLogs(repoRoot)` on `d63dbce`: 240 files scanned, `script/utils/utils.ts` among them, **0 findings**. The URL reaches `infoContent` before it reaches `consola.box`, and `infoContent` is not in `RPC_IDENTIFIERS`. So a variable hop is a third channel, alongside the two the docblock declares (bash, error-object embedding) — which matters precisely because that docblock exists so a green run isn't read as a closed class.

CodeRabbit raised this as an outside-diff comment on line 115 in its first review. It has no reply.

Its proposed fix is one line and I verified it — the box prints `[redacted-url]`, and `Network: Mainnet` still resolves because the `rpcUrl.includes('shasta')` check one line above keeps the raw value:

```ts
import { redactUrls } from './redactUrls'
// ...
RPC URL: ${redactUrls(rpcUrl)}
```

I swept the rest of the class: the only other raw interpolation of an endpoint identifier under `script/`/`tasks/` is `demoScripts/demoPaxosTransit.ts:199`, which is the declared `EXEMPT` local anvil. One site, seven callers.

Minor, same file: `utils.ts:76` throws ``invalid uri or network not supported by node provider : ${uri}`` with the keyed endpoint in the message. It needs an unsubstituted `{{` placeholder to reach, and error construction is a declared EXSC-974 limit — but it is a one-liner in a file this change would already touch.

0xDEnYO:
@coderabbitai review

script/deploy/tron/deploy-core-facets.ts
0xDEnYO: Both fixed in cba9381f3, and the sweep turned up two more of the same shape.

`displayNetworkInfo` now takes `redactUrls(rpcUrl)`. Ran the real function with a synthetic key:

```text
 │  Network: Mainnet         │
 │  RPC URL: [redacted-url]  │
```

`Network: Mainnet` still resolves — as you noted, the `shasta` check reads the raw parameter, which the redaction does not move. `utils.ts:76`'s `node_url` throw is redacted too. Both are pinned by tests in `script/utils/utils.test.ts` and mutation-checked: reverting either redaction fails exactly one named test, restore is byte-identical.

On the scanner: the variable hop is not a fourth channel — it is the one the docblock already declares as *"anything needing dataflow: a value renamed, destructured to another name, or handed to a helper that logs it."* `infoContent` is the "renamed" case. That the declared limit had a live instance behind it is the fair hit, and it is now closed.

Two more found while checking the class, same asymmetry you spotted (a redacted line beside a raw one, same capture, same function):

- `executeAndCapture` redacted the captured stderr and printed the captured stdout raw one line above it.
- `parseExecuteCommandResult` did the same on its error path.

Both are prints only, so nothing that parses `RAW_RETURN_DATA` moves.

Also corrected the rule this PR added: `[CONV:REDACT-RPC-URL]` still ended *"nothing checks bash automatically"*, which this PR made untrue, and it prescribed only `redactRpcUrl` — so the next agent editing `script/emergency/` would have "fixed" `bgRedactUrl` to the helper that script must not depend on. `300-bash.md` now carries the pointer, since a `.sh` edit never loads the TypeScript-globbed rule.

One open item I have **not** touched, flagging rather than deciding: `bgSendPause` and `bgPauserWalletCall` in the break-glass script run `cast` with stderr neither captured nor redirected, so a failed pause send prints the keyed URL verbatim — on the one path that only ever runs during a real incident. Reproduced with a stubbed `cast`. It is a genuine gap in this PR's stated goal, but closing it means capturing output in frozen incident-critical code, which is a behaviour change I would rather Daniel signs off than land quietly under a redaction PR. Happy to do it here if you would rather it not wait.

gvladika (approved):

melianessa (approved):

PR #2369, opened 2026-09-11
feat(safe): make the check ledger the sign-time composition point (EXSC-993)

0xDEnYO:

Fixes [EXSC-993](https://linear.app/lifi-linear/issue/EXSC-993/wp-41-wiring-make-the-check-ledger-the-sign-time-composition-point) — WP-4.1 wiring · Make the check ledger the sign-time composition point

WP-4.1's `check-ledger.ts` is the composition point of the design: `createCheckLedger` → `recordCheck` → `rollUpChecks` → `summariseLedger` over a fixed `AnchorId` vocabulary. On `origin/main` before this branch, `recordCheck` had **zero production callers**. Three merged gates sat unreached and the functions encoding the grading rules — `summariseLedger`, `isTriageRelaxationAllowed` — were unreachable in a real run. Each gate printed its own block in its own idiom and nothing computed "can this proposal proceed" across them, which is precisely what WP-4.1 exists to prevent.

**The translation layer is its own module.** `confirm-check-registry.ts` maps a gate's own verdict to a ledger row. It does not belong in `check-ledger.ts`, which must not know about any particular gate, nor in the gates, which must stay usable without a ledger. A third file is the only place that dependency direction works.

**The ledger is created once the run's network set is known**, not at startup. The denominator has to be the real fleet: a check that never ran on a network must show as a missing row rather than quietly shrink the total it is measured against.

**The status mapping is an exhaustive `Record`, not a switch with a default.** A `TargetStateStatus` added later fails `tsc -p` instead of defaulting to green. Worth knowing where that guarantee stops: CI runs `tsc-files` over *changed* files only, and `pinned-target-state.ts` does not import this module, so a PR that adds a status and nothing else still lands green and throws at runtime instead. Record-derived statuses are anchored `A-MONGO` and so cannot produce a pass — `recordCheck` coerces them, and a test proves the coercion rather than assuming it.

The cross-check test is the one worth reading: the mapping and `ITargetStateVerdict.cleared` encode "may proceed" independently, so it asserts they agree on **every** status. A drift between them would either block a clean proposal or green a blocked one, and nothing else would notice.

Verified at `131c30a8f`. The wiring is live: `createCheckLedger` at `confirm-safe-tx.ts:1473`, `recordCheck` at `confirm-safe-tx.ts:152` and `:1171`. `renderCheckLedger` is deliberately **not** called — see below.

Suites on this branch:

- `forge test --evm-version cancun` — 1686 passed, 0 failed (92 suites)
- `bun test script/` — 4351 pass, 0 fail (157 files) — this is the repo's TS suite per `package.json`
- `bunx tsc-files --noEmit` over the changed files — clean

**Round 3 — what each row is judged against.** `expected` was derived from the row's anchor, but an anchor records where the evidence came from, not whether a version comparison happened. Four statuses reach `compareSemanticVersions` and nine can decide a row, so keying on `A-LOCAL` fixed the two passing rows and left seven — `not-previously-targeted` among them — claiming an ordering against `origin/main` their branch never established.

It is now carried per status through the same worst-first reduction as `anchor` and `detail`, as one of three sentences, each stating a **requirement** rather than describing the row. That distinction is load-bearing: `matches-main`, `ahead-of-main` and `not-previously-targeted` all tie at `needs-ack`, so on the ordinary cut shape — Add a new facet beside a Replace of a live one — a diagnosing sentence made the row's truth depend on which element calldata listed first. The three share one sentence, so the tie is harmless without tie-breaking machinery. A diagnosis would also have been wrong wherever `origin/main` declares nothing for another reason, such as a facet deprecation that removed the entry — the case a signer most needs flagged.

Pinned over all thirteen statuses against the module's own constants, so a status added later is a compile error in both the mapping and the test table.

The render's **absence** is pinned too (`check-ledger-wiring-placement.test.ts`). Restoring the call site was one line that left the whole suite green. `summariseLedger` is pinned beside the renderer because the inversion lives there — the render is a thin wrapper over it and this file already imports from that module, so printing its verdict directly would restore the display without naming the renderer. EXSC-994 deletes that assertion in the commit that lands the corrected verdict.

**Correction to an earlier version of this description.** It said `renderIntegrityAsserts` was a "parallel path" to be dropped before this PR went ready. That was wrong, and the reason matters: `confirm-integrity-asserts.ts:1124` calls `createCheckLedger` and builds its **own** ledger instance, which `renderIntegrityAsserts` renders. It is not a second view of the same data — it is a second ledger. Deleting that render would remove every integrity-assert verdict from the signer's screen rather than deduplicate a display.

So the integrity-assert block is currently the only ledger block on screen, and unifying the two ledgers is real work with an order to it: integrity asserts must first record into the run-level ledger, and only then can the separate render go. That is EXSC-994's scope, alongside the four unwired gates, because it is the same question — which ledger is *the* ledger. Doing it here would also collide with EXSC-994, which is stacked on this branch and editing the same function.

**The ledger is composed but not rendered, and that is the point of the last round.** `summariseLedger` counts only a `pass` as verified. Every `TargetStateStatus` a correct Add/Replace cut produces — `matches-main`, `ahead-of-main`, `not-previously-targeted` — maps to `needs-ack`, and the operator's actual acknowledgement goes to the separate `acknowledgementLedger` (`confirm-safe-tx.ts:1053`), which this ledger never sees. The two consequences point opposite ways:

- a correct multi-network upgrade closes `0/N network results verified` with a `NEEDS REVIEW` row per network, asking the operator to review changes they reviewed and signed minutes earlier;
- a run that graded nothing — no pending transaction, nothing actionable, signer not an owner — records only synthetic passes, which the renderer suppresses, and closes `VERDICT: ALL CHECKS GREEN`.

Together: the only way to earn a green ledger is to grade nothing. Neither can green a proposal that was actually signed — `targetState.cleared` gates that independently — so this is not a security hole. But a composition point that shows yellow on every good run and green on every empty one is how a security display stops being read, so it does not go in front of a signer in this state.

So this PR lands the composition and withholds the verdict. [EXSC-994](https://linear.app/lifi-linear/issue/EXSC-994) owns the display, including whether it should exist at all, and two judgements that go with it: whether `matches-main`/`ahead-of-main` should be anchored `A-MAIN` rather than `A-MONGO` (`check-ledger.ts:31` documents the anchor as where the *expectation* came from, which points that way — but the proposed version is resolved through the proposer-written deployment record, so promoting it to an anchor that can decide a pass weakens the guarantee rather than fixing a mislabel), and draining `proposalChecks` in a `finally` so a mid-network throw stops discarding verdicts already graded.

**Not in scope:** wiring the four gates that still have no production caller (`calldata-address-check`, `executability-simulation`, `rpc-quorum`, `timelock-cancel-decision`). That is a follow-up and depends on this landing first.


0xDEnYO:
@coderabbitai review

gvladika (commented):
Reviewed the wiring end-to-end and drove the real modules to see what a signer actually gets.

**Verified locally** (worktree at `edd701ba`):

| Check | Result |
|---|---|
| `bun test` on the 4 ledger test files | 130 pass, 0 fail |
| `bunx tsc-files --noEmit` on all 5 changed files | clean |
| `bunx eslint` on the same | clean |
| `bun test script/` | 4344 pass, 5 fail — all in `verify-immutable-registry CLI`, unrelated to this PR and caused by my sandbox worktree re-resolving dependencies. CI's `run-ts-tests` is green. |

The wiring itself is correct and the tests are unusually good — the exhaustive `Record`, the cross-check against the real `STATUSES_CLEARED_TO_PROCEED` rather than a copy, the worst-first reduction with its documented tie-break, and the `recordNothingToGrade` / `recordCouldNotGrade` split all do what they claim.

Five inline comments. The first one is the only one I would want settled before this display goes in front of signers; the rest are small.

One process note: **CodeRabbit never actually reviewed this PR** — both the initial run and the `@coderabbitai review` retry came back rate-limited. There is no bot backstop on it right now, so it is worth re-triggering.

Minor things not worth their own thread:
- `recordEveryCheck` writes one `expected` string across every registered check; it will read oddly once EXSC-994 adds the four unwired gates.
- `STATUSES_CLEARED_TO_PROCEED` is exported solely for the test, which is fine here — cross-checking against the real set is the point.

script/deploy/safe/confirm-check-registry.ts
```diff
@@ -0,0 +1,220 @@
...
+ * The three unresolvable statuses reach `A-UNRESOLVED` because nothing
+ * answered at all: an action that is not Add, Replace or Remove, calldata that
+ * could not be read, or an anchor that could not be reached.
+ *
+ * Keyed exhaustively so a status added to `TargetStateStatus` fails to compile
+ * here rather than falling through to a default that would grade it green.
+ */
+const STATUS_MAPPING: Readonly<Record<TargetStateStatus, IStatusMapping>> = {
+  'matches-main': { status: 'needs-ack', anchor: 'A-MONGO' },
+  'ahead-of-main': { status: 'needs-ack', anchor: 'A-MONGO' },
```
gvladika: **The verdict line ends up inverted: green means "nothing was checked".**

These three statuses — `matches-main`, `ahead-of-main`, `not-previously-targeted` — cover every real Add/Replace cut, and all three map to `needs-ack`. Since `pass` is the only state that counts toward "verified", I drove both shapes through the real modules:

```text
--- A: normal upgrade rollout, everything correct, 3 networks ---
✗ Intent      0/1 checks green · 0/3 network results verified · 3 needs review
      mainnet    NEEDS REVIEW  ... → review the change and acknowledge it
      arbitrum   NEEDS REVIEW  ... → review the change and acknowledge it
      optimism   NEEDS REVIEW  ... → review the change and acknowledge it
VERDICT: ACKNOWLEDGEMENT REQUIRED — 3 results awaiting review · 0/3 network results verified

--- B: run where every network had nothing actionable ---
✓ Intent      1/1 checks green · 3/3 network results verified
VERDICT: ALL CHECKS GREEN — 1/1 checks, 3/3 network results verified
```

Case B is the limitation the description discloses and defers to EXSC-994. Case A is not mentioned anywhere and is the everyday path: a correct 40-network rollout closes with `0/40 network results verified` and 40 cyan rows telling the operator to review changes they reviewed and signed minutes earlier. Together the two mean the only way to get a green ledger is to grade nothing.

Two details that make it worse:

1. The `needs-ack` is never discharged. The operator's real acknowledgement goes to the separate `acknowledgementLedger` (`confirm-safe-tx.ts:1054`); this ledger never sees it, so the closing line asks for a review that already happened.
2. The two modules disagree about what `anchor` means. `check-ledger.ts:31` documents it as *"where a result's expectation came from"*; this mapping picks the weakest evidence anywhere in the chain. Under the first reading, `matches-main` / `ahead-of-main` are graded against `origin/main` → `A-MAIN`, which *can* decide a pass, and a clean rollout goes green.

None of this can green a proposal that was actually signed — `targetState.cleared` still gates that — so it is not a security hole. But a composition point that shows yellow on every good run and green on every empty one is how a security display stops being read. I would want one of: settle the anchor semantics so a fully-graded clean network counts as verified; feed the run's acknowledgement back into the row before rendering; or land EXSC-994 first. Whichever way it goes, the description should carry the case-A half of the limitation too, not just case B.

script/deploy/safe/confirm-safe-tx.ts
```diff
@@ -1079,6 +1159,18 @@ const processTxs = async (
         consola.error('Error executing with deployer:', error)
       }
   }
+
+  // One row per network, written once every proposal on it has been graded and
+  // reduced worst-first. A `ready` network always carries at least one proposal,
+  // so the empty branch is the unreachable case made explicit rather than left
+  // to roll up as a missing row and block the run.
+  if (checkLedger) {
```
gvladika, line 1175: **A throw inside `processTxs` discards verdicts that were already computed.**

`proposalChecks` is local and only drained here, at the very end of the function. Any throw mid-network — Safe client, RPC, anything in the per-proposal body — loses the rows already graded, and the network then reports *"no result recorded"* instead of the `downgrade` mismatch it had actually found.

Both outcomes block, so severity is preserved, but the reason is not, and a mismatch is the more actionable line to lose. Wrapping the proposal loop in `try { … } finally { … }` with this block in the `finally` keeps the evidence for a few lines of diff.

script/deploy/safe/confirm-safe-tx.ts
```diff
@@ -1467,6 +1581,19 @@ const main = defineCommand({
...
+      // the run, while the roll-up below only counts what the operator acted on.
+      // It reports and does not gate — signing already happened per proposal,
+      // refused there by `targetState.cleared` and by the sign-time gates.
+      //
+      // Rendered whatever the ledger holds. A ledger with no results renders a
+      // BLOCKED verdict counting every expected network as unverified, which is
+      // the report a run that aborted before its first proposal most needs to
+      // print — for aborts after the ledger exists; an earlier one leaves it
+      // undefined and prints nothing.
+      if (checkLedger)
```
gvladika: **Only unguarded call in a `finally` whose whole purpose is running on an aborted run.**

Everything around it is `.catch(() => undefined)`-guarded for exactly that reason. `summariseLedger` throws by contract on a degenerate ledger (`check-ledger.ts`, the "Refusing to summarise a ledger that verifies nothing" branch). `createCheckLedger` makes that unreachable today, but if it ever fires here it masks the run's real error *and* suppresses the Change Review Summary and Execution Summary printed below it — the two things the operator needs when a run aborts.

A `try/catch` with a `consola.warn` costs nothing and keeps the rest of the block reachable.

script/deploy/safe/check-ledger-wiring-placement.test.ts
```diff
@@ -0,0 +1,285 @@
...
+    expect(networkLoop).toBeGreaterThan(-1)
+    // Created from the run's own network set, and before anything is graded: a
+    // ledger built later would take its denominator from whatever had already
+    // been reached.
+    expect(created).toBeLessThan(networkLoop)
+    // The filter body, not just the call: `toContain` on the opening line stays
+    // green while the predicate grows a condition that drops networks from the
+    // denominator, which is the one thing fixing it early is meant to prevent.
+    expect(SOURCE).toMatch(
+      /expectedNetworks: networks\.filter\(\(network\): network is string =>\s*Boolean\(network\)\s*\),/
```
gvladika, line 117: **Source-text assertions pinned to formatting.**

The justification in the file header is real — `confirm-safe-tx.ts` calls `runMain` at module scope (`:1636`), so it cannot be imported — and this file argues its own case better than most. But assertions like this one, and `/if \(checkLedger\)\s*\n\s*renderCheckLedger\(/` further down, fail on a renamed lambda parameter or a prettier reflow with no behaviour change at all.

The cheaper structural fix is to move `recordNothingToGrade`, `recordCouldNotGrade` and the per-network reduce into `confirm-check-registry.ts`, taking the ledger as a parameter instead of closing over the module-level one. They become directly unit-testable, the status/anchor pins in this file stop being necessary at all, and the source-grep shrinks to "these functions are called".

Worth doing while the surface is one gate — EXSC-994 adds four more to the same wiring.

script/deploy/safe/check-ledger-wiring-placement.test.ts
```diff
@@ -0,0 +1,285 @@
...
+    const declined = body.indexOf("if (action === 'Do Nothing') continue")
+
+    expect(declined).toBeGreaterThan(-1)
+    expect(graded).toBeLessThan(declined)
+  })
+
+  it('records nothing from inside the loop', () => {
+    const { body } = proposalLoop()
+
+    // The bug: `recordCheck` per proposal. `rollUpChecks` treats two records for
```
gvladika: **Comment archaeology, per `.agents/rules/000-global-standards.md`.**

Density is high across the new files — 94 comment lines of 220 in `confirm-check-registry.ts`, 101 of 285 here. Most of it is genuine *why* and earns its place. But several comments restate the debugging history, which the rule explicitly routes to the PR body or the ticket rather than the source:

- this one — *"The bug: `recordCheck` per proposal…"*
- `:12` — *"…which left the whole suite green while three merged gates sat unreached"*
- `confirm-check-registry.test.ts:222` — *"Before the deployment-record statuses became `needs-ack` this ordering was unreachable"*

The invariant each one protects is worth stating; the history of how it was discovered is not. Suggest keeping the first half of each and dropping the second.

script/deploy/safe/confirm-check-registry.ts
```diff
@@ -0,0 +1,220 @@
...
+
+  const failing = verdict.findings.filter(
+    (finding) => STATUS_MAPPING[finding.status].status !== 'pass'
+  )
+
+  return {
+    checkId: TARGET_STATE_CHECK_ID,
+    network,
+    status,
+    expected: 'every installed version at or ahead of origin/main',
```
gvladika: Minor: a `removal` or `no-diamond-cut` row carries this `expected` even though neither read `origin/main` at all — the mapping comments above say as much. Invisible today because the renderer suppresses pass rows, so it only surfaces if one of them is ever superseded into a visible row. Worth making the expectation conditional on what was actually compared, or leaving a note that it is deliberately generic.

0xDEnYO:
@coderabbitai full review

script/deploy/safe/confirm-check-registry.ts
0xDEnYO: Agreed, and case A is the one that matters — I had only found the empty-run half and documented that, which described the rarer failure and missed the everyday one.

Taken the third option: the render is unwired in `b5799792f`, so no verdict line reaches a signer until EXSC-994 settles what it should say. The recording, the registry and the cross-check tests stay. The PR body now carries both halves and states the conclusion plainly — the only way to earn a green ledger today is to grade nothing.

On the anchor: I would not make that change without you agreeing to it explicitly, because I think it is a safety call rather than a mislabel. `matches-main` resolves the *proposed* version through the deployment record, which the proposer writes — so promoting it to an anchor that can decide a pass lets the proposer partly grade their own side of the comparison, which is what `A-MONGO` exists to prevent. Your reading of `check-ledger.ts:31` is right about intent, which is exactly why the two readings need reconciling rather than one being applied. Carried into EXSC-994 alongside the acknowledgement-feedback option, which is the one I would pick since it fixes case A without touching the anchor.

Daniel also raised whether this display is worth having at all, so EXSC-994 owns that question too, not just the fix.

script/deploy/safe/confirm-safe-tx.ts
0xDEnYO: Correct, and no try/catch exists anywhere in that 630-line loop, so a throw escapes and the accumulated rows go with it.

Not fixed here: wrapping the loop means reindenting its whole body, and with the render unwired the lost reason has no observable consequence yet. It is queued on EXSC-994 with the display it serves, so the diff lands next to the thing that makes it visible rather than as 600 lines of whitespace in a PR you are re-reading.

script/deploy/safe/confirm-safe-tx.ts
0xDEnYO: Moot now — the render is gone from the `finally` in `b5799792f`, so there is no unguarded call left in that block. If EXSC-994 puts it back, it goes back guarded.

script/deploy/safe/check-ledger-wiring-placement.test.ts
0xDEnYO: Fair, and the two assertions you point at are gone: the render ones went with the render, and I had tightened the denominator grep in the previous round for a different reason — it passed against any filter body, so `Boolean(network && txsByNetwork[...])` kept it green.

Not taking the structural move here. I agree with it and with your reasoning for doing it while the surface is one gate, but it rewrites the wiring this PR exists to introduce, and EXSC-994 is stacked on this branch editing the same function — so doing it here maximises the collision. Noted on that ticket as the shape to land before four more gates arrive.

script/deploy/safe/check-ledger-wiring-placement.test.ts
0xDEnYO: Right on all three, and my own review pass got this wrong — it checked density against the sibling files and cleared it, which is not the test the rule states.

Trimmed in `b5799792f`: each keeps the invariant and drops the history. The header list lost the "which left the whole suite green" clause and the render bullet, the per-proposal comment now states what `rollUpChecks` does rather than what the bug was, and `confirm-check-registry.test.ts` drops the "before the deployment-record statuses became needs-ack" sentence.

script/deploy/safe/confirm-check-registry.ts
0xDEnYO: Fixed in `b5799792f` — `expected` is now conditional on the anchor, so an `A-LOCAL` row says the cut was read and found to install nothing requiring a version comparison instead of claiming a comparison against `origin/main` its branch never made.

script/deploy/safe/check-ledger-wiring-placement.test.ts
0xDEnYO: Correction to my reply above: only the render regex is gone. The assertion you anchored this on — the `expectedNetworks` filter-body match at `:115` — is unchanged, and tightening it last round made your point stronger, not weaker: it now pins the predicate's exact spelling, so a renamed lambda parameter or a prettier reflow reds it with no behaviour change at all.

Still not taking the structural move in this PR, for the collision reason above. But it is an open item on EXSC-994, not something this round addressed.

0xDEnYO:
@coderabbitai review

script/deploy/safe/confirm-check-registry.ts
```diff
@@ -0,0 +1,293 @@
...
+}
+
+/**
+ * What each status says the network should have.
+ *
+ * Every one is a requirement, never a description of the row — `expected` is
+ * rendered verbatim after the word "expected", and a sentence that diagnoses
+ * the situation instead ("a first deployment") is false as soon as one cut
+ * carries two elements. It also makes the reduction below safe: findings of
+ * equal rank keep the first in calldata order, which only ever costs
```
gvladika, line 54: **The tie-order invariant holds for `needs-ack` but not for `error`.**

This paragraph, and the new *"reads the same whichever tied finding calldata listed first"* test, set the standard: findings of equal rank keep the first in calldata order, so tied statuses must carry the same sentence or the row's truth depends on element order.

That holds for the `needs-ack` group (all `ORDERING_HOLDS`) and the `pass` group (all `NOTHING_TO_COMPARE`), and `fail` has only one member. The `error` group is mixed: `version-not-comparable` is `ORDERING_HOLDS` while the other five are `EVERY_ELEMENT_COMPARED`. Driven through the real reducer:

```text
[version-not-comparable , contract-unidentified] -> expected="no installed version behind what origin/main declares"
[contract-unidentified , version-not-comparable] -> expected="every installed element compared against origin/main"
```

The anchor moves with it, since `rank < worstRank` is strict and both sit at the same rank:

```text
[version-not-comparable , unrecognised-cut-action] -> anchor=A-MONGO
[unrecognised-cut-action , version-not-comparable] -> anchor=A-UNRESOLVED
```

So a two-element cut that fails to grade for two different reasons reports a different anchor depending on which element calldata listed first — the exact shape the new test rules out one rank up. The anchor half predates this round; the `expected` half arrives with it.

The paragraph above hedges this as *"only ever costs specificity when both sentences are requirements"*, which is true for `expected` — both readings are honest — but it does not cover the anchor, which is a claim about evidence rather than about the requirement, and `A-MONGO` vs `A-UNRESOLVED` are not interchangeable.

Not urgent while nothing reads the rows. Two cheap options when EXSC-994 gets there: order the `SEVERITY` comparison within a rank by anchor strength so the weakest evidence always wins, or extend the tie-order test to the `error` pairs and pick one sentence for the group. I would take the first — it makes the existing "never claim a stronger anchor than the thing that decided it" comment true in the tied case too, which is what it already reads as promising.

gvladika (commented):
Re-reviewed at `131c30a8f`.

**Verified locally:** 132 pass / 0 fail across the four ledger test files; `bunx tsc-files --noEmit` and `bunx eslint` clean on all changed files; CI green (only `skipping` jobs are non-pass). CodeRabbit has now run and posted nothing inline.

**Where my six comments landed:**

| # | Finding | Status |
|---|---|---|
| 1 | Verdict inverted | Addressed by withholding — `b5799792f` unwires the render, so no verdict reaches a signer. The inversion itself is EXSC-994's. Thread left open deliberately: the defect still exists in the composed rows. |
| 2 | Throw discards graded verdicts | Deferred to EXSC-994. Agreed — with nothing reading the ledger the lost reason has no observable consequence, and the reindent belongs next to the consumer. |
| 3 | Unguarded render in `finally` | Moot, render is gone. Resolved. |
| 4 | Comment archaeology | Fixed in `b5799792f`, all three spots. Resolved. |
| 5 | Source-grep brittleness | Partly — the render assertions went with the render; the `expectedNetworks` filter-body regex is unchanged, as your own correction notes. Structural move deferred to EXSC-994; agreed on the collision reasoning. Thread left open. |
| 6 | `expected` claimed an untested comparison | Fixed, and more thoroughly than I asked — per-status sentences, pinned against the module's own constants, with a tie-order test. Resolved. |

On the anchor question: your reasoning is the stronger one and I am not asking for the change. `matches-main` resolves the proposed version through the proposer-written record, so promoting it to an anchor that can decide a pass would hand the proposer one side of its own comparison — that is a real weakening, not a mislabel correction. Reconciling `check-ledger.ts:31`'s wording with what the registry actually does is the part worth carrying to EXSC-994, and the acknowledgement-feedback option fixes case A without touching it. Agreed with your pick.

One new finding from this round, inline — the tie-order invariant you just pinned holds for the `needs-ack` group but not for the `error` group. Small, and invisible while nothing reads the rows, but it is the same property so it may as well be consistent.

One standing note, not a finding: with the render unwired, `render-check-ledger.ts` (315 lines) has no production caller at all — only its own test. That is the honest cost of staging it this way and it is the right trade against showing an inverted verdict, but it does mean the repo carries a composed-but-unread ledger plus an unused renderer until EXSC-994 lands. Worth keeping that ticket close behind this one.

script/deploy/safe/confirm-safe-tx.ts
gvladika: Moot — the render is gone from the finally block, so there is no unguarded call left there. Resolving.

script/deploy/safe/check-ledger-wiring-placement.test.ts
gvladika: Confirmed trimmed at all three spots — each keeps the invariant and drops the history. Resolving.

script/deploy/safe/confirm-check-registry.ts
gvladika: Confirmed fixed, and better than asked: per-status sentences pinned against the module's own constants, plus a tie-order test. Resolving.

gvladika (approved):

PR #2370, opened 2026-09-11
feat(safe): wire the four orphan gates onto the check ledger (EXSC-994)

0xDEnYO:

Fixes [EXSC-994](https://linear.app/lifi-linear/issue/EXSC-994/wire-the-four-orphan-gates-onto-the-check-ledger)

#2369 (EXSC-993) has merged, so this now targets `main` directly. It was stacked on `feat/exsc-wire-check-ledger` while that was open; the history carries those merges.

Four modules had zero production importers on `main` — their decision functions were unit-tested, but nothing a signer runs ever took the decision. Each one is wired to the place where its verdict can actually change an outcome, and three of the four now record onto the run-level check ledger.

**The ledger rows, and the anchors they rest on**

`executability` (semantic, `A-CHAIN`) and `rpc-quorum` (semantic, `A-CHAIN`) are new registered checks. The five always-on proposal-integrity assertions plus `INT-TIMELOCK-DELAY` are now mirrored onto the run-level ledger too — `confirm-integrity-asserts.ts` was building a second, separate ledger, so the "composition point" only ever composed target-state.

The distinction the whole change turns on is what a gate with nothing to judge records. A gate that *read* its evidence and found nothing to do passes on the anchor it read — `INT-TIMELOCK-DELAY` on a proposal carrying no schedule is a pass on `A-LOCAL`, the same shape as target-state's `no-diamond-cut`. A gate that could not open the envelope errors on `A-UNRESOLVED` and blocks. Getting those two backwards is how a fully green run verifies nothing, so each mapping is tested in both directions.

**`rpc-quorum` registers report-only** and never records a `fail`. Its hard-block needs ≥2 independent providers on every production chain, and roughly 18 are still single-endpoint; enforcing it fleet-wide would convert missing redundancy into a refusal to sign. A shortfall is recorded as an acknowledgement instead — on the signer's screen, not blocking the run.

**`calldata-address-check` deliberately has no ledger row.** Its only decidable source is the deploy log (`A-MONGO`), which `recordCheck` coerces a pass on to `error`, and no committed file supplies the address→identity `expectations` an anchor would need. A row for it would therefore report `BLOCKED` on every run. It is wired as a rendered gate instead, so it surfaces before signing as WP-3.3 asks, and its ledger row stays blocked on an address-bearing anchor (WP-2.1).

**`timelock-cancel-decision` is placed in the executor, report-only.** Its consumer is `execute-pending-timelock-tx.ts`, not the confirmation CLI. It sits on the reverting rows, where "execute / cancel / hold / block" is a live question and the executor currently blocks on a revert count alone — not on the healthy path, where a cancel matrix has nothing to decide. It does not drive the action: the `integrity` leg needs an execute-time re-derivation of the attested build, which does not exist, and `unsupported` short-circuits to `hold` + `page` before any other branch, so enforcing it would hold and page every operation on every pass. The identity leg is real — the controller is asked what its own call list hashes to and that is compared with the id the row was scheduled under. A test proves that nothing the executor can assemble reaches `cancel`.

**Order is pinned over the recorded ledger sequence**, not over a source-text probe. `recordProposalChecks` is one ordered step, so the sequence is a property of the rows a test can read back rather than of where calls sit in a 1400-line CLI — that assertion anchors on no source literal, which is what broke two earlier PRs when #2350 deleted a string one anchored on. The placement tests still do: the two new placement files add ~13 `source.indexOf('literal')` probes for the call sites the ledger sequence cannot see, in the pattern `codehash-gate-placement.test.ts` already uses.

**Collectors.** The decision modules were already pure; what was missing between them and a signer was the reads. Three small collectors do that and nothing else — every read that fails is left absent rather than defaulted, because the decision modules grade an absent observation as unchecked and an unchecked one as an error. A collector that filled in a plausible value would turn "nobody asked" into "the chain said yes". `diamond-cut-calls.ts` gained `selectors`, `initCalldata`, `raw`, and an optional target/caller context so a timelock-wrapped cut can be simulated from the timelock against the diamond, replaying the cut's own bytes — replaying the envelope would only prove the proposal can be queued.

**Kept, deliberately:** `renderIntegrityAsserts` at its existing per-proposal call site. The run-level ledger renders after the whole run, i.e. after the signing decision, so dropping the per-proposal render would move integrity verdicts to after the point the signer needs them.

**Tron** is out of the EVM simulator's scope, which is recorded as an acknowledgement naming that limit — distinct from an EVM network the simulator covers but could not reach, which stays an `error` and blocks.

**What these rows do and do not do.** They report. `confirm-safe-tx.ts` calls no
`summariseLedger`, no `assertProposalWouldExecute`, no `assertRpcQuorum` and no
`assertCalldataAddressesResolve` — the only refusals that stop a signature remain the codehash
gate and the integrity assertions. An `executability` mismatch puts a red row and a `VERDICT:
BLOCKED` line in front of the signer; it does not itself prevent signing. Promoting any of them
to a refusal is a separate decision, because each one's false-positive rate on the real fleet is
still unmeasured.

**Two findings a reviewer will not infer from the diff.**

*A fallback transport turns a revert into a green row.* An earlier revision of this PR routed the
executability simulation through viem's `fallback`. viem stops failing over only for a revert whose
message matches `"execution reverted"` — a node answering `-32000 "Reverted 0x…"` (Erigon,
Nethermind, Hardhat) is treated as unreachable, so failover continues and **the next endpoint's
success becomes the answer**. A proposal that reverts on the primary could record `succeeded`. The
simulation now walks endpoints itself: a revert from any endpoint settles the question, and only
infrastructure failures move to the next one. This is why `staticCall` is deliberately not behind a
fallback transport while every other read is.

*The gate that refuses a signature had no failover.* `codehash-sign-gate-deps.ts` read
`chain.rpcUrls.default.http[0]` — the primary alone, with the 75 configured `_FALLBACKS` unused —
so a throttled primary made the codehash gate unverifiable and blocked the signer. It now keeps the
primary's answer where the primary answers, and where it cannot, reads the remaining endpoints
together through `collectProviderObservations` + `evaluateRpcQuorum` — no single fallback decides a
read that gates a signature. Not `getFallbackTransportForChain`: that helper is used by the
executability simulators in `confirm-safe-tx.ts`, while this gate got its own primary-then-quorum
reader. `MIN_INDEPENDENT_PROVIDERS` is 2, so this recovers chains with two independent fallbacks;
a single-fallback chain still blocks when its primary is down. Never worse than reading the primary
alone, and not broader than that.

**Known and accepted, tracked as EXSC-1003.** A run that graded nothing closes `9/9 ALL CHECKS
GREEN` while a correct rollout closes `8/9 ACKNOWLEDGEMENT REQUIRED`, because `target-state` grades
`needs-ack` on `matches-main`. EXSC-993 withheld the render for this; registering the other rows
narrowed it from `0/N` vs `N/N` but did not reverse the direction. The render is restored anyway:
an `executability` row that records `error` blocks, and a block the signer never sees is worse than
no check at all.

## Falsification

Every row `proposalCheckResults` produces was removed in turn and the suite confirmed red each
time — a test that still passes with the call removed is observing nothing, which is the failure
this ticket exists to fix.

Mutation-tested beyond that, after review passes found checks the suite could not see fail.
Hardcoding `initCalldata` to `'0x'`, dropping the singular `schedule` envelope's target, defaulting
a failed `owner` or `facetAddress` read, narrowing revert detection back to `"execution reverted"`,
and returning the codehash read to the primary endpoint alone all left the suite green at some
point. Each now fails.

One reported CRITICAL turned out to be defence in depth rather than a live hole: flipping
`unresolved()` to `pass` leaves the suite green only because `recordCheck` coerces a pass on a
reporting-only anchor back to `error`. The shape that would actually matter — a pass on a deciding
anchor — fails eight tests.

**The rows were falsified; the call sites were not.** Deleting a gate's *call site* — as opposed to
its row — left the suite green for all four gates, which is the same blindness in a different
place: nothing observed that the CLI still reaches the module. `proposalCheckResults` cannot see
it, because a verdict it is never handed falls through to the unresolved branch and records an
`error`, so the run blocks for an evidence reason and the gate reads as wired. Three placement
files close it, and each assertion is pinned by the mutation it exists to catch:

| mutation | killed by |
| -- | -- |
| delete the `renderCalldataAddresses` call | `calldata-address-gate-placement.test.ts` |
| delete the whole calldata gate block | `calldata-address-gate-placement.test.ts` (5 tests) |
| index the deploy record on an empty address list | `calldata-address-gate-placement.test.ts` |
| swallow a failed calldata read silently | `calldata-address-gate-placement.test.ts` |
| drop `executability` from the recorder's arguments | `check-ledger-wiring-placement.test.ts` |
| drop `rpcQuorum` from the recorder's arguments | `check-ledger-wiring-placement.test.ts` |
| default a failed `executability` read in the `catch` | `check-ledger-wiring-placement.test.ts` |
| default a failed `rpcQuorum` read in the `catch` | `check-ledger-wiring-placement.test.ts` |
| delete the executor's cancel-matrix block | `timelock-cancel-placement.test.ts` (3 tests) |
| branch the executor on the matrix verdict | `timelock-cancel-placement.test.ts` |
| grade the matrix on a revert count it did not observe | `timelock-cancel-placement.test.ts` |

`timelock-cancel-placement.test.ts` was named for a placement it did not assert — it drove the
decision function only. Its new block reads the executor's source, because
`execute-pending-timelock-tx.ts` calls `runMain` at module scope and its reverting path needs a
chain, a Safe and MongoDB.

The report-only property is asserted too, not just described: every read of the matrix's `decision`
in the executor is counted, and each must be the render call or `decision.notes`. A branch on
`decision.action` raises the read count above the permitted one and fails.

## What the review round after that found

Two defects in what the ledger and the executor *report*, both of the shape this ticket names —
a gate that could not read its evidence presenting the failure as an observed fact.

**A failed on-chain state read was graded as `unset`.** `readOperationState` in the executor
returned `'unset'` from its `catch`. Every member of that union is a state the controller can
really be in, and the matrix renders `unset` as `operation is unset on-chain`, so a read that
threw put a fact on the operator's screen that nothing established. Its two sibling legs do have
could-not-answer values (`undefined`, `'unknown'`); this one has none, so it now throws and the
caller reports the matrix as unevaluated. The decision module's `operationState` union is a merged
module and was deliberately left alone — see the escalation below.

**The rpc-quorum row's comment denied the case it lands on.** `agreed-absent` sets
`reachesQuorum: false`, so it reaches the acknowledgement branch, whose comment said nothing
decided the read "either the providers disagreed, or there were not enough of them". The providers
did agree — on an empty value. Comment corrected; the anchor question is escalated below.

Also corrected, all doc-vs-code drift rather than behaviour: the file header counted two green
emitters where this PR leaves more; `buildDeploymentIndex`'s docblock claimed the source names the
deploy log "only when" the records came from it, while both branches stamp it (`available` is what
carries the distinction); `IExecutabilityChainReader` claimed every read returns `undefined` when
`staticCall` returns an observation; `INT-TIMELOCK-DELAY` is registered for an undecodable payload
as well as a schedule; and the two new docs rows named one anchor each for outcomes resting on two.
The docs row's "roughly eighteen single-endpoint chains" is now stated as a condition with its
source named — the per-chain arrays live in MongoDB, so the number is not measurable from this repo
and the code comment disagreed with it.

Four vacuous assertions were paired with a present: two absence-only checks that passed on an empty
result, and an `expect(refuses || error)` that could not tell the revert it was about from an
unrelated collection error. One dead `?? ZERO` default was removed — unreachable today, but it
would have read every selector against the zero address, a wrong answer where this module's rule is
a missing one. That one is **not** falsifiable by construction: the branch cannot be reached, so no
test distinguishes it, and the mutation reverting it survives.

## Escalated, not fixed

Four questions this PR deliberately does not answer, each recorded on EXSC-994:

1. **Should `timelock-cancel-decision`'s `operationState` be able to express an unread state?** It
   is a merged module (EXSC-872). Widening the union is the complete fix; throwing is the in-scope
   one.
2. **What anchor does an `agreed-absent` quorum rest on?** The read happened (`A-CHAIN`) but
   established nothing the caller wanted (`A-UNRESOLVED`, what it records today). Splitting the
   branch changes printed output in a gate.
3. **`getAddress` is unguarded in `ownerReads` and the single-diamond narrowing**, so a malformed
   diamond aborts the whole collection rather than leaving one read absent — which contradicts the
   module header, though both outcomes are fail-closed. `remember()` guards it; these two do not.
4. **The path-grammar examples in `calldata-address-check.ts` and `executability-simulation.ts`
   name outputs no collector emits** (`scheduleBatch[n]`, and a `diamondCut` with no ordinal). Both
   files are outside this PR's diff.

## Review gate, round 2 — what it found in round 1's own fixes

The placement assertions added above had three false-pass routes. Each was demonstrated, not
argued, and each is now closed and re-falsified:

| the regression that slipped through | why | now |
| -- | -- | -- |
| `rpcQuorum ||= { reachesQuorum: true }` in the `catch` — a fabricated green quorum | the absent-read assertion matched only a bare `=`, so `??=` and `||=` were invisible | killed |
| `const { action } = decision; if (action === 'cancel') return` in the executor | the report-only assertion counted `decision.` property reads, which destructuring does not produce | killed |
| `references: []` or `references.filter(...)` handed to the gate | the calldata assertions pinned the collector call and the record index, never what was graded | killed |

The third is the one worth naming: an emptied reference list is not a hypothetical refactor, it is
the obvious way to quiet a gate, and the assertions would have blessed it.

Also corrected in that round: the absent-read test's own docblock claimed it guards against a
verdict field the recorder is never given, which the type system already refuses — every
`IProposalCheckVerdicts` field is a required key. What it actually guards is a field that is
present and fed something other than the gate's verdict.

`renderCancelRecommendation` read `would op-form-unsupported → hold`, i.e. reason where the verb
belongs; it now reads `recommends hold — op-form-unsupported`. The test that pinned the word
`would` now pins the property its own comment names — that the line must not read as an action
already carried out.

Two claims this PR made about itself were wrong and are fixed rather than defended:
`docs/MultisigSigningProcess.md` said an executability mismatch "requires an acknowledgement"; the
run ledger renders in the `finally` block, after every signature, so it requires nothing. And the
RPC-quorum row sourced the endpoint count to MongoDB, while the code reads
`chain.rpcUrls.default.http` — the signer's own environment, which `fetch-rpcs.ts` materialises
from Mongo. An operator who has not run that fetch sees one endpoint whatever Mongo holds, and the
row now says so.


0xDEnYO:
@coderabbitai review

0xDEnYO:
@coderabbitai review

0xDEnYO:
@coderabbitai review

0xDEnYO:
@coderabbitai review

0xDEnYO:
@coderabbitai review

0xDEnYO:
@coderabbitai full review

script/deploy/safe/calldata-address-collector.ts
```diff
@@ -0,0 +1,129 @@
...
+const referencesOfCall = (
+  call: IDiamondCutCall,
+  ordinal: number
+): IAddressReference[] => {
+  const path = `call[${call.callIndex}].diamondCut[${ordinal}]`
+  const references: IAddressReference[] = []
+
+  for (const [at, cut] of call.cuts.entries()) {
+    const role = ROLE_BY_ACTION[cut.action]
+    if (role === undefined) continue
```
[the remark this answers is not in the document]
0xDEnYO: Fixed in dcc881c2f. An action outside 0-2 is no longer skipped: the cut is pushed onto `undecodable` as `call[i].diamondCut[n].cuts[m] (action N)`, and `evaluateCalldataAddresses` already turns any `undecodable` entry into an error, so the unclassifiable cut blocks instead of leaving its facet address ungraded. Kept report-only — no new role and no change to the evaluator.

Falsified: reverting the push back to `continue` fails `collectAddressReferences > leaves an unknown cut action unmapped rather than defaulting it`, which now asserts the `undecodable` entry alongside the surviving init reference.

script/deploy/safe/confirm-check-registry.ts
```diff
@@ -218,3 +220,319 @@ export const worstResultPerCheck = (
...
+  return {
+    checkId: EXECUTABILITY_CHECK_ID,
+    network,
+    status: 'pass',
+    expected: 'no payload reverts',
+    actual:
+      verdict.notSimulated.length > 0
+        ? `no revert found; ${verdict.notSimulated.length} payload(s) have no revert model`
+        : 'no revert found in any payload',
+    anchor: 'A-CHAIN',
```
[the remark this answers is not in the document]
0xDEnYO: Fixed in dcc881c2f. A non-empty `notSimulated` now records `needs-ack` on `A-UNRESOLVED` with the count in `actual`; `pass`/`A-CHAIN` is reserved for complete coverage. The check is `semantic`, so `needs-ack` reaches `requiresAcknowledgement` and never `blocking` — report-only is unchanged.

Falsified in both directions by `executabilityCheckResult > grades partial simulation coverage an acknowledgement, not a pass` beside the existing complete-coverage pass test.

script/deploy/safe/confirm-check-registry.ts
```diff
@@ -218,3 +220,319 @@ export const worstResultPerCheck = (
...
+
+    if (checkId === CHECK_TIMELOCK_DELAY)
+      return {
+        checkId,
+        network,
+        status: 'pass' as const,
+        expected: "a schedule's delay is at least the timelock's live minimum",
+        actual: 'this proposal carries no timelock schedule',
+        anchor: 'A-LOCAL' as const,
+      }
```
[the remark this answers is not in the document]
0xDEnYO: Fixed in dcc881c2f, taking your diff. The pass is now gated on `!run.registered.includes(CHECK_TIMELOCK_DELAY)`, so "never registered" (the assertions read the calldata and found no schedule) is a pass on `A-LOCAL`, while "registered and produced nothing" falls through to `unresolved(...)` and blocks.

Falsified both ways: reverting the guard fails the new `a registered delay check that never reported errors instead`, and inverting it fails that plus the no-schedule pass.

script/deploy/safe/executability-collector.ts
```diff
@@ -0,0 +1,389 @@
...
+  const codeReads = [...addresses.entries()].map(async ([key, address]) => {
+    const held = await reader.hasCode(address)
+    if (held !== undefined) hasCode.set(key, held)
+  })
+
+  const ownerReads = cutPayloads.map(async (payload) => {
+    const diamond = getAddress(payload.diamond)
+    const owner = await reader.owner(diamond)
+    if (owner !== undefined) owners.set(normalise(payload.diamond), owner)
+  })
```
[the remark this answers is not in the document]
0xDEnYO: Fixed in dcc881c2f. Every address parse in `readObservations` now goes through one `parsed()` guard that returns `undefined` instead of throwing, so a malformed diamond leaves its own owner read absent rather than rejecting the `Promise.all` and discarding every read that did land. Absent still grades as unchecked and blocks, so nothing is softened.

The module header claimed per-read absence while two parses sat outside the guard; it now describes what the code does. Falsified by `a malformed diamond loses its own read, not the whole batch`, which asserts the absent owner read beside a non-empty `hasCode` map.

0xDEnYO:
@coderabbitai review

0xDEnYO:
@coderabbitai review

script/deploy/safe/executability-collector.ts
```diff
@@ -0,0 +1,429 @@
...
+      try {
+        await simulator.call(call)
+        return { outcome: 'succeeded' }
+      } catch (error) {
+        const message = error instanceof Error ? error.message : String(error)
+        // A revert is the answer, and the first endpoint to give one settles
+        // it: asking another endpoint could only disagree with a fact this one
+        // established.
+        if (/revert/i.test(message))
+          return { outcome: 'reverted', revertReason: redactUrls(message) }
```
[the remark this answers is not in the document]
0xDEnYO: Fixed in `c0b416b60`, and you were right that the polarity was wrong.

Failing over is now the narrow case rather than the default: `isEndpointUnavailable` is a closed list of transport-level failures (viem's `HttpRequestError`/`TimeoutError`/`RpcRequestError`/etc. by error *name*, plus the HTTP 429/5xx and socket patterns), and **everything else stops the walk** and is reported as the payload's own answer. An invalid opcode, an out-of-gas, a `CallExecutionError` with no `revert` in it — none of them can now be overwritten by a later endpoint returning `succeeded`.

Falsified: `an execution failure stops the walk even without the word revert` drives `invalid opcode: INVALID`, `out of gas` and a bare `CallExecutionError` through a two-endpoint reader whose second endpoint succeeds, and asserts the outcome is never `succeeded`. Narrowing the classifier back to `/execution reverted/i` fails two tests.

script/deploy/safe/confirm-safe-tx.ts
```diff
@@ -832,13 +883,182 @@ const processTxs = async (
...
+          transport:
+            transports.length === 1 ? onlyTransport : fallback(transports),
+        })
+
+        // One client per endpoint for the simulation itself. A fallback
+        // transport decides revert-versus-unreachable by the node's wording, so
+        // the payload's own answer has to be read endpoint by endpoint instead.
+        const simulators = [...overrideEndpoints, ...endpoints].map((url) =>
+          createPublicClient({ chain, transport: http(url) })
+        )
```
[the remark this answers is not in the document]
0xDEnYO: Fixed in `c0b416b60`. The simulators are now built through `getTransportConfigFromRpcUrl`, carrying each endpoint's `fetchOptions`, `retryCount` and `retryDelay`, and an endpoint the helper rejects is dropped from the list rather than simulated against.

Your failure mode was the sharp end of it: with raw `http(url)` an auth-requiring endpoint fails to authenticate, that error is not a revert, so every simulator returned `errored` and the row recorded `error` — the gate reporting its own misconfiguration as "this proposal could not be simulated". Dropping a rejected endpoint rather than walking it matters for the same reason.

script/deploy/safe/codehash-sign-gate-deps.ts
```diff
@@ -526,6 +526,38 @@ export interface ISignTimeCodehashDeps extends IVerifyCutDeps {
...
+ */
+export const createDeployedCodeReader =
+  (
+    resolveChain: (network: string) => Chain = getViemChainForNetworkName
+  ): ((address: string, network: string) => Promise<string>) =>
+  async (address, network) => {
+    const chain = resolveChain(network)
+    const client = createPublicClient({
+      chain,
+      transport: getFallbackTransportForChain(chain),
```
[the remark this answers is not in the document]
0xDEnYO: Agreed, and fixed in `c0b416b60` — this was a real regression I introduced, not a trade-off worth keeping.

Plain failover on this read widened the set of endpoints whose answer can make a wrong codehash look right, on the one check that refuses a signature. The two cases are now separated:

- **The primary answers → its answer stands.** The same single endpoint the gate has always trusted, so no new trust is added on the common path.
- **The primary cannot answer → no single fallback decides.** The remaining endpoints are read together through `collectProviderObservations` + `evaluateRpcQuorum`, and the answer is taken only when independent providers agree on it at the same block. `agreed-absent` is treated as agreement and returned as `0x`, since the gate is what grades that against the rebuild.

Never worse than reading the primary alone: where that blocked, this either blocks the same way or proceeds on corroborated agreement.

Falsified in both the directions you named. `refuses a lone fallback answer when the primary is down` and `refuses when the fallbacks disagree` both go red against (a) accepting any fallback answer uncorroborated, and (b) the plain `getFallbackTransportForChain` version you flagged.

Side effect worth noting: this gives `rpc-quorum.ts` its first production caller on a read that actually gates, which is what WP-4.2 asked for.

script/deploy/safe/rpc-quorum-collector.ts
```diff
@@ -0,0 +1,119 @@
...
+ *
+ * @param address - The address whose code is read.
+ * @param chainId - The chain the endpoints serve, so a misrouted endpoint fails loudly.
+ * @returns A reader for {@link collectProviderObservations}.
+ */
+export const createCodeReader =
+  (address: Address, chainId: number): TEndpointReader =>
+  async (endpointUrl) => {
+    const client = createPublicClient({
+      transport: http(endpointUrl, { timeout: ENDPOINT_TIMEOUT_MS }),
```
gvladika: **Blocking:** this bypasses `getTransportConfigFromRpcUrl`, which is what turns `https://user:pass@host` into an `Authorization: Basic` header (tested at `script/utils/viemScriptHelpers.test.ts:119`).

Bun's `fetch` silently drops URL-embedded credentials — I confirmed against a local listener: no header sent, no error raised. So a credentialed endpoint is queried unauthenticated and answers 401, and the observation is recorded as an `error`.

It also bypasses the helper's refusal of credentials over cleartext `http:`, plus the TronGrid extras and retry config.

Routing this through `getTransportConfigFromRpcUrl` — as the executability simulators in `confirm-safe-tx.ts` already do — should be a small change. See the consequence on `codehash-sign-gate-deps.ts`.

script/deploy/safe/codehash-sign-gate-deps.ts
```diff
@@ -526,6 +532,78 @@ export interface ISignTimeCodehashDeps extends IVerifyCutDeps {
...
+    )
+
+    // `agreed-absent` is agreement: the providers concur that nothing is
+    // deployed there. The gate grades that against the rebuild, and it is the
+    // loudest thing this read can report.
+    if (verdict.reachesQuorum) return verdict.agreedValue ?? '0x'
+    if (verdict.status === 'agreed-absent') return '0x'
+
+    throw new Error(
+      `The primary RPC for ${network} could not answer and its fallbacks did not agree on the code at ${address} (${verdict.status}), so nothing here is verified`
```
gvladika, line 603: This is where the `createCodeReader` bug above lands, and it's the blocking gate.

On a chain with credentialed fallbacks: primary throttled → every fallback 401s → quorum unreached → this throws → signer blocked. That's the failure this PR adds the failover to remove.

Separately: `MIN_INDEPENDENT_PROVIDERS = 2`, so the failover only helps where two independent fallbacks agree. Single-fallback chains still block when the primary is down. The docblock's "never worse than reading the primary alone" is accurate — it's the PR body that reads as broader coverage.

script/deploy/safe/execute-pending-timelock-tx.ts
```diff
@@ -1867,7 +1925,75 @@ async function executeOperation(
...
+                  // operator's screen that no read established. The caller's
+                  // catch reports the matrix as unevaluated instead.
+                  throw new Error(
+                    `the operation's on-chain state could not be read: ${
+                      error instanceof Error ? error.message : String(error)
+                    }`
+                  )
+                }
+              },
+              readCancellerAuthority: async () => {
```
gvladika: This read is unreachable. `isProven` requires `verdictProvenance === 'anchors'` (`timelock-cancel-decision.ts:162`) and `buildCancelDecisionInput` hardcodes `'unknown'`, so `isProven` is permanently false from this call site and the `cancellerAuthority` branch at `timelock-cancel-decision.ts:241` can never be reached.

Two sequential RPC round-trips per reverting row (`CANCELLER_ROLE`, then `hasRole`) for a value that cannot affect any branch — on a path that is already handling a failure. Suggest dropping `readCancellerAuthority` and passing `'unknown'`, with a comment naming what has to land before the read is worth making.

(The three reads here are also awaited sequentially inside the object literal; moot if this one goes.)

script/deploy/safe/timelock-cancel-placement.ts
```diff
@@ -0,0 +1,146 @@
...
+): ICancelDecisionInput => ({
+  // No execute-time re-derivation of the attested build exists, so this leg is
+  // declared absent rather than reported as a passing comparison.
+  integrity: 'unsupported',
+  opIdentity: gradeOperationIdentity(signals),
+  // Nothing here produced a divergence verdict from an anchor, so the matrix
+  // must not treat an identity mismatch as proven and cancel on it.
+  verdictProvenance: 'unknown',
+  agreeingProviders: 0,
+  executability: 'would-revert',
```
gvladika, line 107: Six of the eleven fields this builds are constants by the time the executor calls it: `integrity`, `verdictProvenance`, `agreeingProviders`, `executability` here, plus `deploymentRecord: 'error'` and `signTimeVerdictRecord: 'missing'` at the call site.

That leaves three reachable outputs, and the default — `hold/op-form-unsupported` — carries no information. The only live signal is whether `hashOperationBatch` matches the scheduled id, which is a string comparison.

146 lines here plus 311 of test to route that. Suggest wiring the id comparison directly (read, compare, warn on mismatch) and leaving the matrix until EXSC-872 makes its inputs real. Keeps the operational value, drops ~450 lines.

Happy to be argued out of this if the intent is to have the wiring in place for when the integrity leg lands — but the reads backing unreachable branches should go either way.

script/deploy/safe/confirm-check-registry.ts
```diff
@@ -291,3 +293,344 @@ export const worstResultPerCheck = (
...
+ *
+ * @param verdict - What `evaluateExecutability` decided.
+ * @param network - The network the verdict is about.
+ * @returns The row to hand to `recordCheck`.
+ */
+export const executabilityCheckResult = (
+  verdict: IExecutabilityVerdict,
+  network: string
+): ICheckResult => {
+  if (verdict.error)
```
gvladika, line 377: `executability-simulation.ts:872` folds every *unchecked* read into `errors`, so a single unanswered `facetAddress` sets `verdict.error` and lands here → `A-UNRESOLVED` → the ledger closes `BLOCKED`.

Stacked on EXSC-1003 (a correct rollout already closes `8/9 ACKNOWLEDGEMENT REQUIRED` because `target-state` grades `needs-ack` on `matches-main`), the closing line is non-green on essentially every run while blocking nothing.

Is that the intended steady state, or should an unchecked read grade separately from a failed one? The value reads do sit behind a fallback transport with retries, so this is less likely than it looks — but a line that's always red gets skipped.

script/deploy/safe/confirm-check-registry.ts
```diff
@@ -291,3 +293,344 @@ export const worstResultPerCheck = (
...
+  // was read, and a queue that moves under it turns the answer over. An
+  // integrity class would hard-block a legitimate proposal on a stale read with
+  // no way for the signer to say so.
+  checkClass: 'semantic',
+  title: 'The proposal would execute rather than revert',
+}
+
+export const RPC_QUORUM_CHECK_ID = 'rpc-quorum'
+
+export const RPC_QUORUM_CHECK: ICheckDefinition = {
```
gvladika, line 312: Question on altitude rather than a defect: this row can never record `fail`, and on single-endpoint chains it records `needs-ack` permanently.

"This chain has one endpoint" is a fleet-configuration fact, not a per-proposal one, so repeating it per proposal per network adds a line to every run that a signer can't act on at signing time. Was the per-proposal row deliberate over a `fetch-rpcs` warning or a fleet report?

script/deploy/safe/executability-collector.ts
```diff
@@ -0,0 +1,461 @@
...
+    onlyDiamond === undefined
+      ? []
+      : [...selectors].map(async (selector) => {
+          const facet = await reader.facetAddress(onlyDiamond, selector as Hex)
+          if (facet !== undefined) selectorFacets.set(selector, facet)
+        })
+
+  await Promise.all([...codeReads, ...ownerReads, ...selectorReads])
+
+  const attempted =
```
gvladika, line 246: Minor: `attempted` counts one owner read per cut payload, including payloads whose diamond failed to parse and were skipped in `ownerReads`. Only gates the `answered === 0` branch, so harmless today — flagging in case the value is ever used for coverage.

gvladika (commented):
One blocker (transport config bypass, inline on `rpc-quorum-collector.ts`), some code I think can come out, and two questions about what the new rows report. Details inline.

Two claims in the description don't match the code:

- "now uses the same `getFallbackTransportForChain` helper" — that helper is only in `confirm-safe-tx.ts`; the codehash gate got a bespoke primary-then-quorum reader.
- "no new `source.indexOf('literal')` assertion" — the three placement files add ~25. True of the order assertion, not of the PR.

script/deploy/safe/rpc-quorum-collector.ts
0xDEnYO: Fixed in `e9790a10a`, and the bypass is real — but the mechanism is narrower than the one you measured, so worth pinning down.

viem's own transport already lifts `user:pass@` out of the URL. Against a stubbed `fetch` on viem 2.55.19, `https://user:pass@host/rpc` arrives as `https://host/rpc` with `Authorization: Basic dXNlcjpwYXNz` — the branch is `node_modules/viem/_esm/utils/rpc/http.js:159`. So that shape was never queried unauthenticated through `http()`.

The hole is what viem keys on: `if (url.username)`. A **password-only** endpoint (`https://:token@host`) has no username, so viem leaves the credential in the URL, sends no header, and bun's `fetch` drops it — 401, recorded as that provider's answer. Confirmed against a real local listener, not just a stub. That shape is anticipated in this repo rather than hypothetical: `script/mongoDb/rpcEndpoints.ts:44` treats `parsed.username || parsed.password` as credential-bearing.

The three bypasses you listed alongside it are all real as stated: the cleartext refusal, the TronGrid extras, and the retry config.

Now routed through the helper. An endpoint the helper rejects throws inside the reader, so `collectProviderObservations` records it as that endpoint's `error` observation — which keeps it in `evaluateRpcQuorum`'s denominator (`independentProviders` counts all observations; only `respondingProviders` drops it) rather than shrinking it.

One consequence I had to undo in `d5e982b72`: forwarding the helper's `retryCount`/`retryDelay` broke the module's 8-second contract, which bounds one attempt and not the call. A TronGrid endpoint carries 8 retries at a 2s exponential backoff, so a throttled one could hold the `Promise.all` a signer is waiting on for minutes — and `codehash-sign-gate-deps.ts` has no Tron exclusion. The reader now pins `retryCount: 1` and keeps `fetchOptions`: a quorum read wants a snapshot of who answers now, and an endpoint needing nine attempts is a non-answer the verdict already grades as one.

Falsified: against the old `http(endpointUrl)` form, `sends a password-only credential as an Authorization header`, `refuses to carry credentials over cleartext http` and `records the cleartext refusal as an error observation` all go red; against an uncapped retry count, `caps the retry budget rather than inheriting the endpoint profile` does.

script/deploy/safe/codehash-sign-gate-deps.ts
0xDEnYO: The `createCodeReader` fix lands here, so the primary-throttled → fallbacks-401 path is closed for the shape that actually had it.

Chasing it also turned up a second route of the same bypass, which I had missed in the earlier round: `confirm-safe-tx.ts` fixed the *simulators* to go through `getTransportConfigFromRpcUrl` in `c0b416b60` but left the `--rpcUrl` override on a bare `http(url)` for the chain-state reader. Same failure, different entry point. Fixed in `d5e982b72`, with an override the helper rejects dropped and warned about rather than silently simulated against.

On `MIN_INDEPENDENT_PROVIDERS = 2` and the docblock-vs-body gap: you are right, and the body was the inaccurate half. Rewritten to say what the docblock says — primary-then-quorum rather than `getFallbackTransportForChain`, recovers chains with two independent fallbacks, single-fallback chains still block when the primary is down, never worse than reading the primary alone and not broader than that.

script/deploy/safe/execute-pending-timelock-tx.ts
0xDEnYO: Dropped in `e9790a10a`. `readCancellerAuthority` is gone from `ICancelRecommendationContext` and from the call site, and `buildCancelDecisionInput` hardcodes `cancellerAuthority: 'unknown'` with the comment naming what has to land first. The two ABI entries that existed only for that read — `hasRole` and `CANCELLER_ROLE` — went with it.

Your reachability argument holds on two independent gates, not one: `buildCancelDecisionInput` hardcodes `verdictProvenance: 'unknown'` *and* `agreeingProviders: 0`, and `isProven` needs both. It is also the only production construction of `ICancelDecisionInput`.

Added one thing you did not ask for, because the silent-failure mode worried me: nothing coupled "someone makes provenance anchored" to "restore the authority read", and if the hardcode survived that day every proven divergence would land on `canceller-authority-missing` — fail-safe, but silently. `declares the authority leg absent only while provenance cannot be anchored` now pins the two together so they have to break in the same change.

script/deploy/safe/timelock-cancel-placement.ts
0xDEnYO: The reads are gone (thread above). On the module itself I would like to keep it, for the reason you named as the one that would change your mind — so here is the argument, and it is a ticket-scope one rather than a taste one.

EXSC-994's acceptance is that each of the four orphan modules has ≥1 non-test production importer, and scope item 2 says `timelock-cancel-decision` is the exception whose consumer is the executor rather than the sign-time ledger. So the placement *is* the deliverable for that module. Wiring the id comparison directly and deleting the rest re-orphans `timelock-cancel-decision.ts`, which is the state this ticket exists to end.

What the module buys for its constants is that each one is constant *by name*: when EXSC-872's integrity leg lands, `integrity` and `verdictProvenance` change value in one file and the matrix starts deciding, with no new wiring and no second construction of `ICancelDecisionInput` to keep in step.

The line count is the part I would push back on hardest. With the authority dimension gone the test file loses three permutation loops, so it is meaningfully smaller than the version you read — and your `~450` was over the whole module plus test, where the reads you actually objected to were about 20 lines of it.

If you still think the matrix should wait for EXSC-872, say so and I will take it out — but it wants to be a scope change on the ticket rather than a quiet deletion, since the acceptance criterion is explicit.

script/deploy/safe/confirm-check-registry.ts
0xDEnYO: Intended, and I do not think unchecked should grade separately — but the "always red" half is a fair worry and I do not have the measurement to close it.

Unchecked and failed are the same fact for a signer: the check did not run. `check-ledger.ts` deliberately gives `error` no acknowledgement path ("an unverified check has no acknowledgement path", and `checkClass` does not soften it), so splitting unchecked off into `needs-ack` manufactures exactly that path for the case where the gate saw least. That is a ledger-semantics change rather than an executability one, so it is not this PR's to make.

What I cannot answer from here is your actual question — how often it fires on a real rollout. Filing that as a follow-up to measure across the fleet before we decide whether the line needs softening, rather than guessing at it now.

script/deploy/safe/confirm-check-registry.ts
0xDEnYO: Deliberate, and the altitude objection is right for single-endpoint chains specifically.

The row is per-proposal because two of the three things it reports are: providers disagreeing on the value, and providers agreeing the address is empty (`agreed-absent`, which is agreement without the fact an integrity read wanted). Both are per-read facts that can change between proposals on the same network. It is only on a single-endpoint chain that it collapses to the fleet-configuration constant you are describing — which is also why the row's `detail` is worded in providers rather than endpoints, so the operator is not sent to re-run a command that would change nothing.

Ticket scope item 3 pins the report-only registration and the acceptance wants the verdict on the ledger, so the row stays in this PR. Whether a single-endpoint chain should get a `fetch-rpcs` warning or a fleet report *instead of* a per-proposal line goes into the same follow-up as the thread above.

script/deploy/safe/executability-collector.ts
0xDEnYO: Fixed in `e9790a10a`. `ownerReadTargets` is resolved before the map, so a payload whose diamond does not parse is never counted as an owner read that went unanswered.

One correction to "harmless today", in both directions. It is reachable: `payload.diamond` is `call.target ?? input.to`, and `input.to` is an unvalidated cast of Safe-API data, so a malformed `to` on a removal-only cut (every `facetAddress` zero, nothing else remembered) makes `attempted` non-zero purely from skipped reads. But the consequence is smaller than a wrong verdict — the absent reads grade `unchecked` and the proposal blocks either way; what changes is that the signer is no longer sent after an endpoint outage that never happened.

Falsified: `does not count a skipped owner read as an unanswered one` goes red against `attempted = addresses.size + cutPayloads.length`.

script/deploy/safe/confirm-check-registry.ts
0xDEnYO: Filed as [EXSC-1006](https://linear.app/lifi-linear/issue/EXSC-1006/measure-how-often-the-sign-time-ledgers-non-green-rows-actually-fire) — measure the distribution on real rollouts first, then decide between leaving it, moving the single-endpoint case to a `fetch-rpcs`/fleet report, or revisiting the ledger's unverified-vs-unread distinction. Acceptance is explicit that no change lands on an intuition about how noisy the line feels.

script/deploy/safe/confirm-check-registry.ts
0xDEnYO: Filed as [EXSC-1006](https://linear.app/lifi-linear/issue/EXSC-1006/measure-how-often-the-sign-time-ledgers-non-green-rows-actually-fire) — measure the distribution on real rollouts first, then decide between leaving it, moving the single-endpoint case to a `fetch-rpcs`/fleet report, or revisiting the ledger's unverified-vs-unread distinction. Acceptance is explicit that no change lands on an intuition about how noisy the line feels.

0xDEnYO:
@coderabbitai review

0xDEnYO:
@coderabbitai review

0xDEnYO:
@coderabbitai review

script/deploy/safe/codehash-sign-gate-deps.ts
```diff
@@ -526,6 +532,78 @@ export interface ISignTimeCodehashDeps extends IVerifyCutDeps {
...
+      try {
+        const { url, fetchOptions, retryCount, retryDelay } =
+          getTransportConfigFromRpcUrl(primary)
+        const client = createPublicClient({
+          chain,
+          transport: http(url, {
+            ...(fetchOptions ? { fetchOptions } : {}),
+            ...(retryCount !== undefined ? { retryCount } : {}),
+            ...(retryDelay !== undefined ? { retryDelay } : {}),
+          }),
```
gvladika, line 579: The retry-profile hazard this PR diagnoses is capped at one of the sites this round touched, and not at this one — the blocking gate.

`rpc-quorum-collector.ts:30-34` names it exactly: "nine TronGrid attempts on an exponential backoff is minutes of wall clock inside the `Promise.all` a signer is waiting on", and then "`codehash-sign-gate-deps.ts` has no Tron exclusion". That file pins `retryCount: 1`. This read forwards the endpoint's profile verbatim.

Measured rather than inferred:

- `getTransportConfigFromRpcUrl('https://api.trongrid.io/jsonrpc')` returns `retryCount: 8, retryDelay: 2000` (`applyTronGridViemTransportExtras`, `MAX_RETRIES + 5`).
- viem's backoff is `~~(1 << count) * retryDelay` (`node_modules/viem/_esm/utils/buildRequest.js:139`), so 8 retries is 2+4+8+16+32+64+128+256 = **510s of sleep**, plus 9 attempts at the transport's 10s default timeout — roughly **10 minutes for one `eth_getCode`**.
- `config/networks.json` gives `tron` an rpcUrl of `https://api.trongrid.io`, which `isTronGridRpcUrl` matches; tron has an active Safe; and `evaluateCodehashSignGate` (`confirm-safe-tx.ts:829`) runs with no Tron exclusion.

So a throttled TronGrid primary holds the one gate that refuses a signature for ~10 minutes before the fallback quorum this PR adds is reached at all.

This is new exposure rather than a pre-existing bound. On `main` this read built its transport from `getTransportConfigFromRpcUrl(...).url` alone and got viem's defaults (3 retries at a 150ms base, ≈41s). Dropping `fetchOptions` there was its own bug and fixing it is the right half of this change — it is `retryCount`/`retryDelay` that want the cap.

Same shape at the other two sites added this round, both previously on viem defaults:

- `read-only-safe-client.ts:35-39` — reached from `getNetworksWithActionableTransactions`, so the stall lands before the signer is shown a network list at all.
- `confirm-safe-tx.ts:944-952` — the `--rpcUrl` override.

To be fair to the diff: `getFallbackTransportForChain` already forwards this profile on `main`, so the TronGrid budget is not new to the fleet. What is new is that it now reaches three reads that previously did not carry it, one of them the blocking gate.

Suggest pinning a sign-time retry budget at all three, the way `rpc-quorum-collector.ts` now does — the argument in its docblock applies here unchanged. Where the TronGrid profile is genuinely wanted it belongs on the deploy and broadcast paths it was written for, not on a read a signer is sitting in front of.

script/deploy/safe/rpc-quorum-collector.ts
```diff
@@ -0,0 +1,155 @@
...
+        ...(fetchOptions ? { fetchOptions } : {}),
+      }),
+    })
+
+    const observed = await client.getChainId()
+    if (observed !== chainId)
+      throw new Error(`endpoint reports chain ${observed}, expected ${chainId}`)
+
+    const block = await client.getBlock()
+    const code = await client.getCode({ address, blockNumber: block.number })
```
gvladika: `ENDPOINT_TIMEOUT_MS` bounds one attempt, and this reader makes three of them.

`getChainId` (line 139), `getBlock` (143) and `getCode` (144) are three separate JSON-RPC round trips, each carrying the full `timeout` + `retryCount`/`retryDelay` budget. Worst case per endpoint is therefore 3 × (8s + 2s + 8s) = **54s**, not the 8s that line 24's docblock claims ("a sign-time fan-out must not stall the operator on one slow endpoint").

`ENDPOINT_RETRY_DELAY_MS`'s docblock (lines 41-44) has the same factor of three: "at most ~2s on top of the timeout, which is the budget the cap above was protecting" is true per call, and there are three calls.

Counted rather than read off the source — one `createCodeReader(...)(url)` on the success path issues 3 `fetch` calls against a stub.

The cap itself is the right change and the Tron case it closes was far worse than this. Two ways to make the file honest:

1. Keep the behaviour and correct the two docblocks to state the real bound — 3 round trips, ~54s worst case. Cheapest, and the number is what a future reader will budget against.
2. Bound the whole read instead of each attempt: build one `AbortSignal.timeout(ENDPOINT_TIMEOUT_MS)` per endpoint and thread it through `fetchOptions.signal`, the shape `getFallbackTransportForChain` already accepts via `options.signal`. viem honours it — `node_modules/viem/_esm/utils/rpc/http.js:39` is `signal: signal_ || (timeout > 0 ? signal : null)`, so a supplied signal replaces the per-attempt one, and `shouldRetry` returns false on an abort, so the read fails fast instead of retrying past its own budget. That makes the 8s docblock true as written.

I'd take (2) if 8s is a real operational requirement and (1) if 54s is acceptable — but the two docblocks shouldn't keep saying 8s either way.

gvladika (commented):
Re-reviewed round 2. All seven of my earlier threads are resolved to my satisfaction — three fixed, one pushback I accept (the cancel-placement scope argument holds now that the unreachable reads are gone), two deferred to EXSC-1006. Verified locally: `bun test script/deploy/safe/` is 2138/0, and `tsc`/`eslint` are clean on every changed file.

Two new findings from this round, both in the transport work. Neither is about correctness of a verdict — they are about how long a signer waits and what the docblocks promise.

script/deploy/safe/rpc-quorum-collector.ts
```diff
@@ -0,0 +1,202 @@
...
+      budgetMs
+    )
+
+    try {
+      const client = createPublicClient({
+        transport: http(url, {
+          timeout: budgetMs,
+          retryCount: ENDPOINT_RETRY_COUNT,
+          retryDelay: ENDPOINT_RETRY_DELAY_MS,
+          fetchOptions: { ...fetchOptions, signal: controller.signal },
```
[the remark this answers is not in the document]
0xDEnYO: Confirmed and fixed in 43a2ce3f3 — reproduced before fixing: a 429 answering `Retry-After: 8` turned a 50ms budget into an **8009ms** read.

One correction to the suggested shape: passing the whole-read signal to the retry loop is not reachable from viem's public API. `createTransport` builds the request as `buildRequest(request, { methods, retryCount, retryDelay, uid })` (`clients/transports/createTransport.js:19`) — `signal` is simply never in that object, and `http()`'s config has no `signal` field either. So `withRetry`'s `wait(delay, { signal })` is uninterruptible no matter what the caller does at transport level.

So the retry moved out of viem instead: the transport is built `retryCount: 0`, and the one retry is taken in `createCodeReader` where the deadline is visible. It refuses to start a wait the budget cannot absorb, and it never consults `Retry-After` — which was the sharper half of this. viem honours that header verbatim (`~~(1 << count) * retryDelay` only applies when it is absent), so an endpoint could name how long the signer waits, and `collectProviderObservations` is a `Promise.all`: one endpoint answering `Retry-After: 600` would have held the whole fan-out for ten minutes on a 20-second budget.

Two tests, both directions — `does not let a Retry-After header extend the budget` (1 attempt, under 5s, on a 200ms budget) and `still retries a 429 when the budget can absorb the wait` (2 attempts on the module's own 2s spacing, not the 30s the endpoint asked for), so the guard is a deadline check rather than a disabled retry.

Falsified: handing the retry back to the transport kills 2 tests; dropping the deadline guard, dropping the `AbortError` short-circuit, and raising the budget each kill 1.

script/deploy/safe/codehash-sign-gate-deps.ts
0xDEnYO: Confirmed every number and fixed in 160b3dbe9. `getTransportConfigFromRpcUrl` now has a sign-time wrapper, `script/deploy/safe/sign-time-transport.ts`, which keeps the URL rewrite and the credential/API-key headers — the half of your point that was the right fix — and replaces only the retry profile with `retryCount: 2 / retryDelay: 1000`.

Two retries rather than the collector's one, deliberately. These are single-endpoint reads, and the codehash gate's primary read falls through to the fallback quorum only when `fallbacks.length > 0`: on a network with no configured fallback, one retry would turn a single 429 into a refusal to sign. 3s of backoff against 510s.

Applied at your three sites plus the executability simulators in `confirm-safe-tx.ts`, which carry the same profile through the same block.

The cap is unconditional rather than TronGrid-only, which is a fleet-wide change worth naming: every network's sign-time reads now get 2/1000 in place of viem's 3/150.

The test pins the hazard against the dependency rather than describing it — `retryCount` 8, `retryDelay` 2000, and the 510s backoff computed from viem's own formula — so if `@lifi/tron-devkit` ever stops carrying that profile the file stops claiming it prevents ten minutes. It reads the tron URL from `networks.json` because both TronGrid hosts are `.env` values the pre-commit secret scanner refuses to see in source; the comment says explicitly that this is not the endpoint a signer reads from, since `getViemChainForNetworkName` builds `chain.rpcUrls` from `ETH_NODE_URI_<NETWORK>` and its fallbacks.

One gap I have not closed, flagged rather than hidden: `getFallbackTransportForChain` at `confirm-safe-tx.ts:965` sits inside the same sign-time block and still forwards the raw profile (measured: 8 / 2000). It is unreachable today — line 919 gates the block on `!isTronNetworkKey(network)` — but that is an invariant in another function, not a property of this read. Same for both `SafeClient.init` transports in `safe-utils.ts`, which have no such guard. Happy to take either here or in a follow-up; say which you prefer.

script/deploy/safe/rpc-quorum-collector.ts
0xDEnYO: Right on the factor of three — `getChainId`, `getBlock` and `getCode` each carried the full budget, so the docblocks were describing a third of the real bound.

Took (2), with one correction to the mechanism. `AbortSignal.timeout` aborts with a `TimeoutError`, and viem's `isAbortError` matches `AbortError` alone (`errors/utils.js:11-16`), so that signal's abort is retried rather than fast-failed — the budget would have been a first estimate. It is an explicit `AbortController` aborting with a `DOMException` named `AbortError`, merged into `fetchOptions` so the endpoint's credential headers survive, and cleared in a `finally` so a pending timer cannot hold a CLI run open.

The constant is 20s rather than 8s, and renamed `ENDPOINT_READ_BUDGET_MS` so the name says what it bounds. 8s across three round trips is a ~6.75x tightening of the effective budget, and a slow-but-honest endpoint dropped from the fan-out costs a provider the verdict counts — below `MIN_INDEPENDENT_PROVIDERS` that is a blocking `insufficient-providers`, i.e. curing a stall by manufacturing a refusal to sign. 20s is a real bound without buying it that way.

Worth being precise that this is still a reduction in patience: the old effective ceiling was ~54s, the new one is 20s. On a two-endpoint network an endpoint that used to fit inside 54s but cannot fit three round trips into 20s now drops out. I think that is the right trade — an endpoint needing >6.6s per round trip is a non-answer — but it is a trade, not a free tightening, so flagging it rather than burying it in the constant.

A second defect fell out of building it, which CodeRabbit independently found: the retry wait was outside the budget, and viem honours `Retry-After` verbatim, so a throttled endpoint could name how long the signer waits. Fixed in 43a2ce3f3 — details in that thread.

script/deploy/safe/codehash-sign-gate-deps.ts
0xDEnYO: Took the first half of this in 7bd2bd61d, and the reading corrected the second half.

`SafeClient.init` builds two transports from the same `provider`, and only one of them should be capped. The public client is reads, it is reached from `ConfirmSafeTxPrefetchQueue` before the signer is shown anything, and `tronshasta`'s primary is a TronGrid host with nothing on that path excluding Tron — so that one was a live ~10 minute stall and is now capped. The wallet transport a few lines below it broadcasts, and the endpoint's retry profile is a broadcast budget: capping it would cut a send's retries, so it deliberately keeps the raw helper. The placement test for this file asserts that split by position rather than by the blanket "no raw helper anywhere in the file" rule the other three files get, because here that rule would be wrong.

`getFallbackTransportForChain` I am leaving alone and ticketing. It is a shared helper whose deploy and broadcast callers legitimately want that profile, so capping it means a new parameter plus a judgement about every caller — a different change reviewed against different criteria — and the call site here is behind `!isTronNetworkKey(network)`. Happy to be overruled if you would rather see it in this PR.

script/deploy/safe/rpc-quorum-collector.ts
```diff
@@ -0,0 +1,232 @@
...
+          // be to sleep, which is the first attempt's failure reported late.
+          if (deadline - Date.now() <= ENDPOINT_RETRY_DELAY_MS) throw error
+          await new Promise((resolve) =>
+            setTimeout(resolve, ENDPOINT_RETRY_DELAY_MS)
+          )
+        }
+    } finally {
+      // A pending timer keeps the process alive after the fan-out has its
+      // answer, which on a CLI is a run that will not exit.
+      clearTimeout(budget)
```
gvladika, line 230: The one guard this round added that nothing falsifies — and it is load-bearing.

Mutation-tested the three new mechanisms in this reader. Two of the three bite, hard:

| mutation | result |
| -- | -- |
| `retryCount: 0` → `2` (transport retries back on) | 4 tests fail |
| `'AbortError'` → `'TimeoutError'` (the `AbortSignal.timeout` mistake) | 1 test fails |
| drop the `deadline - Date.now() <= ENDPOINT_RETRY_DELAY_MS` guard | 1 test fails |
| **delete `clearTimeout(budget)`** | **18 pass, 0 fail** |

The comment above it states a consequence — "a pending timer keeps the process alive after the fan-out has its answer, which on a CLI is a run that will not exit" — and the consequence is exactly right. Driven against a stub that answers all three round trips successfully:

```
### as written                 EXITED after 5ms
### with clearTimeout removed  EXITED after 20002ms
```

So the success path holds a CLI run open for the full `ENDPOINT_READ_BUDGET_MS` per endpoint, on every signing run that reaches the quorum reader. That is the failure mode this PR's own standard is written against: a claim in a comment that no test observes, on a line whose deletion is invisible to a green suite.

Cheap to close without a 20s test — stub `globalThis.setTimeout`/`clearTimeout`, assert the handle the reader created is cleared, and drive it twice: once on the success path and once on the throw path, since the `finally` covers both and only one of them is the obvious one to write.

Worth saying that this is the *only* thing I found unfalsified in this round — `retryCount: 0`, the `AbortError` naming and the deadline guard are all pinned by tests that fail for the right reason, and reverting each of the four sign-time call sites to `getTransportConfigFromRpcUrl` fails `sign-time-transport.test.ts` individually. The budget work is right; this is one line short of matching it.

script/deploy/safe/safe-utils.ts
```diff
@@ -546,13 +547,19 @@ export class SafeClient {
     let chain: Chain | undefined = undefined
 
     if (typeof provider === 'string') {
+      // Reads only, and an operator waits on every one of them — this client is
```
gvladika, line 550: This comment is true of one caller and false of seven, and the module docblock disclaims what the change actually does.

`sign-time-transport.ts` draws the boundary explicitly: "the deploy and broadcast paths keep the endpoint's own profile, which is what it was tuned for." But the cap is applied inside `SafeClient.init`, which is not a sign-time constructor. Via `initializeSafeClient` it is reached by:

- `confirm-safe-tx-prefetch.ts` (`getOrInitializeSafeClient`) — the caller this comment describes, correctly;
- `propose-to-safe.ts`, `safe/safeScriptHelpers.ts`, `tasks/unpauseAllDiamonds.ts`, and the five `propose*ChainIdMappings.ts` task scripts — none of which are sign time.

So "this client is built before the signer is shown a network list" is one of eight call paths. The split you describe — read client capped, wallet client not — is the right split, and the positional test pins it; the issue is the justification, which reads as though the read client were sign-time-only when it is the shared read client for the propose path too.

On whether it matters, rather than just whether it is stated accurately — measured both directions:

- **Non-Tron: strictly more patience.** `applyTronGridViemTransportExtras` returns `base` untouched for every non-TronGrid host, so these endpoints carried no profile at all and got viem's defaults. 2 × 1000ms = 3s of backoff replaces 3 × 150ms ≈ 1.05s. No regression anywhere on the fleet.
- **Tron propose reads: less.** `SafeClient`'s `publicClient` serves `getChainId` and the `readContract` calls at `safe-utils.ts:652-791` (nonce, owners, threshold). On a throttled TronGrid those drop 510s → 3s, so a propose run can now fail a nonce read where it previously ground through.

That is fail-closed and retriable by the operator, so I am not asking for it to be reverted — and I checked the sharper version and it does not apply: `createChainExecutor` hands Tron a `TronChainExecutor(tronWalletClient, networkKey)` with no `publicClient`, so the post-broadcast `waitForTransactionReceipt` at `executors/evm-executor.ts:93` is never on a TronGrid profile. (Had it been, it would have mattered: viem rejects the whole wait on any error that is not `TransactionNotFoundError`/`TransactionReceiptNotFoundError` — `waitForTransactionReceipt.js:196-198` — which turns a shortened transport budget into an execution that looks failed after it broadcast.)

So: a trade, not a free tightening, on a path the module says it does not touch. You flagged exactly this shape for the 54s → 20s change in the quorum reader rather than burying it in the constant — same treatment here. Either narrow the cap to the sign-time caller, or keep it fleet-wide and say in the comment that it also re-budgets the propose path, with the non-Tron direction (more patience, not less) stated so the next reader does not have to measure it again.

gvladika (commented):
Round 3. Re-reviewed the three commits since 02c2f506e (160b3dbe9, 43a2ce3f3, 7bd2bd61d) — the merge from main is not in scope.

**Both of my round-2 comments are properly closed.** The retry cap keeps the URL rewrite and the credential/API-key headers and replaces only the profile, which is the correct half to cut — verified against `applyTronGridViemTransportExtras`, whose return is `{url, fetchOptions, retryCount: MAX_RETRIES + 5, retryDelay: RETRY_DELAY}`. And you were right to correct my mechanism on the second one: `AbortSignal.timeout` raises `TimeoutError`, viem's `isAbortError` matches `AbortError` alone, so the explicit controller is what makes the budget a bound rather than an estimate. The `Retry-After` defect you found while building it is the more valuable catch of the two — an endpoint naming how long the signer waits is the same class of bug as the fallback transport turning a revert green.

**Verified, not read off the diff:**

| mutation | result |
| -- | -- |
| `retryCount: 0` → `2` in the quorum transport | 4 fail |
| `'AbortError'` → `'TimeoutError'` | 1 fail |
| drop the deadline-absorbs-the-wait guard | 1 fail |
| revert `safe-utils.ts` public client to the raw helper | 1 fail |
| revert `codehash-sign-gate-deps.ts` to the raw helper | 1 fail |
| `SIGN_TIME_RETRY_COUNT` 2 → 1 | 2 fail |
| drop `fetchOptions` from the cap | 1 fail |
| **delete `clearTimeout(budget)`** | **survives** |

`bun test script/deploy/safe/` → 2272 pass, 0 fail. `tsc-files --noEmit` clean on all five changed files.

**Two comments**, neither blocking:

1. The surviving mutation. The `finally` is load-bearing — demonstrated at 5ms vs 20002ms — and nothing observes it.
2. The `SafeClient.init` cap reaches the propose path, which `sign-time-transport.ts`'s docblock says it does not. Measured both directions; non-Tron is strictly more patience, Tron propose reads are less. A trade to state, not to revert.

And an answer on your open scope question in the codehash thread: ticket `getFallbackTransportForChain`, with a better reason than the one you gave.

The budget work itself is right, and the falsification discipline in this round is the strongest it has been across the three — one line short of matching it.

script/deploy/safe/codehash-sign-gate-deps.ts
gvladika: Agreed on both, and the reading changes the second one's priority.

**`SafeClient.init`** — the read/broadcast split is the right call and the positional test is the right way to pin it. One note on the justification rather than the code, left on the file itself.

**`getFallbackTransportForChain`** — ticket it, and the residual is smaller than your reply claims. You justified it as "unreachable today, but that is an invariant in another function, not a property of this read", which is a fair reason to be uneasy. The stronger reason not to be: `applyTronGridViemTransportExtras` returns `base` untouched for any host `isTronGridRpcUrl` does not match, so the inflated profile does not exist on a non-TronGrid endpoint at all. A non-Tron chain reaching that helper gets viem's defaults whatever `!isTronNetworkKey(network)` does. The guard is not the load-bearing part — the host check inside the helper is, and that one *is* a property of the read.

For it to bite you would need a non-Tron network configured with a TronGrid endpoint in `ETH_NODE_URI_<NETWORK>`, which is a misconfiguration rather than a state the fleet reaches. And it is the only sign-time caller: on `main` the helper's sole consumer is `healthCheck.ts:177`, which is not sign time. So the follow-up is genuinely low priority, and I would rather it stayed out of this PR than have a new parameter and eight caller judgements land under this ticket.

Same answer for `safe-utils.ts`'s wallet transport, for the same reason — it broadcasts, and that budget is what the profile was written for.

gvladika (approved):

gvladika (dismissed):

0xDEnYO:
Finding 1 — option (b), with the justification now in the code (779bb1083).

The run-level ledger renders inside the `finally` block, i.e. after every signature has been given. The per-proposal `renderIntegrityAsserts` is the only place those verdicts reach the signer *before* the signing decision, so dropping it would move integrity verdicts past the point they can act on. The duplication is the cost of that ordering. Comment added at the call site.

Findings 2–4 are acknowledged as informational/low with no functional impact; 4 is the pre-existing EXSC-1003 interaction, tracked there.

gvladika (approved):

PR #2371, opened 2026-09-11
feat(tests): withhold endpoints, provider keys and webhooks from spawned children (EXSC-988)

0xDEnYO:

Fixes [EXSC-988](https://linear.app/lifi-linear/issue/EXSC-988)

#2364 established that deleting a name from a spawned child's environment does not
withhold it — bun re-loads the repo `.env` inside the child for every name the passed
environment leaves unset — and fixed it by *setting* dummies. Its match covered the
signing keys, the seed phrase and the two proposal-store URIs.

**Classes added:** keyed RPC endpoints (`ETH_NODE_URI`), provider and explorer keys
(`API_KEY`, `ACCESS_KEY`, `SYNC_TOKEN`), Slack webhooks (`WEBHOOK`).

**The guard and the fix now share one definition.** The class table lives in
`spawn-env.ts`; `spawn-env-credentials.test.ts` derives its source-scanning pattern
from the same cores, so widening one widens both and a class dropped from the table
goes red in the scan *and* in the withholding. A guard that no longer matches what the
fix withholds reports a clean tree while the protection is gone, which is the whole
point of the pairing.

**Substring cores, not a name list.** A list cannot see a name that is not on it: the
`ETH_NODE_URI_*` entries grow by one per network and the explorer keys by one per
chain. The sweep walks the keys of the environment it is handed and matches by core, so
a network added tomorrow is covered the day it is added. Cores span the whole
credential-bearing part of a name, never the vendor alone — `MONGODB_URI` not
`MONGODB`, `SYNC_TOKEN` not `TOKEN` — so a logging toggle and a token-contract list
stay out.

Three names match a core and hold no secret, so they are withheld from the sweep *and*
from the scan: `NO_ETHERSCAN_API_KEY_REQUIRED` names a marker the verification helper
compares against (and exports as the explorer key when set, so a value would look like
a present key), and the two anvil entries are the publicly known local-node key and
`127.0.0.1`. Replacing any of those changes what a child does rather than withholding
anything from it.

## Scope of the coverage

Measured against the real store, which holds 225 names:

| class | names |
|---|---|
| `ETH_NODE_URI*` endpoints | 87 |
| provider / explorer keys | 63 |
| signing keys + proposal stores (#2364) | 18 |
| webhooks | 9 |
| **withheld** | **177** |
| non-credentials (paths, flags, limits, lists, salts) | 48 |

This applies to the two callers of `withholdCredentials`
(`strict-flag-placement.test.ts`, `ticket-gate-placement.test.ts`). It is a no-op in
CI: `ts-unit-tests.yml` passes no `env:` block and runners have no `.env`, so there is
nothing to withhold there. The protection bites on a developer machine, which is where
the store actually is — `bun test` loads `.env` into the parent, so the sweep sees all
225 names.

## What this does not cover

Stated here because a scanner that walks names cannot see what is not named:

- **A credential whose name carries no core** (a `FOO_SECRET_SAUCE`) is invisible to
  both halves. Adding a new kind of secret means adding its core — pinned by a test.
- **A name the spawning process does not hold** cannot be swept out of an environment
  it is not in, and the child re-loads the file for exactly those names. Only the five
  pinned names are immune, and the set is asserted rather than the members, because a
  loop over the list cannot see a name dropped from it. This matters most for
  `SAFE_SIGNER_PRIVATE_KEY`, which the store leaves unset — the sweep never sees it, so
  the pin is its only cover.
- **Five other tests spawn children and never call `withholdCredentials`**, building
  the environment by hand instead: `ticket-flag-channel`, `funnel-deploy-gate`,
  `update-deployment-logs`, `helperFunctions.gate`, `ledger-refusal-placement`. They
  are unchanged here, and each has its own load-bearing assertions — `funnel-deploy-gate`
  asserts `'Private key is missing'`, so setting a malformed key there would break it.
  The enforceable invariant is "a test that spawns calls the helper", and nothing
  asserts it yet. Recorded in the rule and tracked in
  [EXSC-996](https://linear.app/lifi-linear/issue/EXSC-996).
- **`GH_TOKEN` / `GITHUB_TOKEN` / `HEXAGATE_PAUSER_PAT` carry no core**, so the sweep
  passes them through. They are the most reachable uncovered class — developers export
  `GH_TOKEN=$(gh auth token)` and `checkCronLiveness.ts` reads it — but a bare `TOKEN`
  core would falsely sweep `ALLOW_TOKEN_CONTRACTS`, so which spelling to add is a
  judgement call rather than a widening. Left for EXSC-996.
- The two `TARGET_STATE_SPREADSHEET_ID_*` entries are the one judgement call in the
  residual 48: a Sheets ID is a locator, not an authenticator, so it is not withheld.

## Sweep of existing deletes

Two statements across the whole tree unset one of the *newly* covered names, both in
`script/utils/utils.test.ts`, both legitimate — that test is in-process and never
spawns, and the unset is what forces the fallback branch whose redaction is asserted
(set, `node_url` returns early and the assertion never runs). Both carry a
`// spawn-env:` marker with the reason. So the widened cores cost two markers, not the
marker tax on every fallback test that #2364 was right to worry about. The tree holds
five marked unsets in total; the other three predate this PR.

## The exposure the ticket asked me to confirm first

`execute-pending-timelock-tx.ts` reads `WEBHOOK_DEV_SC_GITHUB_CI_NOTIFICATIONS` and
POSTs to the live channel in `handleRevertedExecution`. Still unreachable from the
probe, and still only by control flow: `readBooleanFlag` refuses at the top of `run()`,
and the `--executeAll`/`--rejectAll` conflict check `process.exit(1)`s three statements
later — both before any Mongo or webhook read. Reaching the POST needs the queue opened
and a batch reverted on-chain past the retry threshold. It is now withheld by design
rather than by ordering.

## One incidental fix

Prettier breaks a `delete` whose *code* passes 80 columns — across the dot, and across
the brackets for a computed key — leaving the holder alone as the statement
(`delete process.env`), which carries no name to match. The scan joined only the gap
after the keyword, so both forms read as clean. It now collapses the whole member
expression, with a test per form. Separately, prettier relocates a trailing comment when
the delete is the body of an `if`/`else`, which detaches a marker from its delete; that
is why the `utils.test.ts` restore braces its `else`, and why the rule says to keep the
unset a standalone statement.

## Note for review

The comments here name env vars, which `200-typescript.md` discourages as volatile
detail. Kept deliberately: in a guard whose subject *is* the set of names, the names are
the functional content, and this matches the file #2364 established.

## Review-gate round

A pre-ready review pass found two defects in the guard itself, both now fixed and
mutation-covered:

- The reviewed-non-credential exemption was matched against the whole statement rather
  than the name being unset, so `if (env.PRIVATE_KEY_ANVIL) delete env.PRIVATE_KEY_PRODUCTION`
  read as clean. It is now keyed on the captured name.
- `env.X = undefined` leaves a name unset for the child's re-load exactly as `delete`
  does, and the guard asserted that *setting* is always safe — so following the rule to
  the letter could reopen the hole it closes. Measured in a hermetic child: `delete` and
  `= undefined` both hand back the fixture value at full length, `= ''` gives length 0.
  The scan now covers `= undefined` (including the line prettier breaks), `= ''` stays
  legitimate, and both behaviours are pinned by a real spawn.

Separately, `ticket-gate-placement.test.ts` deleted `SAFE_PROPOSAL_TICKET` under a
comment claiming it was cleared. It is in `.env.example`, so the child re-loaded a
developer's own ticket and decided the gate cases for them — now set empty.

### DIFF
--- a/.agents/rules/402-typescript-tests.md
+++ b/.agents/rules/402-typescript-tests.md
@@ -18,6 +18,34 @@ paths:
 - **Restore after each test**: when stubbing globals (e.g. `globalThis.fetch`), save the original in `beforeEach`/`beforeAll` and restore it in `afterEach`/`afterAll` to avoid cross-test pollution.
 - Prefer true unit tests that isolate the code under test; use mocks for any outbound calls (fetch, contract calls, file system if needed) so failures reflect logic bugs, not environment or network issues.
 
+## Withholding credentials from a spawned child ([CONV:TEST-SPAWN-CREDENTIALS])
+
+A test that spawns a child must **set** every credential name to a dummy, never
+`delete` it. Bun re-loads the repo `.env` inside the child for every name the
+passed environment leaves unset, so deleting a name hands the real value back at
+full length. Use `withholdCredentials()` from
+`script/deploy/safe/spawn-env.ts` rather than rolling your own.
+
+A name is a credential when it carries one of the cores in that module's class
+table — today `PRIVATE_KEY`, `MNEMONIC`, `MONGODB_URI`, `ETH_NODE_URI`,
+`API_KEY`, `ACCESS_KEY`, `SYNC_TOKEN`, `WEBHOOK`. The cores are anchored on the
+credential-bearing suffix, not the vendor, so a name that merely shares a prefix
+is not swept up; a name that holds no secret despite matching a core belongs in
+the same module's reviewed non-credential set, with the reason.
+
+`script/spawn-env-credentials.test.ts` derives its source scan from those same
+cores, so the guard and the withholding widen together. Two consequences:
+
+- **Deleting one of these names is red**, including in a test that never spawns.
+  When the delete is legitimate — an in-process test, or a child whose `cwd`
+  holds no `.env` — annotate it with a trailing `// spawn-env: <reason>` comment
+  on the delete's own line. The marker only counts on that line, so keep the
+  reason short enough to survive prettier at 80 columns; restructure the
+  statement rather than letting the formatter move the comment off it.
+- **A credential whose name carries no core is invisible** to both the scan and
+  the sweep. Adding a new kind of secret to the store means adding its core,
+  not assuming the pattern grew.
+
 ## Asserting rejections ([CONV:TEST-ASSERT-REJECTS])
 
 - **Never assert rejections via `expect(...).rejects`** — awaiting Bun's matcher trips `@typescript-eslint/await-thenable` because it isn't a real Promise. Use a local `async function expectRejects(promise, match)` that catches and matches the error message, as in `script/deploy/safe/parked-tasks.test.ts`.
--- a/.agents/rules/402-typescript-tests.md
+++ b/.agents/rules/402-typescript-tests.md
@@ -26,12 +26,13 @@ passed environment leaves unset, so deleting a name hands the real value back at
 full length. Use `withholdCredentials()` from
 `script/deploy/safe/spawn-env.ts` rather than rolling your own.
 
-A name is a credential when it carries one of the cores in that module's class
-table — today `PRIVATE_KEY`, `MNEMONIC`, `MONGODB_URI`, `ETH_NODE_URI`,
-`API_KEY`, `ACCESS_KEY`, `SYNC_TOKEN`, `WEBHOOK`. The cores are anchored on the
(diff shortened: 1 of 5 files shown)


0xDEnYO:
@coderabbitai review

gvladika (approved):

PR #2372, opened 2026-09-11
feat(attestations): mint a build-provenance attestation over a reproducible build manifest (EXSC-952)

0xDEnYO:

Ref EXSC-952 — [WP-5.2 · CI attestation minting](https://linear.app/lifi-linear/issue/EXSC-952)

`Ref` rather than `Fixes`: this is one slice of WP-5.2, not the whole package. The ticket's
own success criteria — the live mint → `gh attestation verify` round trip, the pinned Sigstore
trust root, compiler binaries pinned by hash, and wiring the WP-2.1 gate to consume the
manifest — are not met here, and an auto-close on merge already had to be undone once on #2351.

`buildAttestations.json` is a reproducible record of what each versioned contract
compiles to: one entry per contract, keyed by source closure and compiler settings,
carrying the masked runtime hash and the immutable offsets a reader needs to mask the
deployed code the same way. On push to `main`, CI mints a Sigstore build-provenance
attestation over the committed bytes, which is what makes the file first-party evidence
rather than a transcription.

**Nothing consumes it yet.** Wiring the WP-2.1 sign-time gate to read it instead of
recompiling is a separate change. This PR is the mint and its reproducibility check.

Design notes worth the reviewer's time:

- **Not path-filtered, on purpose.** A path filter turns this into a job that skips, and
  a skipped job reports success — which is how a gate goes quiet without anyone noticing.
  The build is the cost of never having to ask whether the check ran.
- **Immutable offsets are stored flat, not keyed by AST id.** An AST id counts source
  units in the order solc loaded them, so the same contract compiled in a different scope
  carries different ids for identical bytes; keyed by id, two honest machines mint two
  manifests and the attested digest stops surviving a rebuild.
- **`coveredProfiles` records what was minted, not what was requested.** A profile
  published as covered while contributing nothing tells a reader to grade an honest build
  under it as a mismatch, so the mint refuses to write the field unless at least one entry
  was produced under it.
- **Only the `default` profile is minted today**, and that is enforced rather than
  documented, at two points: `manifestEntryFrom` refuses any profile carrying a
  `zksolcVersion`, and **aborts the mint** on any artifact whose own `evmVersion` is missing or contradicts the
  profile's pin — `solc_floor` declares no `out`, so a floor build lands in the default
  profile's tree and would otherwise mint under the wrong label. A conflict aborts rather
  than skips because it is a statement about the tree, not about one contract: minting the
  rest would publish a manifest short by however many disagreed. A zksolc build is pinned
  by solc-fork and LLVM versions that live in the bytecode trailer alone,
  which the shared normalisation strips — so an entry minted under it would file two zk
  toolchains under one hash. Code built under any other toolchain must be treated as
  unverifiable rather than unattested.
- **Verification and minting are separate jobs.** A job's permissions cover every step in
  it, including ones a step-level `if` skips, so the signing scopes must not be live while
  `bun install` and repo scripts — both PR-editable — are running.
- **The lineage string states the toolchain only.** It used to say "built in CI" on every
  row, which this code also produces on a laptop.

## Known gap, called out deliberately

Neither `verify-build-manifest` (the job that runs on PRs) nor `mint` is in
`main protection`'s required status checks, and neither feeds a `-required` aggregator —
so a PR whose manifest has drifted turns the job red and can still merge. Worse, `mint`
has `needs: verify-build-manifest`, so a drifted manifest that lands on `main` fails the
push-run's verify and **skips the mint** — leaving `main` carrying a commit with no
attestation at all. The workflow is deliberately not path-filtered, so the job never skips
and can be registered directly rather than behind an aggregator. That registration is a
repo-admin change and is deliberately not bundled here.

The guard compares the evm version and not the solc pin, which is enough only while no two
non-zk profiles share an evm version (today: 0.8.29/cancun, 0.8.17/london, and a zksolc
profile the mint refuses outright). Nothing enforces that. A second non-zk cancun profile
would let a tree built under it pass the guard while labelled `default`, and closing that
means comparing solc too — which needs a normalisation decision first, since the artifact
reports `0.8.29+commit.ab55807c` where the profile pins `0.8.29`. That belongs with the
change that adds such a profile.

The profile guard corroborates the artifact against `foundry.toml` as it reads *now*, which a
deploy wave rewrites: `deployGroupingHelpers.sh` `sed`s `[profile.default]`'s `solc_version`
and `evm_version` to the london or cancun group's values and builds into the same `out/`. In
that window `default` genuinely pins london, the artifacts are london, the two agree and the
guard cannot fire. `restoreFoundryToml` is `trap`ped on EXIT in all three callers
(`deployContractToNetworks.sh`, `proposeContractToNetworks.sh`, `multiNetworkExecution.sh`), so
the exposure is a mint run against a tree whose restore never ran.

The manifest also carries no machine-readable `provenance` field. `IAttestedBuild.provenance`
is required elsewhere precisely because a CI mint and a local rebuild are otherwise
indistinguishable; deciding what stamps it belongs with the gate that will consume the file,
against a `MANIFEST_SCHEMA` bump.


0xDEnYO:
@coderabbitai review

0xDEnYO:
@coderabbitai review

.github/workflows/mintBuildAttestations.yml
```diff
@@ -0,0 +1,99 @@
...
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
+
+jobs:
+  mint:
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    permissions:
+      contents: read # required to fetch repository contents
+      id-token: write # required by attest-build-provenance to obtain the OIDC token it signs with
+      attestations: write # required to write the attestation to this repository
```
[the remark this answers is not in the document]
0xDEnYO: Taken, fixed in 3c5f27cd8. Split into a read-only `verify` job (checkout, bun install, `attestations:check`) and a `mint` job that holds `id-token`/`attestations: write`, runs only on push to main, and executes nothing out of the checkout but the attest action itself.

script/deploy/codehash/build-manifest.ts
```diff
@@ -0,0 +1,274 @@
...
+      // them by the key alone would leave the bytes decided by the order the
+      // runner emitted them, which is the one thing this file cannot afford.
+      const leftEntry = JSON.stringify(a)
+      const rightEntry = JSON.stringify(b)
+      if (leftEntry === rightEntry) return 0
+      return leftEntry < rightEntry ? -1 : 1
+    }),
+  }
+
+  return `${JSON.stringify(manifest, null, 2)}\n`
```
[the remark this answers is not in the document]
0xDEnYO: Taken, fixed in 3c5f27cd8. `hashedSettings` is now recursively canonicalised before it is embedded, which also makes the entry tie-breaker deterministic since it serialises the entry. Regenerating changed the committed manifest, confirming solc's order was not already sorted. Added a regression test that two orderings of equivalent settings produce identical bytes; verified it fails without the fix.

tasks/buildAttestationManifest.ts
```diff
@@ -0,0 +1,338 @@
...
+import {
+  manifestEntryFrom,
+  serialiseManifest,
+  type IManifestEntry,
+  type IMintProfile,
+} from '../script/deploy/codehash/build-manifest'
+import { parseBuildProfiles } from '../script/deploy/codehash/lineage-scope'
+
+const TARGET = 'script/deploy/resources/buildAttestations.json'
+const SOURCE_DIRS = ['src/Facets', 'src/Periphery', 'src/Security']
```
gvladika: **`LiFiDiamond` is silently outside the manifest.**

This list omits `src/` itself. The repo's own resolver, `script/deploy/shared/getContractVersion.ts:22-27`, searches four locations — these three plus `src/`.

`src/LiFiDiamond.sol` carries `@custom:version 1.0.0`, compiles to 254 bytes of runtime code, is present in `deployRequirements.json`, and appears in 84 deployment logs. It is not in the minted manifest (I checked: `[.entries[].key.contractName] | index("LiFiDiamond")` → absent).

What makes this more than a missing row is that it is not in the `skipped` report either — `versionedContracts()` never sees the file, so nothing tells you it is gone. That is the same failure this file works hard to prevent one level up: `coveredProfiles`, the barren-profile hard failure and the module header on `build-manifest.ts` all exist so a reader can distinguish "not minted" from "minted and empty". Here a reader cannot distinguish "LiFiDiamond wasn't built" from "LiFiDiamond isn't a thing".

Fix looks like one token — `readdirSync` is non-recursive, so adding `'src'` picks up `LiFiDiamond.sol` and nothing else, and the artifact path already keys off the basename:

```suggestion
const SOURCE_DIRS = ['src', 'src/Facets', 'src/Periphery', 'src/Security']
```

Expect the manifest to go 59 → 60 entries.

tasks/buildAttestationManifest.ts
```diff
@@ -0,0 +1,338 @@
...
+  profile: string
+  outDir: string
+}
+
+/**
+ * Parses `--builds name=dir,name=dir` into the trees to read.
+ * @param argv - process arguments
+ * @returns One entry per profile built
+ */
+const parseBuilds = (argv: string[]): IBuildInput[] => {
```
gvladika: This parses a comma-separated `<profile>=<dir>` list, with its own error messages, for exactly one call site carrying exactly one value (`--builds default=out`, hardcoded twice in `package.json`).

Given the PR deliberately mints only `default`, a single `--out <dir>` flag would cover every caller that exists. Worth folding into the same cut as the zk plumbing below — when a second profile actually lands, its shape will be decided by what that profile needs rather than guessed at now.

tasks/buildAttestationManifest.ts
```diff
@@ -0,0 +1,338 @@
...
+/**
+ * Prints where the committed manifest and this build part company.
+ *
+ * Without this the failure says only that two files differ, which on a
+ * cross-machine mismatch leaves no way to tell a stale commit from a build that
+ * is not reproducible — and those call for opposite responses.
+ * @param committed - the manifest as committed
+ * @param built - the manifest this checkout produces
+ */
+const entryNames = (text: string): string[] => {
```
gvladika: This JSDoc block is attached to the wrong function.

"Prints where the committed manifest and this build part company", with `@param committed` / `@param built`, describes `reportDrift` — but it sits above `entryNames`, which takes one parameter and prints nothing. `reportDrift` at line 166 has no doc at all.

gvladika (commented):
Reviewed the code and independently verified the core claim rather than taking it on trust.

**What I checked**

- **Reproducibility holds cross-platform.** Built locally on macOS (`forge build --ast`, 471 files, cold) and ran `--check` against the committed file: `is up to date (59 entries)`, zero skips. Byte-identical to what Ubuntu CI produced. That is the load-bearing property and it survives a different OS and a different solc binary.
- 14/14 new unit tests pass; `eslint` clean on all three TS files.
- All three pinned action SHAs resolve correctly (`attest-build-provenance` v4.2.2 → `4d10147…`, `checkout` v4.1.7, `setup-bun` v2.1.2).
- Prettier only globs `ts,js,sol`, so no formatter will fight the generator over the JSON — that risk is closed.
- `protect-critical-code` is red only for the InfoSec-manager approval, as expected for a `.github/workflows/` change.

**On the approach generally**

It is sound, and the key design is the best part — every field in `IAttestationKey` is there because omitting it lets two different builds share one entry, and each is a real failure mode rather than a hypothetical.

I went in sceptical of the Sigstore layer, on the theory that signed commits plus branch protection already prove the manifest is what is on `main`. That is wrong, and `attestation-provenance.ts` says why: the threat model this gate already committed to is a **compromised signer host**, and a host like that can present any file contents it likes. A manifest read off local disk is worth nothing on its own; a signature bound to a workflow identity is something that host cannot forge. The attestation is load-bearing, and gating `mint` on `verify` sharpens it from "these bytes were on main" to "these bytes reproduce from this commit's source". The expensive-looking parts — byte-stable serialisation, the permissions split, refusing to path-filter — are all downstream of that, and I would defend them.

Inline comments cover the specifics. The one correctness issue is the missing `LiFiDiamond`; the rest are trims and two things that are cheaper to settle before the consumer lands.

**One architectural decision worth re-opening before this merges**

Committing the manifest is what creates the churn treadmill. I measured it: 53 of 59 entries have `LibAsset.sol` in their source closure and 41 have `ILiFi.sol`, so a one-line change to either rewrites most of a 2,879-line tracked file. Every contract PR then has to regenerate and commit, and two concurrent contract PRs conflict on it in a way git cannot merge — the second rebases and regenerates, every time.

I had assumed that was the unavoidable price of having something to attest. It is not: `attest-build-provenance` will sign a file the mint job *generates* just as happily as one that is committed. Build on `main` → generate → attest → publish as a run or release artifact. Same signature, same threat coverage, nothing tracked, no churn.

And the compromised-host model actually pushes that way. If the signer must fetch the manifest and verify its attestation out-of-band anyway — which that model says they must, since local disk is untrusted — then the copy in the checkout is never the copy that gets used. Committing it buys reviewability, and at 2,879 lines of hashes nobody is reviewing that; what is reviewed is the source, and `verify` is what binds source to manifest.

The honest counter is that a committed baseline gives `verify` something to diff against at PR time, and that diff is a real review signal ("this source change moved these hashes"). You can get the same by diffing against the last attested manifest from `main`, but that is more machinery than `--check` against a tracked file.

Genuine trade-off, and I lean publish-don't-commit — but mostly I think it should be a decision rather than something inherited. Once the WP-2.1 gate reads the file, changing it means migrating the consumer too.

tasks/buildAttestationManifest.ts
```diff
@@ -0,0 +1,338 @@
...
+    )
+  for (const name of stale.sort())
+    console.error(
+      `  the committed manifest carries ${name}, this checkout does not build it`
+    )
+
+  return missing.length > 0 || stale.length > 0
+}
+
+const reportDrift = (committed: string, built: string): void => {
```
gvladika: This hand-rolls a positional line diff of a file that is in git, in a failure path whose message already says to run `bun attestations:mint`. Once you have regenerated, `git diff` shows exactly what moved — and does it better, since the comment on `reportEntrySetDrift` correctly notes that a positional diff degrades into a cascade of false renames the moment entries shift.

`reportEntrySetDrift` earns its place: "which contract entered or left" is a question a line diff genuinely cannot answer. This one I would drop and let the failure message point at `git diff`.

tasks/buildAttestationManifest.ts
```diff
@@ -0,0 +1,338 @@
...
+        )
+        continue
+      }
+
+      const identity = identityFromArtifactMetadata(settings)
+      const key: IAttestationKey = {
+        contractName: contract.name,
+        sourceId: identity.sourceId,
+        version: contract.version,
+        closureHash: sourceClosureHash(
```
gvladika, line 201: Two escape hatches from the skip path here.

The loop above carefully accumulates `skipped` for a missing artifact, empty runtime code, absent metadata and an unnormalisable artifact. Then `identityFromArtifactMetadata` (line 232) and `sourceClosureHash` (line 237) both **throw**, out of a `main()` with no try/catch. An artifact naming 0 or 2 compilation targets, or one whose settings canonicalise to `{}`, aborts the whole mint with a raw stack trace instead of one line in the skip report.

Worse for diagnosis: `identityFromArtifactMetadata`'s message ("build artifact names 2 compilation targets…") does not name the contract, so you would be reading a stack trace to work out which of 59 it choked on.

Either route both into `skipped`, or keep them deliberately fatal — that is defensible, they do mean the artifact is malformed — but then say so in a comment and name the contract in the error.

script/deploy/codehash/build-manifest.ts
```diff
@@ -0,0 +1,300 @@
...
+ * only that CI produced exactly these bytes.
+ *
+ * Two properties follow, and both are load-bearing:
+ *
+ * **The serialised form must be byte-stable.** The attestation binds a digest,
+ * so a regeneration that differs by a space is a manifest the attestation no
+ * longer covers. Entries are therefore ordered by their own key rather than by
+ * the order a build runner happened to emit them.
+ *
+ * **A profile nobody minted is not an empty attested set.** Code built under a
```
gvladika, line 19: A note on the prose rather than the code.

The "an unminted profile is not an attested-empty set" argument is written out in full in this module header, again on the `coveredProfiles` field below, again in two test comments, again in the workflow header, and again in the PR description. That is five copies of one argument, and they will drift.

`.agents/rules/000-global-standards.md` is direct about this: never restate the history, debate, or alternatives behind a line — that belongs in the PR description or the ticket. The reasoning here is good and worth keeping *once*; I would make this header the canonical statement and let the other four sites point at it or say nothing.

Same applies to the byte-stability argument, which appears in the header, on `serialiseManifest`, and in two tests.

script/deploy/codehash/build-manifest.ts
```diff
@@ -0,0 +1,300 @@
...
+  lineage: string
+  /** Length of the runtime code as built, before stripping or masking. */
+  rawByteLength: number
+  /**
+   * keccak of the exact runtime bytes, present only where the comparison pins
+   * them. On zkEVM the solc-fork/LLVM version lives in the trailer alone, so
+   * stripping it is exactly what makes fork drift invisible; elsewhere the
+   * trailer holds a source digest that moves for a changed comment.
+   */
+  rawHash?: string
```
gvladika: The zkEVM plumbing in this file is unreachable in production, by this PR's own design.

`rawHash` here, `IMintProfile.zksolcVersion`, the zk branch in `describeLineage`, and the `isZk` fork at line 220 all exist for a profile the PR explicitly refuses to mint: `coveredProfiles` is `["default"]`, and the "artifact has no metadata" skip in the task means a zksolc artifact can never produce an entry anyway. The tests covering it are therefore tests of code that cannot run.

`.agents/rules/000-global-standards.md` and the repo philosophy are explicit on this ("no premature abstraction", "no phantom features"). I would cut it and let the zksync PR add it back — at which point its shape gets decided by what zksolc artifacts actually look like, rather than by what we expect them to look like today.

Not a blocker, and I recognise it mirrors shapes already in `attestation-key.ts` / `rebuild-attestations.ts`, so there is a consistency argument for it. Your call.

script/deploy/codehash/build-manifest.ts
```diff
@@ -0,0 +1,300 @@
...
+    )
+
+  const conflicts = findAttestationConflicts(entries)
+  if (conflicts.length > 0)
+    throw new ManifestConflictError(conflicts.map((c) => c.serialisedKey))
+
+  const manifest: IBuildManifest = {
+    schema: MANIFEST_SCHEMA,
+    coveredProfiles: [...new Set(coveredProfiles)].sort(),
+    entries: [...entries].sort((a, b) => {
```
gvladika, line 316: Sorting on `serialiseAttestationKey` makes the file's order read as arbitrary to a human.

That function is length-prefixed (`10:ERC20Proxy`), so the manifest comes out ordered by contract-name **length** first, then alphabetically within each length bucket. The committed file opens: `ERC20Proxy, MayanFacet, SquidFacet, AcrossFacet, GardenFacet, GasZipFacet, GlacisFacet, LidoWrapper…`

Byte-stability only requires *a* total order, not this one. Sorting on `contractName` then `version`, falling back to the serialised key for a genuine tie, is equally stable and makes a 2,879-line file something a reviewer can actually scan — which is the stated reason for committing it at all.

.github/workflows/mintBuildAttestations.yml
```diff
@@ -0,0 +1,117 @@
...
+
+# Default-deny at the workflow level; grant only what each job needs.
+permissions: {}
+
+concurrency:
+  group: mint-attestations-${{ github.event_name }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
+
+jobs:
+  verify:
```
gvladika: Two things about this job, both cheaper to fix now than later.

**The name is too generic for a required status check.** Every other gating job in this repo is descriptive — `verify-immutable-registry`, `verify-signed-commits`, `validate-scripts`, `audit-verification`, `version-control`. Branch rulesets match required checks by name string, so a bare `verify` is a landmine for the next workflow that wants a job called `verify`, and renaming after registration means re-editing the ruleset. `verify-build-manifest` costs nothing today.

**The consequence of it not being required is worse than the PR description states.** The description says a PR whose manifest has drifted can still merge. The part it does not say: `mint` has `needs: verify`, so a drifted manifest that lands on `main` fails the push-run's `verify`, which **skips the mint**. `main` then carries a commit with no attestation at all, and the only signal is a red workflow nobody is gated on.

The repo already has the convention for closing this (`solc-floor-build.yml:102`, the `*-required` aggregator). Here you do not even need an aggregator: the workflow is deliberately not path-filtered, so `verify` never skips and can be registered directly. That reduces it to a one-line ruleset change rather than a code change — which argues for doing it in this cycle instead of deferring.

.github/workflows/mintBuildAttestations.yml
```diff
@@ -0,0 +1,117 @@
...
+      - name: Checkout repository
+        uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
+        with:
+          persist-credentials: false
+
+      - name: Mint the attestation
+        id: attest
+        uses: actions/attest-build-provenance@4d101475d8b20a2381f78447822ac1eab6504dd8 # v4.2.2
+        with:
+          subject-path: script/deploy/resources/buildAttestations.json
```
gvladika, line 109: This is the first workflow in the repo to hold `attestations: write`, and nothing in the tree says how the minted attestation is meant to be checked — there is no `gh attestation verify` anywhere in `script/`, `docs/` or `.github/`.

That matters more than it looks. A bare

```
gh attestation verify buildAttestations.json --repo lifinance/contracts
```

proves only that *some* workflow in this repo signed those bytes. For the signature to carry the meaning this file's header claims, the consumer has to pin the signer:

```
gh attestation verify buildAttestations.json \
  --repo lifinance/contracts \
  --signer-workflow lifinance/contracts/.github/workflows/mintBuildAttestations.yml
```

Today no other workflow has the permission, so the gap is theoretical. But the whole point of the manifest is to be read later by someone who was not part of this review, and the instruction for reading it safely should ship with the thing being signed — not with the consumer, a PR or two from now.

A short block in the workflow header would do, or better a paragraph in `docs/` next to the codehash-gate material.

tasks/buildAttestationManifest.ts
0xDEnYO: Fixed in e4ec5b4be. `SOURCE_DIRS` now leads with `src`, and the comment on it names `getContractVersion.ts` as the list it has to stay in step with, since that coupling is the thing that will rot.

60 entries, as you predicted. `LiFiDiamond@1.0.0`, 254 bytes.

I went looking for any second instance of the same class rather than just the one you found — enumerating from `deployments/*.json` (88 names) and `deployRequirements.json` (63) instead of from the four directories the code already reads. 28 and 15 names respectively are absent from the manifest, and every one has no source under `src/` at all (`AmarokFacet`, `CircleBridgeFacet`, `LiFiDiamondImmutable`, `HopFacet`, …), so there is nothing to mint. No versioned in-repo contract is outside it today.

Two ways to be invisible do remain, both with no live instance: `readdirSync` is non-recursive, so a contract in a nested directory would be missing from the manifest and the skip report both; and a `.sol` with no `@custom:version` is dropped silently. The comment now says so instead of claiming the hole is closed.

tasks/buildAttestationManifest.ts
0xDEnYO: Taken, e4ec5b4be. `--out <dir>`, profile constant.

It paid for more than the parser: `coveredProfiles` no longer needs deriving from the entries, so the `barren` bookkeeping and its error message went too — with one profile, the existing `entries.length === 0` throw already proves the profile contributed, and that is what the remaining comment says.

tasks/buildAttestationManifest.ts
0xDEnYO: Fixed in e4ec5b4be. The block moved to `reportDrift`'s replacement and `entryNames` got its own; `reportDrift` itself is gone, see the next thread.

tasks/buildAttestationManifest.ts
0xDEnYO: Dropped in e4ec5b4be. The failure now says `bun attestations:mint && git diff -- <target>`.

`reportEntrySetDrift` stays, and its comment no longer defines itself against the function you had me delete — it states its own case: entries are sorted, so a contract entering or leaving shifts every line after it, which is the one question `git diff` cannot answer at a glance.

tasks/buildAttestationManifest.ts
0xDEnYO: Took the second option — deliberately fatal, said so, and named the contract. A malformed artifact is not a legitimate skip the way an interface or a zksolc artifact is, and minting the rest around it publishes a manifest nobody noticed was short.

Verified rather than assumed. Corrupting a real `LiFiDiamond.json` to name two compilation targets now gives:

```text
Error: LiFiDiamond (out/LiFiDiamond.sol/LiFiDiamond.json): build artifact names 2 compilation targets, so which source it compiled cannot be established
```

Same for empty settings and a malformed source digest. One correction to my own first comment on it, caught while proving each site can fire: I had listed `sourceClosureHash`'s duplicate-path refusal among what this catches, and it cannot — `metadata.sources` is a JSON object, so `Object.entries` yields unique paths by construction and `JSON.parse` collapses a literal duplicate key. That branch is dead with respect to this caller. The comment in 80e0313f5 no longer claims it.

One left unfixed and out of scope here: an artifact with a source entry *missing* `keccak256` fails as `Cannot read properties of undefined (reading 'length')` from `strip0x`. The rethrow still names the contract, but the message is opaque. That is in `attestation-key.ts`/`hex.ts`, shared with the sign-time gate, so I would rather not touch it in this PR.

script/deploy/codehash/build-manifest.ts
0xDEnYO: Taken, e4ec5b4be. This header is the one statement; the field doc, the two test comments and the workflow header now point here or say nothing. Same for byte-stability — trimmed off `serialiseManifest` and out of the test file header.

You were right again about a copy I added in the same round: the `--signer-workflow` explanation from the thread below went into both the workflow header and the docs section. 80e0313f5 leaves it in `docs/MultisigSigningProcess.md` §8.1 only, with a pointer from the workflow.

script/deploy/codehash/build-manifest.ts
0xDEnYO: Cut, but not quite the way you proposed — the shape of the cut is the part worth arguing with.

Removing `isZk` entirely would have left `normalizeRuntimeCode(…, { isZk: false })` as a bare assertion this module is in no position to make. So instead of deleting the concept, `manifestEntryFrom` now **refuses** a profile carrying `zksolcVersion`, and `rawHash`, the zk lineage branch and the fork are all gone. Same amount of speculative code removed, and the PR's stated limitation is enforced rather than documented.

Proved both directions against the real `foundry.toml` rather than trusting it: `parseBuildProfiles` attaches the `zksolc = "1.5.15"` pin only to the profile literally named `zksync`, so

- `zksync` → `ok=false`, `profile zksync is zkEVM, which this mint does not cover`
- `default` → `ok=true`, and the full mint still produces 60 entries with **zero** skip lines.

The test that used to pin `rawHash` now pins the refusal, and it fails against the pre-commit implementation.

script/deploy/codehash/build-manifest.ts
0xDEnYO: Taken. `contractName` → `version`, falling back to the serialised key and then the whole entry. The file now opens `AccessManagerFacet, AcrossFacet, AcrossFacetPackedV4, AcrossFacetV4, AcrossV4SwapFacet, AllBridgeFacet, …` — byte-for-byte `LC_ALL=C sort` of the names.

Your point about length-prefixing sent me to check whether the tests actually pinned any of this, and two did not. Mutation-tested each tier against the suite:

- version tier deleted → the version test still passed, because with name and sourceId equal the key's third field *is* the version. Rewritten so the two orders disagree (newer version at an earlier-sorting path); it now fails without the tier.
- the two lower tie-breakers replaced with `return 0` → 15/15 still green, while the serialised bytes became input-order dependent, which the module header calls the one thing this file cannot afford. Added a test for entries the key cannot separate (same key, agreeing bytecode, different profile).

Both mutations now die, and each kills only its own test. On real data tier 1 decides every comparison — 60 entries, 60 unique names — so the rest is defensive, but it is no longer defensive *and* unpinned.

.github/workflows/mintBuildAttestations.yml
0xDEnYO: Both taken.

Renamed to `verify-build-manifest` in e4ec5b4be; it has already run green under the new name. Checked the ruleset rather than assuming: the 11 required contexts are `validate-json, audit-verification, solc-floor-build-required, protect-critical-code, validate-scripts-required, unit-tests-required, version-control, protect_audit_labels, check-new-network-health, smoke-test-required, ts-tests-required`. The old bare `verify` was not among them, so the rename breaks nothing.

The consequence you spelled out — `mint` needing `verify-build-manifest`, so a drifted manifest on `main` skips the mint and leaves a commit with no attestation — was the part the PR description got wrong, and it is now in the description in those terms. Registering the check is the one thing I have left for the merge: it is a ruleset edit that blocks every currently-open PR the moment it lands, so it belongs immediately after this merges rather than before.

.github/workflows/mintBuildAttestations.yml
0xDEnYO: Agreed, and took the better version — a paragraph in `docs/`, not just a workflow comment. New §8.1 in `docs/MultisigSigningProcess.md`, next to the codehash-gate material, carrying the pinned invocation:

```bash
gh attestation verify script/deploy/resources/buildAttestations.json \
  --repo lifinance/contracts \
  --signer-workflow lifinance/contracts/.github/workflows/mintBuildAttestations.yml
```

It states why the bare form is the weaker claim, and that the set of workflows holding `attestations: write` is not fixed — which is what makes the gap more than theoretical over time. §8 also gains rows for the workflow and the generator.

I first duplicated the same explanation into the workflow header, which is exactly what you objected to two threads up; 80e0313f5 reduces that to a pointer.

0xDEnYO:
@coderabbitai review

0xDEnYO:
On the architecture: **keeping it committed**, deliberately rather than by inheritance.

Your measurement is the strongest argument against, and I am not disputing it — 53 of 59 closures carry `LibAsset.sol`, so the churn and the unmergeable conflicts are real and recurring. The reason I am still landing it committed is the one you named as the honest counter: `verify-build-manifest` needs something to diff at PR time, and "this source change moved these hashes" is a signal at the moment a human is actually looking. Diffing against the last attested manifest from `main` gets the same information, but it puts a network fetch and an attestation verification inside the PR gate, which is machinery this PR does not have and a new way for the gate to go amber.

Where I think your argument lands is narrower than publish-vs-commit, and I agree with it: the copy in the checkout is never the copy a signer may trust. That is now written down — `docs/MultisigSigningProcess.md` §8.1 says the consumer fetches and verifies the attestation with `--signer-workflow`, not that it reads the file off local disk. So the committed file is the review artifact and the drift baseline; the attested bytes are what anything security-relevant consumes. Those being two different reads of the same content is the part worth being explicit about, and it was not before.

The churn stays a real cost and I would revisit it if it starts costing rebases weekly — at which point the migration is the workflow plus `--check`, not the consumer, because the consumer was never going to read the tracked copy anyway.

Everything else in your review is fixed across `e4ec5b4be` and `80e0313f5`; replies are on the individual threads. Two of your points made me go back and check my own work harder than I had: the length-prefix observation turned up two tests that passed against a mutated implementation, and the `LiFiDiamond` one turned up the fact that I had documented a refusal in the fatal path that cannot actually fire.

.github/workflows/mintBuildAttestations.yml
```diff
@@ -0,0 +1,121 @@
...
+
+      - name: Report the mint
+        if: steps.attest.outcome == 'success'
```
gvladika: `steps.attest.outcome` cannot be anything but `success` at this point.

The attest step above has no `continue-on-error`, so a failure ends the job and this step never runs at all. The condition reads as though a failed mint could still be reported, which is the one outcome it cannot have. Dropping the `if:` changes no behaviour and removes the implication.

gvladika (commented):
Re-reviewed at `80e0313f5`. All ten threads land. I checked each against the tree rather than against the replies, and read the two commits you pushed beyond what I asked for — the zk refusal and the sort tiers — as fresh code.

Verified here, not taken on trust:

- **Reproducibility on a machine that did not mint it.** `bun attestations:check` in my checkout: `script/deploy/resources/buildAttestations.json is up to date (60 entries)`. That is the claim the PR rests on, and it holds off your machine.
- **No second blind spot.** I enumerated `deployments/*.json` and `deployRequirements.json` independently of your reply: 45 names in the logs are absent from the manifest, and not one of them has a source anywhere under `src/`. `SOURCE_DIRS` is now exactly the four paths `getContractVersion.ts` resolves.
- **The non-recursive scan is the right call, not just the cheap one.** I checked what recursion would drag in before accepting the documented hole: 13 versioned sources under `src/Libraries` and `src/Helpers` compile to non-empty runtime code — nine 85-byte library stubs, `SwapperV2` / `Validatable` / `LiFiData`, and `TransferrableOwnership` at 1135 bytes — and would mint as entries for units nothing deploys standalone. An explicit list plus a comment naming the coupling beats that.
- **Sort tiers.** The file opens `AccessManagerFacet, AcrossFacet, AcrossFacetPackedV4, AcrossFacetV4, AcrossV4SwapFacet, AllBridgeFacet, …`, and `[.entries[].key.contractName]` equals its own `sort`. Both new tests pick cases where the key order and the intended order disagree, so they fail if the comparator goes back to the key — which is exactly what the old test did not do.
- **The zk cut costs nothing downstream.** `rawHash` was only ever on `IManifestEntry`, not on `IMintedAttestation`, so no consumer loses a field; the `rawHash` sites in `rebuild-attestations.ts`, `attested-set.ts` and the sign gate are their own types and untouched. `tsc` is clean on both changed TS files, and 394 codehash tests pass.
- **`actions/attest-build-provenance@4d10147…` really is v4.2.2** — pin and comment agree.
- `protect-critical-code` is red only for the missing InfoSec-manager approval.

One nit inline. One thing I want tracked rather than remembered:

**The ruleset edit.** I accept the ordering argument — registering `verify-build-manifest` as required before this merges blocks every open PR until it rebases onto a main that carries the workflow. What I do not want is "immediately after this merges" being the only record of it. Until it is registered, the failure mode is the one your own description spells out: a drifted manifest lands on main, the push-run's verify fails, `mint` is skipped, and main carries a commit with no attestation behind a red check nobody is gated on. Please open an EXSC ticket for the ruleset change and link it here, so the gap has an owner if the merge and the edit come apart.

Nothing else outstanding — approving once that link is here.

0xDEnYO:
@coderabbitai review

script/deploy/codehash/build-manifest.ts
```diff
@@ -0,0 +1,322 @@
...
+    }
+
+  // The profile is what the caller says the tree was built under; this is the
+  // build's own account of itself. They part company because `solc_floor`
+  // declares no `out` of its own and so writes into the default profile's
+  // tree — run the floor build, mint, and every entry is filed under `default`
+  // while `coveredProfiles` says the toolchain was covered. Refusing here keeps
+  // the profile label something the artifact corroborates.
+  const builtFor = hashedSettings['evmVersion']
+  if (builtFor !== profile.evmVersion)
```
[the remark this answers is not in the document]
0xDEnYO: Took the first half, and it was load-bearing — thanks.

A profile conflict is now `disposition: 'abort'` and the task throws instead of pushing to `skipped`. I measured the hole before fixing it: on a mixed tree (real cancun `out/` with a real london `solc_floor` build over the top) the old code wrote a 59-entry manifest and `--check` reconciled it against a committed file carrying the same omission. The `entries.length === 0` throw did not fire, because a path-scoped floor build (`FOUNDRY_PROFILE=solc_floor forge build`, which `foundry.toml` documents) is not all-or-nothing.

`disposition` is required rather than optional so a refusal added later has to choose — the silent default would be the one that keeps minting. `manifest-abort.cli.test.ts` spawns the real task, since a field nothing reads would leave the unit tests green.

Leaving the solc comparison out, deliberately. The two sides are not comparable as written: the artifact reports `0.8.29+commit.ab55807c` where the profile pins `0.8.29`, so a direct comparison false-reds the whole fleet and doing it properly means deciding how the commit suffix is treated. It also adds no coverage today — the profiles are 0.8.29/cancun, 0.8.17/london and 0.8.29/cancun+zksolc, so every non-zk pair already differs in evm version and the guard above catches them. It starts mattering when a profile is added that differs in solc alone, and that is the change that should carry it.

script/deploy/codehash/manifest-abort.cli.test.ts
```diff
@@ -0,0 +1,97 @@
...
+    sources: {
+      [`src/${contract.file}`]: { keccak256: `0x${'ab'.repeat(32)}` },
+    },
+  },
+})
+
+describe('the mint refuses to publish a manifest short by a profile conflict', () => {
+  it('exits non-zero naming the artifact, instead of skipping past it', () => {
+    const contract = firstVersionedContract()
+    const outDir = fs.mkdtempSync(path.join(tmpdir(), 'mint-abort-'))
```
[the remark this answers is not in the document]
0xDEnYO: Fixed in 49ed5ac3e — `afterEach` with `rmSync(outDir, { recursive: true, force: true })`, matching `deployRegistrationPropagation.test.ts`. Verified by count rather than by reading: two consecutive runs leave zero `mint-abort-*` directories in `TMPDIR`, where before each run left one behind.

.github/workflows/mintBuildAttestations.yml
0xDEnYO: Agreed and dropped, in 1abb62717 — apologies for the late reply, it rode along in a commit whose subject names only the settings refusal, so it went unattributed.

The `if:` is gone from the "Report the mint" step; `grep steps.attest.outcome` over the workflow now returns nothing. As you say, the attest step carries no `continue-on-error`, so the condition could only ever be true.

tasks/buildAttestationManifest.ts
```diff
@@ -0,0 +1,293 @@
...
+      `  this checkout builds ${name}, the committed manifest does not carry it`
+    )
+  for (const name of stale.sort())
+    console.error(
+      `  the committed manifest carries ${name}, this checkout does not build it`
+    )
+}
+
+const main = (): void => {
+  const outDir = parseOutDir(process.argv)
```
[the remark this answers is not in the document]
0xDEnYO: Taken, across 1abb62717 and b7eec5f77 — this thread is where the guard came from.

`manifestEntryFrom` now compares the artifact's own `metadata.settings.evmVersion` against the profile pin, so a `solc_floor` build landing in the shared `out/` is refused rather than filed under `default`. Both halves you named are covered: a mismatching version, and an artifact that reports none at all (which is a separate refusal, since "said nothing" and "said something wrong" are different claims and the reason string is all the skip report shows).

A refusal is now `disposition: 'abort'` and stops the mint rather than skipping the contract — on a mixed tree the skip path wrote a 59-entry manifest that `--check` reconciled against a committed file with the same omission.

Regression tests as asked, in `build-manifest.test.ts` and `manifest-abort.cli.test.ts` — the latter spawns the real task, since a disposition nothing acts on leaves the unit tests green. Falsified against real artifacts: a real london/0.8.17 build is refused, the real cancun `out/` tree still mints all 60 entries.

gvladika (approved):

melianessa (approved):

PR #2377, opened 2026-09-11
feat(clear-signing): generate a testsv2 fixture for every display.formats entry

gvladika:

Fixes [EXSC-1007](https://linear.app/lifi-linear/issue/EXSC-1007/clear-signing-generate-testsv2-fixtures-so-every-displayformats-entry)

The registry added a per-format test coverage check on 2026-09-10 ([#2965](https://github.com/ethereum/clear-signing-erc7730-registry/pull/2965)): it derives the selector of every `display.formats` entry and fails the descriptor unless some test case in `testsv2/` calls it. Our sync bot only ever wrote the descriptor, so it structurally could not pass — sync PR [#2955](https://github.com/ethereum/clear-signing-erc7730-registry/pull/2955) broke with 62 of 69 functions uncovered, and every future sync that adds a format would break the same way. The fixture has to travel with the descriptor.

Generating rather than hand-writing is what makes 69 cases tractable, and two properties of the registry's own contract make it safe: `rawTx` is explicitly an **unsigned** transaction (its runners do not verify signatures), so calldata encoded from the format signature is a complete fixture with no chain access and no keys; and the reference runner (`sourcifyeth/clear-signing-test-runner`, the same one the registry's CI runs) prints what it rendered, so expectations come from the implementation under test rather than from my reading of the descriptor. Values come from templates for `_bridgeData`, `_swapData` and the top-level `swapTokens*` parameters; everything else is derived from the ABI type, since only displayed fields reach a screen.

The reviewed fixture lives in this repo, at `config/clearSigningTests.json`, and the registry copy is an output of it: the sync reads the repo file, writes the registry path, and never reads the registry's copy back. The sync branch on our fork cannot hold reviewed state — it is recreated from upstream and force-pushed on every run, and anyone may delete it — so a new format is reviewed through a normal PR here, alongside the proposal change that made it necessary. The cost is that an edit made directly on the registry is replaced rather than merged; it shows up as a diff in the sync PR, and the fix is to make the same change here.

A generated fixture's real risk is that it passes while asserting nothing, so three guards are explicit. `--check` fails when a field the descriptor **displays** would render a zero value: this caught a real bug during development, where the 13 `BRIDGE_EXTRA_RECEIVERS` bridges were asserting `Non-EVM Recipient: 0x0…0` — green in all three runners, worthless as a test of where funds go. It also found that the top-level `swapTokens*` parameters had no template at all. `--check --tests` fails when a format has no case in the reviewed fixture; without it that failure lands on the next sync, after merge, rather than on the PR that introduces the format. Both run in `verifyClearSigning.yml` because they are purely static. The third guard refuses to write a fixture whose expectation is still `PENDING`, so a new facet's format cannot reach the registry as a case that fails the registry's own runners.

Two checks cannot run at PR time and so run in the sync. It re-runs the zero-value half of `--check` against the merged descriptor, which is the only place formats the registry preserves but we never proposed can be checked, since PR CI cannot see upstream. And every failure of the sync job now posts to Slack: it runs unattended on push-to-main and a monthly cron, and its most likely failure leaves the registry descriptor silently frozen.

Existing cases are kept verbatim — they carry real transactions and expectations a reviewer already signed off — and `--results` never overwrites a block that is not `PENDING`: if the renderer disagrees with a reviewed expectation, that is a finding to investigate, not something to paper over. The `jq -c` before prettier is not incidental: prettier keeps an object expanded if its input was, so without minifying first the fixture carries a whole-file reformat diff over the reviewed cases — the same problem EXSC-781 solved for the descriptor with `erc7730 format`.

One gap is known and documented rather than fixed: a descriptor change that alters an **already-covered** format leaves that case's expectation untouched, so a renamed label can leave the fixture asserting a label the descriptor no longer emits. The generator preserves reviewed expectations rather than guessing at new ones and cannot re-render without the reference runner, so that case needs a deliberate pass through the render loop in `docs/ClearSigningProposal.md`, and nothing flags it automatically.

### DIFF
--- a/.github/workflows/syncLedgerClearSigning.yml
+++ b/.github/workflows/syncLedgerClearSigning.yml
@@ -106,6 +106,7 @@ jobs:
           LEDGER_FORK: ${{ secrets.LEDGER_FORK || 'lifinance/clear-signing-erc7730-registry' }} # optional override
           LEDGER_UPSTREAM: ethereum/clear-signing-erc7730-registry # Ethereum Foundation-hosted (migrated from LedgerHQ in 2026)
           LEDGER_FILE_PATH: registry/lifi/calldata-LIFIDiamond.json
+          LEDGER_TESTS_PATH: registry/lifi/testsv2/calldata-LIFIDiamond.tests.json
           SYNC_BRANCH: sync/lifi-clear-signing
           PR_TITLE: 'lifi: sync calldata-LIFIDiamond.json (deployments + display.formats)'
           # Dry-run only applies on workflow_dispatch. Cron + push triggers always
@@ -165,6 +166,24 @@ jobs:
           echo -e "\033[32mFormatting ${LEDGER_FILE_PATH} with erc7730 (registry canonical formatter)...\033[0m"
           erc7730 format "ledger-registry/${LEDGER_FILE_PATH}"
 
+          # The registry fails a descriptor whose display.formats entries are not
+          # each exercised by a test case, so the fixture has to travel with the
+          # descriptor. Existing cases are kept verbatim: they carry real
+          # transactions and expectations a reviewer has already signed off.
+          echo -e "\033[32mGenerating ${LEDGER_TESTS_PATH} (one case per display.formats entry)...\033[0m"
+          bunx tsx tasks/generateClearSigningTests.ts \
+            --descriptor "ledger-registry/${LEDGER_FILE_PATH}" \
+            --existing "ledger-registry/${LEDGER_TESTS_PATH}" \
+            --out /tmp/lifi-tests.json
+
+          # The registry stores these prettier-formatted at printWidth 120.
+          # Prettier keeps an object expanded if its input was, so the generator's
+          # indented output has to be minified first for prettier to re-compact
+          # it — without this the fixture carries a whole-file reformat diff.
+          jq -c . /tmp/lifi-tests.json > /tmp/lifi-tests.min.json
+          bunx prettier --parser json --print-width 120 /tmp/lifi-tests.min.json \
+            > "ledger-registry/${LEDGER_TESTS_PATH}"
+
           cd ledger-registry
           if [ -z "$(git status --porcelain)" ]; then
             echo -e "\033[32mNo changes detected. Nothing to sync.\033[0m"
@@ -182,13 +201,13 @@ jobs:
             # lines, which sends SIGPIPE to `git diff` and fails the step.
             # First-run dry-runs have a multi-thousand-line diff (one block per
             # display.formats entry), so this matters in practice.
-            git --no-pager diff -- "${LEDGER_FILE_PATH}" | sed -n '1,200p'
+            git --no-pager diff -- "${LEDGER_FILE_PATH}" "${LEDGER_TESTS_PATH}" | sed -n '1,200p'
             echo ""
             echo -e "\033[33m[dry-run] Skipping commit/push/PR-create/Slack. Done.\033[0m"
             exit 0
           fi
 
-          git add "${LEDGER_FILE_PATH}"
+          git add "${LEDGER_FILE_PATH}" "${LEDGER_TESTS_PATH}"
           git -c user.name="lifi-action-bot" -c user.email="lifi-action-bot@users.noreply.github.com" \
             commit -m "${PR_TITLE}" -m "Auto-generated from lifinance/contracts (deployments and display.formats from config/clearSigningProposal.json; deprecated context.contract.abi stripped)."
 
--- a/.github/workflows/verifyClearSigning.yml
+++ b/.github/workflows/verifyClearSigning.yml
@@ -20,6 +20,7 @@ on:
       # without touching a facet.
       - 'src/Libraries/**'
       - 'tasks/buildClearSigningProposal.ts'
+      - 'tasks/generateClearSigningTests.ts'
       - 'config/clearSigningProposal.json'
       - '.github/workflows/verifyClearSigning.yml'
       - '.github/actions/setup-foundry/**'
@@ -62,6 +63,14 @@ jobs:
         # the canonical shape or the generator is extended to handle the new
         # shape.
         run: bunx tsx tasks/buildClearSigningProposal.ts
+
+      - name: Verify every displayed field can be given a real test value
+        # The registry requires one test case per display.formats entry, and a
+        # case whose displayed field renders 0x0 passes every runner while
+        # asserting nothing. This catches a new format whose displayed component
+        # the fixture generator has no value for — before the sync bot pushes a
+        # descriptor whose fixtures would be hollow.
+        run: bunx tsx tasks/generateClearSigningTests.ts --descriptor config/clearSigningProposal.json --check
 
       - name: Verify config/clearSigningProposal.json is up-to-date
         # The generator just wrote a fresh proposal. If it differs from the
--- a/tasks/generateClearSigningTests.ts
+++ b/tasks/generateClearSigningTests.ts
@@ -1,0 +1,562 @@
+import fs from 'fs'
+
+import { defineCommand, runMain } from 'citty'
+import type { AbiFunction, AbiParameter } from 'viem'
+import {
+  encodeFunctionData,
+  parseAbiItem,
+  parseTransaction,
+  serializeTransaction,
+  toFunctionSelector,
+} from 'viem'
+
+const ZERO_ADDRESS = '0x0000000000000000000000000000000000000000'
+
+// Marks an expectation the reference runner has not filled in yet. Only these
+// are ever overwritten by --results.
+const PENDING_INTENT = 'PENDING'
+
+// 13 bridges display their own recipient (`Non-EVM Recipient`, `GasZip
+// Recipient`, …) alongside `_bridgeData.receiver`. A type-derived zero would
+// make those fixtures assert that the destination screen shows 0x0 — so these
+// components carry an explicit value. Deliberately synthetic: it must not read
+// as a real account anyone could mistake for a live destination.
+const NON_EVM_RECEIVER =
+  '0x7c9e6679a2b1c4f0d38e4b5a6c7d8e9f0a1b2c3d4e5f60718293a4b5c6d7e8f9'
+
+const RECIPIENT_COMPONENTS = new Set([
+  'receiverAddress',
+  'recipient',
+  'nonEVMReceiver',
+  'receiver',
+])
+
+// Tokens referenced by the generated calldata. The runner resolves symbol and
+// decimals from here, so fixtures need no network access.
+const USDC = '0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48'
+const WETH = '0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2'
+const RECEIVER = '0x78b59874BF0Fb404D88D9e688fdC56eE3E085e6b'
+
+const DATA_PROVIDER_TOKENS = {
+  [USDC.toLowerCase()]: { decimals: 6, name: 'USD Coin', symbol: 'USDC' },
+  [WETH.toLowerCase()]: { decimals: 18, name: 'Wrapped Ether', symbol: 'WETH' },
+}
+
+// `_bridgeData` and `_swapData[0]` always reach a screen. Bridge-specific
+// structs are mostly hidden, except the BRIDGE_EXTRA_RECEIVERS recipients —
+// see docs/ClearSigningProposal.md.
+const BRIDGE_DATA: Record<string, unknown> = {
+  transactionId:
+    '0x1b7b4d1c9e0e4a2f8c3d5e6f7a8b9c0d1e2f3a4b5c6d7e8f9a0b1c2d3e4f5a6b',
+  bridge: 'lifi',
+  integrator: 'lifi-api',
+  referrer: ZERO_ADDRESS,
+  sendingAssetId: USDC,
+  receiver: RECEIVER,
+  minAmount: 1_000_000_000n,
+  destinationChainId: 10n,
+  hasSourceSwaps: false,
+  hasDestinationCall: false,
+}
+
+const SWAP_DATA: Record<string, unknown> = {
+  callTo: ZERO_ADDRESS,
+  approveTo: ZERO_ADDRESS,
+  sendingAssetId: WETH,
+  receivingAssetId: USDC,
+  fromAmount: 500_000_000_000_000_000n,
+  callData: '0x',
+  requiresDeposit: true,
+}
+
+// The swapTokens* family displays top-level parameters rather than struct
+// components, so those need values too — `--check` reports any that are missed.
+const TOP_LEVEL_ARGS: Record<string, unknown> = {
+  _transactionId: BRIDGE_DATA.transactionId,
+  _integrator: 'lifi-api',
+  _referrer: ZERO_ADDRESS,
+  _receiver: RECEIVER,
+  _minAmount: 1_000_000_000n,
+  _minAmountOut: 1_000_000_000n,
+}
+
+interface IDescriptorField {
+  path: string
+  label: string
+  visible?: string
+}
+
+interface IDescriptorFormat {
+  fields?: IDescriptorField[]
+}
+
+interface IDescriptorFile {
+  context?: {
+    contract?: { deployments?: Array<{ chainId: number; address: string }> }
+  }
+  display?: { formats?: Record<string, IDescriptorFormat> }
+  formats?: Record<string, IDescriptorFormat>
+}
+
+interface IRenderedCase {
+  description: string
+  status: string
+  rendered?: {
+    intent: string
+    interpolatedIntent?: string
+    owner?: string
+    fields: Array<{ label: string; value: unknown }>
+  }
+  message?: string
+}
+
+interface ITestCase {
+  description: string
+  rawTx: string
+  expected: Record<string, unknown>
+}
+
+/**
+ * Builds a zero value for an ABI type.
+ *
+ * Undisplayed arguments only need to encode, so deriving them from the type
+ * keeps the fixture free of incidental data that could mask a descriptor change
+ * surfacing a field nobody vetted. `--check` guards the other direction: a field
+ * that IS displayed must never be left at its zero value.
+ */
+function defaultForType(param: AbiParameter): unknown {
+  const { type } = param
+
+  const arrayMatch = type.match(/^(.*)\[(\d*)\]$/)
+  if (arrayMatch) {
+    const [, inner, size] = arrayMatch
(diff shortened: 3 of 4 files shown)


tasks/generateClearSigningTests.ts
```diff
@@ -0,0 +1,562 @@
...
+
+  const data = encodeFunctionData({ abi: [abiItem], args })
+
+  // Unsigned: the runners do not verify signatures, so fixtures stay
+  // reproducible without a key. See the registry README, "Reference test cases".
+  return serializeTransaction({
+    chainId,
+    type: 'eip1559',
+    to: diamond as `0x${string}`,
+    value: 0n,
```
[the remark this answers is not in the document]
gvladika: Valid — fixed in 520bee456.

Both native-in formats (`swapTokensSingleV3NativeToERC20`, `swapTokensMultipleV3NativeToERC20`) display `@.value`, and `buildRawTx` hardcoded `value: 0n` while `checkVisibleFields` skipped every `@.` path, so nothing would have caught it.

Nothing published is hollow: both selectors are covered in the committed fixture by pre-existing cases carrying real transactions (`Swap - chain 1 (3)` and `(6)`, values `4691322694977235` and `2098039615199859` wei), so the generator never emitted a case for either. Regenerating against the committed fixture before and after this change is byte-identical. The bug was latent — it would have fired on a from-scratch regeneration or on a new format displaying `@.value`.

The fix:

- `nativeValueFor(format)` returns `NATIVE_VALUE` when a visible field reads `@.value`, else `0n`; `buildRawTx` takes it as a parameter. Generating from scratch now yields a non-zero value on exactly those 2 of 69 cases.
- `checkVisibleFields` no longer skips `@.` paths: `@.value` must be non-zero, and any other envelope path is reported as having no generated value, since `buildRawTx` models only `value`.

Verified by mutation: setting `NATIVE_VALUE = 0n` makes `--check` exit 1 naming both formats; restoring it passes.

tasks/generateClearSigningTests.ts
```diff
@@ -0,0 +1,562 @@
...
+    tests.push({
+      description: `${abiItem.name} - chain ${chainId}`,
+      rawTx,
+      expected: { intent: PENDING_INTENT, owner: 'LI.FI', fields: [] },
+    })
+  }
+
+  return {
+    $schema: '../../../specs/erc7730-tests-v2.schema.json',
+    descriptor: '../calldata-LIFIDiamond.json',
```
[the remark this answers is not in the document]
gvladika: Valid — fixed in 520bee456, in the docs rather than the generator.

Confirmed against the runner: `runner.ts` resolves the descriptor as `resolve(dirname(testsPath), input.descriptor)`, and `descriptor-index.ts` then walks up from there to find the registry root. So a fixture at `/tmp/tests.json` resolves `../calldata-LIFIDiamond.json` to `/calldata-LIFIDiamond.json` and fails. The documented loop was wrong; the actual render pass during development was run against a fixture staged in a registry working copy.

Agreed on not deriving `descriptor`/`$schema` from `args.out` — the sync workflow writes to `/tmp` and copies to the registry path, so they must stay relative to the final layout. The loop now writes the fixture to its real path in a registry clone and points the runner there. `--existing` and `--out` can be the same file: the fixture is read in full before it is written, and step 3 re-reads the `PENDING` blocks step 1 wrote, then fills them from `--results`.

.github/workflows/syncLedgerClearSigning.yml
```diff
@@ -165,6 +166,24 @@ jobs:
...
 
+          # The registry fails a descriptor whose display.formats entries are not
+          # each exercised by a test case, so the fixture has to travel with the
+          # descriptor. Existing cases are kept verbatim: they carry real
+          # transactions and expectations a reviewer has already signed off.
+          echo -e "\033[32mGenerating ${LEDGER_TESTS_PATH} (one case per display.formats entry)...\033[0m"
+          bunx tsx tasks/generateClearSigningTests.ts \
+            --descriptor "ledger-registry/${LEDGER_FILE_PATH}" \
+            --existing "ledger-registry/${LEDGER_TESTS_PATH}" \
+            --out /tmp/lifi-tests.json
```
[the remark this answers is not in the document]
gvladika: The analysis is right, but that failure is the intended design, so I've documented it rather than changed the behaviour (520bee456).

You're correct that with neither `--results` nor `--allowPending` the gate throws before the fixture is written. That is the point: a new format has no reviewed expectation, and the only ways past the gate are to invent one or to publish `PENDING` — both produce a fixture that fails the registry's own runners. The step runs under `set -euo pipefail`, so the sync aborts before the commit and nothing partial reaches the fork.

On the two remedies you propose — the committed reviewed fixture already is the input: `--existing "ledger-registry/${LEDGER_TESTS_PATH}"` reads the fixture on the fork's sync branch, and every selector it covers is preserved verbatim and never regenerated. The reference-renderer stage with explicit review also already exists, as the 3-step loop in `docs/ClearSigningProposal.md`. It is deliberately a human step rather than a workflow stage: its whole value is a person reading what the renderer produced, which a bot cannot do. And `--allowPending` is never passed here, as you say it should not be.

What was genuinely missing is that the failure is actionable. It surfaces in a bot-run job whose reader has no context, and the generator's error text is written for a developer at a terminal ("re-run with `--allowPending`…") — which, read in a workflow log, suggests re-running the sync, the one thing that cannot fix it. The workflow header now carries the failure mode explicitly: what it means, that re-running will not help, and that the remedy is to run the render loop against a working copy of the fork and commit the reviewed fixture to the sync branch.

Idempotence and validation behaviour are unchanged; regenerating against the committed fixture is byte-identical before and after this commit.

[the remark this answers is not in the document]
gvladika: Correcting my earlier reply on this thread: I said "`--existing "ledger-registry/${LEDGER_TESTS_PATH}"` reads the fixture on the fork's sync branch." That is wrong, and it made the rest of my answer wrong with it.

Line 153 does `git checkout -B "${SYNC_BRANCH}" "upstream/${UPSTREAM_DEFAULT_BRANCH}"`, so by the time the generator runs, the working tree is **upstream's**, not the fork's. `--existing` has always read upstream's copy. Measured against the live repos:

```
upstream/master fixture          →  7 cases
registry descriptor              → 69 formats
sync branch fixture (PR #2955)   → 69 cases
```

So ~62 formats are uncovered on every run, the PENDING gate throws, and the step fails on every trigger — exactly your finding. Reproduced with upstream's real 7-case file as `--existing`:

```
62 format(s) have no expectation yet: startBridgeTokensViaAcross - chain 1, …
```

And the remedy I documented in the header ("commit the reviewed fixture to the fork's sync branch") could not work either, because `git push --force-with-lease` at line 216 overwrites that branch on the next run. Your second comment on the header was right about that too.

Fixed in 2f5a7e8c5. The branch's own copy is read off `origin/${SYNC_BRANCH}` — a remote-tracking ref, untouched by the reset — before the `cd ..`, and replayed through a new `--overlay`:

```bash
if git cat-file -e "origin/${SYNC_BRANCH}:${LEDGER_TESTS_PATH}" 2>/dev/null; then
  git show "origin/${SYNC_BRANCH}:${LEDGER_TESTS_PATH}" > "${IN_FLIGHT_TESTS}"
```

The overlay wins where both fixtures cover a selector. That branch is this PR's own head, so it carries any edit a reviewer makes here, and it was rendered against the descriptor the run is pushing — upstream's copy can predate a label change. That is not hypothetical: diffing the two showed upstream still asserting `Minimum Amount to receive` / `Receiver` where the new descriptor emits `Minimum to Receive` / `Recipient`. Preferring upstream would have reinstated the stale labels.

Verified by simulating the real sync against both live repos — upstream's 7 cases as `--existing`, the sync branch's 69 as `--overlay` — which reproduces the fixture on registry PR #2955 byte for byte. The no-overlay path (everything already merged) is byte-identical too, and an absent branch degrades to `{"tests":[]}`.

What this does **not** fix is your separate point that an already-covered format whose descriptor changed keeps its old expectation. Neither copy can be re-rendered without the reference runner, so that still needs a manual pass through the loop; the docs now say so explicitly rather than implying the sync self-corrects.

tasks/generateClearSigningTests.ts
```diff
@@ -0,0 +1,664 @@
...
+ * head, so it carries any edit a reviewer made there, and it was rendered
+ * against the descriptor this run is pushing — upstream's copy can predate a
+ * label change and assert labels the new descriptor no longer emits. Order
+ * follows the primary fixture, with the overlay's extra cases appended.
+ */
+function mergeCases(existing: ITestCase[], overlay: ITestCase[]): ITestCase[] {
+  const bySelector = new Map<string, ITestCase>()
+  for (const test of overlay) {
+    const selector = selectorOf(test)
+    if (selector && !bySelector.has(selector)) bySelector.set(selector, test)
```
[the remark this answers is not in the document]
gvladika: Valid on both counts — fixed in df1f26da9.

`bySelector.set` guarded by `!has` meant a second reviewed overlay case for a selector was dropped before it could ever be emitted, and `taken` left the primary's trailing duplicates in place after the first was replaced, so a stale upstream case outlived the overlay meant to supersede it. Losing a reviewed overlay case is precisely what this function exists to prevent, so the direction of that first bug was the wrong one to get wrong.

Nothing hits it in the current data — upstream has 7 distinct selectors and the sync branch 69, no duplicates on either side — but nothing stops a reviewer covering one format with several scenarios, which is the case you describe.

Now grouped, as you suggested: `groupBySelector()` collects every case per selector, and `mergeCases()` moves whole groups — the overlay's group lands at the position where the primary's group started, and the primary's remaining cases for that selector are what it supersedes. Overlay-only groups are appended intact.

One addition beyond the suggestion: a case with no calldata (a plain transfer) exercises no selector, so the overlay has no way to address it. Rather than let those fall out of the map silently, they are carried through, deduped by description so a run does not re-append what the previous one already carried forward.

Verified on a synthetic fixture with two primary and two overlay cases sharing selector `0x5fd9ae2e`:

```
both overlay cases present:        Swap - chain 1 (1)
                                   REVIEWED extra scenario for same selector
stale primary duplicate retained:  0
total:                             70  (69 formats + the extra reviewed scenario)
```

Before the fix the second overlay case was absent and the stale primary duplicate was still there. Both real paths — upstream's 7 cases as `--existing` with the sync branch's 69 as `--overlay`, and the no-overlay path — still reproduce the fixture on registry PR #2955 byte for byte.

On the regression test: this file has no unit tests yet, and adding one needs a refactor first — the module calls `runMain(main)` at import, so importing it to test `mergeCases` would execute the CLI. That is the open checklist item on this PR rather than something to bolt on here; `groupBySelector`, `mergeCases`, `selectorOf`, `nativeValueFor`, `resolvePath` and `isZeroish` are all pure and are what such a suite should cover.

0xDEnYO (dismissed):

tasks/generateClearSigningTests.ts
```diff
@@ -0,0 +1,669 @@
...
+  const covered = new Set<string>()
+  for (const test of readExistingTests(testsPath)) {
+    const selector = selectorOf(test)
+    if (selector) covered.add(selector)
+  }
+
+  const problems: string[] = []
+  for (const formatKey of Object.keys(formats)) {
+    const selector = toFunctionSelector(`function ${formatKey}`).toLowerCase()
+    if (!covered.has(selector))
```
0xDEnYO: `covered` here means "some case has this selector", so once a format has a case its expectation is never compared against the descriptor again. A label rename then leaves the fixture asserting a label the descriptor no longer emits — I renamed one locally and both `--check` and `--check --tests` stay green.

That's the gap the PR body documents as accepted, but it has already bitten once: 6 of the 7 upstream cases in this diff had to be re-rendered for exactly this reason — `Minimum Amount to receive` → `Minimum to Receive`, `Receiver` → `Recipient`, plus the added `interpolatedIntent`. Next time nothing re-renders them, and the failure lands on the unattended sync after merge.

Both sides are in-repo at PR time, so it's checkable right here. For each covered selector, compare the descriptor's visible field labels to that case's `expected.fields[*].label`. I ran it against the committed fixture: **68/68 match exactly, in order** — no false positives today. Needs `covered` to become a `Map<selector, test>` rather than a `Set`:

```ts
const want = (format.fields ?? [])
  .filter((f) => (f.visible ?? 'always') !== 'never')
  .map((f) => f.label)
const got = (test.expected.fields ?? []).map((f) => f.label)
if (JSON.stringify(want) !== JSON.stringify(got))
  problems.push(
    `${formatKey}: fixture asserts [${got}], descriptor displays [${want}]`
  )
```

Won't catch value drift, but labels are the common case and the one that's already happened.

tasks/generateClearSigningTests.ts
```diff
@@ -0,0 +1,669 @@
...
+
+  // `_swapData` is a tuple[] on the bridge facets and a plain tuple on the
+  // swapTokens* family.
+  if (param.name === '_swapData') {
+    const element = { ...param, type: 'tuple' } as AbiParameter
+    const value = fromTemplate(element, SWAP_DATA)
+    return param.type.endsWith(']') ? [value] : value
+  }
+
+  if (param.name && param.name in TOP_LEVEL_ARGS)
```
0xDEnYO: Nit: `in` walks the prototype, so a parameter named `toString`/`valueOf`/`constructor` would resolve to a prototype member rather than falling through to `defaultForType`. Unreachable with today's ABI, but same pattern at line 202 (`key in template`) and 327 (`address in tokens`). `Object.hasOwn` is the drop-in.

.github/workflows/verifyClearSigning.yml
```diff
@@ -63,6 +68,17 @@ jobs:
         # shape.
         run: bunx tsx tasks/buildClearSigningProposal.ts
 
+      - name: Verify every format has a test case that asserts something
```
0xDEnYO: Nit: this runs before the up-to-date diff below, so a dev who edits a facet and forgets to regenerate gets "N format(s) have no reviewed test case → run the render loop" before the much cheaper "run `buildClearSigningProposal.ts`". Both need doing, but the expensive remedy surfaces first. Swapping the two steps reads better.

0xDEnYO (commented):
Reviewed and verified locally rather than by reading: built the merged descriptor via the sync's own path, ran it through `sourcifyeth/clear-signing-test-runner` (**69/69 pass**) and the registry's own `check-selector-coverage.js` (**passes** — the gate that broke #2955 with 62/69 uncovered), mutated both new guards to confirm they fire and exit 1, and checked that `jq -c | prettier --print-width 120` reproduces the registry's existing fixture byte-for-byte. All 7 upstream cases and the `dataProvider.tokens` block survive intact.

The approach is sound and the runbook in `docs/ClearSigningProposal.md` is runnable as written — I ran it.

One finding worth a round, inline on `checkCoverage`: the label-drift gap is cheaper to close than the docs conclude, and it has already happened once inside this diff. Three nits below, none blocking.

(The red `protect-critical-code` is the InfoSec approval gate on `.github/workflows/**`, not a defect.)

One more nit, on a line too far outside the diff to anchor: `docs/ClearSigningProposal.md:227`, the "Pipeline integration" diagram still lists three `verifyClearSigning` gates and says "PR merges only if all three pass". This PR adds a fourth.

tasks/generateClearSigningTests.ts
gvladika: Fixed in d2c42f91b. `covered` is a `Map<selector, ITestCase>` now, and for every covered format `checkCoverage` compares the descriptor's visible labels, in order, against that case's `expected.fields[*].label`.

Confirmed your reading before implementing: the gap is real, and I reproduced your 68/68 independently against the committed fixture — no false positives. Also mutated it three ways: renaming a label and swapping two fields each fail and exit 1; adding a `visible: never` field does not (the filter holds). Regeneration is still byte-identical at `printWidth: 120`.

Two notes on the implementation:

- `ITestCase.expected` was `Record<string, unknown>`, so the snippet needed a type to read `fields` off. Added `IExpectation` with the known keys plus an index signature, rather than casting at the call site.
- PENDING cases are skipped — they have no rendered labels yet, so there is nothing to drift.

Value drift is still unguarded, as you said. `docs/ClearSigningProposal.md` now says so explicitly instead of describing the whole thing as accepted.

tasks/generateClearSigningTests.ts
gvladika: Fixed in e5b011bb1, all three sites.

`Object.hasOwn` is ES2022 and `tsconfig.json` sets `target`/`lib` to es2020, so `tsc` rejected it. Rather than bump the repo-wide lib in this PR, the three sites go through a local `hasOwn()` wrapping `Object.prototype.hasOwnProperty.call` — same prototype-proofing, es2020-safe.

.github/workflows/verifyClearSigning.yml
gvladika: Fixed in 0ac8a9a5c — the up-to-date diff now runs first, with a comment on why. Also worth noting a stale proposal is usually what causes the coverage failure, so the cheap remedy tends to resolve both.

The gate's own failure message was wrong once labels joined it ("N format(s) have no reviewed test case"), so that now covers both cases.

gvladika:
All four points addressed — three commits, each scoped to one of them:

| | |
|---|---|
| `d2c42f91b` | label-drift check in `checkCoverage` (+ docs) |
| `e5b011bb1` | `Object.hasOwn` nit, all three sites |
| `0ac8a9a5c` | CI step ordering |

**The unanchored nit** (`docs/ClearSigningProposal.md:227`): fixed in `d2c42f91b`. The diagram lists the fourth gate and says "all four pass". The paragraph above it also described label drift as accepted with "nothing flags it automatically" — no longer true, so it now says what *is* caught (labels, in order, both sides in-repo) and what is not (values, which need the reference runner).

**One thing neither of us flagged.** The fixture has 69 cases but the descriptor has 68 formats. The extra is `0x4630a0d8` = `swapTokensGeneric(bytes32,string,string,address,uint256,(address,address,address,address,uint256,bytes,bool)[])`, carried as "Swap - chain 1 (7)" from upstream. Both `checkCoverage` and the new label comparison iterate *formats*, so nothing looks at it in either direction — it will ride along indefinitely without being asserted against anything.

Not a defect and not something I changed here: dropping it would edit a reviewed upstream case, which felt out of scope for a review round. Flagging it as a deliberate call — happy to remove it in a follow-up if you would rather the fixture track the descriptor exactly.

**Verification** (beyond the reference runner run you already did): 68/68 labels match on the committed fixture, so the new gate is green with no false positives; renaming a label and swapping two field positions each fail it and exit 1; adding a `visible: never` field does not; `jq -c | prettier --print-width 120` regeneration is still byte-identical. `tsc`, `eslint`, `prettier` and `actionlint` clean on all three files.

0xDEnYO (approved):

melianessa (approved):

PR #2378, opened 2026-09-12
feat(safe): rehearse the verify path read-only, proving it cannot write (EXSC-995)

0xDEnYO:

Fixes [EXSC-995](https://linear.app/lifi-linear/issue/EXSC-995)

Also closes [EXSC-931](https://linear.app/lifi-linear/issue/EXSC-931) — the junk proposal at nonce 31 on the production Tron Safe is gone, cleared with the teardown added here. Spun out: [EXSC-1009](https://linear.app/lifi-linear/issue/EXSC-1009).

## What this is, and when you would run it

One command that runs the confirm-time gate chain against real proposals and shows what a signer would see — without signing, proposing or executing.

```bash
bun verify-pending-proposals --networks arbitrum,base,optimism
```

(`--status` defaults to `pending`; the script opens the Safe tunnel first, as the other Safe commands do. With nothing in flight it says so in one line and exits 0.)

**Run it on a rollout wave, after the proposals are in the store and before anyone signs.** It answers: how many rows does the evidence settle on its own, how many block outright, and how many need a human to say "yes, this is the change we meant". If a 40-chain wave is about to ask someone for 40 acknowledgements, that is better known at rollout time than at signing time.

**Be honest about what is useful today.** One gate of nine is merged, so a run exercises `target-state` and reports the other eight absent. That is on-spec — EXSC-995's build order was explicitly "build the harness before the pipeline, report absent gates as absent" — but it means the runner is scaffolding until [EXSC-994](https://linear.app/lifi-linear/issue/EXSC-994) wires the rest. There are also no pending proposals anywhere right now, so its first real use is the next wave.

The two pieces that earn their keep immediately:

- **`rehearsal-teardown.ts`** — the only thing in the repo that hunts a proposal by nonce under the store's collation. It closed EXSC-931.
- **the grading-anchor check** — a fresh worktree silently turns `target-state` into a refusal on every proposal. Details below; it cost most of a session to find.

## The preflight asserts; it does not intend

The risk a read-only rehearsal carries is not signing — nobody does that by accident. It is an incidental write: the acknowledgement ledger, signed-set persistence, the reconcile back-fill. So the collection is sealed behind an **allow-list of reads** (a deny-list is outgrown by the next driver release, and the method it fails to name is the one that writes), the seal is proven by reaching for the refused members against a canary, and the live handle is tied to that proof by identity.

The proof probes a canary rather than the live handle on purpose: calling `insertOne` on a real collection to find out whether it is sealed would insert a row on exactly the run where the guard was missing. `proveSealRefusesWrites` takes the sealer as a parameter so the proof can itself be falsified — a proof with no failing branch measures nothing.

Four escapes the review gate found, each verified against a real driver handle before and after the fix:

- the seal refused only callables, so `.client` came back as the live `MongoClient`, one call from an unsealed collection;
- an allow-listed read re-exported the write surface one hop later — a `FindCursor` carries `.client`, a `ListIndexesCursor` carries `.parent`, the raw collection itself;
- `AbstractCursor._initialize` resolves to an object carrying a live `Server`, whose `command()` runs an arbitrary command — so "every function on a cursor is a read" was false, and cursor methods are allow-listed by name too;
- `util.inspect` reads a proxy's target directly without consulting a single trap, so `console.log(sealed)` printed the whole `MongoClient`, credentials object included. The seal is now proxied over an empty object, which is the only thing that stops it.

`aggregate` also left the allow-list, since an `$out`/`$merge` pipeline writes. The `getOwnPropertyDescriptor` trap redacts rather than throwing: refusing there took down `Object.keys` and `for…in` on a real `Collection`, whose only own properties are `s` and `client`. Spread and `Object.entries` still refuse, and should — they read the values.

This surfaced a write nobody asks for: **`getSafeMongoCollection()` calls `createIndex` twice on connect**, so opening the store the ordinary way mutates it. The rehearsal opens it by another route.

**Not EXSC-980.** That is the full ceremony on a staging network, and its preflight refuses unless the Safe is outside the production set — it would refuse exactly this run. The two designs are deliberately not merged.

## The teardown is the safety story for dummy proposals

A pending row occupies a Safe nonce and Safes execute in order, so a dummy at N blocks every real proposal after it — but only if it is left behind. `findProposalsAtNonce` matches under the store's collation, which is not optional: one Safe has two spellings in this collection (`propose-to-safe-tron.ts` stores lowercase hex, `initializeSafeClient` stores checksummed). Measured against production, the checksummed spelling of the Tron Safe matches **0 rows uncollated and 34 collated**.

`tearDownProposalsAtNonce` lowercases the network before deleting. The hunt tolerates a miscased network because its collation applies to the whole query, but `deletePendingProposals` matches `network` with an uncollated `$eq` — so `--network Tron --delete` would list the rows, delete none, and report success while the nonce stayed blocked.

## Three things the run refuses to call a result

Each of these was a green that turned out to be an artifact, found by falsifying the harness against real data rather than by a test.

**A corpus whose anchor is in its own future.** The gates grade against the target state `origin/main` declares *now*. Right for a proposal about to be signed; meaningless for one that executed months ago — it installed what main declared then, so today it reads as a downgrade. A run over executed rows was printing 47 refusals and a signer workload as if they were findings. It now reports only what such a corpus establishes (the chain survives real shapes; two passes agree; which gates are wired) and names `--status pending` as the corpus whose verdicts a signer can act on.

**A missing grading anchor.** The deployment record is read from `.cache/deployments_production.json`, which is gitignored, so a fresh worktree does not have it. `resolveDeployedContractByAddress` then returns `unrecorded` for every address and every cut element grades `contract-unidentified`, with nothing saying why. With the file absent: 1066 of 3191 rows refused, all one class. With it present the same corpus produces `downgrade`, `matches-main`, `ahead-of-main` and `not-previously-targeted`. The anchors are now checked before anything is graded and the run refuses, naming the remedy; `--grade-without-anchors` grades anyway and says every refusal is suspect. The gate checks the file parses as a record array, not merely that it exists — the loader memoises `null` for both, so a truncated cache produced the same wall of false refusals. *The signing CLI largely avoids this*: it refreshes the cache from MongoDB at startup, though that refresh is itself conditional on `MONGODB_URI` and swallows failure at debug level. The rehearsal never reaches that warm-up, because it opens the store by another route to avoid the index write.

**A budget over a single refusal class.** When only one class is reachable and this release has already ruled it correct, every refusal lands in that bucket and `unexplained` is 0 for any corpus on any day. That constant is what EXSC-977's promotion criterion reads to decide a gate may start blocking, so it is withheld rather than printed with a caveat.

## Determinism, and the corruption probe

Two passes over the same input must agree; a difference means something in the chain reads a moving value. Each pass builds its **own** pinned-state reader — sharing one memoising reader across both (which an earlier commit did, for speed) meant pass two replayed pass one's `origin/main` read, so the check could not fire on the one genuinely moving input it targets.

`gradeCorruptionProbe` is graded **per row**: every row clean before corruption must refuse after it. Asking whether *any* row refused reported success before the damage was applied, because most rows already refuse for their own reasons.

Both halves of it took two goes, and real data caught both.

The **population** is narrowed to rows the mutation can reach: one that already stopped on the bytes proves nothing by stopping again, and one the gate graded as having no cut to compare can never refuse at all. `needs-ack` rows are deliberately in — the payload was read and understood and only the intent is open, so if corruption makes it unreadable the chain noticed. Excluding them emptied the population fleet-wide and the probe reported "not exercised" everywhere while looking fixed.

The **mutation** replaces the cut's facet address wherever it appears. Overwriting the first argument word — the obvious thing — worked only for a direct `diamondCut`, where that word is the array offset. For the timelock-wrapped `schedule` that most production rollouts use it is `target`, leaving the inner cut untouched, so the gate graded the corrupted payload byte-identically. The survivors split exactly by outer selector: `diamondCut` 18 refused / 0 survived, `schedule` 0 / 12. Substituting the address reaches the inner cut at any nesting depth without this module knowing the wrapper's shape.

## Also

`ADDRESS_COLLATION` is now exported from `safe-utils.ts` rather than copied — that file's own comment says the index and the read that feeds it must not disagree, and a third copy is exactly that risk.

Every failing verdict exits non-zero. Network keys are lowercased before they reach Mongo, where a miscased key matched nothing and read as a clean run.

### DIFF
--- a/package.json
+++ b/package.json
@@ -71,6 +71,7 @@
     "codegen": "plop",
     "compile": "forge build",
     "confirm-safe-tx": "bash script/deploy/safe/with-safe-tunnel.sh && bun build:typechain-and-abi && bunx tsx ./script/deploy/safe/confirm-safe-tx.ts",
+    "verify-pending-proposals": "bash script/deploy/safe/with-safe-tunnel.sh && bunx tsx ./script/deploy/safe/verify-rehearsal.ts",
     "safe:tunnel": "bash script/deploy/safe/with-safe-tunnel.sh",
     "deploy-safe": "bunx tsx ./script/deploy/safe/deploy-safe.ts",
     "coverage": "rm -rf coverage && rm -f lcov-filtered.info && rm -f lcov.info && forge coverage --report lcov --evm-version 'cancun' --ir-minimum && bun script/utils/filter_lcov.ts lcov.info lcov-filtered.info 'test/' 'script/' && genhtml lcov-filtered.info --ignore-errors inconsistent,corrupt --branch-coverage --output-dir coverage && open coverage/index.html",
--- a/script/deploy/safe/rehearsal-report.test.ts
+++ b/script/deploy/safe/rehearsal-report.test.ts
@@ -1,0 +1,74 @@
+import {
+  describe,
+  expect,
+  it,
+  // eslint-disable-next-line import/no-unresolved
+} from 'bun:test'
+
+import {
+  REHEARSAL_GATE_ROSTER,
+  buildGateReport,
+  renderGateReport,
+} from './rehearsal-report'
+
+describe('buildGateReport', () => {
+  it('names a rostered gate that did not register as absent', () => {
+    const report = buildGateReport({
+      registered: ['target-state'],
+      rowCounts: { 'target-state': 3 },
+    })
+
+    const executability = report.find(
+      (gate) => gate.checkId === 'executability'
+    )
+    expect(executability?.presence).toBe('absent')
+  })
+
+  it('never drops a rostered gate, however few registered', () => {
+    const report = buildGateReport({ registered: [], rowCounts: {} })
+
+    expect(report.map((gate) => gate.checkId).sort()).toEqual(
+      [...REHEARSAL_GATE_ROSTER.map((gate) => gate.checkId)].sort()
+    )
+    expect(report.every((gate) => gate.presence === 'absent')).toBe(true)
+  })
+
+  it('separates a gate that registered but reported nothing from one that ran', () => {
+    const report = buildGateReport({
+      registered: ['target-state', 'executability'],
+      rowCounts: { 'target-state': 2, executability: 0 },
+    })
+
+    expect(
+      report.find((gate) => gate.checkId === 'target-state')?.presence
+    ).toBe('present')
+    expect(
+      report.find((gate) => gate.checkId === 'executability')?.presence
+    ).toBe('registered-but-silent')
+  })
+
+  it('reports a registered gate the roster does not name', () => {
+    const report = buildGateReport({
+      registered: ['a-gate-nobody-rostered'],
+      rowCounts: { 'a-gate-nobody-rostered': 1 },
+    })
+
+    const surprise = report.find(
+      (gate) => gate.checkId === 'a-gate-nobody-rostered'
+    )
+    expect(surprise?.presence).toBe('present')
+    expect(surprise?.rostered).toBe(false)
+  })
+})
+
+describe('renderGateReport', () => {
+  it('prints every gate including the absent ones', () => {
+    const rendered = renderGateReport(
+      buildGateReport({ registered: ['target-state'], rowCounts: {} })
+    )
+
+    for (const gate of REHEARSAL_GATE_ROSTER)
+      expect(rendered).toContain(gate.checkId)
+    expect(rendered).toMatch(/absent/)
+  })
+})
--- a/script/deploy/safe/rehearsal-report.test.ts
+++ b/script/deploy/safe/rehearsal-report.test.ts
@@ -1,3 +1,7 @@
+import { mkdirSync, mkdtempSync, writeFileSync } from 'node:fs'
+import { tmpdir } from 'node:os'
+import { join } from 'node:path'
+
 import {
   describe,
   expect,
@@ -11,6 +15,8 @@ import {
   buildGateReport,
   checkGradingAnchors,
   renderGateReport,
+  renderGradingAnchors,
+  renderSignerWorkload,
   summariseSignerWorkload,
   verdictsAreActionable,
 } from './rehearsal-report'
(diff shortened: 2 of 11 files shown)


0xDEnYO:
Closing as superseded: this work is already on `main`.

Every file this PR adds landed in `4b869080b` (#2393, the Signing 2.0 integration merge, 2026-09-18): `verify-rehearsal.ts`, `rehearsal-report.ts`, `rehearsal-run.ts`, `rehearsal-teardown.ts`, `rehearsal-write-guard.ts`.

The diff above is measured against a merge base from 2026-09-11, which is why it still reads as +3,340. Against current `main` the branch is behind, not ahead:

| File | `main` vs this branch |
|---|---|
| `verify-rehearsal.ts` | identical |
| `rehearsal-teardown.ts` | identical |
| `rehearsal-write-guard.ts` | identical |
| `rehearsal-report.ts` | `main` has the newer copy (adds the `not-applicable` handling) |
| `rehearsal-run.ts` | `main` has the newer copy (adds the `not-applicable` status) |

Merging it would regress the last two files, so there is nothing left to take from it.

Most of the review findings raised here were applied before the integration merge and are correct on `main`: the second-pass-only keys in `comparePasses`, the `not-measured` verdict masking a real disagreement, the teardown lookup restricted to nonce-holding statuses with deletion limited to `pending`, the non-zero exit on a failed verification outcome, and the grading-anchor test that now writes a real cache file instead of asserting against an absent one.

Three were never applied and are still live on `main`. They are carried into follow-ups rather than lost with this PR:

- `rehearsal-teardown.ts` — `run` discards the `IDeleteResult[]` from `tearDownProposalsAtNonce`, so a `skipped-signed` row exits 0 and reports a clean teardown while the nonce stays blocked.
- `rehearsal-report.ts` — `renderSignerWorkload` omits `task.proposal`, so two proposals needing judgement on the same check and network render as identical lines.
- `verify-rehearsal.ts` — `openReadOnlyProposalStore` passes `SC_MONGODB_URI` to `MongoClient` with no transport check, so a cleartext connection to a remote host is accepted silently. Worth noting for whoever picks this up: the fix suggested on the thread keys on the host being loopback, which does not hold here, because a localhost tunnel to the production store presents as loopback. That guard wants to key on the credentials the URI carries.

PR #2379, opened 2026-09-12
chore(deps): bump tsx to 4.23.13 to silence Node 26 DEP0205

0xDEnYO:

None — dependency bump, covered by the `trivial` carve-out in the ticket-linkage metric (`.agents/context.md`).

Running any `bunx tsx` script on Node 26 prints:

```text
(node:19518) [DEP0205] DeprecationWarning: `module.register()` is deprecated. Use `module.registerHooks()` instead.
(Use `node --trace-deprecation ...` to show where the warning was created)
```

The call is not ours — `git grep module.register` across the repo returns nothing. It comes from `tsx` 4.21.0's ESM loader; Node 26 deprecated `module.register()` as DEP0205. `tsx` 4.23.13 no longer uses it, so the warning is gone.

**Blast radius is local only.** CI never saw this warning: the workflows run under bun, and the two that use Node pin `node-version: 20` (`jsonChecker.yml`, `updateLouperViem.yml`), where DEP0205 does not exist. This is purely noise reduction for local script runs and agent sessions.

**Transitive change:** `tsx` widened its `esbuild` range from `~0.27.0` to `~0.28.0`, so the lockfile also moves `esbuild` 0.27.7 → 0.28.2 plus its 26 platform-specific optional packages. Nothing else in `bun.lock` changes — the diff is exactly `tsx`, `esbuild`, and those 26 `@esbuild/*` entries.

**Verification**

- Reproduced the warning on `tsx` 4.21.0 and confirmed its absence on 4.23.13, same script, same Node 26.7.0.
- Loaded `script/utils/deploymentHelpers.ts` (pulls in viem) under both versions — identical behaviour, no transpile drift.
- Ran a real repo task end-to-end (`tasks/buildClearSigningProposal.ts`) under 4.23.13; it executed and hit its own `forge build` precondition guard as expected.
- Lockfile regenerated with `bun install --lockfile-only` so the shared `node_modules` symlink was left untouched.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -48,7 +48,7 @@
         "solhint": "^5.0.5",
         "solhint-plugin-prettier": "^0.0.5",
         "tsc-files": "^1.1.4",
-        "tsx": "4.21.0",
+        "tsx": "4.23.13",
         "typechain": "^8.0.0",
         "typescript": "^5.4.5",
       },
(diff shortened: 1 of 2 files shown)


0xDEnYO:
@coderabbitai review

gvladika (approved):

PR #2380, opened 2026-09-12
feat(safe): refuse a network the run cannot check, before a ledger exists

0xDEnYO:
## What

A precondition every check shares is not a check. One unset `ETH_NODE_URI_<NETWORK>` produced a ledger of ten unverified rows — each telling the signer to re-run a check that cannot run — above a single stack trace naming the variable. Ten pieces of unactionable advice above the one actionable line (F6/F8 in the rehearsal findings).

This adds a preflight that resolves every candidate network's endpoint **before** the ownership reads and **before** the ledger is created:

- configured (`ETH_NODE_URI_<NETWORK>`, or `--rpc-url`)
- reachable (one `eth_chainId`)
- answering the chain id `config/networks.json` declares — the dangerous case, since every later read would answer truthfully about a different chain

A refused network is named once with its cause and its remedy, contributes no rows, and is left out of the ledger's denominator. When no network can start, the run exits `78` (`EX_CONFIG`) without building a ledger at all — a ledger that verified nothing is not a weaker verdict, it is no verdict.

It is a **preflight, not a gate**. The gate letters describe the proposal and red in the signer view means *do not sign*; an unset variable is neither, so it gets its own vocabulary (`CANNOT START — …`) and its own exit code.

## Where to look hardest

- **The three orderings** are the whole design and none is observable from the module, so they are pinned as source-order assertions in `signer-preflight-placement.test.ts`: before the ledger; before `getNetworksWithActionableTransactions`, which swallows a failed read as "not actionable" (a true statement with a false explanation); and the ledger fed only what the preflight cleared.
- **A partial run must not read as a clean one.** Refused networks are absent from the ledger's denominator, so its verdict says nothing about them — the run therefore names them again after the ledger and sets the exit code.
- **No endpoint can reach the output.** The deps return booleans and numbers rather than URLs, and the node's error text is redacted on the way in: viem embeds the URL it called in every message, and a provider URL carries an API key.

## Verification

- `bun test script/deploy/safe/signer-preflight.test.ts script/deploy/safe/signer-preflight-placement.test.ts` → 18 pass.
- Six mutations, each killed by the intended assertion: ledger fed every candidate again; the early return building a ledger anyway; the end-of-run coverage line dropped; stopping at the first failing network; trusting the endpoint's own chain id; passing the node's error through unredacted.
- `bun test script/deploy/safe/` → 2377 pass / 8 fail; the 8 are the pre-existing `storage-authority` registration failures already present at the base commit.

## Not in this PR

The preflight checks endpoints only. The deployment record and the pinned target-state ref are preconditions of *particular* checks rather than of every check, so by the dividing line above they belong as those checks' own unverified reasons — but reporting them here too would save a signer a round trip, and the collection shape takes another probe as one entry. Design and the full case table: `28-executability-case-table.md` §7 in the signing-process project.

Wiring is pinned by source order, not exercised end to end: spawning `confirm-safe-tx` from a test would reconcile the store and go on to sign.

### DIFF
--- a/script/deploy/safe/confirm-safe-tx.ts
+++ b/script/deploy/safe/confirm-safe-tx.ts
@@ -183,9 +183,9 @@ import {
   toSignedCodehashEntries,
 } from './signed-set-record'
 import {
-  preflight,
+  networkPreflight,
   PREFLIGHT_EXIT_CODE,
-  renderPreflight,
+  renderNetworkPreflight,
 } from './signer-preflight'
 import {
   checkSummary,
@@ -1881,14 +1881,13 @@ const main = defineCommand({
         )
       }
 
-      // Before the ownership reads, and before a ledger exists. A network
-      // without a working endpoint cannot be graded, so it is named once here
-      // with its cause rather than appearing later as a column of unverified
-      // checks that describe one unset variable. Placed ahead of
-      // `getNetworksWithActionableTransactions` because that one swallows a
-      // failed read as "not actionable", which is a true statement with a false
-      // explanation.
-      const preflightVerdict = await preflight(candidateNetworks, {
+      // Before the ownership reads and before a ledger exists, so a network
+      // that cannot be graded is named once with its cause instead of becoming
+      // a column of unverified checks. `getNetworksWithActionableTransactions`
+      // still probes it — it derives its own network list — but reports a failed
+      // read as "not actionable", a true statement with a false explanation, so
+      // the refusal has to be printed before it speaks.
+      const preflightVerdict = await networkPreflight(candidateNetworks, {
         endpointConfigured: (network) =>
           Boolean(args.rpcUrl?.trim()) ||
           Boolean(process.env[getRPCEnvVarName(network)]?.trim()),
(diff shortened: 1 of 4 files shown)


0xDEnYO:
Folded into #2384 ("various code fixes"), which is based on `signing2-base-0913` = main + #2353 + #2370 + #2372.

Both commits are in: 357ac2b5d and 457bab390 ("make the preflight's own output safe and its claims true", pushed 09:03 today — picked up after it landed). Diffing `signer-preflight.ts` between this branch's tip and #2384 leaves exactly one intended difference:

- the preflight owns `PREFLIGHT_WIDTH = 76` instead of importing `VIEW_WIDTH` from `./signer-view`. The signer view lives in the sibling UI/UX PR #2383, so the two consolidated PRs stay independent of each other; the duplication collapses on the integration branch, where both halves exist. The width is pinned by value in a test, so a change to it cannot ride along unseen.

Everything else — `redactErrorReason` + `sanitizeProvenanceText`, the `networkPreflight`/`renderNetworkPreflight` rename, the test hardening — is carried verbatim.

Closing in favour of #2384. The branch `claude/exsc-signer-preflight` is left in place; if you have work after 457bab390, push it and say so and I will pick it up into #2384.

PR #2382, opened 2026-09-12
fix(safe): distinguish the two integrity sections, and let gate G ask instead of refusing

0xDEnYO:

EXSC-686 — F7 and F9 from the 2026-09-12 signing rehearsal.

**Base is `signing2-integration-0912`, not `main`, because neither defect exists on main**: the
`'Integrity'` section and the `storage-authority` check both arrive via `ba397aaea` on
`claude/exsc-storage-authority-row` and reach here through the integration merge.

## F7 — two adjacent sections both named for integrity

`renderCheckLedger` groups by the exact section string, so the signer saw:

```text
✓ proposal integrity    6/6 checks green · 6/6 network results verified
✓ Integrity             1/1 checks green · 1/1 network results verified
```

Renamed so neither contains "integrity", removing the collision at the root rather than
disambiguating two near-synonyms: `'proposal integrity'` → `'Proposal'` (gates A–F ask what the
proposal *is*), `'Integrity'` → `'Deployed state'` (G and K both compare the live chain against
what main declares). Pinned by a test that refuses any section name containing another, which
fires on the original pair.

## F9 — gate G blocked the honest control with a remedy no signer could follow

`LiFiDiamond.owner`'s expected value is sourced `from: 'deployments'` → anchor `A-MONGO` →
`REPORTING_ONLY_ANCHORS` → `coerceStatus` turns the *pass* into an `error` → a blocking row with
no acknowledgement path. Because the all-matched loop downgrades the row to the weakest anchor in
the set, **every** proposal touching `LiFiDiamond` hit it. The signer was told to "re-run this
check", which could never change the outcome.

`rpc-quorum` already solves the same "cannot decide" case via `checkClass: 'semantic'` +
`needs-ack`. **Flipping gate G to `semantic` was measured and rejected as a false green**: a
semantic `fail` routes to the acknowledgement path, so a genuine storage-authority swap — the
exact attack gate G exists to catch — would become a prompt the signer clicks past. That contrast
is now pinned by a test.

Instead, a per-definition opt-in (`undecidableIsAcknowledgeable`) narrowed to `needs-ack` on an
anchor that can only report, keeping `checkClass: 'integrity'`.

Two hardenings against the obvious version of this fix:

1. Gating on `REPORTING_ONLY_ANCHORS` is **wrong** — `A-UNRESOLVED` is a member, so that version
   would let an expectation of unknown provenance reach the acknowledgement path. Narrowed to
   `ACKNOWLEDGEABLE_REPORTING_ANCHORS = {A-MONGO, A-PROPOSAL}`.
2. The existing all-matched loop assigns `anchor` on each hit, so **calldata order** decided
   between `A-UNRESOLVED` and `A-MONGO`. An unresolved expectation sitting before a
   record-sourced one would have been acknowledgeable. The loop now resolves `A-UNRESOLVED`
   first; the test for that fails without the change.

`undecidableIsAcknowledgeable` is digested into `buildReviewAttestation`, for the same reason
`checkClass` is — so it cannot be added off the record.

- [ ] For new facets: n/a

- [x] Arbitrary calls to external contracts — none changed
- [x] Privileged calls / storage modifications — no contract code in this PR
- [x] Preliminary audit — n/a, TypeScript only

## Verification

16 tests fail before the change and pass after. Full `script/deploy/safe/`:
**2375 pass, 8 fail** — and those 8 are **pre-existing on `signing2-integration-0912`**
(measured independently at `ee737146a`: 2359 pass, 8 fail, the same 8). This PR neither fixes
nor breaks them. `tsc-files --noEmit`, `eslint`, `prettier --check` all clean.

The false-green battery asserts `hardBlocked === true` **and**
`requiresAcknowledgement.length === 0` for: live ≠ declared on `deployments`; live ≠ declared on
`globalConfig`; read failed / value undefined; no contract declares an authority; and an
`A-UNRESOLVED` expectation in two shapes. All pass.

## Known gap, deliberately not fixed here

Gate G is still graded only inside `recordSignedSet` — i.e. **after** the signature — so the
acknowledgement this PR adds arrives after the decision, and "Do Nothing" leaves the row missing.
That placement is also the single root cause of all 8 pre-existing failures. Moving the
observation into `proposalCheckResults` is ~5 points and requires deciding what gate G means when
there is nothing to observe, which widens "what may proceed" and wants an owner rather than a
side-effect of this PR.

### DIFF
--- a/script/deploy/safe/check-ledger.test.ts
+++ b/script/deploy/safe/check-ledger.test.ts
@@ -1042,3 +1042,107 @@ describe('nothing that would soften the verdict may erase a mismatch', () => {
     expect(rollup?.unverified).toBe(1)
   })
 })
+
+describe('undecidableIsAcknowledgeable', () => {
+  // A gate that reads live state against an expectation the proposer writes can
+  // neither decide a green nor honestly refuse: the operator has no way to make
+  // a record-sourced expectation into a repo-sourced one. The opt-in gives that
+  // one case a signer to answer it, and nothing else.
+  const AUTHORITIES: ICheckDefinition = {
+    checkId: 'storage-authority',
+    section: 'Deployed state',
+    checkClass: 'integrity',
+    gate: 'Z',
+    title: 'Storage authorities',
+    undecidableIsAcknowledgeable: true,
+  }
+
+  const row = (over: Partial<ICheckResult> = {}): ICheckResult => ({
+    checkId: 'storage-authority',
+    network: 'mainnet',
+    status: 'needs-ack',
+    expected: 'the address main declares',
+    actual: 'the address main declares',
+    anchor: 'A-MONGO',
+    ...over,
+  })
+
+  const verdictOf = (
+    definition: ICheckDefinition,
+    over: Partial<ICheckResult> = {}
+  ) => {
+    const ledger = ledgerOf(['mainnet'], [definition])
+    recordCheck(ledger, row(over))
+    return summariseLedger(ledger)
+  }
+
+  it('keeps needs-ack on an integrity check whose anchor only reports', () => {
+    const ledger = ledgerOf(['mainnet'], [AUTHORITIES])
+
+    expect(recordCheck(ledger, row()).status).toBe('needs-ack')
+
+    const verdict = verdictOf(AUTHORITIES)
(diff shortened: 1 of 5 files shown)


0xDEnYO:
@coderabbitai review

PR #2383, opened 2026-09-13
feat(safe): various UI/UX changes to the signer's view (EXSC-1010, F7, F9)

0xDEnYO:
> [!IMPORTANT]
> ## ⛔ DO NOT MERGE AS-IS — this PR must be rebased onto `main` first
>
> This branch targets `signing2-base-0913`, a daily aggregation branch that is **never
> merged**. Merging this PR into that base does not put any of it on `main`.
>
> Worse, the base already contains the commits of **#2353, #2370 and #2372**, so simply
> flipping this PR's base to `main` would merge those three PRs' content without their own
> reviews — the diff goes from 36 files / +7,032 to **72 files / +18,997**.
>
> **Required order, in full:**
> 1. #2353, #2370 and #2372 merge to `main`.
> 2. `claude/signing2-ui-ux-0913` is rebased onto `main`.
> 3. This PR's base flips to `main`, and the diff drops back to this PR's own work.
>
> Until step 1 completes, nothing in this PR can reach `main`. Blocker state measured
> 2026-09-13 — **none of the three is currently mergeable**:
>
> | PR | review | failing checks |
> |---|---|---|
> | #2353 | no decision | 1 |
> | #2370 | **CHANGES_REQUESTED** | 0 |
> | #2372 | no decision | 5 |
>
> The rebase cannot be short-cut by splitting this PR up: its own commits overlap #2370 in
> 8 files and #2353 in 3 (`confirm-safe-tx.ts` and `confirm-check-registry.ts` among them),
> and the executability-panel commits do not apply to `main` standalone — they depend on
> `signer-view.ts` and `signer-zones.ts`, which do not exist there.

# Which Linear task belongs to this PR?

EXSC-1010, plus the F7 and F9 findings from the overnight signing rehearsal.

# Why did I implement it this way?

This is one of two consolidation PRs that gather the loose Signing 2.0 work into a
reviewable shape. It carries everything that changes **what a signer reads**; its sibling
carries everything that changes **what the run decides**. A change goes whole into the PR
that owns its reason — no fix is split across the two.

What is here:

- The three-zone signer view (`signer-view.ts`, `signer-zones.ts`): every check named as a
  lettered gate, one line per gate with a link and a closing verdict, and the check id and
  anchor code dropped from the signer's rows.
- The executability verdict rendered as a per-call panel, with a revert reason a signer can
  read and printed once rather than twice.
- **F7** — the two check-ledger sections were both named for integrity, so a signer could
  not tell which one a row belonged to.
[...]

PR #2381, opened 2026-09-12
fix(safe): stop a run that graded nothing from closing all-green

0xDEnYO:
# Which Linear task belongs to this PR?

EXSC-686 (found during the 2026-09-12 signing rehearsal)

# Why did I implement it this way?

A run that graded **nothing** closed all-green.

`recordNothingToGrade` wrote `status: 'pass'` on `A-LOCAL` for every registered check on a
network the run skipped. A pass is a verified network result, so `rollUpChecks`'s
`green: passed === expected` was satisfied by `0 === 0`, and the verdict claimed
`ALL CHECKS GREEN — N/N network results verified` over an empty result set.

How this was found, stated precisely, because the two branches differ. The faulty **recording**
is what `main` has. The rendered verdict was observed on `signing2-integration-0912`, which
registers ten checks and wires the ledger; on `main` `confirm-safe-tx.ts` never calls
`renderCheckLedger` at all (`b5799792f`, pending EXSC-994) and `CONFIRM_CHECK_DEFINITIONS` holds
one check, so no `10/10` line is producible there. What was measured on the real production
Arbitrum store is that `list-pending-proposals --network arbitrum --json` returned `count: 1`
at the moment the walk reported `No actionable pending transactions` — a genuine pending
proposal, graded by nothing. The green verdict over that empty set was rendered on the
integration branch.

Rather than special-casing the renderer, this adds a `not-applicable` `CheckStatus` that is
strictly weaker than the `pass` it replaces: it blocks nothing (nothing is wrong) and satisfies
no counter (nothing was checked). `green` becomes `graded > 0 && passed === graded`, so
`0 === 0` can no longer be green, and the narrowed denominator never renders without the
skipped count beside it.

`not-applicable` is also added to `SEVERITY`. Omitting it yields `indexOf` **-1**, which sorts
it ahead of `fail` in `worstResultPerCheck` — "nothing to grade" would have displaced a real
mismatch on a network that had both. That hole is closed here rather than left latent.

Note on scope: `confirm-safe-tx.ts` does not currently render the ledger on `main`
(`b5799792f` removed the render pending EXSC-994, pinned by
`check-ledger-wiring-placement.test.ts`). The faulty **recording** is on main and is what this
fixes; whoever re-wires the render under EXSC-994 inherits the corrected verdict. The render is
deliberately left un-wired here.

# Checklist before requesting a review

- [x] I have performed a self-review of my code
- [x] This pull request is as small as possible and only tackles one problem
[...]

PR #2384, opened 2026-09-13
fix(safe): various code fixes — gate G ordering, vacuous greens, signer preflight (EXSC-686)

0xDEnYO:

EXSC-686. Carries PR #2381 and PR #2380, plus two fixes found while consolidating them.

Fixes EXSC-1011
Fixes EXSC-1016

This is one of two consolidation PRs that gather the loose Signing 2.0 work into a
reviewable shape. It carries everything that changes **what the run decides**; its sibling
(#2383) carries everything that changes **what a signer reads**. A change goes whole into
the PR that owns its reason — no fix is split across the two.

## The one to look hardest at

**`fix(safe): grade gate G before the signature, not after it`** is a Monday merge-blocker,
not a cleanup.

The storage-authority row was pushed from inside `recordSignedSet`, which runs after the
signature is stored (`confirm-safe-tx.ts`, old line 458). A row recorded there can describe
what was signed; it can no longer refuse it. So `proposalCheckResults` emitted no row for a
check `CONFIRM_CHECK_DEFINITIONS` registers, and the run-level ledger counted it missing.

This is an interaction between **#2353**, which registers the check, and **#2370**, which
consolidated the recorders. Verified on each branch separately — #2353 alone is 32 pass / 0
fail, #2370 alone is 49 pass / 0 fail; merged, 8 fail. Whoever merged second on Monday would
have had to choose between landing red CI on `main` and dropping the registration, and
dropping it would have disabled gate G silently while CI went green.

The observation is split out of `recordSignedSet` into `observeSetForProposal` and cached
per proposal, so the chain read still happens once and persistence still runs after signing.

**What gate G decides is deliberately unchanged.** Every path that previously recorded
nothing — calldata that is not a schedule batch, a chain outside the gate's coverage, a read
that threw — still leaves the run without a graded row, and still blocks. This PR moves
*when* the check is graded, not *what* it grades, so there is no direction in which it can
turn a refusal into a pass. F17 (gate G hard-blocking a proposal that touches no contract
declaring a storage authority, e.g. a timelock `updateDelay`) is therefore still open.

## The rest

- **PR #2381** — a run that graded nothing no longer closes all-green; a `not-applicable` no
  longer erases a finding; a prepared network with no proposal records as unverified.
- **PR #2380** — the signer preflight refuses a network the run cannot check, before a
  ledger exists. Cherry-picked; it owns its own render width rather than importing the
  signer view's, because it prints on runs that never reach a ledger and the view lives in
  the sibling PR. That duplication collapses on the integration branch, where both halves
  exist.
- **`refuse a signature on an integrity run that graded nothing`** —
  `assertIntegrityAssertsAllowSigning` decided on `hardBlocked` alone. That answers "did
  anything object", and a run whose every result was `not-applicable` has nothing to object
  with, so it reached the signer indistinguishable from a clean pass — the exact misreading
  `nothingGraded` was added to `ILedgerVerdict` to prevent, by its own docblock. Not
  reachable today (no assertion returns that status); reachable the moment one does, and
  this is the one function in the repo where a ledger status decides whether a transaction
  gets signed. `STATUS_BUCKETS` also had no `not-applicable` entry — it is a
  `ReadonlyMap<CheckStatus, …>`, which TypeScript does not exhaustiveness-check, so widening
  the union compiled silently.
- A latent substring trap: `not.toContain('0/10 network results verified')` is satisfied by
  `"10/10 network results verified"`, so that assertion stopped asserting anything once the
  ledger reached ten rollups. Now anchored on a digit boundary.

## Verification

Both new guards were falsified rather than just run: removing the `nothingGraded` term fails
exactly the new refusal test, and moving the gate-G grading back inside `recordSignedSet`
fails exactly the new placement test. Full suite on this branch: **2302 pass / 0 fail**
(the base it sits on is 2247 / 8).

## Base branch

Targets `signing2-base-0913`, which is `main` plus #2353, #2370 and #2372 already merged.
Once those land, this branch is replayed onto `main` and the base flips. It should merge
directly after them, because it is what keeps `main` green.

## Three criticals found on the preflight after it was folded in

A review gate found three on PR #2380's final commit; all are fixed here, each verified against the
code rather than taken on report.

1. **The probe and the checks disagreed about which endpoints exist.** The preflight read
   `chain.rpcUrls.default.http[0]`; the executability gate reads the whole list through
   `getFallbackTransportForChain`. A dead primary with a healthy spare was refused with exit 78
   before anything ran. Measured from the real store by name only: **75 of the 87 networks that
   declare a primary also declare fallbacks** — 86%, so the two disagreed on the common case rather
   than a corner. `buildReadOnlyClient` now reads through every endpoint a network declares, which
   also makes the ownership read (`safe-utils.ts:2659`) and `reconcile.ts:251` survive a sick
   primary. Daniel ruled for this scope explicitly: fixing only the preflight would have been worse
   than the bug, since the network would start on a healthy spare and every real read would still
   hit the dead primary.

2. **A refusal that named the wrong cause.** `endpointConfigured` treated `--rpc-url` as
   configuration, but the chain is resolved through `getViemChainForNetworkName` *before* the
   override is read, and that resolve needs the variable. With the override set and the variable
   unset, the signer was told to check an endpoint that had never been contacted.

3. **An unbounded, serial probe.** Several unreachable networks printed nothing for minutes. A
   correct refusal that arrives looking like a hang is what makes an operator kill the run and lose
   the reason. Probes now run concurrently, each bounded by `PREFLIGHT_PROBE_TIMEOUT_MS` carried as
   an `AbortSignal` — so the budget bounds the request itself, retries included, not just the wait.
   Findings are still assembled in the caller's order.

**Two limits, stated rather than left to be discovered.** The probe calls `eth_chainId`, the most
permissive method a node serves, so a rate-limited or method-restricted endpoint passes the
preflight and then fails every `eth_call` behind it — this closes one cause of mass-unverified rows,
not the class. And exit 78 also fires on a partial run where some networks were refused and others
signed; nothing parses that code today, but it is a real behaviour change. Both are now in
`docs/MultisigSigningProcess.md` §4.3, which previously had no mention of the preflight,
`CANNOT START` or exit 78 in 794 lines.

**One coverage gap found while verifying.** The handover reported six killed mutations. Re-running
two, *dropped signal* survived: the signal was asserted only on the `rpcUrl` override branch, while
the configured-endpoint branch — the default for every real run, and the one the budget depends on —
had none. Pinned in `test(safe): pin the abort signal on the configured endpoint path`; the mutation
now dies.

---

## One verdict per gate: a result, or a reason it stood down

Three gates only have subjects when a proposal *installs* something — G storage
authorities, H facet version, K deployed bytecode. On a `diamondCut` that only removes,
each said "nothing here for me" in a different vocabulary, and two of them said it as a
green pass.

- **H** mapped `removal` and `no-diamond-cut` to `status: 'pass'`, and the mirrored
  `INT-TIMELOCK-DELAY` row did the same for a payload carrying no schedule. Both now
  record `not-applicable` with the reason on the row. A comparison that never happened
  no longer reaches the verified numerator.
- **K** was not on the run ledger at all — hence `11 gates · 10 owe a result`. It is now
  registered like the other ten, so `CONFIRM_CHECK_DEFINITIONS` is the single roster the
  footer counts. Its refusal stays in `assertCodehashSignGateAllowsSigning`: the new row
  reports that verdict, it does not decide it.
- **G** already recorded `not-applicable`; the renderer was suppressing the row, so the
  reason never reached the screen. It prints now.
- `renderIntegrityAsserts` said `· NO CLAIM`. Same word as everywhere else now.

**The distinction the change turns on.** `madeNoClaim` on the codehash gate covers two
opposite facts, so `ICodehashSignGate` carries `unopened` structurally. A payload read to
the end that installs nothing is `not-applicable` and blocks nothing; a payload whose
frames would not open is `error` on `A-UNRESOLVED` and blocks — the same input gate G
already refuses on, which makes this a second lock on that door rather than a new one.
Splitting on `summary` instead would have rested it on a sentence the proposer's calldata
shapes.

`targetStateCheckResult`'s status seed moved from `pass` to `not-applicable`: `worstOf`
keeps the lower-ranked side and `not-applicable` is weakest in `SEVERITY`, so a `pass`
seed would have outranked every finding that maps to it and a cut installing nothing
would still have reduced to green. Pinned by a mixed-cut test — a removal paired with an
install must reduce to the install's verdict.

The section line and the closing verdict now divide by one shared predicate
(`stoodDown`), so a gate with nothing to do leaves both denominators. A run where every
applicable gate is green and the rest stood down closes `ALL APPLICABLE CHECKS GREEN`
rather than `COVERAGE INCOMPLETE`; a gate that graded nothing because a network never
answered still reads as the shortfall it is.

Verified from the rendered report, not only from the suite — an Arbitrum Remove closes
green with four reasons on screen and nothing blocking, and the same proposal with one
unopenable frame closes `BLOCKED` with G *and* K unverified. Both mutations that would
re-open the hole (ignore `unopened`; restore the `pass` seed) were run and killed.

---

## The checks for the next proposal run while you read this one

The network prefetch prepared the next network while the operator reviewed the current
one. Inside a network the proposals were still strictly sequential, and every per-proposal
chain read — the codehash gate, the integrity assertions, the executability simulation, the
RPC quorum, the sign-time set — ran inline immediately before `consola.prompt('Select
action:')`, which blocks on a human. The whole per-proposal cost was paid with the signer
waiting, and nothing was computed during the one interval where the machine is idle.

Those reads move into `computeProposalEvidence` and are taken one proposal ahead through a
single-slot queue. One ahead and no further: the win is the human-read interval, and depth
past it only widens the window in which a read can go stale.

**What a prefetched bundle is checked against.** The Safe's on-chain nonce, plus a counter
this run bumps immediately before every `executeTransaction` — resolved once before the
reads start and again before they are handed over, and used only if the two agree. The
counter is bumped *before* the call because a broadcast that throws may still have reached
the chain. An anchor that cannot be resolved at either end is not an agreement: the bundle
is discarded and re-taken inline, which is what the signer would have got with no prefetch
at all.

**Which gates carry forward, and why.** A signature is stored against one proposal's row
and changes no chain state, so a bundle read before it is still true after it — that is
what makes prefetching worth anything, since signing is the common action. Everything that
reads the signature set or the signer's position in the queue — the option list,
`canExecute`, the C8/C9 nonce interlocks — stays in the loop, where it is cheap and where
it has to be current.

The sign-time set is the one verdict that survives a discard, on its own cache. That rests
on what a Safe execution does here — it executes `scheduleBatch`, enqueuing a timelock
operation, and changes neither the code nor the storage authorities gate G reads, so the
observation is the same on both sides of it. The assumption is stated at the cache rather
than left implicit, because it is the one that would fail first if a proposal ever mutated
the diamond directly instead of scheduling.

**A read that is not fresh says so.** A served bundle prints how old its reads are and
names what was re-validated — the Safe's nonce, and that this run has broadcast nothing
since — rather than claiming the reads are current. A discarded one prints why it was
thrown away.

**A background failure is a blocking row, not a missing one.** A prefetch that throws lands
as a bundle whose verdicts are all absent, which `proposalCheckResults` turns into
`unresolved` rows that block — so every check the registry owes still reports. A bundle
that never arrived would produce no row, and a check that reports nothing rolls up as one
the run was never owed, which is exactly how a failed prefetch would go unnoticed.
`rpc-quorum` is the deliberate exception: an unmade quorum read is `needs-ack`, because
that gate reports on endpoint redundancy and must not block a run on it, so ten of the
eleven block and one is acknowledgeable. Driven in `confirm-check-registry.test.ts`.

The reads also stopped writing to the terminal: they write to a deferred console carried on
the bundle and flushed when the proposal it belongs to is displayed. A warning about the
next proposal, printed under this one's verdicts, describes nothing the signer is looking
at. That covers what `computeProposalEvidence` itself writes, not the shared helpers it
calls — `getFallbackTransportForChain` and `SafeClient`'s own read errors still write
directly, and threading a sink through them reaches call sites outside this run.

Placement guards that pinned these reads to the loop body were widened to the loop **plus**
`computeProposalEvidence`, not relaxed — the reads are still taken once per proposal and
still while a refusal is available. The narrow window is kept for the narrow questions
(nothing in the loop records; the push sits above the operator's own `continue`).

Two of the new guards were proven dead by mutation and rebuilt: the broadcast-counter guard
compared two `indexOf` positions, which a bump behind a never-true condition satisfies just
as well, and the unset-anchor case asserted only that a recompute happened, which deleting
the branch that produces it left green. Both now fail on the mutation that killed them.

One latent bug fixed in passing: `readDeploymentRecords` guarded its one-read-per-run cache
with a flag set before the await, so a second caller arriving mid-read would have seen the
not-yet-assigned records as "the deploy log is unavailable" and graded every address as one
nobody deployed. Sequential callers could not hit it; concurrent ones can. It now memoises
the promise.


0xDEnYO:
@coderabbitai review

0xDEnYO:
Pushed here: the capped fan-out fix, plus `origin/main` (now carrying #2390).

**The fan-out bug was live on this branch.** `buildReadOnlyClient` fanned out via `getFallbackTransportForChain`, which keeps each endpoint's own retry profile — TronGrid's is 8 retries on a 2s exponential backoff, and this client decides whether a network has actionable transactions at all, so that wait lands before the signer sees a network list. The endpoints now resolve through `getSignTimeTransportConfig` and compose with `fallback`, keeping both the fan-out and the cap. The old guard couldn't see it: it bans `getTransportConfigFromRpcUrl` by name and the uncapped fan-out reaches it indirectly, so the file carried neither banned name.

**Three things only surface when this lands together with #2383 — whoever merges second owns them:**

1. The prefetched quorum read needs `createPinnedBlock` (it lives on #2383, not here). Without it the fan-out compares endpoints at different block heights and reads the drift as a fork.
2. `storageAuthorityCheckResult` on a matched deployment-record expectation: this PR's test asserts `pass`; #2383's `undecidableIsAcknowledgeable` makes it `needs-ack`. #2383's is the right call — gate G reads an expectation the proposer wrote, so it can neither claim the green nor refuse every honest cut.
3. `ALL_GATE_DEFINITIONS` appends `CODEHASH_CHECK` on the assumption it isn't registered. This PR registers it. Already fixed on #2383 (keyed by `checkId`).

The three-way composition (this + #2383 + #2390) is verified locally on `signing2-integration-0914`: 5140 TS tests green. Happy to push that branch if it's useful as a reference for the merge.

0xDEnYO:
Folded into #2393, which carries this branch's full history — keeping the stack in sync cost more than reviewing the parts separately saved. Closing here; review moves to #2393.

PR #2385, opened 2026-09-13
fix(safe): print nothing when the check ledger graded nothing (EXSC-1016)

0xDEnYO:

EXSC-1016, a finding from the 2026-09-13 signing rehearsal.

A run with no pending transaction printed seventeen lines of check ledger — header, five
section lines, one line per gate A–J — each saying `nothing to grade · 1 network not
applicable`, above a verdict saying the same thing a third time.

`green` is `graded > 0 && passed === graded`, so a check that graded nothing is not green,
and `renderSection` expands every check that is not green. That expansion rule exists for
the failure case: suppressing a non-green check would leave a `✗` section whose shortfall
has no row explaining it. A `·` section has no shortfall to explain, and falls through it.

`renderCheckLedger` now returns the closing line alone when the whole run graded nothing:

```text
VERDICT: NOTHING TO REVIEW — nothing was graded, so nothing was verified · 1 network had nothing to grade · no proposal was graded on arbitrum — no pending transaction was fetched
```

The suppression is safe by construction rather than by care: `nothingGraded` is false the
moment any result is recorded, `error` and missing included, so a run whose gates *failed*
to run still prints in full and still blocks. Nothing about a graded run changes.

## Verification

Replayed with the real `CONFIRM_CHECK_DEFINITIONS` and the exact row `confirm-safe-tx`
records for a network with no pending transaction: 17 lines → 1, byte-identical closing
line. Both directions falsified — disabling the early return fails the new test, forcing it
always-on fails 41 tests. 2548 pass in `script/deploy/safe/`.

One existing test asserted on the section line of a fully-skipped run; that line no longer
exists, so it is replaced by an exact-length assertion on the whole render, paired with a
new test that a run with one graded network still prints its sections and its non-green
checks.


0xDEnYO:
Folded into #2384, which already carries the `nothingGraded` / NOTHING TO REVIEW work this builds on — commit 10257be54 on `claude/signing2-code-fixes-0913`.

PR #2387, opened 2026-09-13
feat(safe): grade periphery registrations at sign time

0xDEnYO:

https://linear.app/lifi-linear/issue/EXSC-1018/registerperipherycontract-calldata-reaches-the-signer-ungraded-by-the

`registerPeripheryContract` calldata reached the signer **ungraded**.
`calldata-address-check.ts` declared `AddressRoleEnum.PeripheryRegistration` and listed it
in `REFUSAL_BEARING_ROLES`, but `calldata-address-collector.ts` only ever emitted
`FacetAdd`/`FacetReplace`/`FacetRemove`/`CutInit` — nothing constructed a reference with
that role. Until #2383 this was partly masked by the inline `(✅ matches deployments)` tick
in zone 1; the zone-1 renderer deliberately dropped it ("zone 1 states, zone 2 grades"), so
the gap became total.

### The name is a lookup key, not an expectation

`ICalldataAddressInput.expectedIdentities` may never carry an expectation derived from the
calldata being judged, and the periphery name is a proposer-written `string` argument — so
it cannot be the anchor. What it can be is the **key**: the record answers *which address
it currently holds under that name on this network*, and that answer was written before the
proposal existed. `IAddressReference.registeredName` carries the key; the anchor stays the
deploy log.

The inverse framing — checking the record's name for the address against the calldata's
name — was rejected because it **passes a superseded deployment**: the old Executor is
still filed under "Executor". 280 of 859 real periphery `(name, network)` groups carry more
than one version, so that is the common case, not a corner.

### The old tick read a source this gate may not decide against

`getPeripheryDeploymentCheckSuffix` resolved names through `deployments/<network>.json`.
That is `DeploymentIndexSourceEnum.RepoDeploymentsFile`, which the check explicitly refuses
as an authority because it merges to `main` only *after* execution — so at sign time it
lacks the very address the proposal is about. This anchors on the MongoDB deploy log
instead, at no extra cost: `confirm-safe-tx` already loads the whole production record.

### Current = latest by `timestamp`, and ties error

`timestamp` is deploy time; `createdAt` is row-write time. They diverge (rows backfilled in
2025 carry 2023 deploy times), so ordering on the write time would rank a backfilled first
deployment above the redeploy that superseded it. Where the record cannot name a single
current entry — a tie on the newest timestamp, or an entry with no usable time — the check
**errors rather than guessing**, reusing `IdentityUnchecked`. Guessing would decide a
refusal on which row the record happened to list first. No new grade was added:
`NameMismatch` carries the decided disagreement.

### Zero is an unregistration, not a malformed registration

`registerPeripheryContract(name, address(0))` unregisters the name, and
`docs/DeploymentLogs.md` names exactly that call as the cleanup proposal a deprecated
periphery contract's registry residue is input for. The role therefore joins
`ZERO_LEGAL_ROLES`. **This flips an existing pinned test** that asserted `IllegalZero` —
written when the role was declarative and unreachable; leaving it would have refused a
documented flow.

### One envelope traversal, not two

A second walk for periphery would silently stop seeing through the next envelope kind added
to `collectDiamondCutCalls` — and a missed registration is an ungraded address, which is
the gap this PR closes. The traversal moved into `collectLeafCalls`;
`collectDiamondCutCalls` is re-expressed over it, keeping the aligned-selector guard
caller-supplied so its observable behaviour is unchanged (all 14 of its own tests and the
funnel-deploy-gate suite pass untouched).

### Proof the gate actually fires

Per the review-gate rule, same-PR tests are not evidence. Exercised against the **real
production deploy log** (`.cache/deployments_production.json`, 4,444 records), synthesising
a real `registerPeripheryContract` call per periphery `(name, network)` pair:

| outcome | count |
|---|---|
| pairs exercised | 636 |
| current address → `Resolved` | 635 |
| superseded address → **REFUSED** `NameMismatch` | 195 |
| cannot decide (timestamp tie) | 1 |
| unexpected | 0 |

The single undecided case is `Permit2Proxy` on `abstract`, where two versions share a
deploy timestamp to the second — the real shape the ambiguity guard exists for, found in
the data before the guard was written. 635 clean passes is the other half of the proof: a
gate that refused everything would satisfy every refusal case above.

The committed tests use the same real tuples from `deployments/_deployments_log_file.json`
(real addresses, real `TIMESTAMP`s) — `Executor` on mainnet has three genuine versions, and
one test pins that the *address-side* grading cannot separate the current from the
superseded one, so a regression in the name anchor cannot hide behind it.

Incidentally: 9 periphery records in the production log carry a malformed address or no
timestamp. Not addressed here.

### A determinism bug this PR's own review found

`timestamp` arrives either as an instant or as a zone-less wall clock
(`2023-07-27 16:43:51`, the shape the export writes). The first version read the
zone-less spelling in **the signer's local zone**, so two signers could order the same
two records differently — seven hours apart between UTC+7 and a UTC runner — and reach
opposite verdicts on identical calldata against an identical record. It is read as UTC
now.

The regression test covers both directions deliberately: a positive offset moves a
zone-less time earlier and a negative one later, so a single-direction fixture leaves the
bug invisible to every signer on the other side of UTC. Reverting the fix fails the suite
under `Asia/Bangkok`, `America/Los_Angeles` and `Pacific/Kiritimati`, and passes under
`UTC`, where the bug cannot manifest.

### What the gate flags on legacy entries

Replaying today's committed registry entries through the gate gives 22 of 734 non-resolved (3.00%).
That is **not** the rate an honest proposal sees, and the distinction matters: a real
registration deploys the contract first, which writes a fresh record, and
`proposePeripheryWithWhitelist.ts` then registers the address from
`deployments/<network>.json` — so the record's newest entry under that name *is* the address
being registered, and it resolves. Nobody re-registers a contract deployed months ago.

What the 22 actually are, all on long-standing entries, measured against a MongoDB cache
refreshed 2026-09-14: 11 registered contracts predating the deploy log with no record at all
(all the same CREATE3 `FeeCollector` address across chains), 9 where the committed registry is
behind a newer deployment — the gate saying the registry is stale, which is true — the 1
`abstract` timestamp tie, and 1 `ERC20Proxy` recorded against `cronos` while registered on
`opbnb`. None is reachable from the deploy-then-register flow, and all are printed rather than
blocking. The `outlaw` → `robinhood` rename previously accounted for 6 more; it has since been
corrected in MongoDB and they are gone.

Re-registering a **superseded** deployment — the rollback path — does read as a mismatch by
construction, since the record's most recent under that name is the contract being rolled
back. Documented at the branch; it would need an anchor for that intent before this could ever
block.

**Tron is unaffected by this PR but not by the gate.** The record stores Tron addresses
base58 while calldata carries 20-byte hex, so every Tron address already grades `Unknown` —
for `FacetAdd`, `FacetReplace` and `CutInit` as much as for periphery. Pre-existing,
report-only, and now tracked as a third affected site on EXSC-1017 rather than widened into
this PR.

### Notes for the reviewer

- Based on `signing2-base-0913` rather than #2383 — the collector and check both live on
  the base, so this is a sibling of #2383/#2384, not a dependant.
- The gate is **report-only** and always has been: `assertCalldataAddressesResolve` has no
  non-test caller and `verdict.refuses` is never read in production. This matters for
  reading the zero-address change — it relaxes nothing that was ever enforced, and the
  refusal it removes was a printed line. The zero refusal *was* deliberate in #2342
  ("refuses on ... the zero address in a role that must name a contract"); it was reasoned
  before the deregistration flow was accounted for, and `docs/DeploymentLogs.md` documents
  that flow.
- The walker refactor is behaviour-preserving **including** `undecodable` ordering, which
  the first version did change: the split discovered envelope failures before
  selector-carrying leaves, so the indices could come back out of order. They are sorted
  now, which is the order the single pre-split walk could only have produced.
- **`run-ts-tests` is red, and none of it is this change.** The failures are exactly the 8 in
  `confirm-check-registry.test.ts` that fail on a clean `signing2-base-0913` checkout
  (`totals.missing` is 1 where the suite expects 0). They have gone unseen because every other
  PR on this base is a draft, and a draft skips `run-ts-tests` while `ts-tests-required` still
  reports `pass` — #2383 and #2384 both show exactly that. This PR is the first on the base
  flipped to ready, so it is the first to actually run the suite. The base needs fixing before
  any of the stack can merge green; it is not addressed here.


0xDEnYO:
@coderabbitai review

0xDEnYO:
Superseded by #2390, which carries the same two commits rebased off the temporary `signing2-base-0913` integration branch and onto #2370 — the PR that brings `calldata-address-collector.ts` to `main`. Closing this one; review #2390 instead.

PR #2390, opened 2026-09-14
feat(safe): grade periphery registrations at sign time (EXSC-1018)

0xDEnYO:

Fixes [EXSC-1018](https://linear.app/lifi-linear/issue/EXSC-1018/registerperipherycontract-calldata-reaches-the-signer-ungraded-by-the)

`registerPeripheryContract` calldata reached the signer **ungraded**.
`calldata-address-check.ts` declared `AddressRoleEnum.PeripheryRegistration` and listed it
in `REFUSAL_BEARING_ROLES`, but `calldata-address-collector.ts` only ever emitted
`FacetAdd`/`FacetReplace`/`FacetRemove`/`CutInit` — nothing constructed a reference with
that role. Until #2383 this was partly masked by the inline `(✅ matches deployments)` tick
in zone 1; the zone-1 renderer deliberately dropped it ("zone 1 states, zone 2 grades"), so
the gap became total.

### The name is a lookup key, not an expectation

`ICalldataAddressInput.expectedIdentities` may never carry an expectation derived from the
calldata being judged, and the periphery name is a proposer-written `string` argument — so
it cannot be the anchor. What it can be is the **key**: the record answers *which address
it currently holds under that name on this network*, and that answer was written before the
proposal existed. `IAddressReference.registeredName` carries the key; the anchor stays the
deploy log.

The inverse framing — checking the record's name for the address against the calldata's
name — was rejected because it **passes a superseded deployment**: the old Executor is
still filed under "Executor". 280 of 859 real periphery `(name, network)` groups carry more
than one version, so that is the common case, not a corner.

### The old tick read a source this gate may not decide against

`getPeripheryDeploymentCheckSuffix` resolved names through `deployments/<network>.json`.
That is `DeploymentIndexSourceEnum.RepoDeploymentsFile`, which the check explicitly refuses
as an authority because it merges to `main` only *after* execution — so at sign time it
lacks the very address the proposal is about. This anchors on the MongoDB deploy log
instead, at no extra cost: `confirm-safe-tx` already loads the whole production record.

### Current = latest by `timestamp`, and ties error

`timestamp` is deploy time; `createdAt` is row-write time. They diverge (rows backfilled in
2025 carry 2023 deploy times), so ordering on the write time would rank a backfilled first
deployment above the redeploy that superseded it. Where the record cannot name a single
current entry — a tie on the newest timestamp, or an entry with no usable time — the check
**errors rather than guessing**, reusing `IdentityUnchecked`. Guessing would decide a
refusal on which row the record happened to list first. No new grade was added:
`NameMismatch` carries the decided disagreement.

### Zero is an unregistration, not a malformed registration

`registerPeripheryContract(name, address(0))` unregisters the name, and
`docs/DeploymentLogs.md` names exactly that call as the cleanup proposal a deprecated
periphery contract's registry residue is input for. The role therefore joins
`ZERO_LEGAL_ROLES`. **This flips an existing pinned test** that asserted `IllegalZero` —
written when the role was declarative and unreachable; leaving it would have refused a
documented flow.

### What review round 2 changed

Three things, all in the periphery path this PR adds:

- The aligned-selector scan ran on every leaf that was not itself a registration, including
  `diamondCut` leaves that had just decoded. `PeripheryRegistryFacet` lists
  `registerPeripheryContract` among its own selectors, so an install cut for it — a core facet —
  was reported as both read and unreadable. The scan now runs only on leaves neither reader
  opens, and the hiding place that exclusion gives up (a registration in a cut's `initCalldata`,
  which the walk does not open) is scanned explicitly.
- The zero-address branch returned before anything read `registeredName`, so on the one call
  shape where the name is the whole payload it was neither checked nor quoted. It now reports
  the record's answer for the name, and stops calling zero "the required value" in a role where
  both zero and non-zero are legal.
- `renderCalldataAddresses` prints a finding only for a contradiction in a refusal-bearing role.
  `NotApplicable` is in neither set, so the line the previous point added was built and dropped —
  the signer saw `✓ 0 of 1 resolved` and nothing else. It has a warn line of its own now. The
  same applies to `FacetRemove` and `CutInit` zeros; that is pre-existing and untouched here.

### One envelope traversal, not two

A second walk for periphery would silently stop seeing through the next envelope kind added
to `collectDiamondCutCalls` — and a missed registration is an ungraded address, which is
the gap this PR closes. The traversal moved into `collectLeafCalls`;
`collectDiamondCutCalls` is re-expressed over it, keeping the aligned-selector guard
caller-supplied so its observable behaviour is unchanged (all 14 of its own tests and the
funnel-deploy-gate suite pass untouched).

### Proof the gate actually fires

Per the review-gate rule, same-PR tests are not evidence. Exercised against the **real
production deploy log** (`.cache/deployments_production.json`, 4,444 records), synthesising
a real `registerPeripheryContract` call per periphery `(name, network)` pair:

| outcome | count |
|---|---|
| pairs exercised | 636 |
| current address → `Resolved` | 635 |
| superseded address → **REFUSED** `NameMismatch` | 195 |
| cannot decide (timestamp tie) | 1 |
| unexpected | 0 |

The single undecided case is `Permit2Proxy` on `abstract`, where two versions share a
deploy timestamp to the second — the real shape the ambiguity guard exists for, found in
the data before the guard was written. 635 clean passes is the other half of the proof: a
gate that refused everything would satisfy every refusal case above.

The committed tests use the same real tuples from `deployments/_deployments_log_file.json`
(real addresses, real `TIMESTAMP`s) — `Executor` on mainnet has three genuine versions, and
one test pins that the *address-side* grading cannot separate the current from the
superseded one, so a regression in the name anchor cannot hide behind it.

Incidentally: 9 periphery records in the production log carry a malformed address or no
timestamp. Not addressed here.

### A determinism bug this PR's own review found

`timestamp` arrives either as an instant or as a zone-less wall clock
(`2023-07-27 16:43:51`, the shape the export writes). The first version read the
zone-less spelling in **the signer's local zone**, so two signers could order the same
two records differently — seven hours apart between UTC+7 and a UTC runner — and reach
opposite verdicts on identical calldata against an identical record. It is read as UTC
now.

The regression test covers both directions deliberately: a positive offset moves a
zone-less time earlier and a negative one later, so a single-direction fixture leaves the
bug invisible to every signer on the other side of UTC. Reverting the fix fails the suite
under `Asia/Bangkok`, `America/Los_Angeles` and `Pacific/Kiritimati`, and passes under
`UTC`, where the bug cannot manifest.

### What the gate flags on legacy entries

Replaying today's committed registry entries through the gate gives 22 of 734 non-resolved (3.00%).
That is **not** the rate an honest proposal sees, and the distinction matters: a real
registration deploys the contract first, which writes a fresh record, and
`proposePeripheryWithWhitelist.ts` then registers the address from
`deployments/<network>.json` — so the record's newest entry under that name *is* the address
being registered, and it resolves. Nobody re-registers a contract deployed months ago.

What the 22 actually are, all on long-standing entries, measured against a MongoDB cache
refreshed 2026-09-14: 11 registered contracts predating the deploy log with no record at all
(all the same CREATE3 `FeeCollector` address across chains), 9 where the committed registry is
behind a newer deployment — the gate saying the registry is stale, which is true — the 1
`abstract` timestamp tie, and 1 `ERC20Proxy` recorded against `cronos` while registered on
`opbnb`. None is reachable from the deploy-then-register flow, and all are printed rather than
blocking. The `outlaw` → `robinhood` rename previously accounted for 6 more; it has since been
corrected in MongoDB and they are gone.

Re-registering a **superseded** deployment — the rollback path — does read as a mismatch by
construction, since the record's most recent under that name is the contract being rolled
back. Documented at the branch; it would need an anchor for that intent before this could ever
block.

**Tron is unaffected by this PR but not by the gate.** The record stores Tron addresses
base58 while calldata carries 20-byte hex, so every Tron address already grades `Unknown` —
for `FacetAdd`, `FacetReplace` and `CutInit` as much as for periphery. Pre-existing,
report-only, and now tracked as a third affected site on EXSC-1017 rather than widened into
this PR.

### Notes for the reviewer

- **Based directly on `main`** (rebased once #2370 landed the collector there). Replaces #2387,
  which was opened against the temporary `signing2-base-0913` integration branch.
- The gate is **report-only** and always has been: `assertCalldataAddressesResolve` has no
  non-test caller and `verdict.refuses` is never read in production. This matters for
  reading the zero-address change — it relaxes nothing that was ever enforced, and the
  refusal it removes was a printed line. The zero refusal *was* deliberate in #2342
  ("refuses on ... the zero address in a role that must name a contract"); it was reasoned
  before the deregistration flow was accounted for, and `docs/DeploymentLogs.md` documents
  that flow.
- The walker refactor is behaviour-preserving **including** `undecodable` ordering, which
  the first version did change: the split discovered envelope failures before
  selector-carrying leaves, so the indices could come back out of order. They are sorted
  now, which is the order the single pre-split walk could only have produced.
- Full `script/deploy/safe/` + `script/deploy/shared/` suite: **2790 pass, 0 fail.**
- Worth recording, since it cost a round to find: on `signing2-base-0913` the same commits left
  8 failures in `confirm-check-registry.test.ts` (`totals.missing` 1 where the suite expects
  0). They are not on #2370 and not on `main` — the integration branch introduces them by
  combining its merges. They stayed invisible because every PR on that base is a draft, and a
  draft skips `run-ts-tests` while `ts-tests-required` still reports `pass` (#2383 and #2384
  both show exactly that). Whoever lands that branch will need to deal with it.


0xDEnYO:
@coderabbitai review

script/deploy/safe/calldata-address-check.ts
```diff
@@ -336,6 +379,181 @@ const normalizeExpectations = (
...
+const currentUnderName = (
+  entries: readonly IDeploymentIndexEntry[],
+  name: string,
+  network: string
+): { entry?: IDeploymentIndexEntry; undecided?: string } => {
+  const wanted = name.trim().toLowerCase()
+  const onNetwork = entries.filter(
+    (entry) =>
+      entry.contractName.trim().toLowerCase() === wanted &&
+      entry.network.trim().toLowerCase() === network.trim().toLowerCase()
```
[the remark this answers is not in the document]
0xDEnYO: Valid — fixed in 6ba8acbc5.

Confirmed the mechanism in `PeripheryRegistryFacet.sol`: `mapping(string => address)` written as `s.contracts[_name]` with no normalisation, so `executor` and `Executor` really are distinct slots and a case-variant call would leave the name the diamond serves untouched.

The name is now compared verbatim and `queriedNames` keeps its casing. Network and address comparisons stay case-insensitive on purpose — the network comes from repo config rather than calldata, and address checksum case is display-only.

Checked that nothing legitimate depended on the fold before changing it: across 1,136 registry names on 71 networks, no registry name differs from the deploy log's `contractName` only by case, and no two record names collide case-insensitively. Case-variant test added; it fails with the fold restored and passes without it.

script/deploy/safe/calldata-address-collector.ts
```diff
@@ -39,6 +50,72 @@ const ROLE_BY_ACTION: Readonly<Record<number, AddressRoleEnum>> = {
...
+  const references: IAddressReference[] = []
+  const unreadable: string[] = []
+  const seen = new Map<number, number>()
+
+  for (const leaf of leaves) {
+    if (leaf.selector !== REGISTER_PERIPHERY_SELECTOR) {
+      // A leaf this walk could not open, which could be carrying a registration.
+      // `collectLeafCalls` reports the envelopes it failed on; this covers the
+      // ones it never recognised as envelopes at all.
+      if (carriesAnySelectorAligned(leaf.data, [REGISTER_PERIPHERY_SELECTOR]))
```
gvladika: **A cut that touches `PeripheryRegistryFacet` now reports a false "CANNOT CHECK".**

This runs the aligned-selector heuristic on every leaf that isn't a periphery registration — including `diamondCut` leaves that were just decoded successfully. A cut that Adds or Replaces `PeripheryRegistryFacet` lists that facet's selectors in `bytes4[] functionSelectors`, and one of them *is* `0x5c2ed36a`, byte-aligned by construction.

Reproduced at `22973257d`:

```
--- cut adding PeripheryRegistryFacet
  undecodable: [ "call[0] (carries a registerPeripheryContract selector this could not read through)" ]
  error: true  refuses: false
  finding: identity-unchecked | call[0].diamondCut[0].cuts[0] (0x1111…) is PeripheryRegistryFacet@1.0.0 on mainnet …
```

The check both read the call and declared it unreadable. `PeripheryRegistryFacet` is in `config/global.json` `coreFacets`, so this is a live proposal shape (new-network onboarding, core-facet updates), not a corner. Report-only today, so it costs a spurious ⛔ — but it trains signers past exactly the line this module argues must never be ignorable, and it becomes a hard block the day the gate is promoted.

`diamondCutCallsIn` doesn't have this problem because it skips cut leaves *before* scanning. The heuristic exists to catch **unopened envelopes**, so the exclusion should be symmetric — scan only the leaves no reader handled:

```ts
const HANDLED: readonly Hex[] = [DIAMOND_CUT_SELECTOR, REGISTER_PERIPHERY_SELECTOR]
// scan only when leaf.selector is in neither reader's set
```

That also collapses the duplicate-report path (a malformed cut is currently flagged twice — once by each reader) and makes the "one envelope traversal, not two" argument hold at the reporting layer, not just the walk.

Nothing in `calldata-address-collector.test.ts` covers this shape.

script/deploy/safe/calldata-address-check.ts
```diff
@@ -204,10 +244,19 @@ const WARN_ONLY_ROLES: ReadonlySet<AddressRoleEnum> = new Set([
...
+ * `registerPeripheryContract(name, address(0))` unregisters the name, and
+ * [docs/DeploymentLogs.md](../../../docs/DeploymentLogs.md) names that call as
+ * the cleanup proposal a deprecated periphery contract's registry residue is
+ * input for. Refusing it would block a documented flow, and the removal it
+ * performs is subtractive, which T2 does not block on.
+ */
 const ZERO_LEGAL_ROLES: ReadonlySet<AddressRoleEnum> = new Set([
   AddressRoleEnum.CutInit,
   AddressRoleEnum.FacetRemove,
+  AddressRoleEnum.PeripheryRegistration,
```
gvladika, line 259: **The unregistration path never looks at the name.**

`isZero` returns `NotApplicable` before anything reads `registeredName`, so on an unregistration the name is neither checked nor queried. But on `registerPeripheryContract(name, address(0))` the name is the *entire* payload — a typo deletes nothing and leaves exactly the registry residue the cleanup proposal in `docs/DeploymentLogs.md` was written to remove. Reproduced at `22973257d`:

```
--- unregister a name nothing is filed under
  error: false  refuses: false
  finding: not-applicable | call[0].registerPeripheryContract[0] is the zero address, which is the required value for periphery-registration
```

(`"Exectuor"`, against a record holding `Executor`.) This is the one call shape where "the name is a lookup key" would earn the most, and it's the one that doesn't use it. At minimum, report what the record holds under that name.

Separately, on the same rendered line: `which is the required value for periphery-registration` is wrong. Zero is *legal* here, not required — that wording was written for `FacetRemove`, where `LibDiamond` genuinely requires it. In the one role where both zero and non-zero are legal, it reads to a signer as if they're looking at the wrong thing.

gvladika (commented):
Re-reviewed at `22973257d`. The byte-for-byte name fix (`6ba8acbc5`) resolves the case-folding gap I raised — `"executor"` no longer resolves against the record's `Executor`. Verified: full `script/deploy/safe` + `script/deploy/shared` suite 2790 pass / 0 fail, `tsc --noEmit` and eslint clean on the changed files.

Two findings still reproduce at this head, both in the periphery path this PR adds. Neither is a security hole — the gate is report-only — but the first is a deterministic false positive on a core-facet proposal shape, and no test covers it.

script/deploy/safe/calldata-address-collector.ts
0xDEnYO: Valid — fixed in 54b9e754c.

Confirmed the mechanism: `PeripheryRegistryFacet` has exactly two external functions, so an install cut's `functionSelectors` is `[0x5c2ed36a, 0xa516f0f3]`, and `carriesAnySelectorAligned` returns `true` on the encoded cut. It is a core facet (`config/global.json`), so it hit every new-network bring-up.

Took your exclusion, with one addition. A blanket skip of cut leaves gives up a hiding place: `collectLeafCalls` does not open a cut's `_calldata`, so before the exclusion the whole-leaf scan was the only thing that would have noticed a registration carried in init calldata. So the scan now runs only on leaves neither reader opens (`HANDLED_SELECTORS`), and each decoded cut's `initCalldata` is scanned explicitly, reported at `…diamondCut[n].init`. The duplicate-report path collapses as you said — a malformed cut is now reported once, by `diamondCutCallsIn`.

Two tests, and each pins a different half: the `PeripheryRegistryFacet` install cut fails with only your exclusion missing, the init-calldata case fails with only the init scan missing. Enumerated the rest of the shapes against real data (undecodable cut leaf, selector present only as `bytes4[]` data, unknown envelope, nesting past `MAX_UNWRAP_DEPTH`, undecodable args, timelock-wrapped init) — all still reported. Non-aligned occurrences stay missed, unchanged and documented at `diamond-cut-calls.ts`.

script/deploy/safe/calldata-address-check.ts
0xDEnYO: Both points valid — fixed in 7612b7178, and the fix needed a third commit to be worth anything (628e52d17).

The name is read now. `describeLegalZero` branches on role: for a periphery registration it quotes the name and the record's answer for it — the address currently held, that it holds nothing, that it cannot decide, or that the name was never queried. Your `"Exectuor"` repro is one of the tests.

It stays `NotApplicable` rather than becoming a mismatch. 11 committed registry entries predate the deploy log and have no record under their name at all, so "nothing answers to this name" does not separate your typo from a legitimate cleanup — the signer gets the name and what answers to it, and decides.

Wording fixed too: "required value" now only for `FacetRemove`/`CutInit`, where `LibDiamond` requires it. A periphery zero reads "legal in this role".

Worth flagging what reviewing your comment turned up: `renderCalldataAddresses` prints a finding only for a contradiction in a refusal-bearing role, plus `errors`/`warnings`. `NotApplicable` is in none of those, so the line I had just written was built and dropped — the signer saw only `✓ 0 of 1 resolved`. Verified against the real production export: the typo case printed nothing at all. It has a warn line of its own now, with a test that fails if the branch is removed. Same green-line trap applies to `FacetRemove`/`CutInit` zeros, which is pre-existing and left alone here.

0xDEnYO:
@coderabbitai review

gvladika (approved):

PR #2391, opened 2026-09-15
feat(immutables): declare every immutable in src/ (EXSC-1022)

0xDEnYO:

[EXSC-1022](https://linear.app/lifi-linear/issue/EXSC-1022/fleet-wide-immutable-registry-authoring-pass-exsc-869-part-ii-was) — the authoring pass EXSC-869 split off and never built. Companion to [EXSC-1021](https://linear.app/lifi-linear/issue/EXSC-1021/the-immutable-value-check-has-no-call-site-layer-1s-unverifiable-is) (#2384), which gives layer 2 its call site; this is what gives it coverage.

`immutableRegistry.json` held one contract and two slots. Every other immutable was `undeclared`, so the value check had nothing to compare and the sign gate refused — which is what forced the EXSC-976 escape-commit runbook into its first live use on the Arc rollout (#2389).

**80 of 80 immutables now carry an entry, 61 flagged authority-bearing.**

Each entry was read off the contract's own constructor, not matched by name. With #2384 landed a mislabelled `configData` key stops being inert: the slot is compared against the wrong config value and blocks signing — a false MISMATCH, the same failure that forced the runbook, in the opposite direction. `GnosisBridgeFacet` is the proof the labels really are wrong sometimes, filing `GNOSIS_BRIDGE_ROUTER` under `_xDaiBridge`. The other label-drift cases resolved here: `FraxFacet.FRAX_HOP` ← `_hop`, `GardenFacet.REGISTRY` ← `_htlcRegistry`, `LiFiIntentEscrowFacetV2.LIFI_INTENT_ESCROW_SETTLER_V2` ← `_inputSettler`, `LidoWrapper.ST_ETH` ← `_stETHAddress`.

**What is declared `unverifiable` rather than given a key.** Every `Receiver*` executor, `Permit2Proxy.LIFI_DIAMOND`, `GasZipPeriphery.LIFI_DIAMOND` and `LiFiDEXAggregator.BENTO_BOX` come from `deployments/<network>.json`, which `deployRequirements` files under `contractAddresses` — and layer 2 resolves expectations through `configData` only. `EcoFacet`/`UnitFacet`'s backend signer and `EmergencyPauseFacet`'s pauser wallet have no `deployRequirements` entry to name at all. Adding those entries changes what the health check asserts on every network that already runs these contracts, so it is its own change; each `reason` states exactly what would make the slot `config`.

**What is `derived`**, per EXSC-992's rules: `EmergencyPauseFacet._emergencyPauseFacetAddress` (`address(this)`), `SupersetFacet.IS_HUB` (`block.chainid == ARBITRUM_CHAIN_ID`), `Permit2Proxy.PERMIT_WITH_WITNESS_TYPEHASH`, `TokenWrapper.CONVERTER`/`USE_CONVERTER`/`SWAP_RATIO_MULTIPLIER`, `ReceiverStargateV2.recoverGas`. Layer 2 reports `derived` as unpriced today, so these still resolve UNVERIFIABLE — evaluating them is EXSC-935's half, not this one's.

The gate stays warn-only. `--strict` and the `push` trigger on `verify-immutable-registry` belong with the batch that closes the remaining `unverifiable` slots, not here.

# Falsification

Not fixtures — real deployed bytecode, real `config/`, run with #2384's layer-2 wiring:

- **48 config-sourced slots verified against code read from chain across 21 networks, 0 disagreements.** Method per target: strip the metadata trailer, mask the immutable offsets, require the masked hash *and* the deployed length to equal a local `forge build src --ast` of `main`, then price every slot. Targets whose deployed version differs from `main` are reported inconclusive rather than counted.
- The harness was falsified before being trusted: flipping any byte outside the trailer and outside every immutable range turns layer 1 red (3/3); flipping a trailer byte does not, which is what stripping is for. Mislabelling `AcrossFacetV4.WRAPPED_NATIVE` as `_spokePool` produces `disagrees`; tampering with a slot's on-chain value is refused as copies disagreeing; deleting an entry returns it to `undeclared`. An earlier pass of this harness compared a misspelled field and so compared `undefined` to `undefined` — these checks are what caught it.
- `bunx tsx script/deploy/immutables/verify-immutable-registry.ts --out-dir out-immutables` after `FOUNDRY_OUT=out-immutables forge build src --ast`: **80 declared, 0 errors, 0 warnings, 0 unenumerated source files.** Run the two commands rather than quoting the number; it moves with every facet that lands.
- The three Arc contracts that forced the runbook, re-graded on chain: `AcrossFacetV4` **MATCH** (`SPOKEPOOL` ← `config/across.json.arc.acrossSpokePool`, `WRAPPED_NATIVE` ← `config/networks.json.arc.wrappedNativeAddress` — the deliberate `0x…0001` dummy is what config declares, so it verifies rather than disagreeing) and `RelayDepositoryFacet` **MATCH**. `ReceiverAcrossV4` stays UNVERIFIABLE on its `EXECUTOR`; it reaches signers through the periphery-registration check (EXSC-1018), not the codehash gate, so no cut is blocked by it today.

**Not closed by this PR, and worth a reviewer's eye:** `FraxFacet.FRAX_TIP_FEE_MANAGER`/`FRAX_PATH_USD` on avalanche and katana, and `SymbiosisFacet.onchainSwapV3`/`onchainSwapV3Gateway` on scroll, price as unpriceable because `config/` carries no value for those networks. The label is right; the config entry is missing.

- [ ] I have added tests that cover the functionality / test the bug — this PR is data, not code, and adds none. `verify-immutable-registry` in CI is the standing check on its shape; correctness of the labels was established by the out-of-band sweep under Falsification, which is not committed here.


0xDEnYO:
Folded into #2384 as `19961ad6c` — same commit, same file, no changes. Closing this so the registry and the call site that consumes it (EXSC-1021) land together; splitting them would put a PR on `main` whose data nothing reads, and one whose gate has no coverage.

CodeRabbit's finding here carries over to #2384: `EcoFacet.BACKEND_SIGNER` is declared `unverifiable`, and its premise is half right. `EcoFacet` has no `_backendSigner` `configData` entry — that is why it is not `config` — but `DeployEcoFacet.s.sol:38-41` reads `.backendSigner.<ENVIRONMENT>` from `global.json`, exactly as `NEARIntentsFacet` does, so the entry is simply missing rather than absent by design. Tracked on EXSC-1022 and answered on #2384.

PR #2392, opened 2026-09-15
feat(safe): give gate G the periphery contracts whose owner lives in storage (EXSC-686)

0xDEnYO:
Stacked on #2384, which carries the install-scoping this change depends on.

## Why

Gate G asserted exactly two contracts — `LiFiDiamond` and `ERC20Proxy` — and grades only the contracts a proposal *installs*. A `diamondCut` installs facets, never the diamond or the proxy, so the subject set was empty on essentially every proposal we sign.

The gap it was built for is real, though, and nothing else covers it. `TransferrableOwnership` holds `owner` and `pendingOwner` in **storage**, not as immutables, so two deployments differing only in who controls them are byte-identical: the codehash gate MATCHes and its immutable layer has nothing to price. Every periphery contract that carries an owner inherits it, and `registerPeripheryContract` is an installing role, so these are exactly the addresses a signer is putting into service.

## What

- 15 periphery contracts added to `DECLARED_STORAGE_AUTHORITIES`, each with `owner` and `pendingOwner`.
- Each `owner` expectation is the `config/global.json` key that contract's `Deploy<name>.s.sol` passes as `_owner`. The deploy script is the authority here, not `deployRequirements.json` — the two disagree on `FeeCollector` (`.feeCollectorOwner` vs `.withdrawWallet`), and a live read across six networks follows the script. The coverage test pins the table against the scripts.
- `pendingOwner` expects zero. A non-zero one lets its holder claim the contract after the proposal executes, and no bytecode check can see it. This needed a new `{ from: 'zeroAddress' }` expectation source, since the resolver previously knew only deployments and global config and returns `undefined` — which the gate holds on — for anything else.
- `authorityExpectationAnchors` treats `zeroAddress` as `A-LOCAL` alongside `globalConfig`: both are written in the repo and read from nowhere, so both may decide a pass rather than only report.
- Renamed for the signer: `Contract ownership`, asserting `every contract this installs owned by the wallet main declares`. The `checkId` is unchanged — it is persisted in signed-set records.

`LiFiDiamond.pauserWallet` is left as-is, though it is an `immutable` on `EmergencyPauseFacet` and therefore registry territory rather than a storage authority. Removing an existing check is a separate decision.

## Look hardest here

**Registration that points at an existing deployment, not a fresh one.** For a freshly deployed contract this gate is correct — the table matches what the deploy scripts construct. But `script/tasks/diamondUpdatePeriphery.sh` does not deploy: it reads addresses from the deploy log and registers them, and `UPDATE_ALL` re-registers every periphery contract already there. On that path the gate reads older deployments, and measured on arbitrum: `Executor` and `TokenWrapper` have no `owner()` at all, and `ERC20Proxy` has no `pendingOwner()` on arbitrum, polygon or base (`execution reverted` with empty data — no matching selector, not a transport failure).

Such a read becomes `readError` → `error` → `A-UNRESOLVED`, which blocks with no acknowledgement path. And the code cannot currently tell that case apart from an unreachable node: `observeAuthorities` catches generically and `describeError` only redacts the message, so both collapse to the same status. Deciding what the gate should do there is the open question on this PR; the safe direction is that it currently over-blocks rather than under-blocks.

## Tests

- `prebroadcast-authorities.coverage.test.ts`: every contract in `src/Periphery` must be classified exactly once — asserted, `NOT_ASSERTED` with a reason, or `NO_OWNER` with a reason. Exclusions are themselves checked against the deploy script, so a reason that stops being true fails rather than quietly parking a contract.
- Falsified: dropping a table entry, pointing FeeCollector back at the `deployRequirements` value, removing a `pendingOwner` row, removing the `zeroAddress` resolver branch, narrowing the deciding-source set, and parking a contract behind a stale exclusion each turn it red.
- `bun test script/deploy/safe/` — 2449 pass, 0 fail.

## Follow-ups, not in this PR

- `deployRequirements.json` declares `FeeCollector._owner` as `.withdrawWallet` while `DeployFeeCollector.s.sol` passes `.feeCollectorOwner`. It also feeds the healthcheck's immutable binding check.
- On arbitrum, `Permit2Proxy` and `GasZipPeriphery` are owned by addresses that appear in neither `config/global.json` nor the deployments file.


0xDEnYO:
Folded into #2393 (merge `1673a91c8`). Resolution: kept #2393's gate letter, section and `undecidableIsAcknowledgeable: true`, took this branch's broader `Contract ownership` title since the gate now reads periphery owners too. Also dropped `.tmpgate/chk1.ts`, a probe script committed here by accident.

PR #2395, opened 2026-09-15
fix(healthcheck): do not report pre-getter reverts on immutable bindings (EXSC-1024)

gvladika:

Fixes [EXSC-1024](https://linear.app/lifi-linear/issue/EXSC-1024/health-check-skip-immutable-binding-read-when-the-live-facet-predates)

The `immutable-bindings-match-config` invariant has warned on 16 production networks in every daily health-check run for at least a week: `GenericSwapFacetV3.NATIVE_ADDRESS() left unverified — read failed`. Those chains run `GenericSwapFacetV3` v1.0.0, which has no `NATIVE_ADDRESS` immutable at all — the `_nativeAddress` constructor arg arrived in v1.0.1 (PR #708), and every v1.0.0 deploy-log entry carries `CONSTRUCTOR_ARGS: 0x`. The read can only revert, so the warning can never clear, and it says nothing a reader can act on. The existing `legacyGetters` fallback does not help: that mechanism exists for renamed getters, and here there is no earlier name to try.

So the registry gains a `getterSinceVersion` annotation next to `getter`. The check still reads every chain; the annotation decides only how a *failure* reads. A revert from a build the deploy log records as older than the annotated version narrates through consola rather than reporting, because a skipped read is not a finding and should not move the run's verdict. Everything else is unchanged: a build that answers is compared whatever the log says about its version, a transient RPC failure is not a revert and keeps warning, and every unknown resolves to "keep reporting" — an address the log does not record, a blank version, a mistyped annotation, or a getter that reverts on a build new enough to have it all warn exactly as today. A check that quietly stops checking is worse than a noisy one.

Classifying the revert rather than skipping the read is deliberate, and it is the second design I tried. `deployments/<network>.diamond.json` `Version` is not chain truth: `updateDiamondJsonBatch` records `getContractVersion(name)` from the working tree at cut time, so a cut from a stale checkout — or a hand-edited or regenerated entry — can label a live address older than its deployed bytecode. Skipping on that would retire an error-severity binding check permanently and silently. Reading anyway costs the same RPC call and cannot misfire in that direction. Address matching already mitigates the common case (a version bump lands at a new address, the log entry stops matching, the binding stays checked), so this covers the residual wrong-version-at-the-live-address case.

The live version comes from `deployments/<network>.diamond.json`, keyed by the registered address and cross-checked against the contract name so a reassigned log line cannot answer with the wrong build's version. I tried the master deployment log first, as the ticket originally proposed, and rejected it: `_deployments_log_file.json` lags and does not record the live `GenericSwapFacetV3` address on 18 of the 71 active networks (arc, tron, stable, tempo, jovay, morph, 0g and every testnet among them), which would have left the mechanism silently inert there. The diamond log resolves all 71. Its one limit is that it versions facets only — periphery is stored as bare name-to-address pairs — so an annotation on a periphery entry resolves to "unknown" and the revert keeps reporting; that is documented on both the resolver and `liveVersionPredatingGetter`.

Verified fleet-wide against the real logs before opening this: the pre-getter set is exactly the 16 warned networks, 55 networks stay fully reported, and none is left unresolved. Worth flagging for the reviewer: this removes the only place the daily run currently mentions those stale chains, and the underlying drift is larger than this one facet — 242 facet/network pairs are behind their target-state version across 67 of 71 networks, with the health check green. Nothing checks that today; I raised EXSC-1025 for it rather than widening this PR.

The annotation decides whether a binding's failure is reported, so three things that were previously cosmetic became load-bearing and are fixed here too. `script/utils/utils.ts` recorded `1.0.0` in the Tron diamond log whenever it could not determine a facet's version — exactly the value that now means "old enough to explain a revert" — so a failed version lookup during a diamondCut would have muted that chain's binding permanently; it now records the version blank, the way `helperFunctions.sh` deliberately already does. `verify-immutable-registry` now holds every `getterSinceVersion` against the contract's own `@custom:version`: a value ahead of it parses and orders cleanly, so it would mute the finding on every chain at once and is the annotation's one failure mode the invariant cannot catch at read time. That gate and the invariant share a single comparator exported from `immutableBindings.ts`, since the gate only protects the invariant while both order versions identically. And `getterSinceVersion` alongside `legacyGetters` now has a stated meaning — the version the value became readable under any of those names, never the version of a rename, since the other reading would mute exactly the builds `legacyGetters` exists to reach. No entry pairs them today; the field says so before one does.


script/deploy/immutables/getter-coverage.test.ts
```diff
@@ -179,3 +184,122 @@ describe('the repo as it stands', () => {
...
+  })
+
+  it('passes on the annotations the repo actually carries today', () => {
+    // GenericSwapFacetV3 is the only one so far: NATIVE_ADDRESS arrived in 1.0.1, and the
+    // contract is at 2.0.0. This is the assertion that fails if someone annotates a version
+    // that does not exist.
+    const declarations = Object.keys(deployRequirements).map((contract) =>
+      declaration({
+        contract,
+        file: `src/Facets/${contract}.sol`,
```
0xDEnYO: This fabricates `src/Facets/<contract>.sol` for every deployRequirements key rather than using the real declarations, so the path is wrong for every periphery/security entry.

Annotate `getterSinceVersion` on any non-facet entry that already has a getter (`Permit2Proxy.PERMIT2`, `GasZipPeriphery.GAS_ZIP_ROUTER`, `ReceiverAcrossV4.SPOKEPOOL`, `TokenWrapper.WRAPPED_TOKEN`, `LidoWrapper.ST_ETH`, …) and `readSource` throws, `declared` is undefined, and this test fails with "declares no @custom:version" — while the real gate in `verify-immutable-registry`, which passes the true declarations, passes. Red for a reason unrelated to whether the annotation is valid. It only holds today because GenericSwapFacetV3 is the single annotated entry and happens to live under `src/Facets/`.

Resolving the file across the three gated trees (or reusing the AST declarations the gate itself gets) would keep this test saying what it claims to say.

script/deploy/immutables/getter-coverage.ts
```diff
@@ -185,3 +185,102 @@ export const verifyGetterCoverage = (
...
+
+/** The shape of a version in this repo; anything else cannot be ordered. */
+const SEMANTIC_VERSION = /^\d+\.\d+\.\d+$/
+
+/** The contract's own `@custom:version` tag, which is what a deployed build reports. */
+const extractDeclaredVersion = (source: string): string | undefined =>
+  /^\/\/\/\s*@custom:version\s+(\d+\.\d+\.\d+)/m.exec(source)?.[1]
+
+/** Orders two versions already known to match {@link SEMANTIC_VERSION}. */
+const compareVersions = (left: string, right: string): number => {
```
0xDEnYO: `SEMANTIC_VERSION` + `compareVersions` here duplicate `CONTRACT_VERSION_PATTERN` + `isVersionBelow` in `script/deploy/shared/immutableBindings.ts:340-354`. The gate and the invariant it protects now order versions through two independent implementations.

The gate's job is to reject an annotation the invariant would order as newer than every deployed build — that only works while both order identically. If one side is later extended (a suffixed version like `1.0.2-tron`, a relaxed pattern) the gate passes an annotation the invariant then treats as a fleet-wide skip, which is exactly the silent failure this gate exists to prevent.

Exporting one comparator from `immutableBindings.ts` and consuming it here removes the drift surface.

0xDEnYO (commented):
Reviewed the mechanism end to end. I reproduced the central claim against the real logs — 16 networks skip, 55 stay checked, 0 unresolved — and confirmed `NATIVE_ADDRESS` arrived exactly in v1.0.1 (`1289a316b`), so the annotation is right. The runtime path resolves every unknown toward "keep checking", which is the correct direction. Four comments below; none is a blocker.

script/deploy/immutables/getter-coverage.ts
```diff
@@ -185,3 +185,102 @@ export const verifyGetterCoverage = (
...
+        errors.push(
+          `${where} sets getterSinceVersion '${String(
+            since
+          )}', which is not a quoted major.minor.patch version. The check cannot order it, so it would leave the binding checked and the annotation inert.`
+        )
+        continue
+      }
+
+      const file = fileOf.get(contractName)
+      if (file === undefined) continue
```
0xDEnYO: This is the one branch in the PR that fails open. Everywhere else an unknown resolves to "keep checking"; here an unknown resolves to "stop gating", silently.

An annotated contract the AST enumeration produced no declaration for — an immutable declared in a base contract (declarations are keyed by the contract that declares the variable), or a contract renamed/moved so the deployRequirements key no longer matches — skips the version check entirely. An annotation ahead of the real version, the single failure mode that exempts a binding on every chain at once, would then ship ungated.

Nothing hits it today: all 39 annotated getters are declared in their own file, I checked. But given what this gate is for, an undeclared contract reads more like an error to report than a `continue`.

script/deploy/healthCheckInvariants.ts
```diff
@@ -2229,6 +2218,24 @@ export const HEALTH_CHECK_INVARIANTS: IHealthCheckInvariant[] = [
         // Not present on this chain — nothing to compare.
         if (!address) continue
 
+        // A build from before the getter existed can only revert, and naming that an unverified
+        // binding describes a pending upgrade rather than anything wrong with this chain's
+        // config. The coverage it costs is recovered by the upgrade itself, so this is narrated
+        // rather than reported: the contract is live here, and every other outcome for a live
+        // contract says something, so dropping it from the log entirely is what would confuse.
+        const versionPredatingGetter = liveVersionPredatingGetter(
```
0xDEnYO: The skip replaces the read rather than classifying its failure, which puts the whole binding check on that chain behind a field the repo writes about itself.

`deployments/<network>.diamond.json` `Version` is not chain truth: `updateDiamondJsonBatch` records `getContractVersion(name)` from the working tree at cut time. Cut from a stale checkout (or hand-edit/regenerate an entry) and the live address ends up labelled older than the deployed bytecode — the read is then skipped permanently and real binding drift never surfaces, on an `error`-severity invariant.

Reading anyway and only downgrading the *revert* when the live version predates the annotation costs the same RPC call and cannot misfire in that direction: a revert plus a predating version narrates exactly as it does now, and a build that does answer is still compared.

Address matching mitigates the common case — a version bump lands at a new address, the log entry stops matching, and the binding stays checked — so this is the residual wrong-version-at-the-live-address case rather than the lagging-log one.

gvladika: Agreed, and changed in f46aef0. The read now happens on every chain and the annotation only classifies the failure: a revert from a build the log records as pre-getter narrates through consola, a build that answers is compared either way, and a transient RPC failure is not a revert so it keeps warning whatever the log says.

The revert-shape test is the same one `readBindingValue` uses for the `legacyGetters` fallback, now a shared `ABSENT_FUNCTION_ERROR_PATTERN` rather than a second copy of the regex.

Two tests cover the direction you named: one asserts a build the log calls `1.0.0` that *answers* still produces a binding error, and one asserts a 503 on a pre-getter build still warns. Mutating the revert guard to `true` fails exactly the second, so neither is vacuous.

script/deploy/immutables/getter-coverage.ts
gvladika: Fixed in 2fec86b, with one adjustment to the diagnosis. The base-contract case you named is a legitimate miss rather than an error: declarations are keyed by the contract that declares the variable, so an annotation on a getter backed by an inherited immutable would report even though nothing is wrong.

So it resolves before it reports — the AST's declaring file first, then `src/Facets|Periphery|Security/<contract>.sol` — and only a contract that resolves nowhere (renamed, moved, key typo) pushes an error. That keeps the renamed/moved case you were actually worried about failing closed without turning inheritance into a false positive.

script/deploy/immutables/getter-coverage.ts
gvladika: Fixed in 2fec86b. `immutableBindings.ts` now exports `compareContractVersions` (3-way, null when either side is unorderable) and `isOrderableContractVersion`; `isVersionBelow` is defined in terms of the former, and the gate imports both. `SEMANTIC_VERSION` and the gate's local `compareVersions` are gone, so the `1.0.2-tron` case you raised can only be accepted or rejected by both at once.

script/deploy/immutables/getter-coverage.test.ts
gvladika: Fixed in 2fec86b. The test passes no declarations now and lets the same resolution the gate uses find the file, instead of fabricating `src/Facets/<key>.sol`. Added a `Permit2Proxy` case so a periphery entry under `src/Periphery` is actually exercised, and one for a contract that resolves nowhere.

Confirmed the failure you described was real before the fix: annotating `Permit2Proxy.PERMIT2` went red with "declares no @custom:version" while `verify-immutable-registry` passed on the same annotation.

0xDEnYO (approved):
All four addressed, verified against 2fec86b.

1. Classify rather than skip — the read now runs on every chain and only a revert-shaped failure from a pre-getter build is narrated; transient failures still warn, and a build that answers is compared whatever the log says. The shared `ABSENT_FUNCTION_ERROR_PATTERN` is the same shape `readBindingValue` uses for the `legacyGetters` fallback, so the two cannot drift apart.
2. One comparator — `compareContractVersions` / `isOrderableContractVersion` exported from `immutableBindings.ts`, `isVersionBelow` defined in terms of it, the gate's local copies gone. The `1.0.2-tron` case can now only be accepted or rejected by both at once.
3. Unresolved contract reports — and your adjustment is the right call: keying declarations by the declaring contract makes the inherited-immutable case a legitimate miss, so resolving through the gated trees first and erroring only when nothing resolves keeps the renamed/moved case failing closed without inventing a false positive.
4. Test no longer fabricates paths — passing no declarations exercises the same resolution a periphery entry falls back to, and the `Permit2Proxy` case makes that explicit rather than incidental.

Only remaining gate is `protect-critical-code`: the PR touches `script/deploy/immutables/**`, so it needs an ISM approval (maxklenk or melianessa). Everything else is green.

maxklenk (approved):

PR #2397, opened 2026-09-15
chore(ci): lint all markdown with the repo's pinned engine

gvladika:

Fixes [EXSC-1029](https://linear.app/lifi-linear/issue/EXSC-1029/fix-8-grandfathered-markdownlint-violations-in-claudemd-and-docs)

The ticket asked only for the 8 grandfathered violations to be fixed, on the rationale that they are a latent tripwire for the next PR touching those files. That rationale turned out not to hold: `lint-staged` already runs `markdownlint-cli2 --fix` on staged `*.md` and re-stages, so the violations self-heal at commit time, and markdownlint is not in the repo's required-status-checks ruleset, so a red run does not block a merge today. I also checked all 8 against GitHub's own renderer (`POST /markdown`, gfm mode) — the lists render as proper `<ul>`, and the fence-then-heading case renders as a correct code block plus `<h2>` — so none of them break a render for any reader.

What the fix is actually good for is the thing the ticket did not claim: reaching zero violations is the precondition for enforcing markdown repo-wide. So this PR does the fix and then spends it — CI now lints every Markdown file rather than only the PR's changed files. That drops `tj-actions/changed-files`, the short-circuit step, and `fetch-depth: 0`, since nothing is diffed any more. Without that flip the fix is a one-off cleanup and the debt simply re-accumulates.

The second change is that CI runs `bun lint:md` with the repo's own linter instead of `DavidAnson/markdownlint-cli2-action`. The action's v18 bundles markdownlint-cli2 0.15.0 (markdownlint 0.36.1) while this repo resolved 0.23.2 (markdownlint 0.41.1) — eight minor versions apart, meaning a green local run was never proof of a green CI run, in either direction. I ran the whole repo under both engines and they agree at zero, so the flip is safe today; the point is that they can no longer silently diverge. The exact `0.23.2` pin is a smaller point than the engine swap, and worth stating precisely: CI's determinism already comes from `bun install --frozen-lockfile` resolving against the locked entry, so the pin does not buy CI anything. What it closes is the local path — under `^0.23.0` a lockfile regeneration or a `bun update` can move a developer within the caret range and put them on a different engine than CI, which is the exact failure this PR is otherwise removing. It also matches the exact pins already used for `tsx`, `@mayanfinance/swap-sdk`, `axios` and `chalk`.

Node 22 is pinned explicitly via `setup-node` rather than inherited from the runner image, because markdownlint-cli2 declares `engines.node >= 22` and `bun lint:md` reaches node through the bin shebang — I confirmed this by running it with only Node 18 on `PATH` and reproducing the `SyntaxError: Invalid regular expression flags` crash. `actions/setup-node` is at v7.0.0 upstream, but this repo pins v4.1.0 in its two existing uses and I matched that; a major bump is a repo-wide Dependabot change, not something to fold in here.

Two follow-ups I deliberately left out. Registering `lint all markdown files` as a required check is a branch-protection change rather than a code change, and un-grandfathering has limited teeth until that happens — worth doing, but it needs a repo admin. And `package.json` still declares `engines.node: ">= 12.18.0"` while its own devDependency needs `>= 22`, with no `.nvmrc`; that is the actual root cause behind the Node 18 breakage and belongs to EXSC-1028 (#2396), so folding it in here would collide.

Review follow-ups. `ignores` in `.markdownlint-cli2.jsonc` was root-anchored, so `node_modules/**` matched only the root copy: in a clone using the repo's `.claude/worktrees/` layout, `bun lint:md` linted every nested session checkout and `lint:md:fix` — same glob, but it writes — rewrote markdown inside them, while CI saw 267 files and passed. So the local↔CI parity this PR is built on did not actually hold. Fixed with `**/node_modules/**` and `.claude/worktrees/**`; `lib/`, `out/` and `cache/` stay root-anchored, since `.claude/worktrees/**` already covers nested checkouts and `**/out/**` would later swallow a legitimate `some/out/*.md` (`safe/*/out/` is already tracked). The workflow header also regained the `known limitations:` bullet required by `.agents/rules/500-github-actions.md`.

Last, `.cursor/rules/README.md` was a symlink to `.agents/rules/README.md`, which commit 2639157a2 moved up to `.agents/README.md` without updating the link. globby skips broken symlinks silently, so it was a hole in this workflow's "lint every Markdown file" guarantee rather than a visible failure — the count is unchanged at 267 with it gone. Deleted rather than repointed, since the README covers rules and commands both and that is why it moved out of `rules/`.


.markdownlint-cli2.jsonc
```diff
@@ -37,9 +37,9 @@
...
   //   - `lint-staged` passes individual staged files
-  //   - CI passes only the PR's changed files (see .github/workflows/markdownlint.yml)
-  // markdownlint-cli2 UNIONS config globs with CLI args, so any non-empty `globs` here
-  // would cause CI to lint the whole repo on every PR that touches a .md file.
+  //   - CI runs `bun lint:md`, i.e. the same full-repo glob (see .github/workflows/markdownlint.yml)
+  // markdownlint-cli2 UNIONS config globs with CLI args, so a non-empty `globs` here would
+  // widen every caller's scope — notably lint-staged, which must stay limited to staged files.
   "globs": [],
   "ignores": [
     "node_modules/**",
```
0xDEnYO: `ignores` is root-anchored, so `node_modules/**` and `lib/**` only match at the repo root. That means the parity this PR is built on doesn't actually hold, and the `lint:md:fix` half of it has write side effects.

In a clone using the repo's documented `.claude/worktrees/` layout, the command CI now enforces:

```text
$ bun lint:md
markdownlint-cli2 v0.23.2 (markdownlint v0.41.1)
Linting: 68699 files
Summary: 66 issues in 33 files
error: script "lint:md" exited with code 1
```

58 of those error lines come from `.claude/worktrees/<session>/…`, including nested `node_modules` — e.g. `.claude/worktrees/beautiful-euclid-62c9ea/node_modules/markdownlint/schema/ValidatingConfiguration.md`. CI sees 267 files and passes, because the runner has no nested worktrees. The tool's own `Finding:` line confirms the anchoring: `!node_modules/**` never matches `.claude/worktrees/x/node_modules/**`.

Two consequences:

1. Nobody using the worktree convention can run the new gate to green locally, and the violations are in files they don't own.
2. `lint:md:fix` shares the same glob and *writes*. Running it before committing a `.md` — which is the standard step, and this PR makes it the practical prerequisite for green CI — rewrites markdown inside other sessions' worktrees and inside their `node_modules`.

Both are fixed here:

```jsonc
"ignores": [
  "**/node_modules/**",
  "**/lib/**",
  ".claude/worktrees/**",
  "out/**",
  ...
]
```

Worth re-running `bun lint:md` from a clone that has worktrees checked out to confirm, rather than from a fresh one.

.github/workflows/markdownlint.yml
```diff
@@ -1,50 +1,66 @@
...
+# - lints EVERY Markdown file in the repo against .markdownlint-cli2.jsonc (excluding that
+#   config's own `ignores`), not just the files changed in the PR
+# - runs the repo's own pinned markdownlint-cli2 via `bun lint:md`, so CI and a local
+#   `bun lint:md` execute byte-identical engine + config. The previous
+#   DavidAnson/markdownlint-cli2-action@v18 bundled markdownlint-cli2 0.15.0 while this repo
+#   pins 0.23.2, so a PR could be green locally and red in CI (or the reverse)
+# - markdownlint-cli2 requires Node >= 22 (its `engines` field) and `bun lint:md` shells out
+#   to node via the bin shebang, so the Node version is pinned explicitly rather than
+#   inherited from whatever the runner image happens to ship
+# - hard-fails the job on any violation (heading hierarchy, fence/list spacing, empty links, etc.)
```
0xDEnYO, line 11: The rewrite dropped the `known limitations:` bullet that the old header carried, and `.agents/rules/500-github-actions.md` lists "Known limitations" as a required part of every workflow header block.

There's a real one to record: `lint all markdown files` is not in the `main protection` ruleset's required checks (verified against `repos/lifinance/contracts/rulesets/20990859`, which has validate-json, audit-verification, solc-floor-build-required, protect-critical-code, validate-scripts-required, unit-tests-required, version-control, protect_audit_labels, check-new-network-health, smoke-test-required, ts-tests-required). So a red run does not block a merge today — which is the one thing a reader needs to know about how much teeth the new gate has. The PR body says this; the file doesn't.

0xDEnYO (commented):
The engine swap, the exact pin, the Node 22 pin and the 8 markdown fixes all hold up — I checked the load-bearing claims rather than taking the body at its word:

- `bun lint:md` really does spawn `node` (PATH-shadowed stub confirms it), so the `setup-node` step is load-bearing and the header note about the bin shebang is accurate.
- Full-repo run on this branch is genuinely at 0 (267 files).
- markdownlint really is absent from the required-status-checks ruleset, so the job rename can't strand a check.
- `bun-version-file: package.json` + `bun install --frozen-lockfile` match the other 15 workflows, and every `uses:` is 40-char pinned.

One blocking issue: the local↔CI parity the PR is built on doesn't survive contact with a clone that has worktrees, and the `--fix` variant of the same glob writes into them. Details inline.

Aside, pre-existing and not yours: `.cursor/rules/README.md` is a tracked symlink to a non-existent `.agents/rules/README.md`. globby silently skips it, so "lint EVERY Markdown file" quietly has one hole.

.markdownlint-cli2.jsonc
gvladika: Confirmed and fixed in 6b3606d.

I reproduced the anchoring rather than taking the `Finding:` line at its word — dropped a probe at `.claude/worktrees/probe-session/node_modules/somepkg/bad.md` plus a plain `.md` sibling, and both got linted. After the fix the run is `Linting: 267 files / Summary: 0 issues`, exit 0 — the same 267 CI sees, so parity holds now.

```jsonc
"ignores": [
  "**/node_modules/**",
  ".claude/worktrees/**",
  "lib/**",
  ...
]
```

One deviation from your suggested list: I left `lib/**`, `out/**` and `cache/**` root-anchored instead of making them `**/`. `.claude/worktrees/**` already excludes nested checkouts wholesale, so the `**/` variants only matter for a checkout nested somewhere *else* — and I tested that case: a stray `stray-wt/` still leaks `stray-wt/docs/bad.md` even with `**/lib/**`, so it shaves the class rather than closing it, while `**/out/**` would silently swallow a legitimate `some/out/*.md` later (we already have tracked `safe/*/out/`). `**/node_modules/**` I did keep broad, since node_modules is never lintable content at any depth and it's the bulk of the 68,699.

Good catch on the `--fix` write side effect specifically — that's the half that would have done real damage.

.github/workflows/markdownlint.yml
gvladika: Added in 6b3606d — `.agents/rules/500-github-actions.md:120` does list "Known limitations" as item 4 of the header block, and the rewrite dropped it.

I re-verified the ruleset independently rather than reusing your list, and it matches: `repos/lifinance/contracts/rulesets/20990859` required checks are validate-json, audit-verification, solc-floor-build-required, protect-critical-code, validate-scripts-required, unit-tests-required, version-control, protect_audit_labels, check-new-network-health, smoke-test-required, ts-tests-required — no markdownlint.

```yaml
# - known limitations: `lint all markdown files` is not in the `main protection` ruleset's
#   required status checks, so a red run does not block a merge today — the gate is advisory
#   until a repo admin registers it
```

gvladika:
Both review items are addressed in 6b3606d (details in the two inline threads). Summary:

1. **Blocker — root-anchored `ignores`.** Reproduced, fixed, re-verified: `bun lint:md` from a clone with nested worktrees is now `Linting: 267 files / Summary: 0 issues`, exit 0. Added `**/node_modules/**` and `.claude/worktrees/**`; kept `lib/`/`out/`/`cache/` root-anchored, reasoning in the thread.
2. **Missing `known limitations:` header bullet.** Added, with the required-checks fact the reader needs.

On your aside — the `.cursor/rules/README.md` symlink: I traced it, and it's not just dangling, it's orphaned residue. Commit 2639157a2 (`refactor(agents): move README from .agents/rules/ to .agents/`) moved the target up one level and updated the in-repo references, but left this symlink pointing at the old path. It's the only dangling symlink under `.cursor/`, `.claude/` or `.agents/`, and `.claude` has no equivalent (`.claude/README.md` is its own regular file, not a link), so the right fix is almost certainly to delete it rather than repoint it.

I've deliberately left it out of this PR — it's from an unrelated commit, and this PR is already claiming to be single-purpose. Happy to send it as a separate `trivial`-labelled one-liner, or fold it in here if you'd rather not have the "lint EVERY Markdown file" header claim carry a silent hole in the meantime. Your call.

gvladika:
Correction to my previous comment — the symlink is folded into this PR after all, in 8eb3bd5.

`.cursor/rules/README.md` deleted rather than repointed. Worth noting for the record: the file count is unchanged at 267 with it gone, which confirms your read that globby was skipping it silently rather than counting it — it was a hole in the header's "lint EVERY Markdown file" claim, not a lint failure waiting to happen.

CI on the branch: 38 pass, 3 skipping, 1 fail. The failure is `protect-critical-code` (InfoSec-manager approval, tripped by any `.github/workflows/` change) — it fails identically on the pre-review commit a429bf6f3, so it's the standing gate rather than anything from these fixes.

0xDEnYO (approved):
All three addressed — verified independently rather than on the thread's word.

**Blocker (ignores anchoring):** reproduced on a real checkout of `8eb3bd5` with probes at `.claude/worktrees/probe-session/node_modules/somepkg/bad.md` and `.claude/worktrees/probe-session/docs/bad.md` — both now excluded, and the `Finding:` line carries `!**/node_modules/** !.claude/worktrees/**`. Run is 267 files / 0 issues, matching CI.

The `lib/`/`out/`/`cache/` deviation is the right call, and I can confirm the reasoning empirically: with a planted `stray-wt/docs/bad.md`, the run went to 268 files and the only 3 issues were that file's — so a stray checkout elsewhere leaks its `docs/` regardless, and `**/lib/**` would not have closed it. Not worth the `**/out/**` false-positive risk against tracked `safe/*/out/`.

**Header:** `known limitations:` bullet present and accurate.

**Symlink:** orphan confirmed — the blob points at `../../.agents/rules/README.md`, which doesn't exist at any ref. Deletion is correct.

CI: 37 pass / 0 fail; the 3 skips are the Solidity jobs, correctly inert on a docs/CI PR. `protect-critical-code` does trigger only on `pull_request_review: [submitted]`, so this approval fires it.

maxklenk (approved):

PR #2401, opened 2026-09-16
fix(safe): let a gate that graded nothing say so once, in its own row

0xDEnYO:

[EXSC-1032](https://linear.app/lifi-linear/issue/EXSC-1032/section-2-says-a-gate-stood-down-four-times-and-two-gates-tick-green)

On a facet removal, section 2 said the same thing four times and twice said it wrong.

**Said four times.** Gates G and H printed a full `expected`/`observed` pair under NOT APPLICABLE, where the `expected` half is boilerplate no proposal can fail. Below them a target-state block printed under no heading, stating it had read `origin/main:script/deploy/_targetState.json` — which a removal never opens, because that branch returns before the anchor is read. Below that a block headed `Codehash gate:` — a name the manifest above does not use for any gate — carried one `· NO CLAIM` line.

An n/a row is now one line, the reason it already carries plus its write-up link. `formatTargetStateLines` returns nothing when every finding consulted no anchor, and heads the block `Gate H · Facet version` when it does print. `renderCodehashSignGate` stands down when it compared nothing and refused nothing, because its ledger row now carries that sentence — and in the manifest, which a free-standing block never was. When it has verdicts to show, it heads them `Gate K · Deployed bytecode`.

**Said wrong.** `✓ 0 of 2 calldata addresses resolved to the deployment record` — a green tick on a lookup that never happened. Both addresses were the zero a removal legally carries, so neither was looked up; the denominator now counts what the record was actually asked about, and a proposal that asked it nothing gets its own line. The renderer's docstring already promised exactly this.

And Gate K reported `ok`. `madeNoClaim` covers only a payload with no `diamondCut` in it, so a removal — a cut that installs no code — fell through to `pass` and graded `0 installed address(es) match an attested build`, a green row on a verified counter earned by nothing, directly under a block saying NO CLAIM. Zero targets is now `not-applicable` on `A-LOCAL`.

**Where the two new headings live.** In the module that prints them, because the registry imports those modules and they cannot import it back. `confirm-check-registry.test.ts` holds each to `gateLabel(...)`, and holds the target-state stand-down set against the registry's real `STATUS_MAPPING` rather than a restatement of it. Both agreement tests were checked by breaking the constants and watching them fail.

**Not in this PR.** The calldata address check still carries no ledger row, so it is the one thing in section 2 the manifest does not count — giving it a gate letter is a design call. Detail blocks still print after all buckets rather than under their own gate's row; they now name their gate, and moving them is a larger change, since the PASSED bucket is skipped entirely and a passing gate's detail would vanish with it.


0xDEnYO:
Folded into #2393 — these commits are on `signing2-integration-0914` now.

PR #2402, opened 2026-09-16
fix(ci): read @custom:version through one parser (EXSC-1033)

gvladika:

Fixes [EXSC-1033](https://linear.app/lifi-linear/issue/EXSC-1033/parse-customversion-through-one-validating-parser-six-parsers-three)

Seven places read the `@custom:version` tag and each carried its own regex. Several stopped at `MAJOR.MINOR.PATCH`, so a value they could not parse whole was truncated to its well-formed prefix instead of rejected — `2.1.3.4` read as `2.1.3` — and a suffixed version read as a different version than the workflow looked audit coverage up under. That is load-bearing on `contracts-tron`, where an overlay is `2.1.3-tron`: the workflow finds the overlay's own audit while the gate resolves `2.1.3`, recomputes upstream's audited closure for source that does not carry the Tron bypass, and blocks on a mismatch against an audit that was never about that source. `audit-gate.ts`'s own doc comment already stated the invariant the split was breaking.

The grammar is now defined once per language — `contract-version.ts` and `extract-contract-version.sh` — with both driven from a single table of cases, so drift between them fails a test rather than surfacing months later as a confusing CI red. Two implementations rather than one because the `version-control` job is pure bash and routing it through `bunx tsx` would add a bun install to the cheapest gate on every PR; this mirrors the existing `verify-foundry-version.sh` seam, where one checker already serves CI, the hook and the deploy path. Every reader now goes through one of the two: the version-control workflow, the audit gate, the immutables getter-coverage check, `getContractVersion`, the deployment backfill, `getCurrentContractVersion` in `helperFunctions.sh`, and the build-attestation mint. A value that is not a version is refused rather than truncated, which is stricter than before at both ends.

The suffix grammar deliberately accepts the fork's own revisions — `2.1.3-tron`, and `2.1.3-tron-r2` / `-r3` for a redeploy of the same overlay, as `docs/TronFork.md` specifies. Suffix segments separate on `.` or `-`, and the suffix is lowercase-only so an overlay identity has exactly one spelling. Because a suffix marks a variant of a release rather than a point before or after it, `readContractVersion` also returns the `MAJOR.MINOR.PATCH` it is built on, and the one caller that orders versions (`getterSinceVersion` in `getter-coverage.ts`) orders on that while reporting the full declared string. Ordering on the whole string would hand `compareContractVersions` a value it returns `null` for, silently retiring that guard on the fork. `compareContractVersions` itself is unchanged — `null` means "keep checking", which stays fail-safe.

Verified the build-attestation manifest is unaffected: over the mint's own source dirs the old grammar and the seam agree on all 60 files, so the committed manifest still matches what this checkout builds.


gvladika:
Dispositions for the CodeRabbit findings, in `28d019ae3`:

**`versionControlAndAuditCheck.yml` — fail when the merge-base version is malformed — taken.** Confirmed as fail-open, and worse than the comment states: the fallthrough sets `OLD_VERSION="0.0.0"`, and since every version differs from `0.0.0` the bump check passes, so an audit-relevant change could ship without the bump that step exists to require. Not reachable today (all 139 `src/**` contracts parse), but it is the one path where the old permissiveness survived my change. My comment there justified falling through for an absent or unreadable old file; that reasoning does not extend to a value that *is* present and is not a version. Status 3 now aborts, status 1 and 2 keep the `0.0.0` default.

**`helperFunctions.sh` — report status 1 as an unreadable file — taken.** The existence check and the read are not the same instant, and the empty result was being reported as "is not a version", which points a caller at the tag rather than the file. Separate branch now.

**`extract-contract-version.sh` — whitespace-only tag — already fixed** in `28cd69029`, before this review. Empty `RAW` exits 2, and `/// @custom:version   ` is pinned in the shared case table with an expected `missing`, so both halves agree.

**`manifest-abort.cli.test.ts` — partially taken.** Declining the first half: making `firstVersionedContract` select fixtures without `readContractVersion` means writing a second `@custom:version` grammar into the tree, which is what this PR removes — and that test's own copy of the old regex is precisely what let the original survive the survey. The fixture discovery only picks a contract to act as a test subject; it asserts nothing about versions, so a parser regression surfaces as "no candidate found", not as a false pass.

Taking the second half: the mint's malformed-version refusal is an error path I added in `287a2287c` with no test. Covered now by a test that runs the mint against a source tree of its own containing `2.1.3.4` — the value the old grammar would have truncated to `2.1.3` and filed a build under — and asserts the message names the file and the whole value.

Verified: manifest-abort 2 pass; actionlint on the workflow is 10 findings, identical to `main`'s (1×SC2001, 9×SC2086); shellcheck, prettier clean; `shfmt` drift in `helperFunctions.sh` is pre-existing at lines 975 and 3744, clear of this edit.

0xDEnYO (approved):

melianessa (approved):

PR #2405, opened 2026-09-17
fix(safe): stop vouching for zkEVM immutables nobody read (EXSC-1035)

0xDEnYO:

[EXSC-1035](https://linear.app/lifi-linear/issue/EXSC-1035/sign-time-constructor-parameter-verification-the-residual-after-exsc) (R1)

Stacked on `signing2-integration-0914` (#2393), where the layer-2 pricing call site lives.

zkEVM does not inline immutables — they are written to `ImmutableSimulator` (`0x…8005`) at construction. So a zk deployment reaches layer 1 with `maskedByteCount: 0`, which is **correct** for the code hash (nothing was excluded; those bytes are real codegen) and says nothing at all about the immutable values.

The gate keyed its layer-2 handoff off the masked count alone (`excludedByteCount > 0`), so the whole zk family skipped it. Read on its own, that path renders the unqualified green — indistinguishable from "this contract holds no immutables", the collapse `summarise`'s own comment says must not happen.

**That green is not reachable today, and this PR does not claim to remove a live one.** A zk target never gets that far: the rebuild runner passes `--out`, which the zk build ignores (it writes `zkout/`), and the zk artifact has no `deployedBytecode`, so `createForgeRebuildRunner` throws and the target is already `UNVERIFIABLE`. Verified by compiling with the repo's own `foundry-zksync` using the exact production argv. So zk blocks today for an unrelated reason — a broken rebuild — and this guard is the correct verdict for the right reason, in place before lookup-based attestations (WP-5.2) make the green reachable.

**Why not just wire `readZkImmutables`, which already exists and is tested:** it cannot be honestly wired yet.

- Zk artifacts carry neither `deployedBytecode` nor an `ast` — verified against all 277 artifacts under `zkout/`. `createImmutablePricer` reads `deployedBytecode.object`, so the zk rebuild cannot even produce an `IRebuiltArtifact`.
- With no AST there is no name→ordinal map from the compilation that produced the graded bytecode. Deriving ordinals from source declaration order is exactly what `zk-immutables.ts` refuses to do, and for a real reason: a wrong index compares one immutable against another's expectation, and that **passes**.
- `readImmutableDeclarations` also filters to the contract's own declarations, while the simulator's index space covers inherited ones in linearized order. No base contract declares an immutable today, so this is latent rather than live — but it would shift every ordinal silently the day one does.

So the gate now grades a zk match `UNVERIFIABLE` for the same reason a masked EVM match is grey: nothing was found wrong, it was not looked at. This is fail-closed and forward-looking — a zk deployment passes once the metadata that names its slots exists, which is the prerequisite work tracked on EXSC-1035.

The flag is derived in `deriveToolchainScope` from `isZkEVM`, i.e. from repo configuration, per the rule `ILineageScope` already states: never from the deployed bytecode, which the proposer controls.

Also corrects the comment in `attested-set.ts` claiming the byte-for-byte path compares "immutables included" — true where they are inlined, false on the chains this PR is about.

**Known limitation, deliberately left:** the guard keys on the chain, not on whether the contract declares immutables, so an immutable-free facet on a zk chain is graded `UNVERIFIABLE` with a reason asserting unchecked values that do not exist. Nothing regresses today (see above), but once zk attestations work this is a standing false refusal on every zk cut — and `false-refusal-budget-run.ts:272` excludes zk profiles, so the budget will not catch it. Narrowing it needs the same per-contract immutable metadata that layer 2 itself is blocked on.

**Operational note:** zk signing stays blocked, as it already is. This does not unblock it and does not newly block it.


0xDEnYO:
Folded into #2393 — landed on `signing2-integration-0914` as 709cbc10e (off-code immutables guard) and 8aa0ef179 (zk rebuild fix).

PR #2408, opened 2026-09-17
feat(target-state): declare membership, make the version an optional pin (EXSC-1040)

0xDEnYO:

Fixes [EXSC-1040](https://linear.app/lifi-linear/issue/EXSC-1040/target-state-declare-membership-make-the-version-an-optional-pin)

`_targetState.json` is declared intent, but its version column was maintained alongside the
rollout it describes — so it restated what the deploy log already recorded, at the cost of a
hand-edit per rollout. Measured on `main` @ `8da9776d1`:

- **The version column carried almost no per-chain intent.** Of 51 contracts declared in
  production, exactly **one** (`CalldataVerificationFacet`, a rollout in flight) had more than
  one version across the fleet. Every other contract was uniform — a lagging copy of
  `@custom:version`.
- **It lagged source on ~11% of entries** — 209 of 1854, all "source ahead"
  (`CalldataVerificationFacet` 71, `FeeForwarder` 69, `PolymerCCTPFacet` 22, `EcoFacet` 13, …).
- **The membership column is real per-chain intent** and nothing else in the repo carries it:
  `ArbitrumBridgeFacet` 1 network, `SquidFacet` 10, `AcrossFacetV4` 24, `SymbiosisFacet` 32,
  `GasZipFacet` 65. (`config/squid.json` is not a substitute — 16 networks vs 10 declared.)

So the keys stay and the value changes meaning: **`"latest"`** (follow the repo's
`@custom:version`) by default, or a **semver pin** acting as a hold-back guard: only that base version may be deployed for the entry or installed by a production `diamondCut`. It does not record what is currently deployed. A rollout now touches the file zero times; a version bump that is not being rolled
out yet touches it zero times. Full semantics: [docs/TargetState.md](docs/TargetState.md).

**Zero pins on migration** — no chain today is deliberately held back.

## Why the sign-time gate had to change in the same PR

This is the part worth reviewing hardest. `readDeclaredVersion` returns `"latest"` as a plain
string, so `compareSemanticVersions(proposed, "latest")` returns `null` → `version-not-comparable`
→ **not** in `STATUSES_CLEARED_TO_PROCEED`. Migrating the file without touching Gate H would have
**refused every production `diamondCut` proposal at sign time.** The coupling runs the other way
too: the new deploy-time pin guard, applied to the un-migrated file, would have blocked deploys on
all 209 stale entries, because it correctly reads them as pins. The file's meaning and both readers
have to flip in one commit.

Gate H now resolves the expected version from the contract's `@custom:version` **at the same
verified `origin/main` ref** the target state is read from — reusing `createGitSourceReader` and
the `readContractVersion` parser unified in #2402. The source read deliberately does not take a
tree-ish of its own: it shares `verifyRemoteAndFetch`, so a proposer's clone cannot author the
anchor.

This makes the gate **stronger**, not weaker. Today the 209 stale entries mean a legitimate v2.0.0
proposal reads `ahead-of-main` → cleared: the check that exists to catch a downgrade does not run
on the common case.

| Entry | Expected | Installs it | Installs anything else |
| --- | --- | --- | --- |
| `"latest"` | `@custom:version` at `origin/main` | `matches-main` | older → `downgrade` **blocks**; newer → `ahead-of-main` clears |
| pin | the pinned version | `matches-pin` | `pinned-mismatch` **blocks** |
| absent | — | `not-previously-targeted`, clears (unchanged) | same |
| source deleted at `origin/main` | unresolvable | — | `expected-version-unresolved` **blocks** |

A pin is graded as **base-version equality**, not an ordering — a *newer* version is still not what the pin asked for, while deployment-record suffixes such as `-tron` compare on their base. Both remedies (fix the pin, fix the proposal) are a PR to `main`.

A version bump landing mid-rollout blocks remaining proposals for the previous version. Operators must either finish that rollout before the bump, or pin the remaining networks before the bump and check out the old ref when deploying. The pin is removed once those networks should follow the repo again.

`countNetworksDeclaring` lost its input (nothing declares versions now) and counts **membership**
instead; a version-matched count would report 0 fleet-wide and read as "no corroboration".

## Deploy-time enforcement

The guard lives in `deploySingleContract.sh`, the chokepoint every deploy path funnels through
(`deployFacetAndAddToDiamond`, `deployContractToNetworks`, `deployAllContracts`), so a pin cannot
be bypassed by entering from another script. A pin can only ever **refuse** — `deploySingleContract`
compiles whatever the repo has, so it could never have installed an older build.

`deployPeripheryContracts` previously *warned and skipped* on a version mismatch and returned 0. It now collects refusals and returns non-zero. Stages 2, 5, and 6 in `deployAllContracts` stop the run after a refused deployment instead of continuing into diamond wiring or periphery registration. This also surfaces pre-existing deployment failures that used to be swallowed.

`printDeploymentsStatusV2` now displays `policy : deployed` and resolves `latest` to the source version for coloring; otherwise every migrated row would have appeared as a mismatch.

## Google Sheet ingest retired

`parseTargetStateGoogleSpreadsheet` and its chain regenerated the whole file from a spreadsheet
nobody in CI could read — one run after this migration would have overwritten every `"latest"`
back to a pinned version and unravelled the scheme silently. Removed with `scriptMaster` use case
10, along with two pieces of dead code found on the way (`updateAllContractsToTargetState`, whose
only caller hard-errors "not yet implemented", and the uncalled `printDeploymentsStatus`).
The retired slots disappear from the menu, while every surviving operation keeps its original number so emergency muscle memory remains valid.

## Judgement calls worth a second opinion

- **`removeNetworkFromTargetStateJSON` deleted.** Its only caller was inside the Google Sheet
  ingest; `/deprecate-network` uses plain `jq del`, so it was orphaned.
- **The Product Target State spreadsheet reminders were kept** in `/deprecate-contract` and
  `/deprecate-network` — that is product bookkeeping, not the retired ingest — with a note that
  nothing reads it into the repo.
- **`scriptMaster` use case 7 kept, not deleted.** It is the membership-editing tool now, and
  defaults to writing `"latest"`.


script/deploy/safe/confirm-check-registry.ts
```diff
@@ -252,6 +252,29 @@ const STATUS_MAPPING: Readonly<Record<TargetStateStatus, IStatusMapping>> = {
...
+    anchor: 'A-MAIN',
+    expected: ORDERING_HOLDS,
+  },
+  // A deliberate pin contradicted. A fail, not an acknowledgement: the pin says
+  // this network is held back on purpose, and clicking past it is how a pin stops
+  // meaning anything.
+  'pinned-mismatch': {
+    status: 'fail',
+    anchor: 'A-MAIN',
+    expected: ORDERING_HOLDS,
```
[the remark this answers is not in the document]
0xDEnYO: Fixed in d645077a4. Added a `VERSION_MATCHES_PIN` condition and used it for both `matches-pin` and `pinned-mismatch` — you are right that a pin is graded as equality, so `ORDERING_HOLDS` described an assertion the code does not make.

script/deploy/safe/pinned-target-state.ts
```diff
@@ -551,10 +778,13 @@ export const createTargetStateDeps = (
   network: string,
   options?: {
     readPinnedState?: () => PinnedTargetStateRead
+    readSourceVersion?: (contractName: string) => SourceVersionRead
     cacheRootDir?: string
   }
 ): ITargetStateDeps => ({
   readPinnedState: options?.readPinnedState ?? createPinnedTargetStateReader(),
+  readSourceVersion:
+    options?.readSourceVersion ?? createPinnedSourceVersionReader(),
```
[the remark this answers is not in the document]
0xDEnYO: Fixed in d645077a4. Both reads now share a `createPinnedAnchor` that verifies the remote, fetches once, and resolves `refs/remotes/origin/main` to a commit SHA; the target state and the source version are then read at that SHA rather than at the ref. Covered by a test that pushes a commit to the origin between the two reads and asserts the second still sees the anchored commit (it fails if the anchor is not shared).

script/deploy/safe/pinned-target-state.test.ts
```diff
@@ -403,8 +588,383 @@ describe('formatTargetStateLines', () => {
...
+    // A merge lands on main between the two reads.
+    const author = fs.mkdtempSync(path.join(os.tmpdir(), 'pinned-anchor-push-'))
+    execFileSync('git', ['clone', origin, author], { stdio: 'ignore' })
+    git(author, ['config', 'user.email', 'other@example.com'])
+    git(author, ['config', 'user.name', 'other'])
+    git(author, ['config', 'commit.gpgsign', 'false'])
+    write(author, 'src/Facets/AFacet.sol', contract('2.0.0'))
+    git(author, ['add', '-A'])
+    git(author, ['commit', '-m', 'bump on main'])
+    git(author, ['push', 'origin', 'main'])
```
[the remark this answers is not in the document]
0xDEnYO: Fixed in 97403b34b. Added `seam.fetch()` after the push so this clone's `refs/remotes/origin/main` moves too — you are right that without it the assertion held for the wrong reason. The test now falsifies independently in both directions: it fails when the anchor is not shared, and it fails when the anchor stores the ref name instead of the SHA `revParse` returns. The fallback to the ref name is also gone; `revParse` is now required on the seam and an unresolvable ref refuses rather than degrading silently.

script/deploy/deployAllContracts.sh
```diff
@@ -245,8 +249,16 @@ deployAllContracts() {
...
           echo "[info] No matching entry found in target state file for NETWORK=$NETWORK, ENVIRONMENT=$ENVIRONMENT, CONTRACT=$FACET_NAME >> no deployment needed"
         else
+          # deployFacetAndAddToDiamond resolves an empty version to the repo's current one
+          if [[ "$TARGET_VERSION" == "$TARGET_STATE_VERSION_LATEST" ]]; then
+            TARGET_VERSION=""
+          fi
+
           # deploy facet and add to diamond
-          deployFacetAndAddToDiamond "$NETWORK" "$ENVIRONMENT" "$FACET_NAME" "$DIAMOND_CONTRACT_NAME" "$TARGET_VERSION"
+          if ! deployFacetAndAddToDiamond "$NETWORK" "$ENVIRONMENT" "$FACET_NAME" "$DIAMOND_CONTRACT_NAME" "$TARGET_VERSION"; then
```
gvladika, line 262: **Stage 5 fails fast while stages 2 and 6 collect.**

This `return 1`s on the first facet that fails, but stage 2 (`deployCoreFacets`) and stage 6 (`deployPeripheryContracts`) both collect every refusal and still attempt the rest. Before this PR the loop ignored the result and continued.

On a new-network bootstrap, one facet failing for any reason — `checkDeployRequirements` missing a config address, a transient RPC error, a pin — now stops the run before periphery deploy, periphery registration, whitelist sync, funding and the health check. Each re-run stops again at the next failure, so N broken facets cost N sequential operator cycles.

Either collect refusals here the way the other two stages do, or make the asymmetry deliberate and say why in the comment above.

script/helperFunctions.sh
```diff
@@ -3591,6 +3018,52 @@ function findContractVersionInTargetState() {
...
+  fi
+
+  local CURRENT_VERSION
+  CURRENT_VERSION=$(getCurrentContractVersion "$CONTRACT")
+
+  # Ordering is defined on MAJOR.MINOR.PATCH only, so a suffixed tag (2.1.3-tron) is
+  # compared on its base — the same reduction the sign-time gate applies. Without it a
+  # pin, which is always bare, could never match a suffixed build and would refuse the
+  # deploy it was written to permit.
+  if [[ "$TARGET_VERSION" == "${CURRENT_VERSION%%-*}" ]]; then
```
gvladika: **The build suffix is stripped from only one side.**

`docs/TargetState.md` says suffixes "are compared by their `2.1.3` base in both the deploy guard and the sign-time gate", and the TS gate does apply `baseSemanticVersion` to both operands. Here only `CURRENT_VERSION` is reduced.

A pin written as `2.1.3-tron` would therefore refuse every deploy — including a deploy of `2.1.3-tron` itself — with a message telling the operator to check out the version they already have.

Latent today, since `pinned-target-state.test.ts` rejects a non-bare pin in the committed file. But the guard and the doc disagree, and the doc is the one a future operator will read.

script/deploy/safe/confirm-check-registry.ts
```diff
@@ -252,6 +256,29 @@ const STATUS_MAPPING: Readonly<Record<TargetStateStatus, IStatusMapping>> = {
     anchor: 'A-LOCAL',
     expected: NOTHING_TO_COMPARE,
   },
+  // A pin matched. Anchored A-MAIN rather than A-MONGO: what decided this row is
+  // the version `main` pins, not the deployment record the proposed version came
+  // from.
+  'matches-pin': {
```
gvladika, line 269: **`matches-pin` is anchored `A-MAIN`, but one side of its comparison came from Mongo.**

The proposed version reaches this row through `deps.resolveDeployed` — the proposer-written MongoDB deployment record. The module header states the rule directly:

> Every one that had to resolve the proposed version through the deployment record is `A-MONGO` — including the three that then compared it against `origin/main`, because the proposer writes that record and so owns one side of the comparison.

`matches-main` and `ahead-of-main` follow that rule; `matches-pin` does not. The pinned version is indeed from `main`, but the value it is compared *against* is not.

No security consequence today: the status is `needs-ack`, so `recordCheck`'s pass-on-reporting-anchor coercion never fires. The concern is that the ledger row claims stronger evidence than the check had, and a later change of this status to `pass` would silently inherit the wrong anchor.

gvladika (commented):
Review of the deploy-time guard, the sign-time gate and the migration. Scope verified: the `_targetState.json` key set is byte-identical to `origin/main` (`jq 'paths(scalars)'` diffs empty) and the only distinct value in the new file is `latest`; all declared production contracts still have source with a `@custom:version` tag, so `expected-version-unresolved` cannot fire on the committed file; no dangling references to any of the nine deleted shell functions.

Six comments below — two medium, three low, one nit. Nothing blocking. Findings 4 and 5 are choices the PR body already argues for; they are flagged for operability rather than as defects.

Two things I checked that are **correct** and worth recording, since both are easy to get wrong:

- `local TARGET_VERSION` and its assignment are split across two lines in `assertTargetStateVersionAllowed`, so `$?` is the command substitution's status and not `local`'s. The classic bash trap is avoided.
- The `TARGET_STATE_PATH` pre-check genuinely closes the `exit 1`-inside-`$( )` hole, and `targetStatePinGuard.test.ts` covers both the missing-file and the unset case.

One pre-existing issue, not a regression: `recordedOn: 'network'` at `pinned-target-state.test.ts:721` is not a member of `DeployedContractLookup` and `tsc --strict` flags it (TS2353 reproduced). The identical construct already exists on `main` at line 96, so these test files are evidently not type-checked.

script/deploy/safe/pinned-target-state.ts
```diff
@@ -191,29 +221,77 @@ export const readDeclaredVersion = (
...
+  contractName: string,
+  readSourceVersion: (contractName: string) => SourceVersionRead
+): ExpectedVersion => {
+  const declared = readDeclaredVersion(state, network, contractName)
+  if (declared === null) return { kind: 'absent' }
+  if (declared !== TARGET_STATE_VERSION_LATEST)
+    return { kind: 'pin', version: declared }
+
+  const read = readSourceVersion(contractName)
+  if (read.ok) return { kind: 'latest', version: read.version }
```
gvladika, line 280: **A version bump merged mid-rollout retroactively invalidates every outstanding proposal.**

With every entry now `latest`, this line makes the expected version a moving target tied to `@custom:version` at `origin/main`.

Concrete: a FeeForwarder rollout is in flight, 30 of 69 chains still unsigned. An unrelated PR bumps FeeForwarder's `@custom:version` on `main`. All 30 pending Safe proposals now grade `downgrade` and cannot be signed. Under the pre-PR scheme those chains' declared version matched the cut and cleared.

The documented unblock is to merge 30 pins to `main` *before* the bump — which needs foresight the operator does not have at sign time. `docs/TargetState.md` states the sequencing rule, but there is no in-band override and no warning at propose time, only a hard refusal hours later at the signing ceremony.

Understood as a deliberate tradeoff. Worth considering either a propose-time check that fails early, or an acknowledgement path for "older than `main` but matches this chain's recorded deployment", so the failure surfaces before the signers are assembled.

script/deploy/deployPeripheryContracts.sh
```diff
@@ -38,37 +38,40 @@ deployPeripheryContracts() {
...
+    # check return code of findContractVersionInTargetState
+    if [[ "$RETURN_VALUE" -ne 0 ]]; then
+      # no matching entry found in target state file, no deployment needed
+      echo "[info] contract $CONTRACT not found in target state file > no deployment needed"
+    else
+      # A pin that disagrees with the repo is refused inside deploySingleContract. The
+      # refusal is collected rather than ignored: a contract target state declares but
+      # nothing deployed is exactly what every health check then reports as missing, and
+      # a run that returns 0 hides it.
+      if ! deploySingleContract "$CONTRACT" "$NETWORK" "$ENVIRONMENT" "$CURRENT_VERSION" false "$DIAMOND_CONTRACT_NAME"; then
```
gvladika, line 64: **This widens what a stage-6 re-entry does on a live production chain.**

The old code deployed a periphery contract only when `CURRENT_VERSION == TARGET_VERSION` and warned/skipped otherwise. This now attempts every declared periphery contract unconditionally, with only the pin guard able to refuse — and after the migration there are no pins.

Combined with the migration erasing the 209 "source ahead" entries (including FeeForwarder on 69 chains), a `deployAllContracts` run entered at stage 6 against an *existing production* network will now deploy — and at stage 7 register — periphery that the old code would have skipped.

That is the intended new semantic, and the collect-and-return-non-zero change here is a clear improvement over the silent `return 0`. The part worth a second look is that the widening is silent: consider an explicit confirmation when stage 6 is entered against a network that already has deployments, or a note in the runbook.

script/helperFunctions.sh
```diff
@@ -3591,6 +3018,52 @@ function findContractVersionInTargetState() {
...
+  if [[ $? -ne 0 ]]; then
+    return 0
+  fi
+
+  if [[ "$TARGET_VERSION" == "$TARGET_STATE_VERSION_LATEST" ]]; then
+    return 0
+  fi
+
+  local CURRENT_VERSION
+  CURRENT_VERSION=$(getCurrentContractVersion "$CONTRACT")
```
gvladika, line 3051: Nit: this return value is used without checking its status, and `error()` writes to **stdout**, so a contract whose source is missing or untagged leaves `CURRENT_VERSION` holding the error text.

The guard then fails closed, which is right. But the refusal reads:

> target state pins X to 1.2.0 on optimism/production but this repo is at [error] '@custom:version' string not found in src/Facets/X.sol

which points the operator at the pin instead of at the real fault. A status check here with a distinct message would land them in the right place.

script/deploy/deployAllContracts.sh
0xDEnYO: Collected, matching stages 2 and 6. The loop now appends to `REFUSED_FACETS` and stage 5 reports the whole set and returns 1 after the loop, so the run still stops before stage 6.

`targetStatePinGuard.test.ts` asserts it on the loop body rather than on the report — my first version of that test asserted the report text and passed with the `return 1` put back inside the loop. (2a029b5, 342bcea)

script/helperFunctions.sh
0xDEnYO: Fixed. The status is checked immediately after the assignment and refuses with its own message naming the failed read, quoting what the read reported rather than presenting it as the repo version:

`cannot read the current version of X, so the target-state pin 1.2.0 on optimism/production cannot be checked. Deploy blocked. The version read reported: [error] '@custom:version' string not found in src/Facets/X.sol`

Covered by a test that drives the version read to fail the way the real one does — on stdout, non-zero — and asserts the refusal does not reach the pin message. (a8b7fb8)

script/helperFunctions.sh
0xDEnYO: Fixed — both sides are reduced now: `"${TARGET_VERSION%%-*}" == "${CURRENT_VERSION%%-*}"`, which is what `docs/TargetState.md` describes and what `baseSemanticVersion` does on both operands in the TS gate.

Two tests: a suffixed pin matches both `2.1.3-tron` and `2.1.3`, and still refuses `2.2.0-tron`. Reverting the pin-side reduction fails the first one. (a8b7fb8)

script/deploy/safe/confirm-check-registry.ts
0xDEnYO: Changed to `A-MONGO`. The module header rule is about which side had to be resolved through the deployment record, and the proposed side here did, so the previous comment was arguing from the wrong side of the comparison. It stays `needs-ack`, and the test now pins the anchor as `A-MONGO`.

`downgrade` and `pinned-mismatch` resolve the proposed version the same way, so they are moving to `A-MONGO` too — the anchor is meant to name the weakest evidence a row rests on, and keeping two exemptions means a comment explaining why the rule does not apply to them, which is exactly the comment that turned out to be wrong here. (2a029b5, plus a follow-up commit)

script/deploy/deployPeripheryContracts.sh
0xDEnYO: Documented rather than prompted. `deployAllContracts` runs non-interactively in the deploy smoke test, so a confirmation on stage-6 entry would hang CI unless it is gated on a TTY, and a prompt that only appears sometimes is a weak guard.

`docs/TargetState.md` now has a "Re-entering a deploy run against a live network" note under **Who reads it**, saying stages 5 and 6 attempt the whole declared set whatever is deployed, that the pre-migration skip is gone, and that the start stage is a deliberate choice rather than a restart convenience. If you would rather have the prompt as well, say so and I will add it behind a TTY check. (2a029b5)

script/deploy/safe/pinned-target-state.ts
0xDEnYO: Agreed on the tradeoff, and no change here. The propose-time check is EXSC-1041 — blocked on #2393, which is where the gate runs early enough to refuse before signers are assembled.

Correcting myself on the acknowledgement path: as stated ("older than main but matches this chain's recorded deployment") it would not cover the mid-rollout case, which is the case that motivated it. A rollout proposal installs the version being rolled out — newer than what the chain has recorded, older than main — so it matches neither side and would still grade `downgrade`. It is a real gap, but it needs its own shape; tracking it separately rather than folding it into EXSC-1041.

script/deploy/safe/confirm-check-registry.ts
0xDEnYO: Follow-up: `downgrade` and `pinned-mismatch` are now `A-MONGO` as well, so the rule holds without exception, and the reason lives in the module header rather than in a per-status comment — a per-status justification is what went wrong here in the first place.

`A-MAIN` now has no producer in the target-state mapping. I left it in the ledger vocabulary: `A-AUDIT` has none either, so the set is deliberately wider than today's producers, and retiring an anchor on behalf of one check would be the larger claim. A test asserts no status reaches `A-MAIN`, so reintroducing one is loud. (cf62da5)

gvladika (approved):

PR #2409, opened 2026-09-17
fix(tron): plan facet upgrade cuts against the live loupe

gvladika:

Fixes [EXSC-1045](https://linear.app/lifi-linear/issue/EXSC-1045/tron-facet-upgrades-propose-an-add-only-diamondcut-reverts-after-the)

`proposeDiamondCut` encoded every Tron facet registration as a single **Add** entry covering all of the new facet's selectors. That is right for a first registration and wrong for an upgrade, which is what all six `script/deploy/tron/deploy-and-register-*.ts` scripts do on a redeploy: `LibDiamond` reverts an Add for a selector the diamond already routes (`FunctionAlreadyExists`, `src/Libraries/LibDiamond.sol:164`), and selectors the new version dropped are never removed, so the superseded facet stays callable. Both failures land *after* the 3h timelock delay has already elapsed — the proposal is signed, scheduled, waited out, and only then fails on `executeBatch`.

The fix plans the cut against what the diamond routes today rather than assuming an empty diamond. A single `facets()` loupe read gives the whole routing table. The outgoing facet is resolved by name from `deployments/<network>.diamond.json` — the diamond log is the only place a facet's *identity* survives a redeploy, since on-chain a diamond knows addresses and selectors but never names — and its live selectors come out of that same table. The new facet's selectors then split into Add (unrouted), Replace (served by the outgoing facet) and Remove (served by the outgoing facet, gone from the new version), emitted as one `diamondCut`, so the upgrade stays a single timelock operation. A facet with no diamond-log entry reduces to the plain Add cut, unchanged.

Three things are refused at proposal time, where they are cheap to resolve, instead of after the delay. An Add whose selector the routing table already attributes to some *other* facet — the cut would revert at execution, and silently taking the selector over would strand that facet's remaining selectors, so the message names the selector and the holder. A diamond-log entry the loupe routes nothing to that is not the address being installed: it no longer identifies what the upgrade supersedes, and planning from it yields precisely the add-only cut this PR exists to remove. And an empty routing table, which means the read failed rather than that the diamond is bare — a live diamond always routes the loupe itself.

A successful Tron proposal now records the facet in `deployments/<network>.diamond.json`, the way the periphery path already does (`deploy-and-register-periphery.ts:1449`). Nothing else writes a facet entry here — `deploy-core-facets.ts` and `register-facets-to-diamond.ts` are the only other writers and neither runs in this flow — so without it the log keeps naming the superseded address once the first upgrade executes, the loupe routes nothing there, and every upgrade after that is refused until someone hand-edits a tracked deployment log.

Because the entry records the proposal and not its execution, a proposal that is rejected or expires comes back with the log naming the address being installed and the loupe routing nothing to it. That is the cut that never landed, not a log the chain disagrees with, so it is matched before the dead-entry refusal and plans as a plain Add. Where the cut did partly land, the Replace entries are dropped instead: `LibDiamond.replaceFunctions` reverts `FunctionAlreadyExists` when the selector already points at the facet the Replace names, and those selectors are already where the cut wants them, so only the ones that never landed are proposed.

`getFacetAddressFromDiamondLog` reads across the deployment roots, which include the parent workspace, so a file that is not a diamond log is skipped rather than believed or allowed to block a proposal: the first root holding one in the shape the writers produce answers, entry or not, and if no root holds one the read refuses rather than report the facet as unrecorded.

This is wired into `proposeDiamondCut` itself rather than kept as a standalone script, so the six `deploy-and-register-*.ts` scripts inherit it with no edits and cannot take the Add-only path again. `register-facets-to-diamond.ts` is a separate batch path that still builds its own Add-only cut and is untouched here: it skips a facet only when the loupe already routes that exact address, so registering a new address for a facet the diamond already routes still reverts `FunctionAlreadyExists`, and the bare `catch` around its `facets()` read turns an RPC failure into "not registered". That is tracked separately in [EXSC-1070](https://linear.app/lifi-linear/issue/EXSC-1070/register-facets-to-diamondts-still-builds-an-add-only-cut-on-the-batch), not this PR. The branch is on `isTronNetworkKey`, so the EVM path is byte-for-byte unchanged — EVM stale-selector removal is the separate, deliberately-unstarted EXSC-428 investigation, and nothing here pre-empts it. The loupe read lives in its own module behind a transport seam and is imported dynamically, matching the existing dynamic proposer imports, so an EVM cut never pulls in the Tron deploy stack and the reader stays testable without a network or a `mock.module` that would leak across the suite. The decisions above sit in the pure planner (`facet-upgrade-cut.ts`) rather than in the encoder, so they are unit-tested without one — including the layer between the loupe read and the arithmetic (`indexFacetRouting` / `resolveOutgoingFacet` / `holderResolver`), exercised with the real Tron address codec and a temporary diamond log so the base58 ↔ hex keying is pinned rather than assumed.

Verified against the live Tron diamond `TU3ymitEKCWQFtASkEeHaPb8NfZcJtCHLt` with the EcoFacet v2.0.0 deploy that surfaced this (EXSC-757): the encoded calldata decodes to Add `0xbff90b61`, `0x762aea18`; Replace `0x0ff754ea`; Remove `0x7e56b7b0`, `0x9e75aa95` — the pre-signature-verification v1.1.0 bridge entry points that the old path would have left routable. `GardenFacet`, which is not registered on Tron, still produces a single Add entry.


script/utils/utils.ts
```diff
@@ -686,6 +687,44 @@ export async function updateDiamondJson(
...
+    if (relativePath.startsWith('..') || isAbsolute(relativePath))
+      throw new Error(`Invalid network name: ${network}`)
+
+    let facets: Record<string, { Name?: string }>
+    try {
+      const parsed = JSON.parse(await readFile(diamondJsonPath, 'utf8'))
+      facets = parsed?.LiFiDiamond?.Facets ?? {}
+    } catch {
+      // Absent or unparseable at this root — try the next one.
+      continue
```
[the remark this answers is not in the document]
gvladika: Fixed in 0c56fe70d — good catch.

The case that matters is the one you named: when the new selectors miss the old ones entirely, the collision guard has nothing to trip on, the Add cut succeeds, and the superseded facet stays routable. That is precisely the failure this planner exists to prevent, so reading a corrupt log as "no log" reintroduced it silently.

Took the `ENOENT`-only `continue`, and report an unparseable log with its path so the fix is obvious from the message. Tests updated: the log-is-absent case still resolves to `null` (a genuine first registration), an unparseable log now throws, and a new case covers a read failure that is not a missing file (`EISDIR`).

gvladika:
> Add a Tron integration test for `proposeDiamondCut`. (`propose-diamond-cut.ts:142-268`)

Partially taken, in `a47395586`.

**What I did.** The collision guard was the only logic in that wiring with no test, so I moved it into the pure planner as `assertAddsAreUnrouted(add, facetName, outgoingLabel, resolveHolder)` behind an injected resolver, and covered it: the exact message, the early exit on the first collision (so a long Add list does not fan out reads after it already knows it will refuse), the replace-only case that reads nothing, and a checksummed holder against the lowercased zero comparison. `encodeFacetUpgradeCutCalldata` now reads as plan → assert → build.

**What I did not do**, and why a full `proposeDiamondCut` integration test is the wrong trade here:

1. It would have to mock `getFacetSelectors`, because `ts-unit-tests.yml` runs `bun test:ts` with no `forge build` step — `out/` does not exist in that job, so the real selector read cannot run in CI.
2. Mocking it means `mock.module('../../utils/utils')`, which is global for a whole `bun test` run, not scoped to one file. That already bit this branch once: an earlier draft mocked `./tronUtils` and took down 12 unrelated tron suites. The transport seam in `diamond-loupe-reads.ts` exists specifically so the loupe readers could be tested without it.

So the wiring that stays uncovered is the glue — the dynamic import, the env/codec lookups, and the `runPropose` handoff — not the arithmetic or the refusal. I verified that glue end to end against the live Tron diamond instead, by decoding the calldata the deploy script actually produced: Add `0xbff90b61`, `0x762aea18`; Replace `0x0ff754ea`; Remove `0x7e56b7b0`, `0x9e75aa95`, init `0x0`. Happy to reconsider if we ever get a scoped module mock.

script/deploy/tron/facet-upgrade-cut.test.ts
```diff
@@ -0,0 +1,202 @@
...
+      reads += 1
+      return ZERO as Address
+    })
+
+    expect(reads).toBe(0)
+  })
+
+  // The loupe returns a checksummed address; the zero comparison is lowercased.
+  it('treats a checksummed holder as a collision', async () => {
+    const checksummed = '0xAaBbCcDdEeFf00112233445566778899AaBbCcDd' as Address
```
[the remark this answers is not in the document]
gvladika: Correct, and worth stating plainly: that literal was hand-alternated, not computed. The canonical EIP-55 form of `0xaabbccddeeff00112233445566778899aabbccdd` is `0xAAbbCCDdeEFf00112233445566778899aABBcCDd`, which is not what I wrote — so the test name claimed more than the fixture delivered.

Fixed in 92c1299f9 with `getAddress`, which also keeps it honest if anyone edits the address. The assertion still exercises what it is for: the derived form has uppercase hex, so it pins the lowercased comparison in `assertAddsAreUnrouted`.

script/deploy/tron/diamond-loupe-reads.ts
```diff
@@ -0,0 +1,60 @@
...
+    rpcUrl
+  )
+  const parsed = parseTroncastFacetsOutput(output)
+
+  // The parser returns [] both for a diamond that routes nothing — which a live
+  // one never does, it routes the loupe itself — and for output whose shape it
+  // could not match. Reading the second as the first plans every selector as an
+  // Add and drops every Remove, which is the failure this planner exists to
+  // prevent, so an empty table is refused rather than believed.
+  if (parsed.length === 0)
```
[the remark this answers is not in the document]
gvladika: Confirmed and fixed in `4a80d4631`. I reproduced it before changing anything — `parseTroncastFacetsOutput` on `[[TG65… [0x0ff754ea]] [TVof… [zzzz]]]` returns a single row, and the emptiness check I had waved that through.

You are right that the short table is the more dangerous shape. Missing rows never reach `holderOf`, so the dropped facet's selectors resolve to the zero address, clear `assertAddsAreUnrouted`, and revert `FunctionAlreadyExists` after the 3h delay — the same failure this PR exists to remove, just reached from the other side.

I put the check at the `readFacetRouting` boundary rather than in the parser: `parseTroncastFacetsOutput` is shared with `healthCheckInvariants.ts`, where a lenient parse is the established behaviour and tightening it would be a separate change with its own blast radius. Every `[T…` row the output opens must now appear in the parsed table, and the error names the addresses that did not. Regression test covers exactly the one-valid-plus-one-malformed case.

script/utils/utils.ts
```diff
@@ -686,6 +687,71 @@ export async function updateDiamondJson(
...
+    try {
+      parsed = JSON.parse(contents)
+    } catch (error) {
+      throw new Error(
+        `Could not parse ${diamondJsonPath}: ${(error as Error).message}`
+      )
+    }
+
+    const facets = parsed?.LiFiDiamond?.Facets
+    if (!facets || typeof facets !== 'object')
```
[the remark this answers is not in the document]
gvladika: Taken, in `4a80d4631` — `typeof [] === 'object'`, so the check I added one commit earlier had a hole the same size as the one it was closing: an array-valued `Facets` clears it, `Object.entries([])` yields nothing, and the Tron caller reads that as a first registration and drops the removals.

Applied the suggested guard and added a test for the array shape. I also reworded the message from "Facets section" to "Facets object", since the shape is now part of what is being asserted.

gvladika:
Re the outside-diff comment on `parseTroncastFacetsOutput.ts:8-22` (review [5245417968](https://github.com/lifinance/contracts/pull/2409#pullrequestreview-5245417968)) — mostly right, and fixed in `6e59f0984`.

**Confirmed.** The dropped-row check I added in `4a80d4631` only saw rows it could recognise, which left two shapes reading as a complete table. I ran them before changing anything:

| output | before |
|---|---|
| `[[A [0x0ff754ea]] [B [zzzz]]]` | refused |
| `[[A [0x0ff754ea]] [TVofq5iFi` (truncated mid-address) | **accepted, 1 row** |
| `[[A [0x0ff754ea]] [B [0x8da5` (truncated mid-selector-list) | refused |
| `[[A [0x0ff754ea]]] error: connection reset` | **accepted, 1 row** |

All four are refused now. The region between the first `[[` and the last `]` is matched against an anchored grammar, so the table has to be rows end to end, and anything printed after it is refused rather than read past. The row count is then compared against what the parser returned, which pins the two grammars together instead of letting them drift.

**Kept out of the shared parser**, per your earlier point: lenient scanning is right for `healthCheckInvariants.ts`, which tolerates a partial read, and wrong here, where a short table plans an Add for a selector the diamond already routes. The strictness belongs at the boundary that cannot tolerate it.

**Declined: the independent second loupe read.** For rows the transport omits with no syntax error, a second `facets()` read has the same failure mode as the first — agreement between two reads of the same endpoint is not evidence of completeness, and disagreement only tells you one of them is wrong. It would double the subprocess cost of every upgrade to detect nothing this check does not already cover. If that case is ever worth closing it needs a different oracle (the diamond log, or a `facetAddresses()` cross-check), not a repeat of the same call — and that is its own change, not this PR.

script/deploy/tron/facet-upgrade-cut.ts
```diff
@@ -0,0 +1,188 @@
...
+  facetAddressHex: Address,
+  outgoing: IOutgoingFacet | null
+): ISelectorPlan {
+  if (!outgoing) return planSelectorCuts(newSelectors, [])
+
+  // Believing the log here would plan every selector as an Add and drop every
+  // Remove — the superseded facet stays routable, which is precisely the bug
+  // this planner exists to prevent. Nothing updates the log on a Tron upgrade,
+  // so a second upgrade through this path is exactly when it happens.
+  if (outgoing.registered.length === 0)
```
0xDEnYO, line 193: This throw runs before the same-address check on line 118, so the re-proposal case the PR description calls out — "a Safe proposal that was rejected or expired" — never reaches its own branch.

Concretely: EcoFacet v2 is deployed, the diamond log is updated to the new address at propose time (the convention `deploy-and-register-periphery.ts:1449` already follows — "Record the pending registration in tron.diamond.json"), and the proposal is then rejected. On the re-run `outgoing` resolves to the new address, `registered` is empty because the cut never executed, and this throws. The right plan there is `planSelectorCuts(newSelectors, [])` — a plain Add — since line 118 would have identified it as the same address.

Worse, the message tells the operator to repoint the log entry, i.e. to rewrite a committed deployment log *back* to a stale address.

Moving the `outgoing.addressHex === facetAddressHex` check above this guard fixes it: same address + nothing routed = first landing of that address.

script/deploy/tron/facet-upgrade-cut.ts
```diff
@@ -0,0 +1,188 @@
...
+  facetName: string,
+  newSelectors: readonly string[],
+  facetAddressHex: Address,
+  outgoing: IOutgoingFacet | null
+): ISelectorPlan {
+  if (!outgoing) return planSelectorCuts(newSelectors, [])
+
+  // Believing the log here would plan every selector as an Add and drop every
+  // Remove — the superseded facet stays routable, which is precisely the bug
+  // this planner exists to prevent. Nothing updates the log on a Tron upgrade,
```
0xDEnYO: "Nothing updates the log on a Tron upgrade" is accurate, and that's what makes this guard fire on *every* upgrade after the first rather than only on the pathological case.

The only writers of facet entries are `deploy-core-facets.ts:347/371` and `register-facets-to-diamond.ts:287` — neither runs in this flow — plus the manual `updateDiamondLog.ts` CLI. So once this PR's first EcoFacet upgrade executes, `tron.diamond.json` still names the v1 address, the loupe routes nothing there, and the next `deploy-and-register-eco-facet.ts` run is refused before it can plan anything. The prescribed fix is a hand-edit of a tracked deployment log.

Calling `updateDiamondJson(facetAddress, facetName, …)` after a successful propose — exactly what the periphery path does — would close the loop and leave this guard for the case it's actually written for.

script/utils/utils.ts
```diff
@@ -686,6 +687,73 @@ export async function updateDiamondJson(
...
+        `Could not parse ${diamondJsonPath}: ${(error as Error).message}`
+      )
+    }
+
+    // Arrays are the same failure wearing an object's typeof: every writer here
+    // keys Facets by address, so a list shape reads as "nothing recorded".
+    const facets = parsed?.LiFiDiamond?.Facets
+    if (!facets || typeof facets !== 'object' || Array.isArray(facets))
+      throw new Error(
+        `${diamondJsonPath} has no LiFiDiamond.Facets object — the log is malformed, not empty`
```
0xDEnYO: This throws on the *first* root whose `<network>.diamond.json` exists but lacks `LiFiDiamond.Facets`, rather than moving on to the next one.

`getDeploymentRoots()` returns `[cwd, cwd/contracts, cwd/..]`. Run from a parent workspace where a sibling checkout's `deployments/tron.diamond.json` has a different shape (or is `{}`), and the upgrade aborts with "the log is malformed, not empty" even though the authoritative log at the next root is fine.

The strictness is right for the root that owns the log — it just shouldn't let an unrelated file block a proposal. `continue` on a non-owning root, or resolve the owning root first and be strict only there.

script/deploy/shared/propose-diamond-cut.ts
```diff
@@ -95,14 +109,136 @@ export async function encodeDiamondCutCalldata(
...
+  const holderOf = new Map<string, Address>()
+  const selectorsOf = new Map<Address, Hex[]>()
+  for (const entry of routing) {
+    const facetHex = toHex(entry.facet)
+    selectorsOf.set(facetHex, entry.selectors)
+    for (const selector of entry.selectors)
+      holderOf.set(selector.toLowerCase(), facetHex)
+  }
+
+  const outgoingBase58 = await getFacetAddressFromDiamondLog(network, facetName)
```
0xDEnYO, line 170: This function is the one part of the change with no test. The 24 new tests cover the pure planner and the loupe reader; the wiring where they meet — `toHex` on base58, the `holderOf`/`selectorsOf` keying, and resolving `outgoing` from the log — isn't exercised anywhere.

That's the layer most likely to hold a silent bug: a map keyed on checksummed hex while lookups lowercase, or `selectorsOf.get(outgoingHex)` missing because `toHex` and the loupe disagree on address form, both produce an under-specified cut that passes every existing test and fails after the 3h delay. The reader already has a transport seam and the planner is pure, so a test that injects a recorded routing table plus a temp diamond log would cover it without a network.

0xDEnYO (commented):
Reviewed the cut arithmetic against `LibDiamond` — `addFunctions` / `replaceFunctions` / `removeFunctions` all line up with what `buildFacetCuts` emits, the Replace-drop on re-proposal is correct (`replaceFunctions` does revert when `oldFacetAddress == _facetAddress`), and the loupe parser's strictness is well-judged. Tests pass and typecheck is clean.

Four inline comments; the first is the only one I'd hold the PR on. One more that has no diff line to hang off:

**`script/deploy/tron/register-facets-to-diamond.ts:156` — the Add-only path is still reachable.** The PR body says a Tron facet upgrade can't take it again, but `registerFacetsBatch` still pushes `action: 0` with the full artifact selector set. It skips a facet only when the loupe already routes *that exact address* (lines 134-148), so registering a new address for a facet the diamond already routes — an upgrade — falls straight through to the Add-only cut and reverts `FunctionAlreadyExists` on the first unchanged selector. The bare `catch (e) {}` at line 142 makes it worse: a failed `facets()` read is read as "not registered", so an RPC blip yields a full Add cut against a populated diamond. Fine to leave for a follow-up, but the claim in the description needs narrowing to the `proposeDiamondCut` callers.

script/deploy/tron/facet-upgrade-cut.ts
gvladika: Confirmed, and fixed in `8ac2b8223`.

Reproduced the ordering: with the log naming the address being installed and nothing routed to it, `planFacetUpgrade` reached the dead-entry throw and never got to the same-address branch. The message then asked for the one edit that must not happen — repointing a committed deployment log back to a superseded address.

The same-address check now runs first. Nothing routed to an address the log names as *this* deployment means the cut never executed, which is the plain Add; a log the chain disagrees with is only the case where the recorded address is some *other* facet. Regression test `plans a plain add when the recorded address never landed` pins it.

Worth stating the boundary this leaves, since you'll hit it before I do: if a proposal is rejected and the facet is then redeployed to a *different* address, the log names the address that never landed and the loupe routes nothing there, so the dead-entry guard fires — correctly, since the entry no longer identifies the outgoing facet, and its message ("point that entry at the address the loupe actually serves") is the right instruction there. Inferring the outgoing facet from which single facet holds the colliding selectors would close it, but that is inference over a routing table, and I'd rather refuse.

script/deploy/tron/facet-upgrade-cut.ts
gvladika: Confirmed, and closed in `033283fb2`.

Checked the writers: `updateDiamondJson*` is called from `deploy-core-facets.ts:347/371`, `register-facets-to-diamond.ts:287` and `deploy-and-register-periphery.ts:1450` — nothing in the six `deploy-and-register-*-facet.ts` flows. So as written, this PR's first EcoFacet upgrade would have executed and left the log naming v1, and the next run would have been refused before planning anything. The guard was firing on the ordinary case, not the pathological one.

`proposeDiamondCut` now calls `updateDiamondJson` after a successful `runPropose` on the Tron branch, so all six callers inherit it, matching where the rest of this change lives. The address is converted back to base58 with `evmHexToTronBase58` per `[CONV:TRON-ADDRESS]`, and the version is left to `updateDiamondJson`'s own `getContractVersion` lookup.

Two properties of this worth naming:

- The entry records the **proposal**, not the execution — same as the periphery path. That is what makes the previous comment's re-proposal case reachable, which is why the two fixes ship together.
- `updateDiamondJson` swallows its own failures (`consola.error`, no throw). Left as is: the proposal has already gone out at that point, so turning a log-write failure into a thrown error would report a successful propose as a failed one. The operator sees the error and can run `updateDiamondLog.ts`.

script/utils/utils.ts
gvladika: Right, and fixed in `5080cb07d`.

The strictness was aimed at "the log exists but is not a log", and I applied it to every root rather than to the one that owns the file. A parent-workspace `deployments/tron.diamond.json` — or a bare `{}` — in front of the checkout that actually holds the log blocked the proposal.

Took the `continue` form. A file that is not a diamond log is skipped, and the first root holding one in the shape the writers produce answers — entry or not, so a facet it does not carry is a first registration rather than a cue to keep looking in a sibling checkout, whose entry would name a facet this diamond never routed. That matches how `pickDeploymentRootForWrites` already picks the owning root.

The fail-closed property is kept where it matters: if *no* root produced a usable log while at least one file existed, the read throws and names each file it could not use, so "unrecorded" still cannot be the answer to a malformed log. Two tests cover the split — `reads past a root whose file is not a diamond log` and `does not consult a later root once a log answers`.

script/deploy/shared/propose-diamond-cut.ts
gvladika: Taken, in `8ac2b8223`. You picked the right layer — this was the gap my earlier "I verified it by hand against the live diamond" answer left open, and hand-verification does not survive the next edit.

The map building and the outgoing lookup moved out of the encoder into three exported functions in the pure planner:

- `indexFacetRouting(routing, toHex)` — both views of the table, every key lowercased inside the function rather than at the lookups, so the failure you named (a map keyed on checksummed hex, read lowercased) cannot be reintroduced at one call site only.
- `resolveOutgoingFacet(outgoingBase58, index, toHex)` — the log's address paired with what it still serves, i.e. the `selectorsOf.get(outgoingHex)` you flagged.
- `holderResolver(index)` — the resolver `assertAddsAreUnrouted` consumes.

`encodeFacetUpgradeCutCalldata` is now those three calls plus the plan, so nothing keyed or resolved lives in the untested function.

On the fixtures: the tests use the **real** Tron codec (`getTronWebCodecOnlyForNetwork('tron')` + `tronAddressToHex`) with real base58 addresses, not a stub — it is offline, so no network and no `mock.module`. `toHex` is deliberately injected as `getAddress(toHex(base58))` in one case so a checksummed codec result is exercised against the lowercase keys. The end-to-end case writes a temporary `deployments/tron.diamond.json`, resolves the outgoing facet through `getFacetAddressFromDiamondLog`, and asserts the full cut is the one this PR verified on-chain: Add `0xbff90b61`/`0x762aea18`, Replace `0x0ff754ea`, Remove `0x7e56b7b0`/`0x9e75aa95`. A third facet in the routing table covers the collision refusal through the same seam.

What is still uncovered is the transport and env lookups either side — `readFacetRouting`'s dynamic import, `getFacetSelectors` (no `out/` in the `bun test:ts` job), `getEnvVar`/codec construction, and the `runPropose` handoff. Everything between them is now pinned.

gvladika:
Re the fifth point in review [5250470155](https://github.com/lifinance/contracts/pull/2409#pullrequestreview-5250470155) — `register-facets-to-diamond.ts:156`, the one with no diff line to hang off.

Confirmed on both halves, and the claim in the description was wrong.

`registerFacetsBatch` skips a facet only when the loupe already routes *that exact address* (lines 133-142), so a new address for a facet the diamond already routes falls through to `action: 0` with the full artifact selector set and reverts `FunctionAlreadyExists` on the first unchanged selector — the same failure this PR fixes, reached by the other door. And the bare `catch (e) {}` at 143-145 reads a failed `facets()` as "not registered", so an RPC blip turns into a full Add cut against a populated diamond, which is worse than the upgrade case because nothing about it looks wrong at proposal time.

The description now says the six `deploy-and-register-*.ts` scripts inherit the fix, names `register-facets-to-diamond.ts` as untouched, and describes both problems, rather than claiming the Add-only path is gone.

Leaving the fix for a follow-up as you suggested — the batch path has its own energy estimation and guarded-send wiring around the cut, so retrofitting the planner there is not a two-line change and does not belong in a PR that is otherwise `proposeDiamondCut`-only.

The four inline comments are addressed in `5080cb07d`, `8ac2b8223` and `033283fb2`; replies on each thread. `bun test:ts` is green (4937 pass, 0 fail), and `bunx eslint` / `bunx tsc-files --noEmit` are clean on every touched file.

gvladika:
Follow-up filed: [EXSC-1070](https://linear.app/lifi-linear/issue/EXSC-1070/register-facets-to-diamondts-still-builds-an-add-only-cut-on-the-batch) — covers both halves (the Add-only batch cut and the bare `catch` around `facets()`), with the acceptance criteria written against reusing `readFacetRouting` / `planFacetUpgrade` rather than re-deriving the arithmetic, and a note that the batch shape wants one routing read for the whole batch. Linked from the PR description.

0xDEnYO (approved):

PR #2413, opened 2026-09-18
fix(deploy): select the london build by profile, not by rewriting foundry.toml (EXSC-1050)

0xDEnYO:

Fixes EXSC-1050

Production deploys of london-EVM networks through the grouped path were refused by the tree-reproducibility guard. `updateFoundryTomlForGroup` selected the london build by `sed`-rewriting `solc_version`/`evm_version` inside `[profile.default]`, and `assertTreeRecordable` lists `foundry.toml` as build-affecting, so every london wave ran against a modified `foundry.toml` and `deploySingleContract` refused before broadcasting (`1 build-affecting path(s) differ from the commit ... foundry.toml`). The cancun group only passed because its rewrite was a no-op against the committed default.

The london build is now selected through the environment. `prepareGroupBuild` exports `FOUNDRY_PROFILE=solc_floor` for the london group and `unset`s it for the cancun and zkEVM groups, and nothing touches `foundry.toml`. Workers inherit the exporting shell's environment (`launchDeployWave` backgrounds `deployToNetworkWorker` from the same shell), so their `forge script` runs compile under the same profile. `getSolcVersion`/`getEvmVersion` read the active profile and fall back to `[profile.default]` for keys it leaves unset, which is how forge resolves them, so a fuse record still says 0.8.17/london: the pair the sign-time codehash rebuild is matched against.

`[profile.solc_floor]` keeps its name and loses only its `skip`, which moves onto the floor workflow's command line so the deploy runners get `script/**` compiled under the profile they deploy with. The name itself is no longer load-bearing: `resolveCheckoutProfile` matches a deployment commit's profile by its **compiler pair**, not by name, and a case in `codehash-sign-gate-deps.test.ts` renames `solc_floor` to `london` in the checkout and asserts the rebuild still resolves it. What the pair-matching does constrain is duplication — two non-zk profiles pinning 0.8.17/london make the resolution ambiguous and the rebuild refuses — so that is what the comment in `foundry.toml` now says. The zk profile is still matched by name, because its zksolc pin attaches by name in `parseBuildProfiles`.

A profile that exists in neither form is still the hazard the section-exists check is for: forge answers an unknown `FOUNDRY_PROFILE` with `[profile.default]`, a warning and exit 0.

```text
$ FOUNDRY_PROFILE=nosuchprofile forge config
Warning: Selected profile `nosuchprofile` does not exist; falling back to the default profile.
evm_version = "cancun"   solc = "0.8.29"
```

So the same silent fallback is now refused rather than trusted: `prepareGroupBuild` checks the section exists before exporting it, and a new case pins that the name the runners export is the one `deriveToolchainScope` resolves for a london network.

Falsification, via `script/deploy/london-profile-selection.test.ts` (a throwaway git checkout with `forge` stubbed to record the profile it was handed, so nothing can broadcast):

- restore the `sed` rewrite on the london branch: the `git status --porcelain -- foundry.toml` assertion fails (2 cases)
- make the getter ignore `FOUNDRY_PROFILE`: the london record case fails (1 case)
- drop the `unset` on the cancun branch: both leak cases fail (2 cases)
- drop the `unset` on the zkEVM branch: the zkEVM leak case fails
- drop the section-exists check: the refusal case fails
- rename the section and follow it in the helper: the suite fails at module load

Left as is, deliberately:

- `script/utils/utils.ts` `getFoundryDefault*`: the only non-test caller is `deploy-safe.ts`, and only as a fallback when `networks.json` carries no `targetEvmVersion`; it does not run inside a grouped wave. Not made profile-aware.
- Nothing else: the single-network path is covered too. `deploySingleContract.sh` calls `selectFoundryProfileForNetwork` ahead of both the compiler-pair guard and `forge script`, so a direct `fuse` deploy in a clean shell builds and records london. An earlier revision of this description said that half of EXSC-1050 was out of scope; it landed in `392f376a`.

The deploy runners need `script/**` compiled under `[profile.solc_floor]`, so the floor workflow keeps it in scope and `script/**/*.sol` is added to its path filter — a 0.8.18+ feature there would otherwise break production deploys with `solc-floor-build` green. The skips are now spelled identically on both sides: `prepareGroupBuild` and `ensureStandardArtifactForSalt` build with `--skip 'test/**'`, the same argument the workflow passes, so nothing a london deploy compiles is outside the gate. The glob matters — `--skip` filters by name and `test` is an alias for `.t.sol`, which is why `ensureStandardArtifactForSalt`'s old `--skip test` still compiled the 27 non-test helpers under `test/solidity/utils`. `scriptMaster.sh`'s `COMPILE_ON_STARTUP` build carries the skip too. The one build left unskipped is `forge script`, which compiles only its target's dependency closure (198 files for `DeployAccessManagerFacet.s.sol`, no `test/**`).

Look hardest at `getFoundryProfileValue` in `script/helperFunctions.sh` (the awk section reader) and at whether the section-exists refusal in `prepareGroupBuild` belongs before or after `assertFoundryVersionOrFail`.

# Review round (Goran, 2026-09-21)

- **zkEVM salt derived from london bytecode.** `ensureStandardArtifactForSalt` only tested that `out/<C>.sol/<C>.json` existed, so a `london -> zkevm` target list with no cancun network between them derived the zk CREATE2 salt — and the deployed address — from london bytecode, with `FOUNDRY_PROFILE` correctly cleared throughout. Fixed at the function rather than at the call sites, per the review: it now compares the artifact's `metadata.compiler.version` / `metadata.settings.evmVersion` against the active profile's pin and rebuilds on a mismatch. Confirmed firing on this checkout's real `out/` after a `solc_floor` build: rc=1 under the default profile, rc=0 under `solc_floor`.
- **`FOUNDRY_PROFILE=zksync` passed the pair guard for a cancun network.** `[profile.zksync]` pins the default's cancun and 0.8.29, so only its `script`/`out` keys differ and the `evm_version` comparison could not see it. Refused by profile name now.
- **An empty `targetEvmVersion` inherited the previous network's profile.** Now cleared — scoped to a profile an earlier network in the loop selected, not the unconditional clear the zkEVM arm does, so an operator-exported profile still reaches `localanvil` uncompared. That is the only row with an empty version and the smoke test runs in a clean shell, so nothing production-addressed goes through it.
- **`foundry.toml` rename paragraph** — superseded by `resolveCheckoutProfile`; restated above and in the file as the duplicate-pair constraint that actually remains.
- **Floor gate stopped covering what the london deploy compiles** — `--skip 'script/**'` dropped, path filter widened. Round 2: the `test/**` half of the same finding is closed from the deploy side instead — both london builds now pass the gate's `--skip 'test/**'`, so the scopes match rather than the gate growing to cover test artifacts no deploy needs. Seven cases in `london-profile-selection.test.ts` hold the two sides together: three compare the argv each build records (strict group build, tolerant group build, salt build) against the skip read out of the workflow, three assert each build site spells the glob **quoted** in its own source, and one pins the job's `FOUNDRY_PROFILE` and path filter — without those last four the invariant is green when the quotes are dropped (the shell would then expand the glob before forge sees it), when the tolerant branch alone is changed, or when the job stops building under the floor profile at all. Each case was falsified by mutating exactly the thing it claims to catch.
- **`diamondRemovalDiff.ts:84` comment** — the `helperFunctions.sh` export it cited is gone; the `FOUNDRY_PROFILE: 'default'` pin is now load-bearing against `prepareGroupBuild`'s `solc_floor` export instead, and says so.

Tests: `bun test script/deploy/london-profile-selection.test.ts` 52 pass (was 31) — covering the artifact-pair check, the zk-profile refusal, the empty-row clearing, and the build-scope parity above.


script/deploy/deploySingleContract.sh
```diff
@@ -100,6 +100,17 @@ deploySingleContract() {
     fi
   fi
 
+  # The record's compiler pair is what the sign-time rebuild compiles under; an empty
+  # one cannot be rebuilt. Checked before the build for the same reason.
+  if ! getSolcVersion "$NETWORK" >/dev/null || ! getEvmVersion "$NETWORK" >/dev/null; then
```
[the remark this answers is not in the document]
0xDEnYO: Fixed in 392f376ad: deploySingleContract now calls selectFoundryProfileForNetwork before the compiler-pair guard, which exports solc_floor for a london network when the shell holds no profile, leaves cancun on the default, and refuses an exported profile whose evm_version is not the network's (driven end to end in london-profile-selection.test.ts against the forge stub).

script/deploy/safe/codehash-sign-gate-deps.ts
```diff
@@ -337,6 +338,77 @@ const assertSubmodulesPinned = (
...
+    )
+  }
+
+  const matching = Object.values(profiles).filter(
+    (candidate) =>
+      candidate.profile !== ZK_PROFILE &&
+      candidate.zksolcVersion === undefined &&
+      candidate.solcVersion === requested.solcVersion &&
+      candidate.evmVersion === requested.evmVersion
+  )
```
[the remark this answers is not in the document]
0xDEnYO: Fixed in 579a32068: parseBuildProfiles takes an inheritFromDefault option that completes a profile declaring exactly one of solc_version/evm_version from [profile.default] (a profile declaring neither stays excluded), and resolveCheckoutProfile matches with it on; tests cover both one-key shapes, a non-matching inherited pair and an ambiguous pair.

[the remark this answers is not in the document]
0xDEnYO: Correction: the inheritance fallback (579a32068) is reverted in b95b87c57 by maintainer decision — not on this PR. Across all 195 revisions of foundry.toml on main the only non-default profiles have been ci, solc_floor and zksync, none of which pins one key without the other, so the described gap cannot fire on any real deployment commit; and a fallback that also completed a no-key profile such as ci would turn every default-pair rebuild into an ambiguous-match refusal. Kept as a known limitation of resolveCheckoutProfile.

gvladika (commented):
Reviewed the profile-selection change end-to-end. The direction is right — the conflict between the `sed` rewrite and `assertTreeRecordable` is real, selecting through `FOUNDRY_PROFILE` is the correct resolution, and the old helpers are removed rather than shimmed. The test suite is unusually good: driving the real bash helpers in a throwaway checkout and asserting `git status --porcelain -- foundry.toml` stays empty pins the actual invariant, not a proxy for it. Scoping the `zksync` profile in `verifyContract` with `env` instead of `export` fixes a genuine leak on the way past.

Six findings inline plus one that has no line in the diff. The two I'd want resolved before merge are the zkEVM salt one and the `FOUNDRY_PROFILE=zksync` guard hole.

**Not inline — `script/deploy/safe/diamondRemovalDiff.ts:84-85` (unchanged by this PR, so no diff line to anchor to).** The comment there still says *"`helperFunctions.sh` exports `zksync` without unsetting it, so a propose script can inherit one"* — this PR removed that export. The `FOUNDRY_PROFILE: 'default'` pin is still correct and is now **more** load-bearing, not less, since `prepareGroupBuild` deliberately exports `solc_floor` into the environment propose workers inherit. Worth updating the reason while the context is fresh.

---

**What I ran against `4706360`:**

| check | result |
|---|---|
| `bun test script/deploy/london-profile-selection.test.ts` | 31 pass, 0 fail |
| `bun test` codehash-sign-gate-deps + assert-zk-toolchain + lineage-scope | 127 pass, 0 fail |
| `shellcheck -S warning script/deploy/resources/deployGroupingHelpers.sh` | clean |
| `FOUNDRY_PROFILE=solc_floor forge build` (no skips, all trees) | exit 0, warnings only — confirms the `skip`-removal claim in the description |
| `forge build --help` | `--skip <SKIP>...` is variadic; the repeated-flag form on the workflow line is valid |
| bash: does `V=x func` persist after return? | no (bash 5.3, non-posix) — so `FOUNDRY_PROFILE=zksync getSolcVersion` is correctly scoped |

**CI:** `protect-critical-code` is a missing security-manager approval, not a code defect. `validate-json` fails on a duplicate `EcoFacet` key at `immutableRegistry.json:122` — that arrived with `b2d9f7cf2` on `main` and was already fixed there by #2422 (`61d7eada9`); this branch is behind `main`, so merging `main` clears it.

script/deploy/resources/deployGroupingHelpers.sh
```diff
@@ -173,101 +166,173 @@ function groupNetworksByExecutionGroup() {
...
-            # Build with new solc version (Foundry will detect if recompilation is needed)
-            logWithTimestamp "Running forge build for Cancun EVM group..."
-            if [[ "$STRICT" == "true" ]]; then
-                forge build || { error "forge build failed for $group group"; return 1; }
-            else
-                forge build || true
-            fi
+            # zk path ensures it per contract (ensureStandardArtifactForSalt) with a plain `forge
+            # build`, so a profile left over from an earlier group would decide that salt's bytecode.
+            unset FOUNDRY_PROFILE
```
gvladika, line 344: **[medium] This guard doesn't guard what the comment above it claims.**

The comment says the `unset` matters because "a profile left over from an earlier group would decide that salt's bytecode" via `ensureStandardArtifactForSalt`'s plain `forge build`. That build never runs in the case being defended against: `ensureStandardArtifactForSalt` (`script/helperFunctions.sh:1829`) returns early when `out/<C>.sol/<C>.json` already exists — and the london wave's `forge build` wrote exactly that file. Confirmed on this branch after a `solc_floor` build:

```console
$ jq -r '.metadata.compiler.version, .metadata.settings.evmVersion' out/AccessManagerFacet.sol/AccessManagerFacet.json
0.8.17+commit.8df45f5f
london
```

So a target list of **london + zkevm with no cancun networks** derives the zk CREATE2 salt — and with it the deployed address — from london bytecode, with `FOUNDRY_PROFILE` correctly cleared the whole time. Mixed lists hide it because the cancun wave overwrites `out/` in between, which is why the wave order `london → cancun → zkevm` in `deployContractToNetworks.sh` masks it in the common case.

This is pre-existing — the `sed` path shared the same `out/` — so it isn't a regression from this PR. But the new call sites in `deployContractToNetworks.sh` (~L390) and `proposeContractToNetworks.sh` (~L442) both carry a comment asserting coverage that isn't there, which is worse than no comment.

Two ways out: run `forge build --skip test` on this arm with the profile cleared, or have `ensureStandardArtifactForSalt` compare the artifact's `metadata.compiler.version` / `metadata.settings.evmVersion` against the active profile instead of testing only for the file's existence. The second closes it for every caller, not just the grouped waves.

script/deploy/resources/deployGroupingHelpers.sh
```diff
@@ -173,101 +166,173 @@ function groupNetworksByExecutionGroup() {
...
+        if [[ "$TARGET_EVM_VERSION" == "$GROUP_LONDON" ]]; then
+            assertLondonProfileDeclared || return 1
+            export FOUNDRY_PROFILE="$PROFILE_LONDON"
+        fi
+        FOUNDRY_PROFILE_SELECTED_FOR_NETWORK="$NETWORK"
+    fi
+
+    local ACTIVE_EVM_VERSION
+    ACTIVE_EVM_VERSION=$(getEvmVersion "$NETWORK") || return 1
+    if [[ "$ACTIVE_EVM_VERSION" != "$TARGET_EVM_VERSION" ]]; then
```
gvladika, line 282: **[medium] `FOUNDRY_PROFILE=zksync` passes this guard for a non-zk cancun network.**

The comparison is on `evm_version` alone, and `[profile.zksync]` also declares `cancun`. Verified against this branch:

```console
== operator exported FOUNDRY_PROFILE=zksync, deploying to cancun network 'base' ==
ACCEPTED -> FOUNDRY_PROFILE=zksync
  solc it would build with: 0.8.29
  out dir it would write:   out/zksync
```

`[profile.zksync]` also sets `script = "script/deploy/zksync"` and `test = "notest"`, so the deploy would look for its script in the zk directory and read/write the wrong artifact tree — while this guard reports the profile as fitting.

No repo code reaches it today (this PR removed the one `export FOUNDRY_PROFILE=zksync` that leaked), but "the operator's shell" is precisely the case this branch exists for, and a zk deploy is the most likely thing to have been run in it just before. Comparing the full pair doesn't help — `zksync` pins 0.8.29 too — so the check has to be on the profile name: refuse when `FOUNDRY_PROFILE` is the zk profile and `isZkEvmNetwork "$NETWORK"` is false.

script/deploy/resources/deployGroupingHelpers.sh
```diff
@@ -173,101 +166,173 @@ function groupNetworksByExecutionGroup() {
...
+    # statement, so it is refused instead of skipping the comparison below.
+    if ! jq -e --arg network "$NETWORK" '.[$network] | has("targetEvmVersion")' "$NETWORKS_JSON_FILE_PATH" > /dev/null; then
+        error "Network '$NETWORK' has no targetEvmVersion in networks.json - refusing to deploy"
+        return 1
+    fi
+
+    local TARGET_EVM_VERSION
+    TARGET_EVM_VERSION=$(jq -r --arg network "$NETWORK" '.[$network].targetEvmVersion' "$NETWORKS_JSON_FILE_PATH")
+
+    if [[ -z "$TARGET_EVM_VERSION" ]]; then
```
gvladika, line 257: **[low] A row with an empty `targetEvmVersion` inherits the previous network's profile.**

This early `return 0` skips the comparison, as documented — but it also skips clearing, so in a `scriptMaster`-style loop the profile survives:

```console
== localanvil (empty targetEvmVersion) after a london network ==
-> FOUNDRY_PROFILE after localanvil=solc_floor
```

The zkEVM branch a few lines up clears for exactly this reason, which makes this the asymmetric one. The comment's reasoning — "states nothing the active profile can contradict" — justifies skipping the *comparison*, but not inheriting a london build silently. Impact is confined to `localanvil` and the deploy smoke test runs in a clean shell, so this is low; symmetry with the zkEVM branch costs one line.

foundry.toml
```diff
@@ -34,23 +34,24 @@ runs = 32
...
-# - optimizer settings are inherited from [profile.default] so the check matches the
-#   codegen that production actually ships (no via_ir).
-# - test/script are skipped: only deployed contracts must honor the floor, and those
-#   suites trip the legacy pipeline's "stack too deep" (a codegen limit, not a
-#   floor-incompatibility signal) which would mask the real check.
+# - Everything else is inherited from [profile.default], so deploys and the floor check use
+#   the codegen production actually ships (no via_ir). No `skip` here: the deploy runners
+#   need script/** compiled under this profile, so the floor workflow passes its skips on
+#   the command line instead.
+# - Do not rename this section. The sign-time rebuild resolves a network's profile by name
```
gvladika: **[low] This paragraph, and the matching one in the PR description, describe code that a later commit on this same branch superseded.**

It says a rename makes forge fall back to `[profile.default]` at older commits and grade every london slot MISMATCH. That was true of `27d0987c0`'s parent — and `27d0987c0` ("resolve the rebuild profile against the checkout's foundry.toml") is what fixed it. `resolveCheckoutProfile` now matches by compiler pair, and this branch's own test renames `solc_floor` → `london` and asserts the rebuild still resolves it (`codehash-sign-gate-deps.test.ts`, "exports the name the checkout gives the pair when it differs from HEAD").

So the stated reason no longer holds, and a future reader following it would conclude the rebuild is more fragile than it is. Per `[CONV:COMMENTS]` the comment should state the constraint that actually remains — `parseBuildProfiles` refuses a renamed **zk** profile because the zksolc pin attaches by name, and two profiles pinning the same pair make `resolveCheckoutProfile` ambiguous — or be dropped.

The PR description's third paragraph ("The name is load-bearing beyond this file and is not mine to change here…", with the `FOUNDRY_PROFILE=london forge config` transcript) needs the same correction before merge; it reads as the rationale for the design and no longer matches the branch tip.

.github/workflows/solc-floor-build.yml
```diff
@@ -92,12 +92,11 @@ jobs:
...
       - name: Build src/ with floor compiler (solc 0.8.17)
-        # The `solc_floor` profile pins solc 0.8.17 + evm london and skips test/script.
-        # A failure here means a src/ contract uses a language/syntax feature newer than
-        # the `^0.8.17` floor it declares.
+        # The `solc_floor` profile pins solc 0.8.17 + evm london. A failure here means a src/
+        # contract uses a language/syntax feature newer than the `^0.8.17` floor it declares.
         env:
           FOUNDRY_PROFILE: solc_floor
-        run: forge build
+        run: forge build --skip 'test/**' --skip 'script/**'
```
gvladika: **[low] The floor gate no longer covers what the london deploy compiles.**

CI skips `test/**` and `script/**` here; the london deploy runs a bare `forge build` under the same profile, so it compiles both. A `script/**` file using a 0.8.18+ feature therefore breaks london production deploys with a green `solc-floor-build`.

Same gap as before this PR — the profile's `skip` covered both sides equally — so not a regression. But the stated reason for moving the skips onto the command line is "the deploy runners need `script/**` compiled under this profile", and that is exactly the argument for CI checking it. Right now the one tree the deploy newly depends on is the one CI stopped compiling.

Dropping `--skip 'script/**'` is free today: I ran the full build on this branch and `FOUNDRY_PROFILE=solc_floor forge build` with no skips at all exits 0 (warnings only, all 471 files), which also independently confirms the description's claim.

script/deploy/resources/deployGroupingHelpers.sh
0xDEnYO: Fixed at the function, per your second option. \`ensureStandardArtifactForSalt\` now compares the artifact's \`metadata.compiler.version\` / \`metadata.settings.evmVersion\` against the active profile's pin and rebuilds on a mismatch, so it closes for every caller rather than only the grouped waves. Confirmed firing on this checkout's real \`out/\` after a \`solc_floor\` build: rc=1 under the default profile, rc=0 under \`solc_floor\`. The two call-site comments that claimed coverage now describe what the function actually does.

script/deploy/resources/deployGroupingHelpers.sh
0xDEnYO: Refused by profile name now, as you suggested — the pair comparison cannot see it. It sits after the network-exists check so a mistyped network is still diagnosed as a typo.

script/deploy/resources/deployGroupingHelpers.sh
0xDEnYO: Cleared, symmetric with the zkEVM arm.

foundry.toml
0xDEnYO: Right — restated as the constraint that remains: a rename is safe because \`resolveCheckoutProfile\` matches by compiler pair, but a second non-zk profile pinning the same pair is ambiguous and refuses the rebuild. The PR description's third paragraph is corrected the same way.

.github/workflows/solc-floor-build.yml
0xDEnYO: \`--skip 'script/**'\` dropped, and \`script/**/*.sol\` added to the path filter so a script-only change still runs the gate. Your bare-build result reproduces here: exit 0, warnings only.

.github/workflows/solc-floor-build.yml
```diff
@@ -91,13 +92,12 @@ jobs:
...
-        # The `solc_floor` profile pins solc 0.8.17 + evm london and skips test/script.
-        # A failure here means a src/ contract uses a language/syntax feature newer than
-        # the `^0.8.17` floor it declares.
+      - name: Build with floor compiler (solc 0.8.17)
+        # The `solc_floor` profile pins solc 0.8.17 + evm london. A failure here means a file
+        # uses a language/syntax feature newer than the `^0.8.17` floor it declares.
         env:
           FOUNDRY_PROFILE: solc_floor
-        run: forge build
+        run: forge build --skip 'test/**'
```
gvladika, line 103: **[medium] The `test/**` half of the round-1 finding is still open, and the header comment now argues against itself.**

The header says *"script/ is in scope because the london deploy runs a bare `forge build` under this profile; only test/ is skipped"*. The same bare `forge build` compiles `test/**` too — `prepareGroupBuild` (`script/deploy/resources/deployGroupingHelpers.sh:355-359`) passes no skips, and the profile carries none:

```console
$ FOUNDRY_PROFILE=solc_floor forge config --json | jq '{test,skip}'
{ "test": "test/solidity", "skip": [] }
```

It is wider than one tree, because `--skip test` is not a directory filter:

```console
$ forge build --help
  --skip <SKIP>...
      Skip building files whose names contain the given filter.
      `test` and `script` are aliases for `.t.sol` and `.s.sol`.
```

So the three scopes are all different:

| build | compiles under `solc_floor` |
|---|---|
| london wave — bare `forge build` | every `.t.sol` **and** the 27 non-`.t.sol` helpers in `test/solidity/utils/` |
| `ensureStandardArtifactForSalt` — `forge build --skip test` | those 27 helpers (only `.t.sol` is skipped) — so on a direct london deploy too |
| this job — `forge build --skip 'test/**'` | none of them |

Confirmed on this branch after a full floor build:

```console
$ jq -r '.metadata.compiler.version, .metadata.settings.evmVersion' out/TestBase.sol/TestBase.json
0.8.17+commit.8df45f5f
london
```

`test/**` is also absent from the `dorny/paths-filter` list above, so a change to `test/solidity/utils/TestBase.sol` does not even trigger this job. A 0.8.18+ feature there breaks london production deploys with `solc-floor-build` green — the same argument that moved `script/**` into scope.

Cheapest fix is on the deploy side, and it makes the two scopes match exactly: have `prepareGroupBuild` build with `--skip 'test/**'` and leave this job as it is. The deploy has no use for test artifacts, and `ensureStandardArtifactForSalt`'s `--skip test` wants the same treatment. Note that swapping this job to `--skip test` instead would *not* be equivalent, given the alias.

The other direction also works and costs nothing: drop the skip here entirely. `FOUNDRY_PROFILE=solc_floor forge build` with no skips exits 0 on this branch (warnings only), same result as last round.

script/deploy/deploySingleContract.sh
```diff
@@ -100,6 +101,27 @@ deploySingleContract() {
     fi
   fi
 
+  # The grouped runners export the profile before launching their workers; a direct
+  # call inherits whatever the shell holds, and forge and the getters below both read it.
+  if ! selectFoundryProfileForNetwork "$NETWORK"; then
```
gvladika, line 106: **[medium] This line contradicts the PR description's scope statement — please correct the description.**

Under *"Left as is, deliberately"* the description says:

> The single-network path (`deploySingleContract.sh` invoked directly, outside a grouped wave) still builds and records under whatever profile the operator's shell carries, so a lone london network such as `fuse` still gets a cancun build. That is the other half of EXSC-1050 and is not fixed here; this PR is the grouped path only.

This call is that path, and it is ahead of both the compiler-pair guard below and the `forge script` run, so a direct `fuse` deploy in a clean shell exports `solc_floor` and both builds and records london. It arrived with `392f376a` ("select the london profile from the network on a direct deploy"), which closed the CodeRabbit finding on this same line — the bullet predates it and was not updated.

Same class as the `foundry.toml` paragraph from the last round: this section is what a reviewer and the audit read as the scope of the change, and as written it hands a follow-up ticket a problem that no longer exists. Either state that the direct path is now covered, or drop the bullet.

gvladika (commented):
Re-reviewed at `78807b0`. Both blockers from the last round are properly closed; two things are left before merge, inline.

| # (last round) | item | status |
|---|---|---|
| 1 | zkEVM salt derived from london bytecode | fixed |
| 2 | `FOUNDRY_PROFILE=zksync` passes the pair guard | fixed |
| 3 | empty `targetEvmVersion` inherits the previous profile | fixed |
| 4 | stale `foundry.toml` rename paragraph | fixed |
| 5 | floor gate vs. what the london deploy compiles | **half fixed** — inline |
| 6 | `diamondRemovalDiff.ts:84` comment | fixed |
| — | PR description scope bullet | **new, inline** |

On (1): `standardArtifactMatchesActiveProfile` is the right shape — it strips solc's `+commit.` suffix before comparing, treats an unreadable or metadata-less artifact as "rebuild", and re-checks after the rebuild so a failed build cannot fall through to a wrong salt. Refusing any profile that redirects `out` is the generalisation that catches `[profile.zksync]`, which the pair alone cannot since it pins the default's 0.8.29/cancun. I also checked the two things the new rebuild could have broken: both zkEVM waves run `launchDeployWave 1`, so it cannot fan out into parallel `forge build`s sharing `out/`, and the standard rebuild writes `out/` while the zk build writes `out/zksync` + `zkcache`, so they do not clobber each other.

On (3): scoping the clear to `FOUNDRY_PROFILE_SELECTED_FOR_NETWORK` rather than copying the zkEVM arm's unconditional clear is a fair narrowing of what I asked for — an operator's own export still reaches `localanvil` uncompared, which is consistent with "a row that states nothing cannot contradict anything".

The new tests are good: the artifact sandbox plants a compiler pair and asserts on the recorded `forge build` invocations rather than on internals, and "still accepts `zksync` for a zkEVM network" is a real negative control for the name-based refusal.

---

**Follow-ups, not blocking this PR.**

**[low] Every guard added here reads `foundry.toml`, so `FOUNDRY_*` key overrides walk past all of them.** `getFoundryProfileValue` awks the file; forge resolves these above the profile:

```console
$ FOUNDRY_EVM_VERSION=paris forge config   -> evm_version = "paris"
$ FOUNDRY_SOLC=0.8.26       forge config   -> solc = "0.8.26"
$ FOUNDRY_OUT=out2          forge config   -> out = "out2"
```

An operator shell carrying `FOUNDRY_EVM_VERSION=london` deploying to a cancun network: `selectFoundryProfileForNetwork` reads `cancun` from the toml, matches `targetEvmVersion`, accepts — and forge ships london bytecode with a record saying cancun, which is the MISMATCH class this PR exists to prevent. Pre-existing (the `sed` path lost to env vars too), but the guards added here are aimed squarely at "the operator's shell must not silently decide a production build", and this is the hole left in that model. A refusal on `FOUNDRY_EVM_VERSION`/`FOUNDRY_SOLC`/`FOUNDRY_OUT` (and the `DAPP_` twins) closes it cheaply; reading `forge config` instead of the toml closes it properly but costs a forge invocation before `assertFoundryVersionOrFail`.

**[note, not a defect] Two salt derivations move addresses.** `DEPLOYSALT = keccak(bytecode || SALT)` and the address is CREATE3 of it, so the profile decides the address. Two cases change: a `london -> zkevm` list previously derived zk salts from london bytecode and now derives them from the default profile's, and a direct london deploy previously derived from cancun and now derives from london. Both are the fix working as intended, and `deploySingleContract.sh:393` treats a changed derived salt as "safe to deploy to a different address" rather than as an error — so if any live deployment was made through either path, the next deploy of that same version lands at a new address instead of being recognised as already deployed. Worth one query against the deployment log before this rolls out.

**Nit.** `PROFILE_ZKSYNC` is now a constant in `deployGroupingHelpers.sh:37`, but `'zksync'` stays hardcoded in `helperFunctions.sh:667` (`getSolcVersion`) and `deploySingleContract.sh:338,470`. Four spellings of a name whose whole point in this PR is that it is matched by name rather than by pair.

---

**What I ran against `78807b0`:**

| check | result |
|---|---|
| `bun test script/deploy/london-profile-selection.test.ts` | 43 pass, 0 fail (was 31) |
| `bun test` codehash-sign-gate-deps + assert-zk-toolchain + assert-foundry-version + `script/deploy/codehash` | 618 pass, 0 fail |
| `shellcheck -S style` on the new `helperFunctions.sh` range and on `deployGroupingHelpers.sh` | clean (pre-existing SC2034/SC2155 elsewhere in `helperFunctions.sh` only) |
| `actionlint .github/workflows/solc-floor-build.yml` | rc=0 |
| `FOUNDRY_PROFILE=solc_floor forge build` (no skips, all trees) | exit 0, warnings only |
| `FOUNDRY_PROFILE=solc_floor forge config` | `skip: []`, `test = test/solidity`, `out = out` |
| `forge build --help` | `test`/`script` are aliases for `.t.sol`/`.s.sol`, not directory filters |
| `gh pr checks 2413` | all green; `validate-json` clears now that `main` is merged in |

.github/workflows/solc-floor-build.yml
0xDEnYO: Closed from the deploy side, as you suggested: `prepareGroupBuild` and `ensureStandardArtifactForSalt` both build with `--skip 'test/**'` now, so the two scopes are identical and the gate keeps its cheap build. The deploy has no use for test artifacts, and this way `test/**` carries no 0.8.17 constraint at all rather than a CI-enforced one.

`forge script` is the one build left unskipped, and it does not reopen the gap — it compiles only the target's dependency closure, not the project: 198 files for `DeployAccessManagerFacet.s.sol` under `solc_floor`, no `test/**` artifact in the out dir.

Header comment says what the skip is matched against and why it is the glob and not the alias. Three cases read the skip out of the workflow and compare it against the argv each build records (`london-profile-selection.test.ts`); each fails when one side is changed alone — verified by mutating the group build, the salt build, and the workflow in turn.

script/deploy/deploySingleContract.sh
0xDEnYO: Right — the bullet predates `392f376a`. Description now states the direct path is covered and points at the call site, so no follow-up ticket inherits a problem that is gone.

.github/workflows/solc-floor-build.yml
0xDEnYO: Follow-up on my own reply: the first version of this only pinned the strict arm. Now also pinned — the tolerant `prepareGroupBuild` arm (what `multiNetworkExecution.sh` calls), the quoting at each site (unquoted, the runner's shell expands the glob before forge sees it, and the sandbox cannot catch that), the job's `FOUNDRY_PROFILE`, and its path filter. `scriptMaster.sh:51`'s `COMPILE_ON_STARTUP` build was the one remaining unskipped build under the profile; it carries the skip now too.

gvladika (approved):

gvladika (approved):

maxklenk (approved):

PR #2418, opened 2026-09-18
fix(deploy): reap finished jobs in wave throttles, stream per-network output

gvladika:

Ref [EXSC-1038](https://linear.app/lifi-linear/issue/EXSC-1038/deploy-pipeline-hangs-after-all-on-chain-work-completes-requiring-a)

`Ref`, not `Fixes`: the reported hang is **not** fixed here and the ticket stays open. This lands the two defects that were provable offline and leaves the unexplained one open.

The ticket named the `$(jobs | wc -l)` throttle as the confirmed root cause and noted that raising `MAX_CONCURRENT_JOBS` to 100 did not help. Measuring it first changed the plan.

`jobs` in a subshell does fail to reap, but the stale entries **do not accumulate** — contrary to what the ticket states. The count pins at exactly 1, on bash 3.2.57 and 5.3.15 alike, after any number of finished jobs:

```
after 20 finished jobs, $(jobs|wc -l) = 1     # and stays 1
while running:          jobs=2  jobs-rp=2     # accurate in flight
```

So the condition can only be permanently true at a concurrency of 1, and every higher value was always fine. Driving the real `launchDeployWave` against the pre-fix code confirms both halves: three networks at concurrency 1 time out with one of three results written, four networks at concurrency 2 take 2s either way.

That makes this a **deadlock fix, not a misthrottle fix** — and it is not academic. `deployContractToNetworks.sh:408` hardcodes `launchDeployWave 1` for the zkEVM group (abstract, lens, zksync), so any deploy covering two or more of those wedges after the first network finishes, and no `MAX_CONCURRENT_JOBS` value can reach it. All six `jobs | wc -l` call sites are converted; the ticket's list of seven was stale line numbers.

Separately, per-network output went through `sed`, which block-buffers when its stdout is not a tty: nothing appeared for a network until its worker exited, so a run wedged eight minutes ago and a run still working were indistinguishable in a redirected log. It is now a bash read/printf loop — a builtin rather than `sed -u` or `stdbuf`, because `-u` is GNU-only and BSD's `-l` means something else entirely to GNU sed, so a flag fix would work on one machine and silently not on the next. Measured over a producer that emits, sleeps 2s, then exits: `sed` has written nothing at the half-second mark, the loop has written the line. Identical final content, which is why this was invisible. The bash loop is ~6x slower than sed — 0.1s per 10k lines, irrelevant next to a deploy.

That "identical final content" holds per network, not across a run, and the difference is worth naming because it changes how a log reads. `proposeToNetworkWorker` ends with a bulk `cat "$PROPOSE_LOG"`; under `sed` that landed as one contiguous block per network, so a reader found a network's whole log body in one piece. Per-line writes mean two networks dumping at the same moment now interleave line by line. The `[network]` prefix keeps every line attributable and liveness is clearly the better trade, but anyone reading a run log should expect the interleaving rather than be surprised by it.

## What is deliberately not here

An earlier revision of this PR carried a stall-report apparatus (`reportStalledWave`, `waitForWaveCapacity`, `WAVE_STALL_REPORT_SECONDS`, ~335 lines with its tests). It has been dropped.

It was built around a leftover child holding the wave's pipe open, and that mechanism does not match what the ticket recorded — *"worker subshells sit spinning on `sleep 0.1` with no forge or cast child process"*. Spinning is an in-worker throttle loop in `helperFunctions.sh`, not a blocked pipe, and the report only instrumented `launchDeployWave`. Its advisory line (`no forge/cast/bun above means the wave is held open by a leftover child`) would have printed a wrong conclusion under exactly the symptom on record. It also converted the free blocking drain into a 5 Hz poll and added ~45s of sleeps to the TS suite.

If the pipe-holding theory is right, it is also a one-line fix rather than something to observe — measured, same leaked-grandchild scenario:

```
worker | prefixer &          -> 4s   (current shape)
worker > >(prefixer) 2>&1 &  -> 0s
```

That is not obviously correct, though: `abortInFlightDeployments` relies on grandchildren being reachable, and waiting on pipe EOF is what guarantees nothing is mid-broadcast when the caller repoints `foundry.toml`. Decoupling them trades a hang for a silent correctness risk, which is a decision for the follow-up, not a drive-by.

## Follow-ups (not in scope here)

- **EXSC-1038 stays open.** The hang at concurrency 10 and 100 is unexplained. The ticket's own next lead is untouched: no `timeout` around the `cast` calls in `saveDiamondPeriphery`, which is the one mechanism that survives this fix and does produce workers spinning at `sleep 0.1`.
- `checkPauserFunds.sh:329` and `verifyEmergencyPauseReadinessGitHub.sh:611` use `wait -n`, which bash 3.2.57 does not support. Every one of these scripts is `#!/bin/bash`, which on macOS is 3.2.57 despite `.agents/rules/300-bash.md` claiming 5.0+. When those throttles engage locally they hot-spin, printing `wait: -n: invalid option` with no sleep at all. Works in CI on bash 5 and nowhere else.


script/deploy/waveThrottle.test.ts
```diff
@@ -0,0 +1,114 @@
...
+    expect(peak).toBe(1)
+    // one at a time: three 1s workers cannot finish in under 3s
+    expect(elapsed).toBeGreaterThanOrEqual(3)
+  }, 60_000)
+
+  it('holds the wave to the configured concurrency', () => {
+    const { elapsed, results, peak } = runWave(2, 4)
+
+    expect(results).toBe(4)
+    expect(peak).toBe(2)
```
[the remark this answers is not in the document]
gvladika: Agreed on the race, fixed in 17792a934 — but with a barrier rather than relaxing the assertion.

`toBeLessThanOrEqual(2)` passes a throttle that permits too *little*, which is the failure mode this ticket is actually about (the pre-fix code deadlocked at a concurrency of 1). Measured over the same worker shape:

| throttle enforces | peak observed | `toBe(2)` | `toBeLessThanOrEqual(2)` |
|---|---|---|---|
| 2 (correct) | 2 | pass | pass |
| 4 (too many) | 4 | fail | fail |
| 1 (too few) | 1 | fail | **pass** |

So each worker now holds at a barrier until the marker count reaches the limit before sampling — the synchronized launch instrumentation you suggested as the alternative. The wait is bounded at 10s, so a throttle that never saturates fails on the sample with a readable peak instead of hanging to the harness timeout.

script/deploy/deployContractToNetworks.sh
```diff
@@ -127,6 +186,61 @@ function abortInFlightDeployments() {
...
+  local MAX_RUNNING="$1"
+  local ELAPSED
+  # `jobs -rp`, not `jobs`: from a command substitution `jobs` prints a copy of
+  # the table that the parent never reaps, so a finished job stays in the count
+  while [[ $(jobs -rp | wc -l) -gt $MAX_RUNNING ]]; do
+    # short poll so a wave that finishes quickly is not held up by the poll itself
+    sleep 0.2
+    ELAPSED=$((SECONDS - WAVE_START))
+    # re-check: a wave that finished during the sleep above is not stalled
+    if [[ $ELAPSED -ge $NEXT_REPORT && $(jobs -rp | wc -l) -gt $MAX_RUNNING ]]; then
```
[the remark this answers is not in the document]
gvladika: Not applying this one — the premise needs job control, which this script never enables.

Your probe sets `set -m`. `deployContractToNetworks.sh` runs with job control off, and without it bash never reaps a child's stopped status (no `WUNTRACED`), so a stopped worker stays `Running` in the job table. Measured on both bash versions this script runs under, with a full pipeline SIGSTOPped:

```
bash 5.3.15  job control: off        bash 3.2.57  job control: off
  74160 74157 T bash                   74359 74357 T /bin/bash
  74161 74157 T cat                    74360 74357 T cat
jobs -rp = 1                         jobs -rp = 1
jobs -sp = 0                         jobs -sp = 0
```

So `jobs -rp` still counts the stopped pipeline, `waitForWaveCapacity 0` keeps waiting rather than falling through, and `reportStalledWave` fires on schedule — and lists the worker with its `T` state, which is what the report's STAT column is there for (its docstring: "a process in D or T state is stuck, not working"). Adding `jobs -sp` would be a branch that can only be non-zero under a `set -m` this script doesn't use.

If we ever turn job control on, this becomes real — worth a comment on that change, not a guard now.

script/deploy/waveStallReport.test.ts
```diff
@@ -0,0 +1,230 @@
...
+    expect(output).not.toContain('must be a positive integer')
+    expect(output).toContain('EXIT:0')
+  }, 60_000)
+})
+
+describe('launchDeployWave stall report', () => {
+  it('names the leftover child holding the wave open, and still completes', () => {
+    const output = runStalledWave(12)
+
+    expect(output).toContain('wave has been running 5s')
```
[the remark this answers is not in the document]
gvladika: Fixed in 7e6acbd61. The assertion now extracts the reported seconds and requires `>= 5`, matching the lower-bounds-only rule the throttle test's header already states for the same reason.

gvladika:
Scope narrowed in d5b1ebbd9 — this PR is now +275/−13 across 6 files, down from +634/−15 across 8.

The stall-report apparatus is gone (`reportStalledWave`, `waitForWaveCapacity`, `resolveStallReportThreshold`, `WAVE_STALL_REPORT_SECONDS`, `waveStallReport.test.ts`). It was diagnostics for a mechanism that does not match what the ticket recorded — *"worker subshells sit spinning on `sleep 0.1` with no forge or cast child process"*. Spinning is an in-worker throttle loop in `helperFunctions.sh`, not a blocked pipe, and the report only instrumented `launchDeployWave`; its advisory line would have printed the wrong conclusion under exactly that symptom. Rationale and the measurements are in the PR description.

What remains is the two defects that are provable offline: `jobs -rp` in all six throttles (a real deadlock at concurrency 1, which is the hardcoded zkEVM wave) and `prefixNetworkOutput` replacing the block-buffering `sed`.

**CodeRabbit:** the comments on `waveStallReport.test.ts` and on `deployContractToNetworks.sh` lines 130 / 132 / 200 / 237 now target deleted code and are moot. The two on `waveThrottle.test.ts` (strict wall-clock upper bound, peak measurement) were already addressed in 17792a934 and survive here — the retained test asserts lower bounds only and proves concurrency with a barrier plus marker-file sampling.

CI is green, and `run-ts-tests` is back down from 3m08s to 2m33s.

script/deploy/waveThrottle.test.ts
```diff
@@ -0,0 +1,125 @@
...
+        sleep 0.1
+        SPINS=$((SPINS + 1))
+      done
+      running >>"$RUN_DIR/samples"
+      sleep 1
+      rm -f "$MARKER"
+      echo "OK" >"$5/$1"
+    }
+    START=$SECONDS
+    launchDeployWave ${concurrency} production SomeFacet 1.0.0 "$RESULT_DIR" ${names}
```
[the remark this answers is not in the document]
gvladika: Moot as of d5b1ebbd9 — the stall-report apparatus was dropped from this PR. `launchDeployWave` no longer reads or validates `WAVE_STALL_REPORT_SECONDS` (`resolveStallReportThreshold` and its `exit 1` are gone, along with `waitForWaveCapacity` and `reportStalledWave`), so there is no inherited value the harness could trip over and nothing for it to set. Zero references to the knob remain in `script/` or `.env.example`.

Rationale for the removal is in the PR description and https://github.com/lifinance/contracts/pull/2418#issuecomment-5757603541.

script/deploy/waveThrottle.test.ts
```diff
@@ -0,0 +1,124 @@
...
+    echo "ELAPSED:$((SECONDS - START))"
+    echo "RESULTS:$(find "$RESULT_DIR" -type f | wc -l | tr -d ' ')"
+    echo "PEAK:$(sort -n "$RUN_DIR/samples" | tail -1)"
+    rm -rf "$RESULT_DIR" "$RUN_DIR"
+  `
+  const output = execFileSync('bash', ['-c', script, 'harness'], {
```
0xDEnYO, line 141: Nit (the one I'd actually consider doing): a throttle regression is detected only by letting this `execFileSync` hit its 60s timeout.

I reverted the throttle to `$(jobs | wc -l)` on this branch and ran the file — it does fail, which is the point, but via `error: spawnSync bash ETIMEDOUT` after 60005ms, and bun then reported `killed 1 dangling process`: the SIGTERM'd harness leaves the backgrounded worker subshell alive. So every legitimate failure costs a CI runner 60s of dead wall-clock plus a leaked process, and the message names a timeout rather than the deadlock.

A bounded abort inside the harness script (give up after ~15s and `echo` the RESULTS/PEAK reached so far) would fail in seconds with a diagnostic that reads as the actual bug.

script/prefixNetworkOutput.test.ts
```diff
@@ -0,0 +1,109 @@
...
+    'script/deploy/deployContractToNetworks.sh',
+    'script/tasks/proposeContractToNetworks.sh',
+  ]
+
+  it.each(LAUNCHERS)('%s tags worker output through the prefixer', (file) => {
+    const source = readFileSync(join(REPO_ROOT, file), 'utf8')
+
+    expect(source).toContain('| prefixNetworkOutput "$WAVE_NETWORK" &')
+    // the buffering this replaced is invisible in a passing deploy, so guard the revert
+    expect(source).not.toMatch(/\|\s*sed\s+"s\/\^\//)
```
0xDEnYO: Nit: this pins one literal spelling. `| sed 's/^/[$WAVE_NETWORK] /'` with single quotes restores the exact EXSC-1038 buffering and passes this assertion; so does an `awk` equivalent. The `toContain` above is similarly satisfied if a third, unprefixed launcher is added later.

Not worth much effort — the streaming test above is what really guards the fix, and it does (verified: it fails when the function is swapped back to `sed`). Mostly flagging so the source-text assertions aren't mistaken for the safety net.

script/deploy/deployContractToNetworks.sh
```diff
@@ -153,15 +153,19 @@ function launchDeployWave() {
...
     # </dev/null makes the no-stdin guarantee explicit - the sourced framework must
-    # never block on an interactive prompt inside a background worker; sed attributes
-    # every line of framework output to its network, since concurrent workers'
-    # otherwise-unprefixed logs interleave on the shared terminal
-    deployToNetworkWorker "$WAVE_NETWORK" "$WAVE_ENVIRONMENT" "$WAVE_CONTRACT" "$WAVE_VERSION" "$WAVE_RESULT_DIR" </dev/null 2>&1 | sed "s/^/[$WAVE_NETWORK] /" &
+    # never block on an interactive prompt inside a background worker;
+    # prefixNetworkOutput attributes every line of framework output to its network,
+    # since concurrent workers' otherwise-unprefixed logs interleave on the shared
+    # terminal, and streams it unbuffered so a stalled worker stops producing output
+    deployToNetworkWorker "$WAVE_NETWORK" "$WAVE_ENVIRONMENT" "$WAVE_CONTRACT" "$WAVE_VERSION" "$WAVE_RESULT_DIR" </dev/null 2>&1 | prefixNetworkOutput "$WAVE_NETWORK" &
```
0xDEnYO, line 168: Nit, writeup only: worth a line in the PR body that this also changes how the logs read.

`proposeToNetworkWorker` ends with `cat "$PROPOSE_LOG"` — one bulk dump. Under `sed` that landed as a contiguous block per network; per-line writes mean two networks dumping at the same moment now interleave line by line. The `[network]` prefix keeps it attributable and liveness is clearly the better trade, but someone reading a run log will no longer find a network's log body in one piece, and that's surprising if it isn't called out.

0xDEnYO (dismissed):
Approving. I re-derived the load-bearing claims rather than taking the writeup's word, and they hold:

- The deadlock is live: `launchDeployWave 1` (deployContractToNetworks.sh:412, zkEVM) and `launchProposeWave 1` (proposeContractToNetworks.sh:458) both wedge permanently after the first network finishes on current `main`.
- `$(jobs|wc -l)` pins at exactly 1 after 20 finished jobs on both bash 3.2.57 and 5.3.15 — so this is a concurrency-1 deadlock, not a misthrottle, as the description says. `jobs -rp` fixes it on 3.2, which is what `#!/bin/bash` resolves to on macOS.
- All six call sites converted; no `$(jobs | wc -l)` throttle left on the branch.
- Subshell job-table inheritance is not a confounder — a forked subshell starts with an empty table, so the throttles inside `saveDiamondPeriphery`/`saveDiamondFacets` count their own jobs.
- Both regression tests genuinely fire: reverting the throttle fails `waveThrottle.test.ts`, and swapping `prefixNetworkOutput` back to `sed` fails 2 of 7 in `prefixNetworkOutput.test.ts` including the streaming one. 9/9 pass as-is.

Three non-blocking nits inline — all test-quality or wording, none touch the shipped fix. Take them or leave them; don't hold the merge for them.

One thing worth chasing in the follow-up: at `MAX_CONCURRENT_JOBS=1` the `saveDiamondPeriphery` throttle (helperFunctions.sh:1245, `sleep 0.1`) had this same deadlock — which matches the ticket's recorded symptom *"worker subshells spinning on `sleep 0.1` with no forge or cast child"* exactly. This PR fixes that path too, so the hang still seen at 10/100 may be a different mechanism than the one EXSC-1038 describes.

script/deploy/waveThrottle.test.ts
gvladika: Done in 12ea930ff. A disowned watchdog inside the harness ends a wedged wave at 15s, kills each worker's process tree with the script's own `killProcessTree`, and reports `ABORTED` alongside the RESULTS/PEAK the wave reached. Reverting the throttle to `$(jobs | wc -l)` now fails in 17s with

```
- "aborted": false        + "aborted": true
- "results": 3            + "results": 1
```

instead of `ETIMEDOUT` after 60005ms, and bun no longer reports a dangling process. The three assertions were collapsed into one `toEqual` so the diff carries how far the wave got rather than a bare boolean. Healthy runs are unchanged at 8.5s for the file.

One thing your suggestion surfaced that is worth recording: bounding it the obvious way — `{ launchDeployWave ...; } &` plus a poll — **silently disarms the test**. Measured, `$(jobs | wc -l)` inside a forked subshell reaps normally under `bash -c`, so the pre-fix throttle completes a concurrency-1 wave there (RESULTS:3, 6s) and the regression goes undetected. Under `bash file.sh` the same wrapped form still wedges, so it is specific to how bun invokes the harness. Hence the watchdog signals the shell instead, `disown` keeps it out of the throttle's own job count, and its stdout is detached so it cannot hold the harness pipe open after a healthy wave. That's in the file header so the next person doesn't "simplify" it back.

script/prefixNetworkOutput.test.ts
gvladika: Fixed in 12ea930ff — you were right that these were decorative.

The assertions now match on the **worker launch** rather than on the prefixer, so anything piped in its place still gets caught:

```
const WORKER_LAUNCH = /^(?!\s*#).*\w+ToNetworkWorker\b.*&\s*$/gm
const THROUGH_PREFIXER = /\|\s*prefixNetworkOutput\s+"\$\w+"\s*&\s*$/
```

and the launcher list is discovered by walking `script/` for files containing such a line, instead of being hardcoded. Verified all three of the holes you named now fail:

- `| sed 's/^/[$WAVE_NETWORK] /' &` (your single-quoted example) → fail
- `| awk '{print "[x] " $0}' &` → fail
- a third launcher added with no prefixer at all → fail, and it is picked up automatically

The keep-list test is `toContain` on the two known launchers rather than `toEqual` on the whole set, so it guards discovery silently returning `[]` (which would make `it.each` register no tests and pass) without making a new launcher an edit here.

Agreed on the framing regardless: the streaming test is the safety net, these are the tripwire.

script/deploy/deployContractToNetworks.sh
gvladika: Added to the PR description, right after the paragraph whose "identical final content" claim it qualifies — since that claim holds per network but not across a run, which is exactly the gap you spotted:

> That "identical final content" holds per network, not across a run, and the difference is worth naming because it changes how a log reads. `proposeToNetworkWorker` ends with a bulk `cat "$PROPOSE_LOG"`; under `sed` that landed as one contiguous block per network, so a reader found a network's whole log body in one piece. Per-line writes mean two networks dumping at the same moment now interleave line by line. The `[network]` prefix keeps every line attributable and liveness is clearly the better trade, but anyone reading a run log should expect the interleaving rather than be surprised by it.

No code change.

0xDEnYO (approved):

PR #2419, opened 2026-09-18
fix(safe): rehearsal gate roster, and refuse store credentials over cleartext (EXSC-1042)

0xDEnYO:

Fixes [EXSC-1042](https://linear.app/lifi-linear/issue/EXSC-1042)

Two fixes in the rehearsal/proposal-store path.

## The gate roster can say a gate is absent

`bun verify-pending-proposals` prints a gate roster whose job is to tell a gate that was never wired into the rehearsal (`absent`) from one that ran and answered for nothing (`registered-but-silent`). The caller passed the whole confirm registry as `registered`, so `absent` was unreachable and every unwired gate rendered as `registered-but-silent rows=0`, reading as a gate that ran.

`REHEARSED_CHECK_IDS` (beside `runGateChain` in `verify-rehearsal.ts`) now names the ids the chain records. The ledger registers exactly those and the report is built from the same constant, so the roster follows the code path rather than a second hand-typed list; a recorder added without an entry throws on its first `recordCheck` instead of rendering as a gate that never ran. The roster in `rehearsal-report.ts` gains the three registry gates it did not name (codehash K, storage-authority G, immutables L), so the report can name them absent with their `(expected from: …)` hint.

## Proposal-store credentials are no longer accepted over cleartext

`SC_MONGODB_URI` went straight to `MongoClient` with no transport validation, so a credentialed URI pointed at a remote host over an unencrypted connection was accepted silently (CWE-319). Raised by CodeRabbit on #2378 against `verify-rehearsal.ts`; that PR is superseded by #2393, so the finding is applied here.

**Not the suggested fix.** CodeRabbit proposed requiring TLS for non-loopback hosts and exempting loopback as local. That reading is wrong in this repo: `lifi-connect prod smart-contracts` forwards *production* Mongo to a `localhost` port, so a loopback host is not evidence of a local store — the same mistake a seeder made when it printed LOOPBACK while writing production data.

`assertStoreCredentialsAreEncrypted` (`mongo-store-transport.ts`) classifies on the credential instead. A URI carrying userinfo reaches a real store and must be encrypted; a URI carrying none has nothing on the wire to protect and passes untouched, remote or not. Loopback appears only as a property of the wire — a connection that cannot leave the machine cannot be read off it — and never as a claim about which store is on the other end, which is why the refusal says "a credentialed store, not a local one". This mirrors `getTransportConfigFromRpcUrl` (`script/utils/viemScriptHelpers.ts:76`), which already refuses RPC credentials over `http:` on the same reasoning; following it is why the guard is credential-gated rather than host-gated.

**In the shared path, not just the reported file.** `getSafeMongoCollection` and three report scripts had the identical gap, and `getSafeMongoCollection` is the one every signer actually goes through — fixing only `verify-rehearsal.ts` would have left the hole where it matters most. All five openers now call the guard before constructing their client.

A test pins that against the tree rather than against today's list: it walks all of `script/`, recognises `new MongoClient(`, the driver's static `MongoClient.connect(` and a namespaced constructor, and compares positions, so a guard called *after* the client is built counts as unguarded. It cannot recognise an opener that reaches `SC_MONGODB_URI` through a helper — that is the residual gap, and it is stated in the test's own docstring rather than claimed away.

**`MONGODB_URI` is deliberately out of scope.** Roughly a dozen openers on the deployment-log store (`parked-tasks.ts`, `timelock-queue.ts`, `signed-set-record.ts`, `script/deploy/update-deployment-logs.ts`, `script/mongoDb/*`, …) have the same gap, and unlike `SC_MONGODB_URI` that store is reached from GitHub Actions with no tunnel — so it is the larger exposure, not the smaller one. It is left out here because turning the guard on for it can refuse a URI that CI uses today, which is a rollout with its own blast radius rather than a line in this PR. The guard already takes the variable name as a parameter for that follow-up.

The URI is never put into a thrown message; the refusal names only the exposed hosts. The credentialed loopback URI documented in `docs/Setup.md:169` (`tls=false`, "TLS terminates at the tunnel") is accepted unchanged, so no signer's `.env` or the setup docs need editing.

### TLS that does not verify the peer is not TLS

Two further findings from CodeRabbit's security review on this PR.

`tls=true` was enough to clear the guard, so `tls=true&tlsInsecure=true` passed — as did `tlsAllowInvalidCertificates` and `tlsAllowInvalidHostnames`. The driver then accepts any certificate or any hostname, and it sends the credentials during authentication, so an impersonated peer collects them before anything the store says could give it away (CWE-295). Those three are now refused when a credentialed URI reaches a non-loopback host. The check sits after the loopback filter, so a relaxed tunnel and an uncredentialed URI are untouched.

Separately, the host group of the URI pattern permitted `@`, so `user:pa@ss@host` parsed as the host `ss@host` and put that half of the password into the refusal message (CWE-209). The pattern now excludes `@`, `?` and `#` from the host, and such a URI is refused on its shape before any message is built.

That second refusal is deliberately stricter than the driver rather than a bug-for-bug match. The spec requires those characters to be percent-encoded in userinfo, but `mongodb-connection-string-url` accepts them anyway by splitting on the last `@`. Guessing which half is the password is what leaked one, so the guard refuses and the message says to encode it.

**Falsification, these two.** Neutering the peer-validation refusal fails 5 of the new tests; restoring the permissive host group fails the other 2. A differential run of the old and new function over 5,760 generated URIs moves nothing from refuse to pass — the only verdict changes are the intended new refusal and unparseable-on-`#`. The three options are also the complete relaxing set the pinned driver honours: `sslValidate`, `tlsDisableCertificateRevocationCheck` and `tlsDisableOCSPEndpointCheck` are all rejected by mongodb v6.17.0 as unsupported URI options, and `rejectUnauthorized=false` arrives as the string `'false'`, which Node's `!== false` check ignores.

**Falsification, the guard as a whole.** Neutering `assertStoreCredentialsAreEncrypted` to return immediately fails 7 of the 13 new tests. Swapping in CodeRabbit's loopback-only classification (drop the `if (!userInfo) return`) fails `lets an uncredentialed URI through, remote or not` — the test that pins the credential classifier specifically. Removing the call from `getSafeMongoCollection` fails the placement test, as do a static-factory opener, a namespaced constructor, an opener in a nested directory, and a guard call moved below the construction. Separately, the ledger-narrowing test fails against the old caller:

```text
error: expect(received).toBe(expected)
Expected: "absent"
Received: "registered-but-silent"
(fail) buildRehearsalGateReport > names every registry gate the chain never wired as absent, with where to expect it
```

Run: `bun test script/deploy/safe/` → 3009 pass, 0 fail. `bunx tsc-files --noEmit` and `bunx eslint` on all changed files exit 0. The guard was also run against the real configured `SC_MONGODB_URI` as a string check (no connection opened): accepted — though that was before the host pattern tightened, so what now stands behind the documented URI is the test pinning the `docs/Setup.md:169` shape, plus a probe confirming that shape still parses.

No signer-facing gate behaviour changes; the rehearsal tool is not described in `docs/MultisigSigningProcess.md`, so no doc update.

# Folded-in PRs

**#2421** (EXSC-995) is closed into this branch — it touched the same two files
(`rehearsal-report.ts`, `rehearsal-teardown.ts`) and would have conflicted here.
It adds: a teardown that fails when the nonce is still held, the proposal named
on every signer workload line, and teardown graded on one snapshot and on what
the delete actually removed.

The only collision was the import block in `rehearsal-report.test.ts`; both
sides were kept.


0xDEnYO:
## Live CLI evidence (EXSC-1042 AC6)

Run against the real proposal store through the lifi-connect tunnel, at `fafdfa596`.

Nothing is pending right now, so this uses `--status executed` for a corpus, which is what the flag is for. The roster is the thing under test here, and it renders the same either way.

```text
bunx tsx ./script/deploy/safe/verify-rehearsal.ts \
  --networks arbitrum,mainnet,optimism,base,polygon --status executed

ℹ Grading anchors
present : .cache/deployments_production.json
✔ Preflight: 18 write surfaces refused; run is read-only
ℹ Rehearsing 538 executed proposal(s) on arbitrum, mainnet, optimism, base, polygon
✔ Deterministic: two passes graded 538 row(s) identically

Gate roster
INT-SAFE-ADDRESS    absent                 rows=0  (expected from: confirm-integrity-asserts.ts, mirrored by EXSC-994)
INT-SAFE-TX-HASH    absent                 rows=0  (expected from: confirm-integrity-asserts.ts, mirrored by EXSC-994)
INT-SIGNATURES      absent                 rows=0  (expected from: confirm-integrity-asserts.ts, mirrored by EXSC-994)
INT-FIXED-FIELDS    absent                 rows=0  (expected from: confirm-integrity-asserts.ts, mirrored by EXSC-994)
INT-TARGET          absent                 rows=0  (expected from: confirm-integrity-asserts.ts, mirrored by EXSC-994)
INT-TIMELOCK-DELAY  absent                 rows=0  (expected from: confirm-integrity-asserts.ts, mirrored by EXSC-994)
codehash            absent                 rows=0  (expected from: confirm-check-registry.ts (merged))
immutables          absent                 rows=0  (expected from: confirm-check-registry.ts (merged))
storage-authority   absent                 rows=0  (expected from: confirm-check-registry.ts (merged))
target-state        present                rows=538  Facet version matches the declared target state
executability       absent                 rows=0  (expected from: EXSC-994)
rpc-quorum          absent                 rows=0  (expected from: EXSC-994)
```

Eleven gates print `absent` with their `(expected from: ...)` hint, and the one the chain wired prints `present rows=538`. On `main` every one of those eleven rows reads `registered-but-silent rows=0`, which is the defect this PR fixes: a gate that was never wired rendering as a gate that ran and found nothing.

The run also exercised the read-only guard against the live store (18 write surfaces refused) and agreed with itself across two passes over 538 real proposal shapes.

Two caveats, both the tool's own and neither affecting the roster:

- `--status executed` grades against the target state `main` declares now rather than the one each proposal was raised against, so the tool withholds verdicts. It establishes that the chain survives 538 real proposal shapes and that the roster renders correctly. It establishes nothing about those proposals.
- A fresh worktree cannot run this at all: `.cache/deployments_production.json` is gitignored, and the tool refuses rather than grading without it. I copied it from the primary checkout.

gvladika (approved):

gvladika (approved):

PR #2421, opened 2026-09-19
fix(safe): fail a teardown that left the nonce held, name the proposal for the signer (EXSC-995)

0xDEnYO:

Fixes [EXSC-995](https://linear.app/lifi-linear/issue/EXSC-995) — two CodeRabbit findings raised on #2378 that were never applied. That PR is now closed as superseded (its code reached main via #2393), and its closing comment records both. Verified live against `origin/main` at `b2d9f7cf2` before any code was written. The Linear connector is unauthorized in this session, so no dedicated follow-up ticket exists yet.

**1 — `rehearsal-teardown.ts` exited 0 with the nonce still blocked.** The CLI discarded the results of `tearDownProposalsAtNonce`. `deletePendingProposals` runs with `force: false`, so a row carrying more than one signature returns `skipped-signed` and is not deleted — and the command still printed nothing and exited 0. A Safe executes strictly in order, so automation reading that exit code proceeds while the nonce stays held and everything queued behind it stalls.

The guard is the new exported `unclearedRows`; the citty handler is a two-line read of it. `tearDownProposalsAtNonce` keeps its non-throwing contract — it now returns the hunt alongside the delete's verdict, so both halves of the decision come from **one snapshot**. Grading `found` from the CLI's own separate hunt against `results` from the function's internal one was a race: a row flipping `pending → submitted` between the two lands in neither list and exits 0.

A row survives three ways, and the set is deliberate:

- **not deletable** — a non-`pending` row is never offered to the delete. That path previously only warned.
- **`skipped-signed`** — the delete saw it and refused.
- **`deleted` with `deletedCount: 0`** — `deletePendingProposals` reported success and removed nothing. `findOne` matches under the collection's collation while `deleteOne` uses an uncollated `$eq`, so a row stored under a different casing is read and then missed. The outcome names the branch taken, not the effect, so `IDeleteResult` now carries `deletedCount` and the guard reads that.

**`not-found` also counts as surviving.** It is ambiguous: a concurrent teardown that won the race leaves the slot free, but so does the collation mismatch above. Indistinguishable from here, and the costs are not symmetric — a false success stalls the whole queue behind the nonce, a false failure costs one look at the log.

**2 — `renderSignerWorkload` dropped `task.proposal`.** `summariseSignerWorkload` dedupes `(checkId, network)` within one ledger only, so the same pair recurs legitimately across proposals and two rows rendered byte-identical. On a realistic corpus (6 proposals × 2 networks × 2 semantic checks) the old renderer collapsed 10 rows into 4 distinct lines. The line now leads with the proposal.

The `slice(0, 10)` truncation flagged alongside it is fixed here rather than left, since it is the same three lines: the remainder is now counted, unindented so it reads as a summary of the list rather than an entry in it.

No signer-facing gate behaviour changes. No contract, deploy or config surface is touched.

**Falsification.** Every new guard was run against a deliberately broken implementation and fails there:

| Break | Failing test |
|---|---|
| drop `task.proposal` from the line | `tells apart two tasks that differ only by proposal` |
| ignore delete outcomes | `reports a signed row the delete refused…`, `counts a row the delete could not find…` |
| ignore `deletedCount` | `counts a delete that reported success but removed nothing` |
| drop the `Set` dedupe | `names a row once when it survives both ways` |

The exit path itself was verified by execution, not by reading: citty 0.1.6's `runMain` calls `process.exit` only on `--help` and on a thrown error, so a resolved handler's `process.exitCode` survives — a repro mirroring this file (required args, `try`/`finally` with an async close, `void runMain`) was observed exiting 1. `process.exitCode` rather than `process.exit(1)` so the `finally` still closes the Mongo client, matching the sibling CLIs in this directory.

Test fixtures were reshaped to production form: `safeTx.signatures` is keyed by lowercased signer with `ISafeSignature` values, as `safe-utils.ts` stores it, rather than single-letter keys over strings.

**Run:** `bun test script/deploy/safe/` → 2982 pass, 0 fail (94 files). `bunx tsc-files --noEmit` and `bunx eslint` on all six changed files exit 0.


0xDEnYO:
@coderabbitai review

0xDEnYO:
Folded into #2419 — its commits are now on that branch, so this PR is closed to keep the review queue to one unit per subsystem. No work is lost.

PR #2422, opened 2026-09-19
fix(deploy): drop the duplicate EcoFacet key from the immutable registry

0xDEnYO:

None. Found while checking CI on #2419.

`validate-json` fails on `main` and on every open PR:

```text
[error] script/deploy/resources/immutableRegistry.json:122 declares 'EcoFacet' again under $.
        Every parser keeps only the last one, so the earlier copy is silently dropped.
[error] 1 duplicate key(s) across 215 file(s).
```

`jsonChecker.yml` runs on every pull request with no path filter, against the merge commit, so main's duplicate reaches every open PR. The duplicate itself is not new. It went unnoticed until #2414 landed the duplicate-key checker this morning, which is why a pre-existing defect is only now red.

The EcoFacet rollout (#2406) prepended a `EcoFacet` block at line 2, of which the file already had a copy at line 122 from #2393. The two blocks are byte-identical, down to trailing commas, so the parsed registry never depended on which one won:

```text
parsed-identical: true
key count: 42 -> 42
```

I removed the copy at line 122 rather than the one at line 2, which is the choice that leaves iteration order untouched. A repeated string key keeps its **first** insertion position while its **last** value wins, so `EcoFacet` currently sits at index 0. Dropping the later block leaves it at 0; dropping the earlier one would move it to index 9. No consumer of this registry is order-dependent — they key by contract name, or sort — so this is about making the diff a true no-op rather than about fixing a bug.

Verified in a worktree off `b2d9f7cf2`:

```text
bunx tsx ./script/deploy/resources/verify-json-duplicate-keys.ts config deployments \
  audit/auditLog.json script/deploy/_targetState.json script/deploy/resources script/deploy/immutables
✔ No duplicate keys in 215 JSON file(s).

bunx tsx ./script/deploy/immutables/verify-immutable-registry.ts
ℹ 80 immutables declared in src/; 80 carry a registry entry, 61 flagged authority-bearing
✔ every immutable in src/ has a registry entry

bun test script/deploy/immutables/   123 pass, 0 fail
bun test script/deploy/codehash/     475 pass, 0 fail
```

Restoring main's copy of the file makes the duplicate checker fail again, so the deletion is what makes it pass.

`EcoFacet` is the only duplicate. The checker accumulates findings across every file rather than stopping at the first, so the run above is itself the exhaustive scan over the paths CI covers.

The file is hand-maintained, not generated: every reference to it under `script/`, `tasks/` and `.github/` reads or validates it, and the only writer is a test writing its own fixtures.

The pre-commit secret scanner warns on a `bytes32` expectation value at line 287. It predates this branch, and this commit adds no lines.

- [ ] I have added tests that cover the functionality / test the bug — the checker added in #2414 is the test; it now passes


0xDEnYO:
@coderabbitai review

gvladika (approved):

PR #2424, opened 2026-09-19
fix(safe): gate G stands down on Tron and grades a failed read apart (EXSC-1043)

0xDEnYO:

[EXSC-1043](https://linear.app/lifi-linear/issue/EXSC-1043) — Gate G reds Tron instead of standing down, and conflates it with a failed read. Child of [EXSC-1062](https://linear.app/lifi-linear/issue/EXSC-1062) (Tron gate coverage).

`observeSetForProposal` in `confirm-safe-tx.ts` returned `undefined` for three different situations — a chain outside the reader's coverage (Tron), a failed chain read, and calldata that schedules no timelock batch — and `proposalCheckResults` graded all three as one blocking `error` row, "no storage-authority read was made for this proposal". On Tron that is an infrastructure-looking red on every proposal for a chain the gate was never written to read, and a real RPC failure was indistinguishable from it.

The read now returns a tagged observation (`observed` | `absent` with a reason), the evidence bundle carries the reason to the recorder as `storageAuthorityAbsence`, and the registry grades each absence on its own terms:

- **out of scope** (Tron) → `needs-ack` on `A-DOCUMENTED`, the row naming the chain. `A-DOCUMENTED` rather than `A-UNRESOLVED` because gate G is an integrity check and its `needs-ack` survives only on an anchor a human can decide about; the coverage gap is the reviewed statement in `resolveGateCoverage`.
- **read failed** → `error` (blocks), with the read failure on the row instead of "nothing was attempted".
- **not scheduled** → `not-applicable` only when the codehash gate, which read the same calldata, found nothing installed and nothing it could not open; otherwise `error`, since something is installed outside the only envelope this gate observes.
- **no reason given** (the unreadable-evidence fallback) keeps the blocking default.

Mirrors the out-of-scope channel gate I already has (`executabilityOutOfScope`). No default is loosened for EVM: the silent absence still blocks, and a read that fails on a covered network still blocks.

Placement is pinned in `check-ledger-wiring-placement.test.ts` (the absence reaches the recorder from `computeProposalEvidence`), and the four distinguishable rows are pinned in `confirm-check-registry.test.ts`. `docs/MultisigSigningProcess.md` §5 gains the gate-G row it was missing.

**Two consequences worth knowing:** a Tron proposal that bypasses the timelock (`propose-to-safe-tron.ts --direct`) is graded `not-scheduled` before the coverage check, so it blocks as "installs outside a timelock schedule" rather than being acknowledged — conservative on purpose. And a per-address RPC failure inside an observation that was made stays on the observation's own row (`NOT READ`, blocking); the new `read-failed` branch covers an observation that could not be made at all (envelope would not decode, no client, deployment file unreadable).

**Falsification on real data (gate review):** a real Tron `scheduleBatch(diamondCut)` built from `deployments/tron.json` reaches the out-of-scope branch and is stored as `needs-ack` on `A-DOCUMENTED` (not coerced, `hardBlocked` false); eleven real non-scheduled calldata shapes (direct cut, periphery registration, `executeBatch`, empty `0x`, unknown selector, removal-only, `_init` refusals, `updateDelay`) each land on the honest row — no installing calldata reaches not-applicable; the four rows render distinct, within 140 columns, with no internal ids.

**Real-row evidence:** see the EXSC-1043 comment of 2026-09-19 — four loopback walks of `confirm-safe-tx --network tron` (a real replayed removal-only cut, a real-facet Replace cut, a tampered Safe address, a foreign signature); gate G is `n/a` on the removal, `needs-ack` naming Tron on the install, and never printed twice. Two misses the walk found are fixed in this PR's last two commits.

**Evidence bar (was open):** the ticket asks for the three rows against a real Tron proposal. That needs the pending Tron rows dumped into the loopback store (proposal-store tunnel), and is tracked on EXSC-1062; this PR stays a draft until the render is attached.


0xDEnYO:

PR #2425, opened 2026-09-19
feat(safe): gates K and L stand down on Tron instead of reporting MISMATCH (EXSC-1062)

0xDEnYO:

[EXSC-1062](https://linear.app/lifi-linear/issue/EXSC-1062) — Tron: decide and build the sign-time gate coverage for TVM. This PR is the K + L leg. The decision table with the measurements behind it is the spike comment on that ticket; the open question (acknowledge vs keep refusing on Tron) is on EXSC-883 and this PR stays a draft until it is answered.

Measured on `origin/main` against the live Tron diamond and the production deployment-record store (read-only), gate K on a Tron facet cut today:

- refuses every cut as **UNVERIFIABLE** ("the production deployment record says nothing about this address") — the lookup is by EVM hex and the Tron records are keyed base58;
- with the lookup corrected in a probe (that correction is not part of this PR), reports a legitimately deployed, TronScan-verified facet (`LiFiIntentEscrowFacetV2` 1.0.0) as **MISMATCH** — the recorded commit exists only in the `contracts-tron` fork, and the toolchain scope resolves Tron to `solc_floor` (0.8.17/london, from `targetEvmVersion` in networks.json) while the record names 0.8.29/cancun. 38 of 44 Tron records carry no commit at all, and no attestation manifest covers TVM.

Neither verdict can ever reach MATCH from this repository, so the refusal is an outage on Tron rather than a gate, and its text sends the signer after a substitution that never happened.

The gate now decides coverage first (`resolveGateCoverage`, the same statement gate G and the pre-broadcast gate use) and, on an uncovered chain, compares nothing and does not refuse; the reason travels on the verdict as `outOfScope`. Gates K and L each grade that as one `needs-ack` row on `A-DOCUMENTED` naming Tron, so the signer takes the documented gap on explicitly. `CODEHASH_CHECK` opts into `undecidableIsAcknowledgeable` for exactly that case; the tests pin that a K `needs-ack` on any other anchor is still stored as `fail`, that a sibling proposal's MISMATCH still decides the network's row, and that the same cut on a covered chain still goes to the rebuild. The free-standing K block prints nothing for the stand-down, since the ledger row carries it.

A real TVM check (fork as source of truth, record-declared profile, base58 lookup, or a TronScan-verified anchor) is a follow-up ticket, not this PR.

**Real-row evidence:** see the EXSC-1062 comment of 2026-09-19 — on the replayed real removal-only Tron cut K and L are `n/a` like on every chain; on a Replace of a real Tron facet they each render one `needs-ack` row naming Tron and the signature is not refused by K. The walk found the first stand-down fired before the decode; fixed in `050e7efa7`.

**Hardening after the walk (`4a910d7d3`, `10e56f5cc`):** the stand-down applies only to a cleanly decoded, cleanly classified cut; a removal carrying an init target, an addition of the zero address, or a registration of the zero address is judged and refused on Tron like anywhere else (pinned by test).

**Evidence bar (was open):** the loopback walk of the dumped real Tron rows (needs the proposal-store tunnel) is tracked on EXSC-1062.


0xDEnYO:
Closing: the premise is disproven, and the honest answer is a real check rather than a stand-down.

A Tron facet's deployed code **does** reproduce from the `contracts-tron` fork. Rebuilt at the commit each production record names, compared as trailer-stripped, immutable-masked code against what is deployed:

| contract | address | commit | verdict |
|---|---|---|---|
| FeeForwarder 2.0.0 | TNZ3fznh… | ca21417ad | MATCH |
| ReceiverOIF 1.0.0 | TSZnVJnh… | 00b0546f3 | MATCH |
| LiFiIntentEscrowFacetV2 1.0.0 | TMck2qdZ… | 00b0546f3 | MATCH |
| AllBridgeFacet 2.2.0 | TR15epdw… | 00b0546f3 | MATCH |
| LayerSwapFacet 1.0.0 | TBFPs7mx… | a2a261f7a | MATCH |
| SymbiosisFacet 2.0.0 | TMY1N6rC… | 3e42b46e1 | **MISMATCH** |

Five of six reproduce exactly; the sixth is a finding a working gate would have caught, and is being investigated on its own.

Why this PR read it as impossible:

1. **The fork's source was already reachable.** The rebuild runner does `git worktree add --detach` out of the local clone, which carries the `tron` remote, so the fork commits check out with no new plumbing. The gate was rebuilding all along.
2. **`config/networks.json` declares Tron `targetEvmVersion: london`,** so the scope resolver picks the `solc_floor` profile at solc 0.8.17. Every Tron deployment record says 0.8.29 and cancun, and the deployed metadata trailer agrees. Under `solc_floor` the rebuilt runtime is a different length and can never match; under `default` it matches byte for byte.
3. **The record lookup missed:** Tron records are keyed base58, a cut carries hex, so the gate reported UNVERIFIABLE before reaching a rebuild.

The replacement makes gate K resolve Tron's lineage from the profile its records actually name and look the record up under the spelling Tron stores, and leaves L following K as it does everywhere. 38 of 44 Tron records carry no commit, so a cut installing one of those stays UNVERIFIABLE — honest, since the record does not say what to rebuild.

PR #2426, opened 2026-09-19
feat(safe): gate J names Tron as outside its reach instead of "no read was made" (EXSC-1062)

0xDEnYO:

[EXSC-1062](https://linear.app/lifi-linear/issue/EXSC-1062) — Tron: decide and build the sign-time gate coverage for TVM. This PR is the J leg (gate I's wording already names Tron). Decision table with the measurements: the spike comment on the ticket.

The quorum read is skipped on Tron (`evmSimulatable` is false there), and the ledger then printed "no quorum read was made for this proposal" — the wording of a covered chain the run failed to read, and it does not name Tron. Measured against both configured Tron providers with the read forced: TronGrid rejects `eth_getCode` at a pinned block ("Invalid parameters"), and the second provider fails the same read, so no two providers can be held to one height on TVM. That is a limit of the chain, not of the run, and the honest row says so.

A Tron proposal now carries `rpcQuorumOutOfScope` beside the executability scope note it already had, and `proposalCheckResults` grades it `needs-ack` on `A-DOCUMENTED` naming the chain. The unmade-read row a covered network gets keeps its own wording and anchor, so the two absences stay apart on the signer's screen; a quorum read that did run is never softened by the note. Gate J already prints once per network (`shownRpcQuorum`), so the Tron row does too. Wiring is pinned in `check-ledger-wiring-placement.test.ts` (both scope notes reach the recorder), the grading in `confirm-check-registry.test.ts`, and §5 of `docs/MultisigSigningProcess.md` records the limit.

**Real-row evidence:** see the EXSC-1062 comment of 2026-09-19 — on all four loopback Tron walks gate J renders once, `needs-ack`, naming Tron's pinned-block limit; gate I's note now sits on the same anchor.

**Evidence bar (was open):** the loopback walk of the dumped real Tron rows (needs the proposal-store tunnel) is tracked on EXSC-1062.


0xDEnYO:

PR #2429, opened 2026-09-19
fix(safe): the calldata address check finds a Tron facet's base58 record (EXSC-1062)

0xDEnYO:

[EXSC-1062](https://linear.app/lifi-linear/issue/EXSC-1062) — Tron gate coverage. Found by the real-row walk: the report-only calldata address check printed `× REFUSED call[0].diamondCut[0].cuts[0] (0x7fC2…) has no deployment record` for `LiFiIntentEscrowFacetV2`, a legitimately deployed and TronScan-verified Tron facet.

A Tron deployment record stores its address in base58 while a cut carries 20-byte EVM hex, so `buildDeploymentIndex` never matched a Tron record against the calldata and every Tron facet read as one nobody deployed. The target-state gate already looks a Tron address up under both spellings; this does the same for the address check without touching its decision module: `withCalldataSpellings` respells every record on the proposal's network the way the calldata spells it, in place (a copy beside the original would read as two addresses at one deploy time, a tie the name-anchored periphery lookup cannot decide — caught by the gate review, `9cb62ef6c`), through a translator the CLI builds from the Tron codec (`tronBase58ToEvm20Hex`). Other networks get no translator and pass through untouched; a record whose address cannot be translated is kept as it was; an unread record set stays unread. The placement test pins the translated set as what `buildDeploymentIndex` receives.


0xDEnYO:

PR #2430, opened 2026-09-21
fix(safe): the Tron signer render is honest per gate (EXSC-1043, EXSC-1062)

0xDEnYO:

Fixes [EXSC-1043](https://linear.app/lifi-linear/issue/EXSC-1043) (gate G) and fixes [EXSC-1062](https://linear.app/lifi-linear/issue/EXSC-1062) (gates I/J and the address check). Supersedes #2424, #2426 and #2429, which were three PRs over the same two files and conflicted pairwise; this is the same code as one review.

A proposal on the Tron Safe rendered as a broken environment: gate G printed an infrastructure-looking red for a chain its reader was never written for, gate J said "no quorum read was made" as though the run had failed to read a chain it could have, and the report-only address check called a deployed Tron facet one with no deployment record. None of that is a statement about the proposal.

Three changes, one per symptom:

- **Gate G** (`confirm-safe-tx.ts`, `confirm-check-registry.ts`). `observeSetForProposal` returned `undefined` for three different situations — a chain outside the reader's coverage, an observation that could not be made, and calldata that schedules no timelock batch — and all three graded as one blocking `error`. The read now returns why the set is absent, and each absence is graded on its own terms: an uncovered chain is acknowledged on `A-DOCUMENTED` with the chain named, and only when the calldata actually installs something (decoded from the calldata, since on Tron the codehash gate also compares nothing); an observation that could not be made stays unverified and blocks with the failure on the row; an unscheduled proposal is `not-applicable` only when the codehash gate read the same calldata and found nothing installed, and blocks otherwise.
- **Gates J and I**. The quorum read is skipped on Tron, and the row then borrowed the wording of a covered chain the run failed to read. Measured against both configured Tron providers with the read forced: TronGrid and dRPC alike reject `eth_getCode` at a pinned block, and both answer at `latest`, so no two providers can be held to one height on TVM — a limit of the chain, not of the run. The row now names Tron on `A-DOCUMENTED`, printed once per network as every J row is, and gate I's existing scope note moves to the same anchor so the two limits on one screen rest on the same evidence class.
- **The calldata address check**. A Tron deployment record stores base58 while a cut carries 20-byte hex, so no Tron record ever matched. Records on the proposal's network are respelt the way the calldata spells them before the lookup — in place, because a copy beside the original reads as two addresses at one deploy time, which the name-anchored periphery lookup reports as an undecidable tie.

No EVM or zkEVM default is loosened anywhere: the silent absence still blocks gate G, a failed read on a covered chain still blocks, an unmade quorum read on a covered chain keeps its own wording and anchor, and networks without a translator pass through the index untouched.

**Proven on real Tron proposals, not fixtures.** The Tron Safe has no pending rows (32 executed, on-chain nonce 30, threshold 3, six owners), so the executed nonce-29 row — a removal-only `scheduleBatch(diamondCut(Remove ×4))` — was replayed at nonce 30 with its `safeTxHash` recomputed on-chain, alongside a Replace of a real Tron facet built from `deployments/tron.json`, a tampered Safe address and a signature from a key that is not an owner. All four walked through `confirm-safe-tx --network tron --ledger=false` against a loopback store. Gate A catches the tampered Safe with the character diff, gate C names the foreign signer, gate G is `n/a` on the removal and `needs-ack` naming Tron on the install, gate J renders once, the address check reports the record, and no row carries an internal identifier. Renders and the row files are attached to the two tickets. Gate C refuses every walk because the replay copies carry no signature — this machine holds only the proposer's key — which is a property of the replay, not of Tron.

Walking those rows found three defects that the unit tests had not: gates K, L and G asked for an acknowledgement on a removal-only cut that installs nothing; gate G then read the codehash gate's empty Tron target list as "installs nothing" on a cut that does install; and the address check reported a real deployed facet as unrecorded. All three are fixed here and pinned by test.

# Review round (Goran, 2026-09-21)

- **Gate G's out-of-scope row overclaimed on bytes nobody read.** `calldataInstallsSomething` consulted only `calls` and `registrations`, so a Tron `scheduleBatch` carrying an unopenable frame with no `diamondCut` selector bytes — which raises no refusal — graded `not-applicable` with "this proposal installs no contract". `unopened` and `refusals` now answer true. Gate K blocked that input independently, so this was a wrong row rather than a signing hole.
- **The predicate is testable now.** Lifted out of the `processTxs` closure to `collectedInstallsSomething` in `safe-decode-utils.ts`, exported and pure over `ICollectedDiamondCuts`, with nine cases — Add/Replace/Remove, an `_init`-only cut, a registration, the zero-address unregistration from the CodeRabbit finding, an unopened frame, and a refusal.
- **A Tron signer gets the base58 back.** `withCalldataSpellings` keeps the record's own spelling on `recordSpelling` while `address` stays respelt for the lookup, and `describeEntry` prints it beside the hex, so the `NameMismatch` and "the record currently holds X" lines can be checked against `deployments/tron.json` and Tronscan.
- **The respelling runs once per network**, not once per proposal, via a promise memoised beside `tronRecordTranslator`.

**RPC measurement behind the gate-J row**, recorded here so a future reader can tell a still-true limit from a stale one — taken 2026-09-20 against the two providers `config/networks.json` configures for `tron`, TronGrid and dRPC: both reject `eth_getCode` at a pinned block height and both answer it at `latest`. That is why no two providers can be held to one height on TVM. Re-measuring is the way to retire the row; nothing in the code re-tests it.

Tests: `bun test script/deploy/safe/` — 3047 pass, 0 fail across 96 files.


script/deploy/safe/confirm-safe-tx.ts
```diff
@@ -576,6 +584,38 @@ const processTxs = async (
...
   const observedSets = new Map<string, IObservedSet>()
 
+  // Read from the calldata alone: a cut that adds or replaces a facet, sets
+  // an init target, or registers a periphery contract puts code into service.
+  // A registration to the zero address is the opposite — it removes one — so it
+  // is not counted, or a proposal that only unregisters would be asked to
+  // acknowledge an installation it does not make.
+  const calldataInstallsSomething = (callData: Hex): boolean => {
+    const collected = collectDiamondCutTargets(callData)
+    return (
```
gvladika: **`out-of-scope` doesn't guard the unknown install set, unlike `not-scheduled`.**

`calldataInstallsSomething` reads `collected.calls` and `collected.registrations` but ignores `collected.unopened` and `collected.refusals`.

On Tron, a `scheduleBatch` carrying a frame the decoder cannot open and no `diamondCut` selector bytes anywhere (so `collectDiamondCutTargets` emits no refusal — that list is keyed on `hex.includes(DIAMOND_CUT_SELECTOR.slice(2))`) yields `installs: false`. Gate G then prints _"this proposal installs no contract, so there is no authority to read"_ as `not-applicable`, when the honest answer is that nobody knows.

This is the exact asymmetry the `not-scheduled` branch guards against: `confirm-check-registry.ts:1404` consults `installSetUnknownPerCodehash`, which treats a non-empty `unopened` as "install set unknown", and the `out-of-scope` branch at `:1370` deliberately declines to. The stated reason (the codehash gate compares nothing on Tron) justifies not reading its `targets`, but not skipping the "could the bytes even be read" question, which is chain-independent.

Not a signing hole — gate K blocks that input on its own via `codehashCheckResult`'s `madeNoClaim && unopened.length > 0` path, so the run still ends `BLOCKED`. But for a PR whose thesis is that each gate says only what it knows, this row is overclaiming.

Suggest folding the unreadable case in, so an undecodable frame is never read as "installs nothing":

```suggestion
    const collected = collectDiamondCutTargets(callData)
    if (collected.unopened.length > 0 || collected.refusals.length > 0)
      return true
    return (
```

gvladika (commented):
Reviewed on the branch: ran `bun test script/deploy/safe/` (2993 pass, 0 fail across 95 files), `bunx tsc-files --noEmit` and `bunx eslint` on the four changed TS files — all clean. The gating wiring checks out: gate G is `integrity` with `undecidableIsAcknowledgeable: true` and `A-DOCUMENTED` is in `ACKNOWLEDGEABLE_REPORTING_ANCHORS`, so `survivesIntegrityRefusal` turns the Tron row into an acknowledgement rather than a block, exactly as described. Gates I and J were already `needs-ack`, so those two changes are wording and anchor only — the PR body is accurate about that. No EVM path is loosened.

Approve with the inline comments below. Only the first one I'd want addressed before merge; it's a small predicate change and it closes the last spot where a Tron row asserts more than it knows. The rest are follow-up-sized.

One housekeeping note: the red `validate-json` check is stale, not this PR — it fails on the duplicate `EcoFacet` key in `script/deploy/resources/immutableRegistry.json`, which #2422 already dropped on `main`. The branch is two commits behind; merging `main` turns it green.

script/deploy/safe/confirm-safe-tx.ts
```diff
@@ -576,6 +584,38 @@ const processTxs = async (
...
+        (call) =>
+          call.init !== ZERO_ADDRESS ||
+          call.cuts.some(
+            (cut) =>
+              cut.action === FacetCutActionEnum.Add ||
+              cut.action === FacetCutActionEnum.Replace
+          )
+      )
+    )
+  }
```
gvladika: **Testability: `calldataInstallsSomething` has no unit test and cannot get one.**

It is a closure inside `processTxs`, so nothing can reach it. It is now the function that decides whether gate G blocks or stands down on a Tron install — and the PR body says walking real proposals is what caught the previous bug here (reading the codehash gate's empty Tron target list as "installs nothing").

The ledger-side branches are well covered — 13 new cases in `confirm-check-registry.test.ts`, including the four-way "grades the three absences and the silent one as four different rows". The gap is the other half of the decision, the one derived from calldata.

Worth lifting it beside `withCalldataSpellings` as an exported pure function over `ICollectedDiamondCuts`; it needs no `processTxs` scope. That also gives the zero-address unregistration case from the CodeRabbit finding a regression test, which right now rests on the manual replay alone.

script/deploy/safe/deployment-record-spellings.ts
```diff
@@ -0,0 +1,39 @@
...
+  records: readonly IDeploymentIndexEntry[] | undefined,
+  network: string,
+  translator: IRecordSpellingTranslator | undefined
+): readonly IDeploymentIndexEntry[] | undefined => {
+  if (!records || !translator) return records
+  const wanted = network.trim().toLowerCase()
+  return records.map((entry) => {
+    if (entry.network.trim().toLowerCase() !== wanted) return entry
+    const spelt = translator.toCalldataSpelling(entry.address)
+    return spelt ? { ...entry, address: spelt } : entry
```
gvladika: **The respelling loses the record's base58, which is what a Tron signer would cross-check.**

Respelling in place is the right call for the lookup — the reasoning about the name-anchored tie is correct, and `deployment-record-spellings.test.ts` pins it. The cost is on the render side: `calldata-address-check.ts` prints `entry.address` verbatim at `:543` (the `NameMismatch` line) and `:610` ("the record currently holds X under that name"), so a Tron signer now reads `0x7fc2…` where `deployments/tron.json` and Tronscan say `TMck2q…`.

Unlike the calldata addresses elsewhere in the signer view, these strings don't go through `tronHexSuffix`, so there's no base58 anywhere on the row. Before this change the periphery path printed base58 (it just never matched), so this is a real, if small, step down in what the signer can verify against an explorer.

Display-only gate, so low stakes. Cheapest fix that keeps the lookup semantics is carrying the original spelling on the entry — e.g. keep `address` respelt and add the record's own spelling as a separate field the two messages above can print beside it — rather than touching the comparison.

script/deploy/safe/confirm-safe-tx.ts
```diff
@@ -1251,7 +1321,7 @@ const processTxs = async (
           ...(undecodable.length > 0 ? { undecodable } : {}),
         },
         buildDeploymentIndex(
-          records,
+          withCalldataSpellings(records, network, tronRecordTranslator),
```
gvladika: Nit: this runs once per proposal, but nothing in it varies per proposal. `records` is memoised as a single promise over the whole production fleet (`readDeploymentRecords` → `deploymentRecordsRead ??=`), and `tronRecordTranslator` is built once per network at line 917 — so every iteration maps the same fleet-sized array to the same result, re-running the TronWeb base58 decode on each Tron row.

Hoisting it next to the translator, or memoising on the records promise, makes it one pass instead of one per proposal. Immaterial in absolute terms; it just reads oddly next to the comment a few lines below the translator that explains hoisting the per-network reads out of the loop for the same reason.

script/deploy/safe/confirm-safe-tx.ts
```diff
@@ -1698,6 +1772,9 @@ const processTxs = async (
       ...(isTronNetworkKey(network)
         ? {
             executabilityOutOfScope: `${network} is executed through its own chain executor, which the EVM simulator does not cover`,
+            // TronGrid answers `eth_getCode` only at `latest`, so the pinned
+            // block every provider is held to cannot be asked for there.
+            rpcQuorumOutOfScope: `${network} serves a code read only at the latest block, so its providers cannot be compared at one height`,
```
gvladika, line 1779: This bakes a one-off measurement into a permanent waiver: the row, and the matching sentence in `docs/MultisigSigningProcess.md`, will keep asking for an acknowledgement on `A-DOCUMENTED` whether or not TronGrid and dRPC still behave this way. Nothing re-tests the claim, and a provider that later serves `eth_getCode` at a pinned block wouldn't change the row.

Fine as it stands — the gate is report-only and both the old and new statuses are `needs-ack`, so nothing turns on it. Just worth recording the measurement (date, endpoints, what was asked) on EXSC-1062 so a future reader can tell a still-true limit from a stale one, rather than leaving the evidence only in the PR description.

script/deploy/safe/confirm-safe-tx.ts
0xDEnYO: Folded in, as suggested. Proven on the exact shape: a real viem-encoded \`scheduleBatch\` whose one payload is an unknown selector, with no \`1f931c1c\` bytes anywhere, gives \`unopened:["0xabcdef12"]\` and empty \`refusals\`/\`calls\`/\`registrations\` — the old predicate answered false, the new one answers true, so the row is \`needs-ack\` rather than \`not-applicable\`.

script/deploy/safe/confirm-safe-tx.ts
0xDEnYO: Lifted beside the collector it reads rather than beside \`withCalldataSpellings\` — it is pure over \`ICollectedDiamondCuts\`, so \`safe-decode-utils\` is where it belongs. Nine cases now, including the zero-address unregistration from the CodeRabbit finding.

script/deploy/safe/deployment-record-spellings.ts
0xDEnYO: Carried on \`recordSpelling\`, exactly as you framed it: \`address\` stays respelt for the lookup, and \`describeEntry\` prints the record's own spelling beside the hex, so both the \`NameMismatch\` line and "the record currently holds X" now show the base58. The \`!== address\` guard matters — a Tron record already stored as hex round-trips to itself and would otherwise print the same string twice.

script/deploy/safe/confirm-safe-tx.ts
0xDEnYO: Hoisted next to the translator and memoised on the records promise, so it is one pass over the fleet instead of one per proposal.

script/deploy/safe/confirm-safe-tx.ts
0xDEnYO: Recorded in the PR description with the date and both endpoints, so the row can be told apart from a stale one. The EXSC-1062 comment is still to come — the Linear connector is not reachable from this session.

gvladika (approved):

PR #2431, opened 2026-09-21
refactor(tron): drop the dead verifyFacetRegistration helper

0xDEnYO:

None — follow-up to #2327 (EXSC-920).

`verifyFacetRegistration` lost its only caller when #2327 removed `registerFacetToDiamond`. Nothing has called it since on `main`, nor on either open Tron branch (#2409, #2427), which both carry it unreferenced at the same line.

Three older open branches (#2092, #2417, #1940) do still contain a caller, in their own copy of `tronUtils.ts`. They are 85–413 commits behind and simply predate #2327, which deleted the calling `registerFacetToDiamond`; merging `main` resolves each to that deletion. No in-flight work depends on the helper.

No post-cut verification is lost with it. The live registration path verifies inline and more thoroughly: `register-facets-to-diamond.ts:256` re-reads the loupe after each cut and only records a facet in `tron.diamond.json` if it actually shows up, and the "Final verification" pass at `:531` diffs the whole facet list against the expected set and warns on anything missing. The deleted helper returned a boolean for one facet.

Removing the function leaves `tryTronFacetLoupeAddressToBase58` unused in this file; `register-facets-to-diamond.ts` imports it separately and is untouched. The file header no longer claims facet/whitelist verification, since `checkOwnershipTron` is the only health check left in it.

- [ ] I have added tests that cover the functionality / test the bug — removal only; no test referenced the function


0xDEnYO:
Folded into #2433 — its commits are now on that branch, so this PR is closed to keep the review queue to one unit per subsystem. No work is lost.

PR #2428, opened 2026-09-19
chore(immutables): drop the duplicated EcoFacet entry from the registry (EXSC-1053)

0xDEnYO:
# Which Linear task belongs to this PR?

[EXSC-1053](https://linear.app/lifi-linear/issue/EXSC-1053) — the duplicate-JSON-key check that PR #2414 added. Found because `validate-json` is red on every open PR.

# Why did I implement it this way?

`script/deploy/resources/immutableRegistry.json` declares `EcoFacet` twice (lines 2 and 122 on `main`), with identical bodies — the second copy arrived with the EcoFacet v2.0.0 rollout merge (#2406). Every JSON parser keeps the last one, so nothing behaved differently, but the check that landed the same day (#2414) now fails `validate-json` on every pull request. This removes the second copy; `verify-json-duplicate-keys.ts` reports 0 duplicates across 215 files afterwards, and the surviving entry is byte-identical to the one removed.

# Checklist before requesting a review

- [x] I have performed a self-review of my code
- [x] This pull request is as small as possible and only tackles one problem
- [ ] I have added tests that cover the functionality / test the bug
- [ ] For new facets: I have checked all points from this list: https://www.notion.so/lifi/New-Facet-Contract-Checklist-157f0ff14ac78095a2b8f999d655622e
- [ ] I have updated any required documentation

# Checklist for reviewer (DO NOT DEPLOY and contracts BEFORE CHECKING THIS!!!)

- [ ] I have checked that any arbitrary calls to external contracts are validated and or restricted
- [ ] I have checked that any privileged calls (i.e. storage modifications) are validated and or restricted
- [ ] I have ensured that any new contracts have had AT A MINIMUM 1 preliminary audit conducted on <date> by <company/auditor>

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2423, opened 2026-09-19
fix(config): drop the duplicate EcoFacet entry from the immutable registry (EXSC-1053)

0xDEnYO:
# Which Linear task belongs to this PR?

Relates to EXSC-1053

# Why did I implement it this way?

`script/deploy/resources/immutableRegistry.json` declares `EcoFacet` twice, at line 2 and line 122, with byte-identical bodies. Every JSON parser keeps the last, so the parsed registry has always been correct and nothing behaved differently — but the duplicate-key check added in #2414 refuses the file, and `validate-json` is red on `main` itself:

```text
$ bunx tsx ./script/deploy/resources/verify-json-duplicate-keys.ts config deployments \
    audit/auditLog.json script/deploy/_targetState.json script/deploy/resources script/deploy/immutables
[error] script/deploy/resources/immutableRegistry.json:122 declares 'EcoFacet' again under $.
[error] 1 duplicate key(s) across 215 file(s).
```

That gates every open PR, which is how it surfaced — #2413 went red on it with no JSON of its own in the diff.

The copy at line 2 is the stray: the file is alphabetical once it is ignored, and the one at 122 sits between `DeBridgeDlnFacet` and `EmergencyPauseFacet` where it belongs. Removing the first leaves the file alphabetical and the parsed registry byte-identical, since that is the one a parser was already discarding.

After: `✔ No duplicate keys in 215 JSON file(s).`, and `bun test script/deploy/immutables/ script/deploy/codehash/immutable-expectations.test.ts` is 156 pass / 0 fail.

# Checklist before requesting a review

- [x] I have performed a self-review of my code
- [x] This pull request is as small as possible and only tackles one problem
- [ ] I have added tests that cover the functionality / test the bug
- [ ] For new facets: I have checked all points from this list: https://www.notion.so/lifi/New-Facet-Contract-Checklist-157f0ff14ac78095a2b8f999d655622e
- [ ] I have updated any required documentation

# Checklist for reviewer (DO NOT DEPLOY and contracts BEFORE CHECKING THIS!!!)

- [ ] I have checked that any arbitrary calls to external contracts are validated and or restricted
- [ ] I have checked that any privileged calls (i.e. storage modifications) are validated and or restricted
- [ ] I have ensured that any new contracts have had AT A MINIMUM 1 preliminary audit conducted on <date> by <company/auditor>

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2432, opened 2026-09-21
fix(deploy): remove the duplicate EcoFacet key from the immutable registry

0xDEnYO:

None — unblocks `validate-json`, which is currently red on every PR opened since #2414.

`script/deploy/resources/immutableRegistry.json` declares `EcoFacet` twice, at the top of the file and again above `EmergencyPauseFacet`. #2406 (EcoFacet v2.0.0 rollout) and #2393 (signing 2.0) each added the entry; both merged, and neither noticed the other. #2414 then added the duplicate-key check that surfaces it, so `validate-json` now fails on PRs that touch nothing near this file — #2427 and #2431 are both red on it.

The two blocks are identical. Every JSON parser keeps the key at its **first** position with the **last** occurrence's value, so removing the second copy is inert: I verified `JSON.parse` output and key order are unchanged, with `EcoFacet` still at index 0.

Verification:

- `verify-json-duplicate-keys.ts` over the same paths the workflow passes it: `✔ No duplicate keys in 215 JSON file(s)` (was 1 duplicate)
- `bun test script/deploy/immutables/ script/deploy/codehash/` — 598 pass, 0 fail

- [ ] I have added tests that cover the functionality / test the bug — the check added in #2414 already covers it


0xDEnYO:
Duplicate of #2422, which is older, ready and carries a byte-identical diff. Closing this one; #2422 is the survivor (same precedent as #2423 and #2428).

PR #2433, opened 2026-09-21
fix(safe): grade Tron under gates K and L (EXSC-1067)

0xDEnYO:

Fixes [EXSC-1067](https://linear.app/lifi-linear/issue/EXSC-1067)

#2430 has merged and this is now based on `main`. Its inline Tron translator is
replaced here by the shared `createTronAddressSpellings`, which adds base58/hex
shape guards: a record field holding a base58-alphabet placeholder (`n/a`,
`true`, `TODO`) used to decode to the zero address and now grades untranslatable.
Differential run over all 55 values in the four Tron deployment files: zero
disagreements on any of the 26 real addresses, in any spelling.

Gate K was believed impossible on Tron and a stand-down was built for it
(#2425, closed). It is possible: a Tron facet's deployed code reproduces exactly
from the `contracts-tron` fork. Six of the 44 Tron production records carry a commit; five of them reach MATCH
on gates K **and** L, on real chain and record data.
The sixth is [EXSC-1068](https://linear.app/lifi-linear/issue/EXSC-1068).

Four things stood between the gate and that answer.

**1 · The source is on the fork, not on `origin`.** `ensureCommitAvailable` only
ever fetched from `origin`, so a Tron commit was readable purely by luck — if
the operator's clone happened to carry the fork as a remote. A network now
resolves its own source remote (`script/deploy/codehash/source-remote.ts`), and
`createAttestationSource` takes that resolution as a required dependency rather
than defaulting to `origin`: a source built without it answers UNVERIFIABLE for
every Tron address while looking complete.

The remote's URL is checked with the existing `isTrustedRemote` /
`REPO_CONTRACTS_TRON` pair before anything is fetched through it. That is not
ceremony — a `tron` remote pointed at a proposer's own fork would have the gate
rebuild from source they wrote and reach MATCH on it, which is exactly what
`isTrustedRemote`'s docstring exists to refuse. A clone with no such remote, one
naming another repository, and one over cleartext are three separate refusals,
each naming the `git remote add` to run.

The resolution happens **after** the record has yielded a commit, not before.
An address the record is silent about, and a record too old to carry a commit,
stay the answers they already were rather than becoming an infrastructure error
about a rebuild nobody was going to run.

**2 · `config/networks.json` said tron was `london`.** So `deriveToolchainScope`
picked `solc_floor` at solc 0.8.17, whose build is a different length and can
never match. Every Tron record and every deployed metadata trailer says 0.8.29 /
cancun / 1000000 runs.

The ticket offered "resolve the lineage from the record's declared pair"
instead. Rejecting that: `rebuild-attestations.ts` already records why it is
unsafe — fourteen production slots declare 0.8.29 for code the compiler stamped
0.8.17, so a profile taken from the record grades honest deploys rogue, and it
hands the proposer a second axis to search on.

The row is simply false, and the repo says so itself:
`audit-recorded-constructor-args.ts` reads Tron artifacts from `OUT_ROOT`, which
resolves to `out/` — `[profile.default]`'s `out` — and its own error text says
*"Run 'forge build' with the default profile before auditing"*. Nothing on the
Tron deploy path sets `FOUNDRY_PROFILE`. Blast radius, checked rather than
assumed:

- `deploy-safe-tron.ts` takes its Safe v1.4.1 artifacts from the **hardcoded**
  `TRON_SAFE_ARTIFACTS_BASE = safe/london/out`, not from the config row — only
  its docstring claimed otherwise, and that is fixed here;
- `deploy-safe.ts` and the bash deploy grouping read the row, but Tron never
  goes through either; it deploys and proposes through `script/deploy/tron/*`;
- `EVM_VERSIONS` is unchanged — fuse, mantle, nibiru, telos and viction keep
  `london`, and `lineage-scope.test.ts` now asserts `fuse` by name so a future
  config edit that moved the last of them fails instead of going vacuous.

`tronshasta` moves with `tron`: same fork, same scripts, same profile.

**3 · Records key Tron addresses in base58, cuts carry hex.** So the record
lookup found nothing and the gate reported UNVERIFIABLE before it ever rebuilt.
#2430 solved the same miss for the report-only address check; this adds the
inverse direction beside it, in `script/deploy/shared/tron-address-spellings.ts`
so `confirm-safe-tx.ts` and the codehash deps build one object instead of two.
The Mongo lookup offers both spellings, and the existing case-insensitive
fallback still applies to the hex only — base58check is case-sensitive, so
folding case there could match an address that is not the one asked about.

**4 · Gate L then blocked what gate K had matched.** `config/` and
`deployments/` write Tron addresses in base58 too, and `paddedToSlot` refused
such a value rather than comparing it — correctly, at the time, because padding
base58 yields a value no slot can hold and that read as a disagreement. It is
now translated through the same module and compared. Network-gated rather than
shape-gated: off Tron nothing spells an address that way, so translating there
would manufacture an expectation from a config entry that is genuinely
unreadable. A value that is still not hex afterwards stays unpriced.

## Evidence

Gate K and L through `verifyCutTargets` with the production wiring — real Mongo
production records, real TronGrid reads, real forge rebuilds from the fork:

| contract | address | verdict |
| -- | -- | -- |
| FeeForwarder 2.0.0 | `0x8a07dd…fcc5` | **MATCH** |
| ReceiverOIF 1.0.0 | `0xb60c1b…a3f4` | **MATCH** |
| LiFiIntentEscrowFacetV2 1.0.0 | `0x7fc2ad…8c7e` | **MATCH** |
| AllBridgeFacet 2.2.0 | `0xa4e495…a3e1` | **MATCH** |
| LayerSwapFacet 1.0.0 | `0x0e07d9…8b3b` | **MATCH** |
| SymbiosisFacet 2.0.0 | `0x7edd65…64da` | MISMATCH (EXSC-1068) |

Four of those reached MATCH only after change 4; before it they were
`UNVERIFIABLE` with "N bytes holding immutables include N with no declared
expectation".

It can still fail. Same wiring, one field changed at a time:

| case | verdict |
| -- | -- |
| the FeeForwarder record moved to `91e5faa5a`, the commit before v2.0.0 landed | **MISMATCH** |
| the same run immediately afterwards, at the commit the record names | MATCH |
| AllBridgeFacet's deployed code judged against the LiFiIntentEscrowFacetV2 record | **MISMATCH** |
| an address with no code (one hex nibble flipped) | **UNVERIFIABLE** |
| `EmergencyPauseFacet` — one of the 38 Tron records carrying no commit | **UNVERIFIABLE**, naming the record and the missing commit |

That last row is the honest outcome the ticket asks to keep, not a gap.

The two remote refusals, against real `git` in a throwaway clone:

```text
clone without the fork remote, tron:
  tron is built and deployed from github.com/lifinance/contracts-tron, and this
  clone has no "tron" remote, so there is no source to rebuild its code from.
  Add it and fetch it: git remote add tron
  https://github.com/lifinance/contracts-tron.git && git fetch tron

clone whose fork remote points elsewhere, tron:
  … but this clone's "tron" remote names github.com/attacker/contracts-tron. …
```

No EVM verdict changes: `mainnet` EcoFacet, `gnosis` TokenWrapper and `arbitrum`
EcoFacet each still MATCH end to end on their real records. The two zkEVM rows I
sampled returned the gate's own "zkEVM toolchain problem, not a codehash
verdict" — `foundry-zksync` is not installed in this worktree — so I am not
claiming a zk verdict from them; no code on the zk path changed, and the unit
suite covers it.

Finally, end to end through `bun confirm-safe-tx --network tron --ledger=false`
against the loopback rehearsal store, on a Replace of a real Tron facet:

```text
  ✓  K    Installed bytecode matches the attested build ...... ok
  ✓  L    Immutable values match what config declares ........ ok
```

On 2026-09-19 the same row rendered `ASKS YOU / REVIEW` for both. Gate C still
refuses, because the replay copy carries no signature — a known artefact of
copying an executed row, not a finding.

## Known limits, named rather than left to be found

**Tron's fleet is not one toolchain.** The scope closes around `default`, which is
right for everything the fork has deployed since it moved to solc 0.8.29 — all six
records carrying a commit reproduce under it. But the deployed
`WhitelistManagerFacet` at `TViNVAJ…` reports **solc 0.8.17** in its own trailer,
and two of the 44 Tron records declare that version. They are harmless today
because they carry no commit, so gate K answers UNVERIFIABLE before a profile is
consulted — and admitting a second profile is not available as a remedy, because
`createImmutableReferencesResolver` throws on a multi-profile scope ("offsets are
per lineage"), which would make every Tron address UNVERIFIABLE instead. Every
network resolves to exactly one profile today and this PR adds a test asserting
that stays true. The constraint this puts on backfilling commits is written up on
EXSC-1064.

**`config/networks.json` syncs into the fork.** `contracts-tron` still carries
`london` for tron and tronshasta, and its `syncUpstreamContracts` workflow merges
upstream `main` wholesale, so this flip will land there and move tron out of its
london deploy group in the repo where Tron is actually deployed. Given the records
say cancun / 0.8.29 was used, that is a correction rather than a regression — but
it is a change on the fork side that nobody asked for, so it is called out here.

**Top-level `@lifi/tron-devkit` import.** `.agents/rules/200-typescript.md` asks for
`await import()` for chain-specific modules. `source-remote.ts` and
`immutable-expectations.ts` import `isTronNetworkKey` and the codec factory at the
top level, because both functions are synchronous and making them async to satisfy
the rule would be a far larger change than the one-line predicate warrants. This
matches every other Tron branch in `script/deploy/safe/`, which imports the devkit
the same way. Worth noting that the rule points `[CONV:TRON-NETWORK-KEY]` at
`script/deploy/shared/tron-network-keys.ts`, a file that no longer exists.

# Folded-in PRs

Three separate PRs were opened for this subsystem by parallel sessions and are
now closed into this branch, so the review is one unit rather than four:

- **#2431** — deletes the dead `verifyFacetRegistration` helper in `tronUtils.ts`.
  Follow-up to #2327, which removed its only caller. No Linear ticket.
- **#2434** — points the Tron rules at `@lifi/tron-devkit` and scopes the
  dynamic-import rule (also EXSC-1067).
- **#2435** — makes the deployment-record read refuse an address whose records
  disagree, instead of picking one by `timestamp` (EXSC-1062).

**Where #2435 and this PR collided, and how it was resolved.** Both rewrote
`createMongoRecordSource`. This PR had added the Tron base58 spellings to an
exact-match tier that ran before a case-insensitive fallback; #2435 deleted that
two-tier shape on purpose, because a corrupt row stored in the other casing sat
behind a clean exact match and was never compared against it.

The merged form keeps both properties in one query: the calldata spelling is
matched case-insensitively, and the base58 spellings are matched exactly, under
a single `$or` whose whole result goes through `resolveDeploymentRecord`. Base58
is never case-folded — base58check is case-sensitive, so folding it would match
an address that is not the one asked about.

Worth a hard look there: `createMongoRecordSource` has no test on either
branch — it is the production default and every test injects `IRecordSource`
instead — so that `$or` is the one thing here resting on reading rather than on
a green assertion.


0xDEnYO:
@coderabbitai review

0xDEnYO:
@coderabbitai review

0xDEnYO:
@coderabbitai review

0xDEnYO:
@coderabbitai review

0xDEnYO:
@coderabbitai review

script/deploy/codehash/immutable-expectations.ts
```diff
@@ -151,6 +152,36 @@ const refused = (reason: string): IPricingRefused => ({
...
+ * @param declared - the value as `config/` writes it
+ * @param network - key in `config/networks.json`
+ */
+const asInlinedSpelling = (declared: string, network: string): string => {
+  const value = declared.trim()
+  if (HEX_VALUE.test(value)) return value
+  try {
+    return (
+      createTronAddressSpellings(network)?.toCalldataSpelling(value) ?? value
+    )
```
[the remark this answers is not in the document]
0xDEnYO: Valid — fixed in 6eb96ab.

The translator now runs before any hex value is passed through, so Tron's native `41`-prefixed spelling resolves to the 20 bytes the compiler inlined instead of faulting the slot at 21. Not reachable from the tree today (no `41…` value in `config/` or `deployments/`), but `tron-address-spellings.ts` admits that spelling by design, so the ordering was wrong regardless.

Taken slightly differently from the proposal: with the translator first, `HEX_VALUE` was only the short-circuit that caused the bug — it declines every non-Tron network and every value it cannot read, so the generic path falls through unchanged — and the constant is deleted rather than left unused. New test `compares a Tron 41-prefixed config value against the slot it fills`; it fails on the previous ordering.

script/deploy/safe/codehash-sign-gate-deps.ts
```diff
@@ -1262,6 +1271,146 @@ let client: MongoClient | undefined
...
+  // Every candidate, not `kept`: the blank-version collapse above drops the row
+  // the verification step rewrote, and that row is the one whose commit was
+  // actually verified. Comparing only what survives the collapse lets a row
+  // that can fill in a version outrank the row that was checked.
+  const commits = new Set(
+    candidates.map((r) => text(r.gitCommitHash)).filter((c) => c !== '')
+  )
+  if (commits.size > 1) refuse('which commit built what is', [...commits])
+
+  return (kept.find((r) => text(r.gitCommitHash) !== '') ?? kept[0]) as T
```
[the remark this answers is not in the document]
0xDEnYO: Valid — fixed in 6eb96ab.

`UNKNOWN` is what the record writer stores when it could not read a commit, and both downstream readers (`refsFor`, the off-code immutables reader) already fold it into absence, so counting it here as a claim was the outlier: a pair holding `UNKNOWN` plus one real SHA refused as a conflict, and `UNKNOWN` could win the `kept.find` over the SHA. Both now go through one `namesCommit` predicate.

New test `treats an UNKNOWN commit as absent rather than as a second claim`, asserted in both orders; it fails without the change.

gvladika (commented):
Two findings that can each produce a false blocking verdict. Both are narrow, and neither is covered by a test. (A third, non-blocking note — `buildRecordQuery` dropping the indexed exact-address match, so every record read is now a regex no index can serve — is a throughput regression only, so I've left it out of the inline set.)

script/deploy/safe/codehash-sign-gate-deps.ts
```diff
@@ -1262,6 +1271,156 @@ let client: MongoClient | undefined
...
+    const commit = text(record.gitCommitHash)
+    return commit !== '' && commit !== UNKNOWN_COMMIT
+  }
+
+  const commits = new Set(
+    candidates.filter(namesCommit).map((r) => text(r.gitCommitHash))
+  )
+  if (commits.size > 1) refuse('which commit built what is', [...commits])
+
+  return (kept.find(namesCommit) ?? kept[0]) as T
```
gvladika: The commit comparison above deliberately runs over `candidates`, but the pick returns from `kept` only — so the row whose commit was verified can be validated and then dropped.

When the blank-version row is the one carrying the commit and its versioned twin carries none, the `named`/`kept` collapse discards the blank row, `kept.find(namesCommit)` finds nothing, and `kept[0]` comes back with `gitCommitHash: ''`. `createAttestationSource` then answers `no-commit` → UNVERIFIABLE, blocking a cut whose single unambiguous commit the resolver had already seen and accepted.

That is precisely the split the comment four lines up describes ("the verification step rewrites the row it just verified and loses `version`"), and the comment argues the collapsed row's commit is the authoritative one — the return statement doesn't honour it.

Not caught by the suite: every `row({ version: '' })` fixture inherits the default `gitCommitHash`, so the versioned twin always has one too and the combination never arises.

script/deploy/shared/tron-address-spellings.ts
```diff
@@ -0,0 +1,103 @@
...
+ * immutable legitimately holding zero would satisfy it.
+ */
+const TRON_BASE58 = /^T[1-9A-HJ-NP-Za-km-z]{33}$/
+
+/**
+ * The hex spellings a record may carry instead: `0x` and 20 bytes, Tron's own
+ * `41` prefix and 20 bytes, or the 20 bytes bare. Admitted because the codec
+ * normalises all three and a record written in any of them names an address.
+ */
+const TRON_HEX = /^(0x|41)?[0-9a-fA-F]{40}$/
```
gvladika, line 51: The bare 20-byte form is admitted here on the grounds that "the codec normalises all three" (line 49). It doesn't — `tronBase58ToEvm20Hex` → `tronAddressToHex` unconditionally strips a leading `41` after `tronWeb.address.toHex`, which is a no-op on an already-unprefixed string. So a bare 20-byte address whose *first byte* is `0x41` has that byte eaten and the result left-padded.

Run against this branch:

```text
shape                              TRON_HEX  input                                       result
bare 20-byte (first byte 0x41)     true      41ababababababababababababababababababab    0x00ababababababababababababababababababab  ✗
0x + 20-byte                       true      0x41ababababababababababababababababababab  0x41ababababababababababababababababababab  ✓
41 + 20-byte (canonical tron hex)  true      4141ababababababababababababababababababab  0x41ababababababababababababababababababab  ✓
```

The other two round-trip; the bare form resolves to a **different address**. Via `withCalldataSpellings` that silently misattributes the record in the calldata address check; via `asInlinedSpelling` it makes gate L compute a wrong expectation and report `disagrees` — a blocking false MISMATCH, which is the one outcome this PR's gate-L work exists to avoid.

1-in-256 of bare-hex records by first byte, so narrow — but the regex admits the shape, so it's reachable. Either drop the bare alternative, or `0x`-prefix the value before handing it to the codec.

script/deploy/safe/codehash-sign-gate-deps.ts
0xDEnYO: Valid — fixed in 120fa18ef.

You're right that the two halves disagree: the commit is compared across every candidate precisely because the collapse drops the row the verification step rewrote, and then the pick ignores that row anyway. `commits.size > 1` has already refused above, so at that point there is exactly one commit claim and it is unambiguous; the chosen row now carries it, which keeps both halves — the version the collapse exists to preserve, and the commit it agreed on.

New test `keeps the commit when the collapse drops the only row naming it`, built from your repro (blank row holds the commit, versioned twin holds none); it fails on the previous return.

It returns a shallow copy in that one case, so I checked what that costs: the sole non-test caller projects the record to three scalars immediately (`createRecordReader`), nothing relies on object identity, and a BSON round-trip confirms a driver document spreads losslessly — no non-enumerables, no getters, `_id`/`Date` survive by reference. Also brute-forced every 2- and 3-row subset of a 24-row fixture space: 56 grafts, zero where the donor's `contractName` differed from the kept row's — `identities.size === 1` over `kept` forces a single name on every non-throwing path.

script/deploy/shared/tron-address-spellings.ts
0xDEnYO: Valid — fixed in 120fa18ef, and the reverse direction you implied was also broken, fixed in bae38483d.

Your table reproduces exactly. `tronWeb.address.toHex` prepends `41` to a `0x`-prefixed value but passes a bare one through, so the unconditional strip in `tronAddressToHex` only ever eats a real byte in the bare form. Took the second of your two options — `0x`-prefix the bare shape before the codec sees it — so the shape the regex admits stays admitted and is read correctly; the docstring's "the codec normalises all three" claim is gone, since it was the thing that wasn't true.

Swept all 256 first bytes × 9 spellings (bare / `0x` / `41`, each lower, upper, whitespace-wrapped): 2304 cases, 3 changed, 0 newly wrong — the three spellings of the one broken case, nothing else. Re-ran all 25 base58 addresses in `deployments/tron.json` through every spelling: no regression, and none of them starts with byte `0x41`, so this was latent rather than live. `41`+40 hex is 42 chars and bare is 40, so the two never collide.

**`forCalldataAddress` had the mirror bug.** `evmHexToTronBase58` skips its `41` prepend when the hex already starts with those digits, so `0x41ab…ab` encoded 20 bytes instead of 21 and came back as `6zEaMLH1Xof…` — not even a `T` address, and it decodes to a different one. A miss rather than a misattribution, but the module advertises both directions and that one was offering a value naming someone else. It now offers the base58 only if it decodes back to the address asked about. I could not make it produce the *correct* spelling here: the 21-byte value that would fix it is rejected by the encoder's own length guard, so the real repair is in `@lifi/tron-devkit` and I'll raise it there.

Two sibling call sites carry the same unfixed pattern — `confirm-integrity-asserts.ts:1044` and `propose-to-safe-tron.ts:256` both feed raw deployment-log values to `tronBase58ToEvm20Hex` with neither the shape guard nor the normalisation. Unreachable today (`deployments/tron.json` holds 25 base58 values and zero hex), so I've left them out of this PR rather than widen it; ticketing separately.

gvladika (approved):

PR #2434, opened 2026-09-21
docs(rules): point Tron helpers at @lifi/tron-devkit, scope the dynamic-import rule (EXSC-1067)

0xDEnYO:

Fixes [EXSC-1067](https://linear.app/lifi-linear/issue/EXSC-1067)

Two stale claims in `.agents/rules/200-typescript.md`, found while working on EXSC-1067. Both are rule-vs-code disagreements, so each was settled against the call sites rather than by rewriting the rule to taste.

**1. `isTronNetworkKey()` pointed at a file that no longer exists.**

`script/deploy/shared/tron-network-keys.ts` was deleted in the `@lifi/tron-devkit` migration (EXSC-253, #1739). All 12 call sites in the repo import the guard from the package. The rule now names `@lifi/tron-devkit`.

**2. "Dynamic imports for chain modules" — narrowed the rule rather than changing the practice.**

The rule said chain-specific modules such as TronWeb must be loaded with `await import()`. Every Tron branch under `script/deploy/safe/` top-level imports `@lifi/tron-devkit`, so the rule and the code disagreed everywhere. The code is right and the rule was overbroad:

- The devkit publishes exactly two entry points, `.` and `./safe`. There is no tronweb-free subpath — every deep-import escape hatch (`@lifi/tron-devkit/dist/utils/networkKeys`, `.../dist/index.js`, `.../package.json`) is refused with `ERR_PACKAGE_PATH_NOT_EXPORTED`. And `.` statically re-exports `TronContractDeployer` and the TronWeb factory (`dist/helpers/tronWebFactory.js` line 1 is `import { TronWeb } from 'tronweb'`), so any import of the package pulls tronweb. Measured under `bunx tsx`: a bare `import { isTronNetworkKey }` loads 10 CJS files from the `tronweb` package itself (55 unique files, 131 load calls, once tronweb's own nested deps are counted), costing ~0.4–0.5 s steady-state in-process. Deferring that import moves the cost, it does not avoid it — the rule as written was unsatisfiable for these helpers.
- `isTronNetworkKey` is the Tron-vs-EVM branch itself, not a Tron-only dependency. `normalizeAddressForNetwork()` in `script/utils/normalizeAddressStringForViem.ts` is a synchronous `export function` and calls the guard as the *first* term of its condition, so it runs on every network and cannot `await`.

Git history supports reading this as rule rot rather than a deliberate constraint: the dynamic-import paragraph was introduced in `0a63eb611` (2026-03-26) and untouched since, two months *before* the migration. At that commit the rule was genuinely correct — `tronWebFactory.ts` was its own module and `isTronNetworkKey` lived in a dependency-free local file with no TronWeb in its import graph. The migration invalidated it. So the rule is now scoped to what it actually governs — in-repo chain-specific implementation modules — with the devkit helpers stated as a top-level import instead.

**Also fixed, same defect class**

Sweeping for the phantom path turned up every other dead pointer left by the same migration. Left as-is they would send an agent editing a Tron script to a non-existent file, which is exactly where these rules fire:

- `202-tron-scripts.md`: `tron-network-keys.ts`, `tronTvmChain.ts`, `tronRpcConfig.ts`, `tronWebFactory.ts`, `tronWebCodecOnly.ts` and `tronAddressHelpers.ts` all repointed to `@lifi/tron-devkit`. The codec line also named `getTronWebCodecOnly()`, which the package does not export and which has no call site anywhere — dropped. Its "Helpers directory" examples (`tronPricing.ts`, `tronWebFactory.ts`, `formatAddressForCliDisplay.ts`) no longer exist in `script/deploy/tron/helpers/`; swapped for files that do. "Do NOT construct `new TronWeb(...)` outside the factory" no longer named a location in this repo now that the factory is external — reworded to "anywhere in repo code".
- `201-safe-decode-scripts.md`: same phantom file in the "adding a new chain" step, plus a note that the key guard now lives in the package, so adding one is a publish-and-bump rather than a local edit.
- `script/deploy/safe/confirm-integrity-asserts.ts`: the repo's only remaining `await import('@lifi/tron-devkit')` sat in a file that already imports the same specifier at module scope, so the deferral was provably a no-op. Folded into the existing top-level import — the new rule's one live counter-example.
- `script/deploy/safe/executors/create-chain-caller.ts`: the lazy-load comment claimed "EVM-only runs never pull in TronWeb", which the measurement disproves — line 7 of that same file top-level imports the devkit. Restated to what actually holds: the lazy import skips the Tron caller's own imports, and explicitly does *not* keep TronWeb out of the process.

Docs plus two small TS changes; no behaviour change. Edits are in the canonical `.agents/` files — the `.claude/` and `.cursor/` copies are symlinks and were not touched.

**Verification**: `bunx eslint` and `bunx tsc-files --noEmit` exit 0 on both touched TS files. `bun test script/deploy/safe/confirm-integrity-asserts.test.ts script/deploy/safe/confirm-integrity-asserts-placement.test.ts` — 73 pass, 0 fail. `bun lint:md:fix` reports 0 issues across 270 files and produced no unrelated churn.

**Unrelated pre-existing red**: `validate-json` fails on `origin/main` itself — `script/deploy/resources/immutableRegistry.json` declares `EcoFacet` twice (lines 2 and 122; the two blocks are byte-identical). #2406 added the second copy and the duplicate-key checker in #2414 landed right after. Not caused by this PR, which touches no JSON, and it is not a required check. Tracked separately.


0xDEnYO:
Folded into #2433 — its commits are now on that branch, so this PR is closed to keep the review queue to one unit per subsystem. No work is lost.

PR #2437, opened 2026-09-21
fix(scripts): make every CLI entry guard fire under bunx tsx (EXSC-1039)

gvladika:

Fixes [EXSC-1039](https://linear.app/lifi-linear/issue/EXSC-1039/bun-healthcheck-runs-zero-checks-and-exits-0-false-all-clear)

Also closes part 1 of [EXSC-971](https://linear.app/lifi-linear/issue/EXSC-971/port-shipped-script-modules-off-bun-only-runtime-apis-importmetamain) (the `import.meta.main` half). The `Bun.*` half of that ticket is untouched and still open.

`tsx` drops `import.meta.main` for `.ts` entry modules, so every CLI guarded with it loaded, evaluated the guard as `undefined`, and exited 0 having run nothing. The reported symptom was `bun healthcheck` (`package.json` → `bunx tsx ./script/deploy/healthCheck.ts`), but the same guard sat in eleven other modules, two of them invoked from CI: `healthCheckAllNetworks.yml` ran the daily production-fleet sweep as a no-op — never writing `warned_count`, which coerces to `0`, so the Slack composer stayed silent and the "Fail if any network failed" step saw success — and `deploymentAddressConsistency.yml` passed its PR gate the same way. A silent `exit 0` is indistinguishable from a real all-clear, which is why this is worse than a crash.

EXSC-1023 already built the fix — `isEntrypoint(import.meta.url)` in `script/utils/is-entrypoint.ts`, which realpaths both sides — but only converted four call sites. This routes the remaining twelve through it. I did not switch `package.json` to bare `bun run`, the other option the ticket floated: `200-typescript.md` says scripts are invoked with `bunx tsx` so the pinned local `tsx` and `node_modules` resolution are used, and the two CI jobs invoke `bunx tsx` directly where a `package.json` edit would not have reached them anyway. Fixing the guard fixes every caller at once. Conversely I left `healthCheckForNewNetworkDeployment.yml` and `deploy-smoke-test.yml` on bare `bun`: moving them onto `tsx` would run `script/utils/utils.ts` under Node, which still has 15 `Bun.*` call sites (EXSC-971 part 2).

While converting I found seven hand-rolled variants of the same check and folded them into the helper too, so there is one spelling to get right rather than eight. Five were realpath copies (`shared/assert-direct-broadcast-gate`, `tasks/proposePeripheryWithWhitelist`, the three `resources/*Reminder` modules). Two were not on anyone's list: `safe/verify-rehearsal` and `safe/rehearsal-teardown` guarded with `process.argv[1].endsWith('verify-rehearsal.ts')`, a suffix match that fails the opposite way — it fires for any entry sharing that basename rather than silently not firing.

`safe/execute-pending-timelock-tx.ts` stays deliberately unguarded. Its comment explains why: no guard, however spelled, should be able to silence the scheduled executor, and logic there that needs coverage moves to `safe/timelock-prefetch.ts` instead. I only updated the wording of that comment and its sibling, not the decision.

The new `script/entrypoint-guard-placement.test.ts` is a source assertion because a behavioural test cannot catch this class: under `bun test` the module is never the entrypoint, so the guard is correctly `false` however it is spelled. It holds `import.meta.main` at zero and asserts `is-entrypoint.ts` is the only non-test module that reads `process.argv[1]`, which keeps both the silent variant and the hand-rolled ones from coming back. I verified both assertions go red — reverting one guard, and adding a tracked module that reads `process.argv[1]` — before going green again.

Verification beyond the suite: `bun healthcheck --network sepolia` now runs the full invariant set (3 warnings, "Deployment checks passed"), confirming it reaches no `Bun.*` call under `tsx`; `bunx tsx script/tasks/checkDeploymentAddressConsistency.ts` now genuinely reports a pass; `contractDependencyReminder` and `assert-direct-broadcast-gate` both fire. `bun test script/` is green at 5795 tests.


gvladika:
Closing unmerged. The guard is genuinely broken under `bunx tsx` locally, but the impact is narrower than this PR's description claimed: both CI jobs I cited were verified to run correctly (healthCheckAllNetworks run 35570774580 sweeps 66 networks; deploymentAddressConsistency run 35348268321 printed a real result pre-fix). That leaves developer-local tooling only, which doesn't justify a 24-file change across the Safe, Tron and deploy scripts. EXSC-1039 closed as won't-do.

PR #2439, opened 2026-09-21
chore(agents): remove dated prompting patterns from rules and skills

gvladika:

Fixes [EXSC-1076](https://linear.app/lifi-linear/issue/EXSC-1076/agent-config-remove-dated-prompting-patterns-from-rules-and-skills)

Ran a prompt audit over the agent surface (~9,200 lines across `.agents/rules`, `.agents/commands`, `context.md` and `CLAUDE.md`), looking for instructions written for older models — not for volume. Current models follow instructions more literally, so leftover text is not merely wasted tokens: emphasis added because an older model under-triggered now causes over-triggering, and a rule restated three times makes the reader reconcile wordings instead of following one. I removed only what I could tie to a named pattern with a reason, and deliberately left working redundancy alone — `200-typescript`'s lint gate still overlaps `099-finish`, and `deprecate-contract` / `deprecate-network` keep their pre-execution checklists, because those validate *input* rather than restate a workflow.

Three changes worth calling out. `003-context-monitor` asked the model to self-estimate "% context full" and to hand off before a reset — a number it cannot measure, contradicting the harness's own auto-summarisation, and nothing checked it. `deploy-contract`'s description opened with ~100 words routing the reader to `multisig-rollout`; that boundary is already asserted from `multisig-rollout`'s own side and survives in `deploy-contract`'s body table, so what is left here is one scope clause. `analyze-tx` stated its analysis rules five times (trace-first principle, quick checklist, critical rules, quality checks, and the execution-flow step pointing back at them) — and two of its `CRITICAL` markers sat inside the user-facing output template, where they would have rendered into the report the requester reads.

Domain content is unchanged throughout: trace over source, `msg.value` from the receipt, the selector cross-check order, premium-RPC-only, the audit-log field formats and the bash strict-mode rules all survive, each stated once and with the reason it exists. Two sweeps the audit started are now finished across the whole surface rather than only where the audit happened to look: the retired "VPN" term is gone from the four rotation/rollout commands that still carried it (`deploy-contract` and `context.md` had already moved to lifi-connect), and the references to `/pr-ready` — a command removed from this repo — are gone from `finish-rollout`, which was still calling it a mandatory gate.

A review pass over this branch caught two defects in the rewrites themselves, both fixed here. The new `099-finish` line named a `self-review-pass` skill that does not exist anywhere in the repo; the name comes from the wallet-rotation design doc, where it is still only a planned gate. And `add-audit`, whose PDF mandate I reduced to a rule plus its verifying step, ended up telling the agent to stop before writing the log entry if the PDF could not be saved while step 7 still wrote that entry first — step 7 now saves and verifies the PDF before touching the log, and re-carries the "filename does not already exist" check that the deleted Validation Checklist used to hold, since the copy step was an unconditional `cp`.

Peer review then found six things, all resolved on the branch. One was a defect the round above introduced rather than fixed: with the PDF now written before the log entry, the filename-collision branch read a stray report as proof the audit was logged, so a run that died in between stranded the audit permanently and looked fine doing it. That branch now compares the existing file against the pasted report to tell an interrupted run from a genuine same-day audit, and `auditLog.json` is the only thing that proves logging — matched on date, auditor, contracts and `auditCommitHash`, not on the step-4 audit ID, which is generated to be unique and so is absent from the log by construction.

The other five were routing and accuracy. `deploy-contract`'s description now leads with its scope instead of its capability, because a bare "deploy X to Y" naming no environment matched the capability sentence first and the mainnet exclusion sat three clauses later; the ~15 words up front buy back the symmetry that `multisig-rollout` was asserting alone. `600-transaction-analysis` names itself an activation gate again — the only justification the rules offer for its `globs: ['**/*']`, and what `010-agents-authoring` and `add-rule-or-skill` both point at. `context.md` counted 32 commands where there are 34, the README advertised a dry-run that `sweep-wallet-funds` says the script does not have, and `self-review-pass` is gone from the wallet-rotation design doc, which was the last reference to it anywhere in the repo.

Two review suggestions I did not take, both argued in the threads. Three descriptions (`manage-wallet-funds`, `interact-tron`, `eip7702-atomic-batch`) keep their quoted trigger phrases removed, since each deleted phrase maps onto a category the rewrite kept — "the dev wallet is out of gas on BSC" onto "gas top-ups", "the sweeper keeps taking the gas" onto "a compromised key being drained by a sweeper bot". And the README's other eleven command rows stay as they are: I checked each against its new description and only `sweep-wallet-funds` had drifted. Both are cheap to reverse if the reviewer prefers.

Not yet verified, and worth a reviewer's attention: the rewrites are argued from the text, not from a run. A live `/analyze-tx` and `/add-audit` would confirm the two runbooks still carry what they need — `/add-audit` especially, since its ordering is what changed twice — and a bare "deploy X to <mainnet>" should route to `multisig-rollout` on the strength of `deploy-contract`'s new leading clause alone.


.agents/commands/add-audit.md
```diff
@@ -132,32 +130,34 @@ When `/add-audit` is invoked with a pasted PDF:
...
-   - **Locate and save PDF file** (use the PDF file that was already pasted into the chat):
+   - **Locate and save the PDF file** (use the PDF file that was already pasted into the chat):
      - **Search for PDF file** in common locations (in order):
        1. User's Downloads: `~/Downloads/` or `/Users/<username>/Downloads/`
        2. User's Desktop: `~/Desktop/` or `/Users/<username>/Desktop/`
        3. Current workspace: `.` or workspace root
        4. Search command: `find ~/Downloads ~/Desktop . -maxdepth 3 -name "*.pdf" -type f 2>/dev/null | grep -i "audit\|report\|repot"`
+     - **If the generated filename already exists** in `audit/reports/`: never overwrite it. A
+       genuine same-day audit of the same contract takes the next `_1` / `_2` suffix (step 5);
+       otherwise this audit is already logged — stop and ask the user which case it is.
```
0xDEnYO: This branch infers "already logged" from the presence of the PDF, but step 7 now writes the PDF **before** the log entry - so the two can disagree.

Run 1 copies the report to `audit/reports/2026.09.21_FooFacet(v1.0.0).pdf`, then the session dies before the `auditLog.json` write. Run 2 with the same PDF regenerates the same filename, sees it exists, concludes "otherwise this audit is already logged" and stops. The audit never gets into `audit/auditLog.json`, and nothing looks wrong because the report file really is there.

The thing that proves it was logged is the log: check `audit/auditLog.json` for the audit ID from step 4 rather than the PDF path. (Pre-change the ordering was the other way round, so this failure mode is new.)

.agents/commands/deploy-contract.md
```diff
@@ -1,6 +1,6 @@
...
+description: Deploys a facet or periphery contract (the version currently in the repo) to one or more networks and registers it in each network's LiFiDiamond — CREATE3 deploy, explorer verification, diamondCut (facets) or diamondUpdatePeriphery (periphery), plus the diamond allowlist sync for diamond-called periphery. This is the staging/testnet deploy path, and the deploy primitive `multisig-rollout` calls. Scope — staging and testnet targets only, or delegation from `multisig-rollout`; a deploy landing in a production (mainnet) diamond goes through `multisig-rollout` instead, which owns the Safe-proposal lifecycle that registration there requires. Tron (`tron`/`tronshasta`) goes to `deploy-contract-tron` — Foundry cannot deploy there. Requires Foundry, gh, and (production only) lifi-connect for MongoDB.
```
0xDEnYO: Agreed the old ~100-word caps-and-ALWAYS block was too much, but I would keep one leading clause rather than none.

As written the first sentence reads as a general-purpose deploy skill ("Deploys a facet or periphery contract ... to one or more networks and registers it in each network's LiFiDiamond") and the mainnet exclusion is three clauses later in `Scope -`. For a bare "deploy EcoFacet to polygon" with no environment named, that is the text selection sees first.

You are right that `multisig-rollout` asserts the boundary from its own side, and it does so with the exact phrasings ("Also triggers on a bare 'deploy <Contract> to <network>' ... production deploys ALWAYS enter here, never `deploy-contract` directly"). So the gate is not gone - it is one-sided now. Given the failure is a production deploy whose registration never enters the Safe-proposal lifecycle, ~15 words up front ("Staging and testnet only; a mainnet target goes to `multisig-rollout`.") buys back the symmetry cheaply. This is also the change your own PR body flags as unverified, so it is the one I would most want a live check on.

.agents/commands/manage-wallet-funds.md
```diff
@@ -1,7 +1,7 @@
...
+description: Moves or provisions funds for any LI.FI wallet whose private key is in `.env`, routing through the LI.FI API — gas top-ups, cross-chain rebalancing, and acquiring a token needed to test a route. Three modes — `bridge` (native gas asset cross-chain, same wallet on both ends), `swap` (same-chain native↔ERC-20 in either direction, same wallet), and `send` (gas asset to a different recipient). `bridge` and `swap` keep custody inside one wallet and run autonomously; `send` changes custody and requires an explicit human `--confirm` that an agent must never set on its own. Resolves the wallet by role (devWallet, refundWallet, deployerWallet, pauserWallet, …) or raw address from `config/global.json` `walletKeys` + `.env`. EVM only — Tron routes to `interact-tron`. Requires `bun` and the wallet's key in `.env`.
```
0xDEnYO, line 4: Four descriptions lose their quoted natural-language triggers in this PR: this one, `deploy-contract`, `interact-tron`, `eip7702-atomic-batch`.

The `description` field is the only text available when the user describes a symptom instead of typing the slash command. "The dev wallet is out of gas on BSC" and "swap native into USDC so I can test" now appear nowhere in this description; `interact-tron` loses "check the balance on tron" / "send TRX to X"; `eip7702-atomic-batch` loses "the sweeper keeps taking the gas before my tx lands".

`010-agents-authoring` (skill authoring) requires `description` to carry both the *what* and the *when*, and those phrasings are the *when*. The rewrites keep the semantics but drop the literal anchors - I would keep one short trigger list per skill.

.agents/rules/099-finish.md
```diff
@@ -12,6 +12,6 @@ alwaysApply: true
...
 - **Health-check invariants**: If you added, removed, or changed a facet or periphery contract, confirm `script/deploy/healthCheckInvariants.ts` was reviewed for needed invariant changes (see `601-healthcheck-invariants`). If it binds an external protocol address immutably, confirm the constructor arg carries a `getter` annotation in `deployRequirements.json`.
-- **PR review feedback**: The local CodeRabbit pre-flight (`/pr-ready`) has been removed — the CLI was hitting rate limits and wasting review cycles across the team. Cloud CodeRabbit still runs in GitHub CI on every PR, unaffected; treat its comments as the review backstop. The `self-review-pass` skill (mechanical/semantic/executable sweep, unrelated to CodeRabbit) remains the mandatory step before pinging for review.
+- **PR review feedback**: Walk the full diff yourself before pinging for review — that is the local gate, and there is no local CodeRabbit step. Cloud CodeRabbit runs in GitHub CI on every PR and is the review backstop.
```
0xDEnYO, line 15: `self-review-pass` does exist - it is a user-level Claude skill, and it is the team's mandated pre-review step (`docs/wallet-rotation-skill-suite.md:131` still names it). "Does not exist anywhere in the repo" is true, but that is a statement about the repo, not about the gate.

The consequence of the rewrite is that prose does not invoke a skill. After this change an agent satisfies this line - and `finish-rollout` Phase 5 step 1, and the `finish-rollout` description - by reading the diff inline, so the structured mechanical/semantic/executable sweep never runs, and nothing in the repo points at it any more.

Either keep the name (and accept that it resolves outside this repo) or drop the gate on purpose and say so. The current wording does neither.

.agents/commands/add-rule-or-skill.md
```diff
@@ -66,7 +66,7 @@ Negation patterns (`!src/**/*.s.sol`) are supported in `globs:` (Cursor) but **n
 ## Scoping & Activation (globs-first)
 
 - Prefer **precise `globs`** over cross-references.
-- Use `alwaysApply: true` (+ omit `paths:`) only for truly universal rules (role/guardrails, architecture, project structure, context monitoring, final checks).
+- Use `alwaysApply: true` (+ omit `paths:`) only for truly universal rules (role/guardrails, architecture, project structure, final checks).
 - Avoid `globs: ['**/*']` unless the rule is an explicit **activation gate** (like tx analysis) or truly universal.
```
0xDEnYO, line 70: This line and `010-agents-authoring.md:67` both still cite the tx-analysis rule as the canonical "activation gate" that justifies `globs: ['**/*']` - but `600-transaction-analysis.md` was rewritten in this PR to drop that framing, while keeping `globs: ['**/*']`.

So an author who follows the pointer finds a routing note with no activation gate in it, and 600's own wildcard glob loses the only justification the rules offer for it. Either keep one sentence in 600 naming itself the activation gate, or update both pointers (here and 010:67) to whatever the exception is now called.

.agents/README.md
```diff
@@ -76,7 +72,7 @@ Custom commands live in `.agents/commands/` (source of truth) and are symlinked
 | `deprecate-contract.md` | `/deprecate-contract <Name> ...` | Deprecate facet/periphery contracts by removing them from the codebase                  |
```
0xDEnYO, line 72: `010-agents-authoring` -> README update triggers: "Command changes: **always** update the Custom Commands table", plus a README review when a rule's `name`/`description`/`globs` change.

Twelve command descriptions changed here (`add-audit`, `analyze-tx`, `create-pr`, `deploy-contract`, `eip7702-atomic-batch`, `interact-tron`, `manage-wallet-funds`, `move-tron-delegation`, `offboard-sc-dev`, `rotate-deployer-wallet`, `rotate-pauser-wallet`, `sweep-wallet-funds`) and `000-global-standards`' frontmatter description changed, but only the `finish-rollout` row and the two 600 lines were touched.

The remaining rows are still broadly accurate, so nothing breaks today - but this is the drift the trigger exists to prevent, and a PR about the hygiene of this surface is the place to honour it.

(Unrelated, spotted while checking the table: the `sweep-wallet-funds` row advertises a "dry-run preview" while its own description says the script has no dry-run flag. Pre-existing, not yours.)

0xDEnYO (commented):
Reviewed the whole diff. The sweeps check out: no dangling references anywhere for `003-context-monitor`, `/pr-ready`, or the deleted `Quick Checklist` / `Critical Rules` / `Quality Checks` / `Validation Checklist` sections; symlink parity is intact (28 rules x 2 = 56); the README rule table matches the 28 files; every `add-audit` checklist item you deleted is genuinely re-carried by steps 3-5, including the `_1`/`_2` suffix rule the new text cites; `300-bash`'s "Pitfalls not covered above" heading is accurate; and `analyze-tx`'s 12 -> 9 rule compression loses no content.

Seven things inline. One is a real ordering bug in the `add-audit` rewrite; the rest are routing/consistency calls, the `deploy-contract` description being the one with a live cost.

Two notes that have no line to sit on:

- `.agents/context.md:11` says `32 command files`; there are 34. Pre-existing (the rule count was wrong on main too and is only correct now because 003 went away), but this PR edits that file and is about the accuracy of exactly this surface, so it is cheap to fix here.
- The `CodeRabbit` check on this PR is green via the rate-limit path ("Review limit reached"), so no bot review actually ran - the green check is not evidence here.

.agents/commands/add-audit.md
gvladika: Confirmed and fixed in d4337ec40.

One correction to the suggested check: the audit ID from step 4 cannot serve as the test. Step 4 generates it to be unique (`auditYYYYMMDD_N` if a same-day entry exists), so a freshly generated ID is absent from the log by construction and the lookup would never match — including in the case you describe, where the interrupted run's ID is exactly the one regenerated.

So the collision branch now does two things instead:

- Compares the existing file with the pasted report (`cmp -s`). Identical means an earlier run was interrupted: keep the PDF and continue to the log entry. Different means a genuine distinct same-day audit: take the next `_1` / `_2` suffix. That is the only signal that separates the two, since neither case leaves a log entry behind.
- The log write is now explicitly the proof of logging, and checks `audits` for an entry describing *this report* (date, auditor, contracts, `auditCommitHash`) rather than for the generated ID. The file carries a note saying why the ID is not the test, so this does not get re-derived wrongly later.

.agents/commands/deploy-contract.md
gvladika: Agreed — fixed in 0bba0fc30.

The description now opens with the scope rather than the capability:

> Staging and testnet targets only — a mainnet target goes to `multisig-rollout`, including a bare "deploy \<Contract\> to \<network\>" that names no environment.

I named the bare-invocation case explicitly since that is the one with the live cost, and dropped the trailing `Scope —` clause, which the lead now carries; what is left of it is the reason (`multisig-rollout` owns the Safe-proposal lifecycle) stated once. Net change is about +10 words.

Still unverified by a live run, as the PR body says. Worth a check once this merges.

.agents/commands/manage-wallet-funds.md
gvladika: Took `deploy-contract` (0bba0fc30 — see the other thread); I would leave the other three as they are, and here is the reasoning rather than a flat no.

The rewrites replaced each trigger list with the category that contains it, so the *when* is still there — just not as quotations:

| deleted phrase | what now carries it |
|---|---|
| "the dev wallet is out of gas on BSC" | "gas top-ups" |
| "rebalance the deployer across chains" | "cross-chain rebalancing" |
| "swap native into USDC so I can test" / "get the funds I need to test X" | "acquiring a token needed to test a route" |
| "check the balance on tron" | "view calls" |
| "send TRX to X" | "transactions (transfer, approve, admin calls)" |
| "what's the base58 for this address" | "address checks and base58↔hex conversion" |
| "the sweeper keeps taking the gas before my tx lands" | "a compromised key being drained by a sweeper bot" |

`010-agents-authoring` asks the description to carry *what* and *when*; I read the category as satisfying that, and the same file also says every line costs tokens. `deploy-contract` is genuinely different — there the failure is a production deploy whose registration skips the Safe-proposal lifecycle, which is worth paying for; the other three fail soft (the user rephrases).

That said, this is an empirical question about selection behaviour and I am arguing it from the text, same as the rest of the PR. If you would rather have the literal anchors back it is ~120 words and I will restore all four — say the word and I will push it rather than leave it to a follow-up.

.agents/rules/099-finish.md
gvladika: I went looking for it and could not find it. On this machine `~/.claude/skills/` holds `grill-me`, `grilling`, `solidity-auditor`, `synced`; `find ~/.claude -iname '*self-review*'` returns nothing, including under `~/.claude/plugins`; and it is not in the session's available-skills list. The only occurrence anywhere was `docs/wallet-rotation-skill-suite.md:131`.

If it resolves for you, it is installed per-machine and not shared, which has the same consequence for anyone else running these skills: the line names a gate that does not fire. So I do not think "keep the name" is available as an option here.

That leaves your second option, and I think the branch already takes it — the line says what to do (`Walk the full diff yourself before pinging for review — that is the local gate`) rather than leaving a hole. What it did not do was clean up after itself, so b72989881 removes the `self-review-pass` reference from `docs/wallet-rotation-skill-suite.md:131` as well. There is now no reference to it left in the repo.

If the structured mechanical/semantic/executable sweep is worth having as a real gate, that is a skill someone has to write and check in — happy to open a ticket for it, but it should not be a name in prose in the meantime.

.agents/commands/add-rule-or-skill.md
gvladika: Confirmed — both pointers were left dangling. Fixed in b72989881 by the first of your two options: 600 names itself again, in one sentence tied to the thing it justifies.

```
This is an **activation gate**: it fires on what the user asked, not on which file is open,
so it carries `globs: ['**/*']` rather than a directory scope.
```

That keeps `010-agents-authoring.md:67` and this line resolving, and puts the wildcard glob's justification in the file that carries the glob — which is where it was missing. Updating the two pointers instead would have left 600's `globs: ['**/*']` unexplained in its own file.

.agents/README.md
gvladika: I checked all twelve rows against their new descriptions rather than rewriting them, and only one was actually wrong — the `sweep-wallet-funds` one you spotted. Fixed in b72989881: "dry-run preview" → "read-only balance preview (the script has no dry-run flag)", which is what its own description says.

The rest I would leave. The summary column is a short gloss, not a mirror of the description, and none of the eleven says anything the rewrites made untrue — `deploy-contract`'s row already reads "Staging/test terminal path … production rollouts go through `multisig-rollout`", for instance. Rewriting accurate rows to track wording changes is churn in a diff, and it is the kind of churn that makes the next person distrust the table rather than trust it.

I read "Command changes: always update the Custom Commands table" as *review the table whenever commands change* — the trigger exists to catch drift, and here the check found one row of drift and it is fixed. If you meant it literally as "rewrite every touched row", that reading should be in `010-agents-authoring` explicitly, because the current wording does not distinguish the two and I would rather not guess at it again next time. Happy to make that edit either way.

Also picked up `.agents/context.md`'s command count in the same commit: 32 → 34.

gvladika:
@coderabbitai review

0xDEnYO (approved):

PR #2442, opened 2026-09-22
test(tron): take the live TronGrid reads out of the TS unit-test suite

gvladika:

Fixes [EXSC-1078](https://linear.app/lifi-linear/issue/EXSC-1078/take-the-live-trongrid-mainnet-reads-out-of-the-ts-unit-test-suite)

The `propose-to-safe-tron` pass case spawned the real CLI with no `RPC_URL_TRON`, so the devkit fell back to the public keyless TronGrid host (`getTronRpcUrl`: override > env > `https://api.trongrid.io`) and the case read Tron mainnet. Its marker was the proposal store's refusal, which the CLI reaches only after four reads — `getMinDelay`, the salt pick's `hashOperationBatch` and `getTimestamp`, and the Safe `nonce` — so the case passed only if all four answered. #2427 took that count from two to four and crossed the keyless host's rate limit, which is what turned it red in CI.

The ticket offered two routes: pass `TRONGRID_API_KEY`, or serve the reads from a local stub. I took a third that is cheaper than either. The gate prints `Production deploy gate passed` *before* the first read, so the read's own failure is already proof that the funnel got past the gate — no chain required. The case now runs against `UNROUTABLE_RPC` like its two refusal siblings and asserts `Could not read getMinDelay from Timelock` instead of the store marker. That also fixes an asymmetry: `TIMELOCK_READ_FAILURE` was asserted absent in two cases with no positive anchor anywhere, so a reword would have left both passing silently. It is now anchored, and the store marker it replaces was already covered directly by `mongo-store-transport.test.ts:154`.

Dropping the store marker from the refusal case loses nothing: the Timelock read is the first step past the gate and the store is opened after it, so a run that never read cannot have stored, and `assertChildIsUsable` trips on `Proposal stored in MongoDB.` for every spawn regardless.

The harness now also deletes `RPC_URL_TRON`. Bun loads the repo env file into the test process and the harness cleared every other credential-shaped name but not this one, so a developer's own endpoint was inherited locally while CI fell through to the public host — which is why the flake only ever showed up in CI.

Verified by running the suite with outbound network denied (macOS `sandbox-exec`, loopback and unix sockets allowed so local stub servers and tsx's IPC still work). Before: 8 failures, one of them this chain read. After: 7, and no case in `script/` reads a chain. The two remaining causes are not chain reads — `calldata-effect-lines.test.ts` resolves selectors missing from the local table through the Sourcify 4byte API, and the CLI probes that spawn `bunx tsx` into a fixture directory resolve `tsx` from the npm registry. The workflow header now names both instead of claiming the suite is fully self-contained; each is worth its own ticket.

Assertion strength was checked by feeding the case a diverged repo: it goes red on the gate-refusal assertion, and the refusal lands before any read, so the new marker assertion fires too.

Not changed: the `TRON_READ_MAX_ATTEMPTS`/`TRON_READ_RETRY_DELAY_MS` backoff on the production propose path, which the ticket asked about. It is not only test-driven — `runPendingTimelockTXs.yml` exercises that path with no `TRONGRID_API_KEY` wired (`healthCheckAllNetworks.yml` is the only workflow that passes it), so the keyless exposure it was written for is real. The separate concern that `isRateLimitError` classifies by message substring belongs in its own ticket: it lives in `script/deploy/shared/rateLimit.ts`, predates #2427, and is shared with `healthCheckInvariants.ts`.

Note on the tests item: no test was added — this PR *is* a change to an existing test, re-pointing the Tron pass case off mainnet while keeping (and anchoring) what it asserts. The suite count is unchanged at 6005 and the file's 15 cases pass with the network denied.


0xDEnYO (approved):

.github/workflows/ts-unit-tests.yml
```diff
@@ -6,8 +6,13 @@
...
 #   can remain a required check on `main` branch protection even when the test job is
 #   skipped (same pattern as `forge-unit-tests.yml`, see `.agents/rules/500-github-actions.md`).
-# - Lightweight: the suite is fully self-contained (no Foundry, submodules, MongoDB, RPC,
-#   or generated typechain bindings), so the job only installs Bun deps and runs the tests.
+# - Lightweight: no Foundry, submodules, MongoDB, RPC, or generated typechain bindings, so
+#   the job only installs Bun deps and runs the tests. Two network dependencies remain, and
+#   neither is a chain read: `calldata-effect-lines.test.ts` resolves selectors that are not
+#   in the local table through the Sourcify 4byte API, and the CLI probes that spawn
+#   `bunx tsx` into a fixture directory resolve `tsx` from the npm registry. Both were
+#   measured by running the suite with outbound network denied, which is how to re-check
```
0xDEnYO, line 14: Good change — naming the two remaining network dependencies is strictly better than the "fully self-contained" claim it replaces, and pointing at how to re-measure it is the part that keeps it honest.

The body says each is worth its own ticket. Were they filed? An accurate comment that outlives the intent behind it is exactly how the previous claim went stale, and this one now carries a measurement date it doesn't state.

0xDEnYO (commented):
Already approved — one follow-up question inline, not a blocker.

For the record, I checked the three load-bearing claims rather than taking the body's word: `RPC_URL_TRON` is genuinely what the devkit resolves (`script/utils/rpc-url-log-scan.ts:53` and its test), the marker string exists verbatim at `propose-to-safe-tron.ts:174`, and the ordering argument for dropping `NEXT_STOP_TRON` holds — `getMinDelay()` is read at :166 and the store is opened at :289, so a run that never read cannot have stored. Net this is stronger than what it replaces: `TIMELOCK_READ_FAILURE` was asserted absent twice with no positive anchor anywhere, and now has one.

Also worth knowing: CodeRabbit reports "Review rate limited" on this PR, so the green check means no review ran rather than a clean one.

maxklenk (approved):

.github/workflows/ts-unit-tests.yml
gvladika: Filed — both, just now:

- **EXSC-1082** — [calldata-effect-lines.test.ts reaches the Sourcify 4byte API](https://linear.app/lifi-linear/issue/EXSC-1082/calldata-effect-linestestts-reaches-the-sourcify-4byte-api-give-the)
- **EXSC-1083** — [CLI probes that spawn `bunx tsx` into a fixture dir resolve tsx from the npm registry](https://linear.app/lifi-linear/issue/EXSC-1083/cli-probes-that-spawn-bunx-tsx-into-a-fixture-dir-resolve-tsx-from-the)

Both are 1-point, assigned, and related to EXSC-1078. Each names the specific call site rather than the symptom, so they don't decay into "the suite touches the network somewhere":

- EXSC-1082: `buildCalldataEffectLines` calls `resolveSelectorsViaFourByte` with no options at `calldata-effect-lines.ts:397` and `:575`, so the `fetcher` seam the resolver already exposes (`selector-registry.ts:374`) is unreachable from the test. Seeding `SELECTOR_SIGNATURE_CACHE_PATH` doesn't close it either — misses are deliberately never persisted (`selector-registry.ts:355`), so an unresolvable selector fetches every run.
- EXSC-1083: it's narrower than "the CLI probes" in the body. Only spawns with a **fixture cwd** are affected — `verify-immutable-registry.test.ts:177` runs with `cwd: root` from `mkdtempSync` at `:157`. The probes using `cwd: REPO_ROOT` (`helperFunctions.gate.test.ts`, `updateWhitelistPeriphery.test.ts`, `manifest-abort.cli.test.ts`, `verify-approvals.test.ts:711`) resolve `tsx` locally and are fine. Sweeping for any others is in the ticket's acceptance.

On the missing date: fair, and you're right that it's the same failure mode as the claim it replaces. Pinning it here rather than leaving it implicit — **measured 2026-09-22 at `5a5e606`**, suite run under macOS `sandbox-exec` with outbound denied and loopback/unix sockets allowed. Each ticket carries that method in a Verification section and has "update the workflow header" in its acceptance list, so the comment gets corrected by whichever ticket lands first instead of relying on someone remembering. I'd rather not add the date to the header in this PR — a date in a comment is the thing that goes stale, and the re-measure instruction is what actually keeps it honest.

Also noting your CodeRabbit point for the record: rate-limited on `5a5e606`, so that green check means no review ran. The earlier run on `5c21df0` did complete clean, and the only change between them is the `RPC_URL_TRON` deletion in the harness.

PR #2443, opened 2026-09-22
fix(tron): flag the deployment record when TronScan verification succeeds (EXSC-757)

gvladika:

https://linear.app/lifi-linear/issue/EXSC-757/deploy-updated-ecofacet-after-tron-fix-and-audit

Found while rolling out EcoFacet v2.0.0 to Tron under that ticket.

`verify-tron-contracts.ts` submits to TronScan and prints a summary, but never
writes MongoDB — so a Tron contract verified through it stays `verified: false`
in the deployment record indefinitely. The EVM path has no such gap:
`deploySingleContract.sh` sets `VERIFIED` inline after verifying. Nothing on the
Tron path did.

Reproduced on this rollout: `EcoFacet` at `TRzEvqhoK6pC58rmUFMgCoEtQgiHYh2FCQ`
verified successfully on TronScan and the record still read `verified=false`.

**Why a new `mark-verified` subcommand rather than reusing `update`.**
`update` matches on `version`, and a verification run knows the address it just
submitted but not the version the record carries. Reading the version from local
source would be wrong on Tron specifically: the fork carries `-tron` overlay
versions that `main` does not, so a `main` read would silently match nothing and
the flag would stay false with a success exit. Resolving the version from the
record itself keeps that out of the caller's hands.

**Why it selects by address and flags every record that names it.** One contract
can hold several records over its life. This rollout produced exactly that — an
abandoned deploy left an earlier `EcoFacet` v2.0.0 on tron — and the abandoned
record can be the *newer* of the two, so "newest record" and "address just
verified" are not the same contract. Selecting by recency would flag an address
nobody submitted while the live one stayed unflagged for good. `resolveRecordsToFlag`
therefore filters on the address and returns **every** record naming it (they all
describe the contract that was just verified), and throws when none does, naming
the recorded addresses so the gap is actionable.

**Why the caller warns rather than fails.** If the record write fails, the
contract is still verified on the explorer. Failing the verification run over
bookkeeping would be the worse outcome, so `verify-tron-contracts.ts` logs a
warning and carries on to the next contract — but collects the unflagged names
and exits non-zero at the end, because nothing else on the Tron path writes that
flag, so a green run is the last chance anyone looks.

The guard lives in its own module because `update-deployment-logs.ts` reads
`MONGODB_URI` and calls `runMain` at load, so importing it from a test would
fire the CLI.

## The TronScan success-message matcher

The other substantive change in this PR, and the riskier one. The matcher was a
substring regex (`/validated|verification success|already (been )?verified/i`);
it is now three patterns anchored to the whole trimmed message. `interpretResponse(true,
'contract validated ok')` deliberately flips from `ok` to `fail`.

The reason is the direction of the error. A match here is persisted to the
deployment record as `verified: true`, so a substring search turns a negated or
in-progress message — "not already verified", "already in the verification
queue" — into a durable false claim that the contract's source is published.

**Where the three wordings come from.** They are the only success wordings
anyone here has observed, and there is no documentation to check them against:
the endpoint is undocumented and was reverse-engineered in #2095, where a live
submission returned status `2001` ("The contract has been validated.") and a
separate code whose message is "Verification success.", and re-submitting an
already-verified contract returned the third. That is the whole basis — not a
published enum.

**Why failing closed is safe: it is recoverable.** The new failure mode is that
a real success wording outside these three reads as a failure, exits 1, and
leaves the record unflagged — the same symptom this PR exists to fix. What makes
that acceptable is that it is loud and non-destructive rather than durable:
nothing is written, the wording is printed, and the contract is verified on
TronScan either way — so a re-run answers "already verified", which the list
accepts. An unseen wording costs one re-run after it is added to the list; a
wrong accept costs a permanently wrong record that nothing will correct. Failing
closed is the cheap direction.

## Tests

`script/deploy/shared/resolve-records-to-flag.test.ts` — 6 tests covering
`resolveRecordsToFlag`: the match case, the abandoned-deploy case where the
unrelated record is newer, several records naming the same address (so none is
left unflagged), no record at all, the message naming the recorded addresses
when none match, and exact comparison, since Tron base58 is case-significant.

`script/deploy/tron/helpers/tronscanVerify.test.ts` — the `interpretResponse`
cases were extended for the matcher change: each accepted wording, a queued
contract, a negated success message, an unrecognised message, a bytecode
mismatch, an HTTP error with a success-looking body, and a non-JSON body (which
is no longer a success — the raw-text fallback case was updated to assert that).

Verified both sets bite:

- dropping the address filter in `resolveRecordsToFlag` fails 3 of its 6 tests;
- un-anchoring the three patterns back to substring matches fails 3 of the 10
  `interpretResponse` tests (negated message, unrecognised message, non-JSON body).

## Manual verification

Ran against the real record. `mark-verified` flipped
`TRzEvqhoK6pC58rmUFMgCoEtQgiHYh2FCQ` to `verified=true`, and refused when given
the abandoned `TTukYpunN8FipnhvpbGhhd5646Er3YUhom` with the expected error.

## Documentation

`docs/TronFork.md` still described TronScan verification as a manual browser
upload, which stopped being true when the CLI landed in #2095; it now documents
the CLI, its flags, and the record flagging, keeping the browser path as a
fallback. `docs/DeploymentLogs.md` gains a section on the `verified` flag and
the standalone `mark-verified` command — there was no written home for that flag
at all.

It also records a residue this PR does not close: `AccessManagerFacet` and
`LiFiDiamond` are skipped by default on Tron (verified before the CLI existed,
sources not reproducible from the current fork checkout), so no run will ever
flag them and nothing else on that path writes the flag. Clearing it is a one-off
`mark-verified` per contract, out of band — noted in the code next to the skip
list and in `docs/DeploymentLogs.md` rather than done here, since it is a data
change, not a code change.

`package.json` gains `mongo-logs:mark-verified` alongside its `sync`/`add`/`update`
siblings. It is spelled `bunx tsx` rather than the siblings' bare `bun`, per
`200-typescript`, and matching how `verify-tron-contracts.ts` already invokes the
same command.


script/deploy/tron/helpers/tronscanVerify.ts
```diff
@@ -20,9 +20,13 @@ export const VERIFY_TIMEOUT_MS = 120_000
...
  * message is "Verification success." — so we treat a success *message* as
  * authoritative rather than relying on a single status code.
+ *
+ * "already" is matched only next to "verified": a bare "already" also appears
+ * in the queue message ("already in the verification queue"), which is not a
+ * verified contract, and a match here is persisted to the deployment record.
  */
 export const TRONSCAN_SUCCESS_MESSAGE_RE =
-  /validated|verification success|already/i
+  /validated|verification success|already (been )?verified/i
```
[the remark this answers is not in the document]
gvladika: Done in 9c88ca5fe. `TRONSCAN_SUCCESS_MESSAGE_RE` is replaced by `TRONSCAN_SUCCESS_MESSAGES`, an allowlist matched against the whole trimmed message, plus `isTronscanSuccessMessage`. "not already verified", "already in the verification queue" and any unrecognised wording now read as failures with the message printed, so nothing false is written to the record. A non-JSON body is no longer a success either — the raw-text fallback test was updated to assert that.

script/deploy/tron/verify-tron-contracts.ts
```diff
@@ -45,6 +47,54 @@ import {
...
+    )
+    return true
+  } catch (err) {
+    const message = err instanceof Error ? err.message : String(err)
+    consola.warn(
+      `${contractName}: verified on TronScan, but the deployment record was not flagged — ${redactUrls(
+        message
+      )}`
+    )
+    return false
```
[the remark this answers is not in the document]
gvladika: Done in 9c88ca5fe, fixed in the child rather than the caller: `mark-verified` now invalidates the deployment cache in its own try/catch that warns and leaves the exit code at 0. The records are written by that point and the cache is derived from them, so a non-zero exit from the child again means exactly what the caller reports — the update did not happen.

script/deploy/tron/helpers/tronscanVerify.ts
```diff
@@ -16,13 +16,34 @@ import { fetchWithTimeout } from '../../../utils/fetchWithTimeout'
...
+ * persisted to the deployment record as `verified: true`, and a substring turns
+ * a negated or in-progress message — "not already verified", "already in the
+ * verification queue" — into a durable claim that the contract is verified.
+ * An unrecognised message is reported as a failure with the wording printed,
+ * which is the cheap direction to be wrong in: nothing false is written, and
+ * the wording is there to be added here.
  */
-export const TRONSCAN_SUCCESS_MESSAGE_RE =
-  /validated|verification success|already/i
+export const TRONSCAN_SUCCESS_MESSAGES: readonly RegExp[] = [
```
0xDEnYO, line 41: This is the riskiest change in the PR and the body doesn't mention it at all — the description is entirely about the mark-verified gap, while this quietly swaps a substring regex for three whole-message patterns and deliberately flips `interpretResponse(true, 'contract validated ok')` from `ok` to `fail`.

The reasoning is right: a substring match writes `verified: true` durably off "not already verified", and that's a bad direction to be wrong in. But the new failure mode is that a real TronScan success wording outside these three patterns reads as a failure, exits 1, and leaves the record unflagged — which is the same symptom this PR exists to fix, just reached differently.

Two asks:

1. Where did the three patterns come from beyond the two shapes named in the doc comment? If they're the only wordings observed, say so — that's a fine basis, but it should be stated rather than implied by the allowlist's existence.
2. Put the recovery argument in the body, because it's what makes this safe and it's currently unwritten: a genuinely-verified contract answers "already been verified" on a re-run, which the allowlist accepts, so an unrecognised wording is loud, non-destructive, and fixed by running the command again. That property is the whole case for failing closed here.

script/deploy/tron/verify-tron-contracts.ts
```diff
@@ -45,6 +47,54 @@ import {
  */
 const ALREADY_VERIFIED = new Set(['AccessManagerFacet', 'LiFiDiamond'])
```
0xDEnYO, line 54: These two stay skipped by default, so their `verified` flags on Tron remain `false` indefinitely — the exact gap this PR closes for everything else, still open for two contracts.

Not a blocker, and I don't think it belongs in this diff. But since nothing else on the Tron path writes that flag, "someone will notice later" isn't true here. Either flag them once with the new subcommand out of band, or note the residue somewhere it'll be found.

0xDEnYO (commented):
Correctness is sound — I checked the load-bearing pieces against the code rather than the description:

- `spawnAndCapture` rejects on a non-zero exit, so the `flagged` boolean is real and this can't report a false green.
- `--env` is the right flag (not `--environment`).
- The `environment = 'production'` hardcode is correct: `deployments/<network>.json` is the production log for every network, staging is `<network>.staging.json` — confirmed against `saveContractAddress`.
- `bunx tsx script/deploy/update-deployment-logs.ts` matches the convention already used in `helperFunctions.sh` and `script/utils/utils.ts`, and the target has no Bun-only APIs.
- Shell-arg escaping is correct, `updateDeployment`'s filter includes the address so the per-record loop can't over-write, and keeping cache-invalidation failure out of the exit code is the right call.

Three things I'd like addressed before merge — two are inline, the third is the PR body itself:

**The body is one revision stale.** It describes `resolveRecordToFlag` (singular) in `resolve-record-to-flag.test.ts` with 5 tests, selecting *one* record. What shipped is `resolveRecordsToFlag` in `resolve-records-to-flag.ts`, 6 tests, returning *every* record naming the address. That's a design change, not a rename — the "why it refuses on an address mismatch" paragraph no longer describes the code it justifies. Please re-sync it; the body is the artifact the next reader reviews from.

Also: "I have updated any required documentation" is unchecked while the PR adds a new CLI subcommand. Either tick it or say why nothing needed updating. And there's no `mongo-logs:mark-verified` in `package.json` alongside `sync`/`add`/`update` — trivial, but the three siblings set an expectation.

One more, unrelated to the diff: CodeRabbit reports "Review rate limited" here, so the check is green but no review actually ran. Worth a `@coderabbitai review` if you want that backstop before merge.

script/deploy/tron/helpers/tronscanVerify.ts
gvladika: Both addressed in e9d40fef5, and the PR body now carries the change — that was the gap you named, since the description was entirely about the mark-verified path.

1. **Provenance.** These three are the only wordings anyone here has observed, and that is now stated rather than implied by the allowlist's existence. There is no published enum to check against: the endpoint is undocumented and was reverse-engineered in #2095, whose body records the basis — a live submission returned status `2001` ("The contract has been validated.") and a separate code whose message is "Verification success.", and the shape was "confirmed against an already-verified contract", which is where the third comes from. The doc comment on `TRONSCAN_SUCCESS_MESSAGES` and the PR body both say so and both name #2095, so the next reader can chase it.

2. **The recovery argument** is in the body under "The TronScan success-message matcher", and condensed in the doc comment: nothing is written, the wording is printed, and the contract is verified on TronScan either way — so a re-run answers "already verified", which the list accepts. An unseen wording costs one re-run after it is added to the list; a wrong accept costs a permanently wrong record that nothing will correct. That asymmetry is the case for failing closed, and you're right that it was load-bearing and unwritten.

I also re-ran the mutation check on the matcher, since the case list grew with the change: un-anchoring the three patterns back to substring matches fails 3 of the 10 `interpretResponse` tests (negated message, unrecognised message, non-JSON body).

script/deploy/tron/verify-tron-contracts.ts
gvladika: Took the "note it somewhere it'll be found" option in e9d40fef5, not the out-of-band flagging — flagging them is a production MongoDB write behind the tunnel, so it's a data change that should be a deliberate act with a person at the keyboard, not something folded into a code PR. Agreed it doesn't belong in this diff either way.

The residue is now recorded in two places: next to the `ALREADY_VERIFIED` set itself, which is where someone reading the skip list would ask the question, and in `docs/DeploymentLogs.md` under a new "Flagging a record as explorer-verified" section, which spells out that these two never reach the flagging step, that their records keep whatever `verified` value they were last given, and that clearing it is a one-off `mark-verified` per contract against the address in `deployments/tron.json`.

That section also covers the rest of your body feedback: the `verified` flag had no written home at all, and `docs/TronFork.md` still described TronScan verification as a manual browser upload, which stopped being true in #2095 — it now documents the CLI, its flags and the record flagging, keeping the browser path as a fallback. Documentation checkbox ticked on that basis.

`mongo-logs:mark-verified` added alongside the three siblings. One deviation worth flagging: it is spelled `bunx tsx` rather than their bare `bun`, per `200-typescript` and matching how `verify-tron-contracts.ts` already invokes the same command. I left the three siblings alone as out of scope.

Happy to run the two `mark-verified` calls once the tunnel is up if you want them cleared now — say the word and I'll do it as a separate, non-PR action.

gvladika:
@coderabbitai review

script/deploy/tron/helpers/tronscanVerify.ts
```diff
@@ -16,13 +16,42 @@ import { fetchWithTimeout } from '../../../utils/fetchWithTimeout'
...
+  /^verification success\.?$/i,
+  /^(the contract has )?already (been )?verified\.?$/i,
+]
+
+/**
+ * Whether TronScan's response message reports a verified contract.
+ * @param message - Server message, as returned or as raw body text
+ */
+export function isTronscanSuccessMessage(message: string): boolean {
+  const trimmed = message.trim()
```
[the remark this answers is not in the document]
gvladika: Not applying this one — the throw it describes can't reach `isTronscanSuccessMessage`, and the behaviour it asks for is already what happens.

`interpretResponse` is the only caller, and `message` is a string by construction there:

```ts
let message = body.slice(0, 200)          // string
try {
  const json = JSON.parse(body) as { data?: { message?: string } }
  if (json.data?.message) message = json.data.message.trim()   // <- inside the try
} catch {
  // Non-JSON body — match against the raw text below.
}
return { ok: httpOk && isTronscanSuccessMessage(message), message }
```

A truthy non-string `data.message` trips `.trim()` on that line, which is inside the `try`, so the `catch` swallows it and `message` keeps the raw body slice — still a string. Checked it rather than reasoning about it; `data.message` as `2001`, `{code: 2001}`, `['ok']` and `true` all return `{ ok: false, message: <raw body> }` and none throws. That is exactly the "return a failed result instead of throwing" outcome the comment asks for, and the raw body is more use to the operator than the `''` the suggested diff would substitute.

The premise is worth pinning even so, since nothing held that behaviour in place and it depends on the trim staying inside the parse block. Added a case in ba9649580 covering those four shapes; moving the `.trim()` out of the `try` fails it.

Adding `typeof message === 'string'` inside `isTronscanSuccessMessage` would be unreachable against a `string` parameter, so it would mean widening the signature to `unknown` to buy a branch no caller can hit.

0xDEnYO (approved):

